PR #3, opened 2012-12-20
Fix to issue when encountering JAR files missing manifest files

rongutierrez:
This was due to NullPointException being thrown when encountering a JAR
file which does not have a manifest file. The change will cause the
program to continue executing instead of just crashing when
encountering the problematic JAR file.


jeremylong:
Thanks Ron. I had found this issue also and had a fix in a different local branch. I just pushed 0.2.3.2 which contains the fix (among other fixes/enhancements)

jmanico:
You're a good man for caring about maintaining this so well....

Aloha,

## 

Jim Manico
@Manicode
(808) 652-3805

On Dec 22, 2012, at 1:49 AM, Jeremy Long notifications@github.com wrote:

Thanks Ron. I had found this issue also and had a fix in a different local
branch. I just pushed 0.2.3.2 which contains the fix (among other
fixes/enhancements)

—
Reply to this email directly or view it on
GitHubhttps://github.com/jeremylong/DependencyCheck/pull/3#issuecomment-11636465.

PR #36, opened 2013-11-26
Add option to whitelist CVE:s

herder:
This PR adds a  comma-separated parameter with the name org.owasp.dependencycheck.whitelisted which contains the name(s) of optional whitelisted CVE:s.

This is useful if we want to suppress warnings for certain CVE:s, for example if they have a bad version specification. ('All versions', for example)


jeremylong:
Thanks. However, when this PR came in I was in the middle of implementing a more full featured mechanism to suppress false positives. I just pushed 1.0.6 which contains the ability to specify a suppression.xml file. The HTML report can be used to copy the necessary data to build the XML suppression file.

Issue #37, opened 2013-12-09
Making failBuildOnCVSS configurable at runtime


phsh:
I need to run org.owasp:dependency-check-maven:1.0.7:check on a number of projects, and don't want to change the pom foreach project.

The setup would require the failBuildOnCVSS to be runtime configurable,

@Parameter(property = "failBuildOnCVSS", defaultValue = "11", required = true)
    private float failBuildOnCVSS = 11;

jeremylong:
I apologize if I am missing something - but because there is a defaultValue of 11 this element does not actually need to be configured in the pom. This value will default to 11 which basically indicates that it will never fail the build as the CVSS scores range from 0-10.

Are you asking for a different way to configure the fail build on CVSS property?

--Jeremy

phsh:
What I wanna do is:

mvn org.owasp:dependency-check-maven:1.0.7:check -Dcheck.failBuildOnCVSS=5

I hope that clarify me needs... Sorry for not being clear enough...

jeremylong:
phsh,

Sorry for the delay in getting back to this. I did not incorporate your PR as you can run the following command to set the failBuildOnCVSS score argument:

mvn dependency-check:check -DfailBuildOnCVSS=4

Please let me know if I am missing something.

--Jeremy

PR #38, opened 2013-12-09
Fix for the "Check fail build on cvss" issue

phsh:
ISSUE: #37

as there is no tests for the maven plugin ATM, I have just made the changes to the plugin.

The coding style of the patch isn't verified against anything, hopefully that won't be a problem.


jeremylong:
Per - thanks for the PR; I've been swamped and haven't had time to merge
this. Hopefully, I will be able to get to this within the week as I have a
couple of other changes in flight.   Coding style wise - I might change you
PR to use the settings object (global container for properties in DC), but
that will be a minor change.

Thanks!

--Jeremy

jeremylong:
Please see https://github.com/jeremylong/DependencyCheck/issues/37#issuecomment-32257684

--Jeremy

PR #41, opened 2014-01-08
NexusAnalyzer

colezlaw:
This is a new analyzer which runs during the `INFORMATION_COLLECTION` phase. It will take the SHA-1 hash of jar files and look those up on a configured Nexus repository. There are two properties (in `dependencycheck.properties`) which govern this behavior:
1. `analyzer.nexus.enabled` is a `boolean` specifying whether the analyzer is even enabled
2. `analyzer.nexus.url` is a `String` specifying the root REST service on which the searches should be executed

The final URL for a search would be something like `http://repository.sonatype.org/service/local/identify/sha1/<SHA-1>`, so the configured URL is `http://repository.sonatype.org/service/local/` since this may be extended in the future to allow for other types of searches.

`analyzer.nexus.enabled` is set to `false` in `src/test/resources/dependencycheck.properties` since there are other unit tests which execute all the Analyzers and we don't want tests to possibly fail because the build happened behind a proxy, etc. Furthermore, two of the test cases in `NexusSearchTest` have the `@Ignore` annotation to prevent those from running as they will actually attempt to communicate with the repository, possibly causing tests to fail. However, leaving it set to `true` when actually running the tool should not cause any analysis failures, but may add exceptions to the analysis exception list for any dependency.


jeremylong:
Should this analyzer not be added to the core? I still hadn't had time to
look closely at it...

--Jeremy

colezlaw:
I goofed and merged your most recent changes rather than a rebase. So now I'm trying to figure out how to get my tree back to the way it was before I did the merge, then rebase, then make a new pull request.

PR #42, opened 2014-01-13
NexusAnalyzer

colezlaw:
This adds a new Analyzer which will search Maven Central for the SHA-1 of the dependency. If it is found, the groupId, artifactId, and version are added as Vendor, Product, and Version evidence, respectively. Furthermore, it adds a new Identifier with the Maven coordinates for the jar identified.

You can disable this analyzer by setting `analyzer.nexus.enabled` to `false`. It is set to `false` in `src/main/resources` in order to keep the results of existing tests consistent. The Nexus repository can be changed using the property `analyzer.nexus.url`. By default, it's set to Maven central, but can be changed, for example, if you have an internal Nexus repository which mirrors central.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,3 +12,5 @@ nb-configuration.xml
 /target/
 #maven-shade-plugin generated pom
 dependency-reduced-pom.xml
+# vim files
+*.swp
--- a/dependency-check-core/src/main/java/org/owasp/dependencycheck/analyzer/NexusAnalyzer.java
+++ b/dependency-check-core/src/main/java/org/owasp/dependencycheck/analyzer/NexusAnalyzer.java
@@ -1,0 +1,182 @@
+/*
+ * This file is part of dependency-check-core.
+ *
+ * Dependency-check-core is free software: you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License as published by the Free
+ * Software Foundation, either version 3 of the License, or (at your option) any
+ * later version.
+ *
+ * Dependency-check-core is distributed in the hope that it will be useful, but
+ * WITHOUT ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+ * FITNESS FOR A PARTICULAR PURPOSE. See the GNU General Public License for more
+ * details.
+ *
+ * You should have received a copy of the GNU General Public License along with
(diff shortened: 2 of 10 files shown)


stevespringett:
Jeremy,

Will this be part of the core?  If so, what will be the default behavior of this analyzer? Will it be enabled by default (or should it be)? Or should this be an optional thing a user would turn on?

If it's going to be part of the core, I'll add support for it in the Jenkins plugin. There needs to be a user interface to configure the Nexus URL and to enable/disable it. I'll make it a global setting so all slave nodes can inherent them. But I'll need to know what the default behavior is going to be.

jeremylong:
colezlaw,

I am merging this into the core library now. However, I do need to ask a license question - I've been planning on changing the license from GPL3 to Apache 2.0. Do you have any issue with that (or would you prefer a different more permissive license)?

--Jeremy

jeremylong:
Steve,

I think this analyzer has another very interesting use case. When this analyzer was first posted as a PR I wasn't sure of its use as any JAR file in Maven central will contain a POM that has the GAV - so why do we need to reach out to Maven Central (or other Nexus server). Then I realized we can add a check for whether or not this is an "official" published version. Checking to ensure the library being used is the one blessed by your organization (as it is published in your companies Nexus server) would be very useful to many organizations.

This idea may cause me to create an Artifactory and Archiva analyzers (if similar functionality exists).

Also - not being a Nexus guru can other things be loaded into Nexus (such as DLLs, eggs, gems, etc.) that don't contain a POM? I kinda thought everything in Nexus had to have a POM with GAV data...

--Jeremy

colezlaw:
I have no problem with the license change. I wrote it for it to be used.

Nexus can be used to hold whatever type of artifact. In fact, Nexus sells under a separate license a bit of code which gives your internal Nexus repo Nuget capabilities. If you search on Maven Central for g:"org.openengsb.wrapped.csharp", there are a handful of more common .NET archives there.

Another benefit to the plugin is for internal repositories. Our internal repository allows devs to deploy non-Maven jars (such as vendor-provided libraries, etc.) so those jars in the wild wouldn't have POM's, nor does Nexus add it to the Jar when it's deployed. It will generate a POM that goes beside the Jar, but not one within the Jar itself. So this could be used to identify (with a little more accuracy) internal or commercial libraries.

PR #55, opened 2014-02-01
Grokassembly work

colezlaw:
This is the initial version of the work to determine version information from a .NET assembly. It includes an executable, GrokAssembly.exe, which takes a single argument - the path to a .NET assembly (exe or dll) and it will spit out some information in XML format which is then parsed.

There is a lot of black magic and witchcraft to extract the EXE, and I'm quite certain it won't work under all environments.

For *nix-like systems, this should work if mono is installed. If it's on the PATH, it should just work. If it's not on the PATH, a new setting, `analyzer.assembly.mono.path` can be used to specify the full path to `mono`


jeremylong:
I've merged this, but had to update this a little to get it to work on my system. The two changes where explicitly closing the FileOutputStream when writing the assembly; if we don't close it on my system it is "in use" and cannot be executed. Additionally, I made the test for the invalid mono path conditional so that it doesn't run on Windows.

Thanks!

colezlaw:
Holy cow - how did I miss closing the stream? I'm not sure how, but
somewhere along the line I did a push and it wouldn't accept it so had to
pull which generated a bunch of conflicts. Rather than trying to re-learn
some git-goofiness, I decided to deal with the conflicts, but I guess I
missed one - I'm usually very particular about how I deal with streams.

And it looks like the check for mono is already guarded - but not by an OS
check but a file separator check.

colezlaw:
Oh! I see what happened! I put the fos.close() in a finally block (which
was correct), but I actually attempt to call GrokAssembly within the same
block. The guard during the initializer was new last night, it didn't used
to be there. If it could _write_ GrokAssembly, it was assumed it would be
possible to execute it.

I'll pull your stuff and then refactor that bit.

And I see what you mean by test - actual unit test, not test at the
beginning of initialize.

Thanks for catching those things.

w

PR #63, opened 2014-02-11
Runtime and Provided dependencies should be excluded from analysis - Add Zippable Extension configurable

hgomez:
No need to scan Runtime and Provided dependencies, as they won't be bundled

### DIFF
--- a/dependency-check-ant/src/main/java/org/owasp/dependencycheck/taskdefs/DependencyCheckTask.java
+++ b/dependency-check-ant/src/main/java/org/owasp/dependencycheck/taskdefs/DependencyCheckTask.java
@@ -20,6 +20,8 @@ package org.owasp.dependencycheck.taskdefs;
 import java.io.File;
 import java.io.IOException;
 import java.io.InputStream;
+import java.util.Arrays;
+import java.util.HashSet;
 import java.util.List;
 import java.util.logging.Level;
 import java.util.logging.Logger;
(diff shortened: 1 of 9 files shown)


hgomez:
Did you had a chance to review this pull-request ?

For Maven part, scaning Provided and Runtime, bring false positives and could be counter productive.

For extra extensions, it's mandatory in our organisation and probably in many others companies.

jeremylong:
I haven't merged the PR yet. I am swamped until this weekend and will look
at this then.

My concern about not scanning provided and runtime is that this could hide
an issue in the environment - much like why scanning db drivers can
highlight an old vulnerable db server may be deployed. But I understand why
a developer would want to suppress these types of issues (the provided and
runtime). This tool has two users; devs and security auditor types each
with there own needs. As such, this feature needs to be configurable. Also
we need to skip test scope (if we aren't already). Another thing that I
need to consider are the optional dependencies.

The extensions will be added this weekend too.

I truly appreciate your feedback and contributions; thanks!

Jeremy

hgomez:
Provided and Runtime are usually used at component build time, for example a webapp.

When these components are bundled in final application, provided & runtime artifacts are furnished by runtime engine, ie Tomcat or JBoss.

And at that time, you should have a second Dependency Check Analysis not only for components but for whole delivery, components and runtime engine.

For much flexibility and control, it would be great to parametize scopes excluded from analysis, it's simple to do. I'll take a look at code after merge and start to works from this.

Cheers

PR #71, opened 2014-02-14
Add Data Mirroring in Maven Plugin

hgomez:
Maven Plugin need a way to provides Data Mirroring to Dependency Check core


jeremylong:
Is the "useDataMirroring" necessary? I would think if the URLs are specified that would indicate we should use the data mirror.  Any objection to me removing the extra flag or do you see a reason to keep it?

hgomez:
Good idea. Please do.

jeremylong:
This was merged by applying a patch and updating the PR prior to committing.

jeremylong:
This has been pushed to 1.1.2-SNAPSHOT

PR #72, opened 2014-02-15
Ignore the test if we can't run GrokAssembly.exe

colezlaw:
This disables all of the GrokAssembly sort of tests if it's not able to set it up properly for some reason.


jeremylong:
This has been merged.

PR #76, opened 2014-02-27
Fix for Issue 75

colezlaw:
This does a couple of things in relation to Issue 75 - both were ones we wanted to do:
- `NexusAnalyzer` will now use the proxy if it's configured _and_ `analyzer.nexus.proxy` is `true` (it's configured to `true` in the default `dependencycheck.properties`)
- A preflight request is added which will check the status of the Nexus repository. If this preflight request fails, the Nexus analyzer will be disabled

This solves an open challenge we had where the user might configure a proxy but might also be using an internal Nexus repository where they might not want to use the proxy. The out-of-the-box configuration is the most appropriate one, and others work somewhat elegantly:
- I'm not behind a proxy, and I want to use Central - a proxy won't be used (because the proxy settings are not turned on, even though `analyzer.nexus.proxy` is `true`)
- I configure a different Nexus URL, but don't set a proxy - as above
- I set a proxy but don't have an internal proxy - because the proxy settings are enabled and the default for `analyzer.nexus.proxy` is `true`, we'll use the proxy and use it to connect to Central
- I set a proxy, but have an internal Nexus repository. If I go through the trouble of configuring that URL in the properties, the appropriate property to disable the proxy for Nexus requests is right there.

### DIFF
--- a/dependency-check-core/src/main/java/org/owasp/dependencycheck/analyzer/NexusAnalyzer.java
+++ b/dependency-check-core/src/main/java/org/owasp/dependencycheck/analyzer/NexusAnalyzer.java
@@ -93,6 +93,10 @@ public class NexusAnalyzer extends AbstractAnalyzer {
             LOGGER.fine(String.format("Nexus Analyzer URL: %s", searchUrl));
             try {
                 searcher = new NexusSearch(new URL(searchUrl));
+                if (! searcher.preflightRequest()) {
+                    LOGGER.warning("There was an issue getting Nexus status. Disabling analyzer.");
+                    enabled = false;
+                }
             } catch (MalformedURLException mue) {
                 // I know that initialize can throw an exception, but we'll
                 // just disable the analyzer if the URL isn't valid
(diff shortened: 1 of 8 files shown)


stevespringett:
The Jenkins plugin now incorporates this functionality but works slightly different. By default, the NexusAnalyzer will use the proxy defined in the Jenkins global configuration. This is the same behavior as the rest of Dependency-Check. This is done by ensuring analyzer.nexus.proxy is defaulted to true.

In the Jenkins global configuration, there is an option to "Bypass proxy for Nexus analyzer". When this is checked, the NexusAnalyzer will not use the Jenkins proxy configuration and analyzer.nexus.proxy will be set to false.

colezlaw:
This is actually consistent with the rest of core. The default value for
analyzer.nexus.proxy is true, so if you specify a proxy and nothing else,
it'll use the proxy. The only time the proxy gets skipped is if you
explicitly set analyzer.nexus.proxy to false.

PR #110, opened 2014-04-16
Try to fix accidental disabling of analyzers during large multi-module maven projects

cschneider4711:
Hi Jeremy,

after some debugging I found out why for some multi-module maven projects the list of dependencies scanned is zero. It appears that if after at least one module didn't result in any files to analyze for a given analyzer, it is completely disabled (without being re-enabled), so that subsequent modules during the same build will have no files (via supportsExtension) checked.

Please check if the change is the correct way to solve the problem that multi-module maven builds are sometimes resulting in zero dependencies being identified. 

Thanks & best regards,
Christian
@cschneider4711


bkimminich:
:+1: ... I gave up trying to find out why this happens couple days ago. Seems I debugged in the completely wrong place.

cschneider4711:
The release 1.2.0 fixes the aforementioned bug now ;)

jeremylong:
Thanks for validating thst this issue is closed in your environment!

Jeremy

PR #113, opened 2014-04-22
all loggers extracted to constants

bkimminich:
I hope I didn't miss any... :-)

### DIFF
--- a/dependency-check-ant/src/main/java/org/owasp/dependencycheck/taskdefs/DependencyCheckTask.java
+++ b/dependency-check-ant/src/main/java/org/owasp/dependencycheck/taskdefs/DependencyCheckTask.java
@@ -62,6 +62,7 @@ public class DependencyCheckTask extends Task {
      * System specific new line character.
      */
     private static final String NEW_LINE = System.getProperty("line.separator", "\n").intern();
+    public static final Logger LOGGER = Logger.getLogger(DependencyCheckTask.class.getName());
 
     /**
      * Construct a new DependencyCheckTask.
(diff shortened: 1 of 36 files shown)


colezlaw:
@bkimminich I submitted a patch a few days ago with a PMD rule to identify all of them and then went through and corrected all that the PMD rule found. Might want to do a fetch and rebase before submitting the patch to make sure we don't end up with conflicts.

bkimminich:
Sorry, I fetched but obviously only from my fork... 8-) ... the downside of late-night coding...
Am 23.04.2014 01:41 schrieb colezlaw notifications@github.com:@bkimminich I submitted a patch a few days ago with a PMD rule to identify all of them and then went through and corrected all that the PMD rule found. Might want to do a fetch and rebase before submitting the patch to make sure we don't end up with conflicts.

colezlaw:
I did notice you caught the ones in `dependency-check-ant` and `dependency-check-cli`. I'm working on trying to move the PMD rule to the top-level and not have to reproduce it in all the modules - if that works I can correct those as well and put them in a single PR.

bkimminich:
Sound good. I fixed my remote upstream issue and am now "back on track" ... will close this PR then. Thx!

jeremylong:
Regarding moving the PMD config to the top level - see what I did with
Checkstyle.

--Jeremy

colezlaw:
Yeah - with checkstyle, you duplicated the configuration in all the child projects (including the external files). I moved the PMD rules up to DependencyCheck, then all the modules reference `../dc-rules/dcrules.xml`. This isn't **ideal** if folks try to break the project out from under the parent directory, but I don't think it builds properly that way anyhow.

Also, unfortunately, the reporting plugin doesn't cascade its configuration to the modules. This is very sad since the reporting plugin is so complicated to get set up properly.

I'll send a PR shortly. If you don't like referencing `..`, then just reject the PR and I'll duplicate the rules throughout the projects and resubmit.

bkimminich:
How about adding a separate Maven module `dependencycheck-codemetrics` that only contains the config/filter files for FindBugs, CheckStyle, PMD etc. inside its `src/main/resources` folder? This module would be added as a dependency to all relevant modules and then these can just refer to the files by their names again.

PR #115, opened 2014-04-25
Enable loading of suppression file from classpath resource

bkimminich:
DependencyCheck supported only suppression files loaded via URL or a file path. Similar analysis plugins (e.g. FindBugs, CheckStyle) are also able to locate filter/suppression files in the classpath. This makes it possible to put these files into a separate Maven artifact and reuse them easily among modules or projects.

I have added this feature to the AbstractSuppressionAnalyzer in a non-invasive way by attempting to load suppressions via classpath only if the previous attempts fail. I built it similar to the URL download way via a temporary file. It allows to keep the API and responsibility or the SuppressionParser as it is. All the validation if the suppressions can be loaded at all remains in the AbstractSuppressionAnalyzer.


bkimminich:
I have reduced the level of redundancy in logging code a bit along the way. Help readability of the class quite a bit imho.

PR #116, opened 2014-04-29
Added test for suppression file via classpath feature

bkimminich:
(+) added test case for loading suppression file via classpath
(+) added test case for failing to locate suppression file entirely
(+) initialize object under test as field in @Before instead of using local variables
(-) redundant test for initialization
(-) setting up suppression via URL in @Before for all test methods, even those where it is not needed

### DIFF
--- a/dependency-check-core/src/test/java/org/owasp/dependencycheck/analyzer/AbstractSuppressionAnalyzerTest.java
+++ b/dependency-check-core/src/test/java/org/owasp/dependencycheck/analyzer/AbstractSuppressionAnalyzerTest.java
@@ -18,9 +18,7 @@
 package org.owasp.dependencycheck.analyzer;
 
 import org.junit.Before;
-import org.junit.Rule;
 import org.junit.Test;
-import org.junit.rules.ExpectedException;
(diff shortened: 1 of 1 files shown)


dependency-check-core/src/test/java/org/owasp/dependencycheck/analyzer/AbstractSuppressionAnalyzerTest.java
```diff
@@ -57,32 +54,51 @@ public void setUp() throws Exception {
...
+        setSupressionFileFromURL();
         instance.initialize();
+        int expCount = 5;
+        List<SuppressionRule> result = instance.getRules();
+        assertEquals(expCount, result.size());
     }
 
     /**
-     * Test of getRules method, of class AbstractSuppressionAnalyzer.
+     * Test of getRules method, of class AbstractSuppressionAnalyzer for suppression file declared as URL.
```
bkimminich, line 74: Copy & paste typo in Javadoc. Should be "...from classpath" instead of "...declared as URL" of course.

PR #121, opened 2014-05-02
Workaround for classloader issues with service loading

c-a-m:
Changes the classloader parameter to the Engine from the current thread to the task classloader.

Another options might be getting the system classloader: http://docs.oracle.com/javase/6/docs/api/java/lang/ClassLoader.html#getSystemClassLoader%28%29

I'll attach the stacktrace I'm seeing to this pull request.


jeremylong:
Part of this PR was applied manually. I am not planning on importing the examples (unless this might be useful in the site documentation?). If you think the examples should be in the site documentation - please move them to the site directory.

Thanks!

c-a-m:
Great. I just wanted the example for you in case it would be help reproduce
the problem.  Thanks for this project.  I just got it running in our team
city ci.

PR #131, opened 2014-06-24
Updated to 1.1 of GrokAssembly.exe to deal with exceptions

colezlaw:
GrokAssembly had a couple of other exception types which might be thrown, which would cause Dr. Watson/debug windows to pop if they happened. This catches two more types of exceptions:
- `TypeLoadException` when attempting to get the company name. If a parent assembly is not available, this can be thrown when attempting to get the custom attributes. By catching this, we can still get the product and version even when the company can't be determined.
- `Exception` at the very end - if some other exception might occur which we're not catching. The results won't be valid, but at least when running in an automated environment, it won't just completely lock up Dependency Check.


jeremylong:
The PR has been merged, a new release will be pushed soon.

PR #133, opened 2014-06-26
Fixed suppression analyzer to load from input stream fixing failure

colezlaw:
This fixes the broken Suppression Analyzer by doing the following:
- Allowing suppression files to be loaded from an `InputStream` in addition to a `File`
- Opening the base suppression list as an `InputStream` and passing that to get parsed


jeremylong:
The PR has been merged, a new release will be pushed soon.

PR #161, opened 2014-10-12
Uses ExpectedException to test for exceptions

hansjoachim:
Hi, thanks for this great project! :)

I was browsing the code and noticed parts of this test suite could be simplified by using ExpectedException. (See [Junit's notes on exception testing](https://github.com/junit-team/junit/wiki/Exception-testing) for details.)

I also took the liberty to trim out some empty setup methods.


jeremylong:
Thanks!

Issue #162, opened 2014-10-15
Spring Security False Positives


Quantas:
Currently our projects depend on org.springframework.security:spring-security-core:3.2.5.  We are getting vulnerability errors for spring security 3.2.5 when there are currently no known vulnerabilities.  I believe the plugin is confusing spring security for spring framework 3.2.5 which does indeed contain vulnerabilities.  Here is the output from the plugin:

spring-security-core-3.2.5.RELEASE.jar (cpe:/a:springsource:spring_framework:3.2.5, cpe:/a:vmware:springsource_spring_framework:3.2.5, cpe:/a:vmware:springsource_spring_security:3.2.5) : CVE-2014-1904, CVE-2014-0054

jeremylong:
A patch has been put in to address this problem. Until the next release you could use the suppression file below:

``` xml
<?xml version="1.0" encoding="UTF-8"?>
<suppressions xmlns="https://www.owasp.org/index.php/OWASP_Dependency_Check_Suppression">
    <suppress>
        <notes><![CDATA[
        This suppresses false positives identified on spring security.
        ]]></notes>
        <filePath regex="true">.*spring-security-[^\\/]*\.jar$</filePath>
        <cpe>cpe:/a:mod_security:mod_security</cpe>
        <cpe>cpe:/a:springsource:spring_framework</cpe>
        <cpe>cpe:/a:vmware:springsource_spring_framework</cpe>
    </suppress>
</suppressions>
```

PR #163, opened 2014-10-22
Inclusion of evidence from the project-pom, configration to exclude internal groupIds

ctrl-alt-dev:
Hi Jeremy,

First of all: thank you for Dependency Check!

I've made two new features of which I hope you'll consider to include in the project. Both are for the maven plugin.
- I noticed that only the (jar) files of the maven dependencies are are used but not the group, artifact and version fields which are present as well. This is mostly the same data retrieved when querying nexus. Using this data instead of nexus speeds up the analysis process considerably. So I added an additional scan method in the Engine that also takes a MavenArtifact and evaluates it like the nexus evidence. I moved the evaluation code from the NexusAnalyzer to Dependency to share it between Engine and NexusAnalyzer.
- The (internal) project I am working on uses a lot of poorly named artifacts which trigger false positives because of name clashes. Since they all start with a common groupId I've added the feature to exclude those from analysis using the excludeInternalGroupIds configuration parameter, like this:

``` xml
    <excludeInternalGroupIds>
        <groupId>some.internal.group.id</groupId>
    </excludeInternalGroupIds>
```

I apologize for the code reformatting in my commits. I found out when it was too late :-)

Best regards,

Erik Hooijmeijer

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,6 +7,9 @@
 # Eclipse project files
 .classpath
 .project
+.settings
+maven-eclipse.xml
+.externalToolBuilders
 # Netbeans configuration
 nb-configuration.xml
 /target/
--- a/dependency-check-core/src/main/java/org/owasp/dependencycheck/Engine.java
+++ b/dependency-check-core/src/main/java/org/owasp/dependencycheck/Engine.java
@@ -27,6 +27,7 @@ import java.util.List;
 import java.util.Set;
 import java.util.logging.Level;
 import java.util.logging.Logger;
+
 import org.owasp.dependencycheck.analyzer.AnalysisPhase;
 import org.owasp.dependencycheck.analyzer.Analyzer;
 import org.owasp.dependencycheck.analyzer.AnalyzerService;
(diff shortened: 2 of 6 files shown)


jeremylong:
Erik,

Thank you! I actually realized what you were talking about, regarding not using the POM information, last week. I had previously been confused by the statement because if the JAR file contains a POM in the META-INF/maven directory it is included as evidence in the scan. While fixing issue #162 I realized my mistake in that the actual POM might be available - but it may not be in the JAR file being scanned. I had put in a quick fix for this, but your approach is likely more reliable. 

That being said, I'm trying hard not to expand the number of dependencies the tool uses. As such, I will likely accept the PR, but then change a bunch of your code so that instead of passing in the MavenArtifact (and as such, making Maven a dependency of the CLI, Ant Task, etc.) I will see if I can come up with a different way of passing in a path to the POM or possibly passing in a Set of Evidence instead of the MavenArtifiact.

--Jeremy

jeremylong:
Also, regarding the exclude internal groupid - once the external POM issue is resolved the same filter could be applied using the standard suppression file mechanism:

``` xml
<suppress base="true">
     <gav regex="true">internal\.groupid:.*</gav>
     <cpe regex="true">.*</cpe>
</suppress>
```

Do you feel the configuration option 'excludeInternalGroupIds' is needed? Or could you just create a suppression file to handle this?

ctrl-alt-dev:
Hi Jeremy,

Thanks for your quick response.

I reused the MavenArtifact from dependency-check-core (part of the Nexus Analyzer) so there is no extra dependency on Maven (which would be bad indeed).

Regarding excluding the internal group id config parameter, we use a master pom that holds the default config for the plugins we use and it would be convenient to have this parameter setup in advance. On the other hand, most projects would require a suppression file, so it would be little effort to add this each time.  

--Erik

PR #178, opened 2014-12-31
Upgrade to junit 4.12

hansjoachim:
- Upgrade to junit 4.12 which was released recently.
- The submodules will inherit this dependy from the parent pom, so removed it from two of the submodules. This way, we'll have the definition in one place and only need to update the version number once.
- Dropped type because "jar" is the default.


jeremylong:
thanks!

PR #179, opened 2015-01-01
Upgraded some more dependencies

hansjoachim:
Continuing on, I've upgraded most of the dependencies to use the latest version. See commit log for details.

Some comments:
- I left the majority in -core alone, since it looked like most of them were included to verify those versions are indeeded marked as vulnerable.
- `mvn clean install` still runs sucessfully, though since I don't have mono installed, the .NET related tests have been ignored (Cool, haven't seen Assume used in practice before). Someone might want to doublecheck those are still working.
- When upgrading h2 to the 1.4.x-series I saw some failing tests because they couldn't connect to the database. From their homepage it looks like this is still in beta (http://www.h2database.com/html/changelog.html) so I went with the latest 1.3.x instead. I didn't look more into it, but should probably do so if a stable release of the 1.4.x-series show up.
- Apache Lucene also has some newer versions, though switching to these caused compilation errors. I might look more into why, but thought I'd deal with the upgrades which didn't require code changes first.

### DIFF
--- a/dependency-check-ant/pom.xml
+++ b/dependency-check-ant/pom.xml
@@ -456,12 +456,12 @@ Copyright (c) 2013 - Jeremy Long. All Rights Reserved.
         <dependency>
             <groupId>org.apache.ant</groupId>
             <artifactId>ant</artifactId>
-            <version>1.9.3</version>
+            <version>1.9.4</version>
         </dependency>
         <dependency>
             <groupId>org.apache.ant</groupId>
             <artifactId>ant-testutil</artifactId>
-            <version>1.9.3</version>
+            <version>1.9.4</version>
             <scope>test</scope>
         </dependency>
     </dependencies>
--- a/dependency-check-ant/src/test/java/org/owasp/dependencycheck/taskdefs/DependencyCheckTaskTest.java
+++ b/dependency-check-ant/src/test/java/org/owasp/dependencycheck/taskdefs/DependencyCheckTaskTest.java
@@ -30,6 +30,9 @@ import org.owasp.dependencycheck.utils.Settings;
  * @author Jeremy Long <jeremy.long@owasp.org>
  */
 public class DependencyCheckTaskTest extends BuildFileTest {
+    //TODO: The use of deprecated class BuildFileTestcan possibly
+    //be replaced with BuildFileRule. However, it doesn't seem to be included
+    //in the ant-testutil jar.
 
     @Before
     @Override
--- a/dependency-check-core/pom.xml
+++ b/dependency-check-core/pom.xml
@@ -728,6 +728,6 @@ Copyright (c) 2012 Jeremy Long. All Rights Reserved.
         </profile>
     </profiles>
     <properties>
-        <apache.lucene.version>4.5.1</apache.lucene.version>
+        <apache.lucene.version>4.6.1</apache.lucene.version>
     </properties>
 </project>
--- a/dependency-check-core/pom.xml
+++ b/dependency-check-core/pom.xml
@@ -728,6 +728,6 @@ Copyright (c) 2012 Jeremy Long. All Rights Reserved.
         </profile>
     </profiles>
     <properties>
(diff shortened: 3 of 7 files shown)


jeremylong:
Thanks again! As an FYI, the dependencies in core that are marked as both
'test' and 'optional' are the ones that are used by other test cases to
validate that the analysis is working correctly.

Last night I think I figured out why another developer was having an
occasional build error with the test cases (getting a null pointer
exception in the hint analyzer). I think a generic method needs to get
added to the BaseTest class for getting file references to optional test
dependencies and in that method we need to use something like junits assume
(I think that is the right api) to ensure that the file exists. That way if
the file we are assuming to exist actually doesn't, the test will work (I
might have the assume API wrong, but there is a way to end the test if a
condition exists). That might be a good enhancement request I should put in
;)

--Jeremy

hansjoachim:
> the dependencies in core that are marked as both
> 'test' and 'optional' are the ones that are used by other test cases to
> validate that the analysis is working correctly.

Ok, that seems to also correspond with the "only used for testing"-comment in pom.xml, not surprisingly. I believe this has upgraded the ones we wish to change.

I've moved a bit forward with Apache Lucene. The next step would be 4.8.x, but among the changes is a new parameter (Version) for the WordDelimiterFilter constructor (see http://svn.apache.org/viewvc?view=revision&revision=1578993 for diff). This cause compilation to fail for some classes, for instance SearchFieldAnalyzer. 

I am not quite sure what we wish to send in for this new parameter. Version is an enum so we could send in the version of Lucene we happen to be using, but it seems like we would then need to manually change all instances once we upgrade to a newer one. And LUCENE_CURRENT is deprecated so that would light up among warnings. We might be able to send in null, but that has its own set of potential problems.

I briefly looked at whether it would be easier to use the WordDelimiterFilterFactory but that would include changing the flags from OR-ing the individual ints to populating a map with <String, String> with the same information. That just seemed harder to read/set up than the current situation, so probably not. 

 There was introduced a Lucene47WordDelimiterFilter with the old behaviour, but since this was created as a deprecated class, I see it as a no-go.

Any preferences/suggestions?

jeremylong:
Looking at the JavaDoc I think you can use [LATEST](https://lucene.apache.org/core/4_10_0/core/org/apache/lucene/util/Version.html#LATEST)
instead of LUCENE_CURRENT.

--Jeremy

jeremylong:
Also, wouldn't 4.10.3 be the latest version of lucene?

<dependency>
    <groupId>org.apache.lucene</groupId>
    <artifactId>lucene-core</artifactId>
    <version>4.10.3</version>
</dependency>

hansjoachim:
> I think you can use LATEST

Ah, that would probably work, though that wasn't part of the enum in 4.8.x yet.

> Also, wouldn't 4.10.3 be the latest version of lucene?

Yes, it would. Though when I tried upgrading to this I ran into compilation errors, so I tried to roll back and deal with each version incrementally so I would only need to solve one problem at a time.

Turns out though, that in 4.10.x (maybe 4.9.x too) WordDelimiterFilter regained a Version-less constructor, thus avoiding this whole problem. So turns out the Version-problem is a non-issue. There 
still was a tiny compilation error elsewhere, but easily dealt with.

With Apache Lucene upgraded to the latest version I believe this is ready for review/merging. The version changes seem to have marked more things as deprecated, and I can look into some of them, though I don't know whether it is preferable to deal with the separately or as part of this request.

jeremylong:
Lucene has been rolled back to 1.7.2 as newer versions require Java 7. There are issues with the Jenkins plugin if dependencies require 1.7 or newer.

hansjoachim:
Yes, I saw that in `git log`. Sorry about that, I had not considered/checked this.

I have an idea though, since I've noticed DependencyCheck has a CI system running at cloudbees. I don't know how much control you have over this setup, but would it perhaps be possible to add a separate job building with JDK6? That would presumably have caught this a lot earlier.

(And possible also a job building with JDK8 to see if it has the same problems as I ran into a while back)

PR #182, opened 2015-01-03
Upgrade plugins to latest version

hansjoachim:
This is still work in progress, though should be ready to start talking about.

Upgrades plugins used to latest version available, see commit messages for details. At the same time I've taken the opportunity to move the plugin names and version numbers into pluginManagement placed in the parent pom. This makes it possible to update the version numbers in a single place.

Some questions/comments:
1. I had some surprises with moving plugins into pluginManagement. Even after doing so, -core would still report that the plugins used the older version. It turned out that these were in fact also declared inside a profile where it was set to use the older version number. I don't know much about this case, but I tried experimenting and it seemed like the profile did _not_ pick up entries from the parent pom's pluginManagement. For the mean time, I've upgraded these version numbers manually, but if possible I would like to define them in a single place.
2. As you might see from the diff, I have tried to move common configurations out to the parent pom to avoid duplication/repetition. 
3. Unfortunately, I have not been able to verify that all plugins still perform as expected after the version upgrade. This is for the most part simply because I am not familiar with all of them, and some are quite hard to verify (at least without -Ddryrun) such as maven-release-plugin. 
4. This is also why I've left maven-site/github-site-plugins alone. I'm not really familiar with the setup nor how to verify this. I skimmed the blogpost linked to, and I will look closer into understanding it in the future. I have looked some at it though, and hope it will be possible to at least put part of the common configuration in a single place. 
5. Required maven version. 
5 A) I've currently required 2.2.1 in the parent pom (and effectively all submodules not overriding this). Then I discovered that the -ant module requires maven 3.0 in order to use the latest version of the maven-shade-plugin, so I changed that module. I am somewhat torn on specifying the highest required version number in the parent pom versus only where it is strictly needed. Any preferences on this.
5 B) I also noticed that although the -maven module requires maven 3.0, it only requires 2.2.1 at least based on plugin usage. I am not sure whether there was a reason for specifying this, so I have left it alone. 
6. Is the maven-gpg-plugin still used? I assume it has been commented out so that regular developers won't need to sign every local build they do, but it may be used for release builds. If it is still used, I'd like to declare the version number under pluginManagement so that we can keep it up to date, while keeping the actual usage out-commented. If it isn't used, I can remove it.

### DIFF
--- a/dependency-check-ant/pom.xml
+++ b/dependency-check-ant/pom.xml
@@ -37,6 +37,9 @@ Copyright (c) 2013 - Jeremy Long. All Rights Reserved.
         </site>
     </distributionManagement>
     <!-- end copy -->
+    <prerequisites>
+        <maven>3.0</maven>
+    </prerequisites>
     <build>
         <resources>
             <resource>
@@ -193,7 +196,7 @@ Copyright (c) 2013 - Jeremy Long. All Rights Reserved.
             <plugin>
                 <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-shade-plugin</artifactId>
-                <version>2.1</version>
+                <version>2.3</version>
                 <configuration>
                     <transformers>
                         <transformer implementation="org.apache.maven.plugins.shade.resource.ServicesResourceTransformer" />
--- a/dependency-check-ant/pom.xml
+++ b/dependency-check-ant/pom.xml
@@ -268,7 +268,6 @@ Copyright (c) 2013 - Jeremy Long. All Rights Reserved.
             <plugin>
                 <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-surefire-plugin</artifactId>
-                <version>2.16</version>
                 <configuration>
                     <systemProperties>
                         <property>
--- a/dependency-check-ant/pom.xml
+++ b/dependency-check-ant/pom.xml
@@ -220,17 +220,6 @@ Copyright (c) 2013 - Jeremy Long. All Rights Reserved.
             <plugin>
                 <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-jar-plugin</artifactId>
-                <version>2.4</version>
-                <configuration>
-                    <archive>
-                        <manifest>
-                            <addDefaultImplementationEntries>true</addDefaultImplementationEntries>
-                        </manifest>
-                    </archive>
-                    <excludes>
-                        <exclude>**/checkstyle*</exclude>
-                    </excludes>
-                </configuration>
             </plugin>
             <plugin>
                 <groupId>org.codehaus.mojo</groupId>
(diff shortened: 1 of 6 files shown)


jeremylong:
Thank you for all the work in cleaning up some of the code base and configuration! I'll try to answer your questions or explain what was done:

1) I have no idea why the profile doesn't use the plugin management section. What happens if you remove the version from those entries? Wouldn't it use the plugin management in that case?

To explain the two profiles:
mysql - there is an option to configure dependency-check to use a more robust database that is shared by all instances within an organization. This profile is used to test that this works using mysql.
allTests - this just adds an additional collection of test dependencies that I use occasionally to manually inspect some conditions.

2-3) Thanks for moving everything to the parent pom. I can find some time over the next week to verify everything is still working as expected.

4) Regarding the maven-site/github-site-plugins - I was having a ton of problems getting the multi-module site to build correctly. I was never able to get an automated publish working. However, I fixed a couple of problems I discovered a month or two ago and never went back to revisit the site configuration. If you feel like removing/replacing this I have no concerns as long as a `mvn site site:stage` ends up creating the site more or less like it is today.

5) In all honesty, this was my first maven plugin. Some of the decisions made early were probably wrong and should be fixed. I originally choose 3.0 because if you are still using 2.x I think you should upgrade because it is [EOL](http://maven.apache.org/maven-2.x-eol.html). I also was unsure of some of the backward compatibility with the maven plugin apis as I only referenced the 3.x documentation.

6) I use the gpg plugin when I build the release version so the plugin version should go to the plugin management section.

hansjoachim:
4) maven-site:

> If you feel like removing/replacing this I have no concerns 

Ok, just wondered. I probably won't do much changes to this setup, at least not until I understand it better. Thanks for providing the command you use to generate the site. I took a look locally, and `mvn site` currently generates the content so I have a rough idea what should be there. Though when I added `site:stage` I got an error message saying "A required class was missing (...) org/apache/commons/lang/StringUtils". Suspecting this might be due to some transient dependency or something affected by my changes, I also checked with v1.2.8 which had the same result. Not sure what's going on there, have you seen anything like this. I tried searching online and found some similar cases. Some suggested it might be some incompatibility between the plugin and version of maven, so I'll investigate that some. 

Oh, and just to double-check, I assume the links to other modules shouldn't work in the generated web pages. For instance the target directory in the root directory has an index.html with links to the different sub-modules, but these don't work because they are generated separately. 

5) Fair enough, just wondered if there was some other critera for setting the requirement. I've now switched to require 3.0 for all. The reasoning is that the -ant module will need it based on its plugins and if someone were to try building it with an older version of maven I think they would much rather be told up-front than after building half of the modules before running into problems.

6) Moved out the version number for pgp-plugin. Still left the commented-out plugin itself though. I see other projects, for instance JUnit (https://github.com/junit-team/junit/blob/master/pom.xml) has placed it in a profile. That way it won't interfere with regular builds, you don't have to remember uncommenting/commenting stuff when running a release, and it would presumably be just another option to specify the profile when invoking `mvn release:prepare/perform`. Might be something to look at.

PR #184, opened 2015-01-05
Replaced deprecated constructors which contained parameter Version

hansjoachim:
Cleaning up most of the constructors now marked deprecated after I upgraded the Apache Lucene version. 
When looking into the code, these ended up toggling behaviour if Lucene version was later than 3.1. 

Some deprecated Lucene-related constructors remain, but these were less straight-forward as they assigned the version number to fields for presumably later usage. The constructors are deprecated though, but I'll look a bit closer into what they are actually used for.

While working on this I found VersionAnalyzer, SearchVersionAnalyzer and VersionTokenizingFilter which seem to have been deprecated for some time, and are currently unused classes in DependencyCheck. Would it be ok to remove these or should they be kept around (for instance if deprecated classes are only removed with major version changes)? The latter case only occured to me after I had removed unused parameter from the constructors, but I can revert that part if you want. :)

### DIFF
--- a/dependency-check-core/src/main/java/org/owasp/dependencycheck/data/cpe/CpeMemoryIndex.java
+++ b/dependency-check-core/src/main/java/org/owasp/dependencycheck/data/cpe/CpeMemoryIndex.java
@@ -125,7 +125,7 @@ public final class CpeMemoryIndex {
                 }
                 indexSearcher = new IndexSearcher(indexReader);
                 searchingAnalyzer = createSearchingAnalyzer();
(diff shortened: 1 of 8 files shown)


jeremylong:
The version tokenizers etc that you found are no longer used.  Originally
we stored the entire CPE identifier in the Lucerne index; but this ended up
causing too many false positives. I'm fine with removing these classes as
there is a good chance the next release will be 1.3 (unless a bug comes up
that needs to get patched).

Jeremy

PR #197, opened 2015-02-17
Removes the test for default Maven repository directory.

ahi:
Error: If the M3_REPO directory is not set to a path that contains _/m2/repository/_ the DependencyCheck doesn't read <library>.pom file. Some dependencies like "spring-core-3.2.0.RELEASE.jar" are not found due to this error.

Fix: Remove the test for default Maven repository directory. The existing check for the existence of <library>.pom file is sufficient.


jeremylong:
I was debating this code when I put it in... Glad it was removed. Thanks for the PR!

PR #198, opened 2015-02-20
Modified NexusAnalyzer to download POM if required

colezlaw:
NexusAnalyzer previously would just get GAV for a
match, but the POM may be separate from the jar
and contain other valuable information. This
includes refactoring of the analyzePom into
PomUtils.


dependency-check-core/src/main/java/org/owasp/dependencycheck/analyzer/NexusAnalyzer.java
```diff
@@ -202,6 +212,38 @@ public void analyzeFileType(Dependency dependency, Engine engine) throws Analysi
...
+                    break;
+                }
+            }
+            if (!pomAnalyzed && ma.getPomUrl() != null) {
+                File pomFile = null;
+                try {
+                    final File baseDir = Settings.getTempDirectory();
+                    pomFile = File.createTempFile("pom", ".xml", baseDir);
+                    if (!pomFile.delete()) {
+                        final String msg = String.format("Unable to fetch pom.xml for %s from Central; "
```
jeremylong: Shouldn't this say from Nexus?

PR #199, opened 2015-02-21
Upgrade jmockit to latest version

hansjoachim:
`mvn clean install` is still passing.

### DIFF
--- a/dependency-check-core/pom.xml
+++ b/dependency-check-core/pom.xml
@@ -354,7 +354,6 @@ Copyright (c) 2012 Jeremy Long. All Rights Reserved.
         <dependency>
             <groupId>org.jmockit</groupId>
             <artifactId>jmockit</artifactId>
-            <version>1.14</version>
             <scope>test</scope>
         </dependency>
         <dependency>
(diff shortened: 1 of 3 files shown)


hansjoachim:
Now placed jmockit under dependencyMangament in the root pom so that the same versio number will be used across all modules. 

(I considered adding dependencyManagament when I upgraded various dependencies last time, but didn't find too much overlap between the modules. Though considering the use of jmockit in -core almost slipped by, it might be worth doing anyways)

jeremylong:
Thanks again.  I need to remember to put any dependencies I add into the dependency-management section...

hansjoachim:
You're welcome. I've thought a bit about this. -core contains various older dependencies which are in place in order to test/verify they are correctly marked as vulnerable. I think the long-term plan would be to leave these alone, but move everything else into dependencyManagement in the root POM. That way it should be possible to check for newer versions in a single place (the root POM), and the outdated dependencies still remaining in -core can be ignored. What do you think?

Also, I noticed lucene 5.0 was released here the other day. Don't know whether you want to push for the new major release in DependencyCheck straight away or wait a while. Anyways, I haven't really had time to look into what's new and if it made any backwards-incompatible changes affecting DC.

PR #200, opened 2015-02-22
Fixes SonarQube Issues

hansjoachim:
Fixes various SonarQube issues, see individual commits for details.

Two notes:
- I marked one TODO to be discussed during the code review, it looks like a map which is assigned values but don't end up being used. Could you double-check whether these should/are indeed used or if it would be safe to remove the map.
- I noted the byte[] arrays in the first commit were used in rather similar ways, but don't know whether it would make sense to pull this out in some common place like a utility method.


jeremylong:
Thanks!  I'll take a closer look at the TODO you added.  BTW - you should add yourself to the contributors list in dependency-check-parent.

hansjoachim:
Ok, I'll try to remember it for the next patch. ;)

PR #203, opened 2015-03-26
Add update task

Bowbaq:
Use case: I want to run this on CircleCI. Right now, running the `check` task downloads the CVE definitions on each run. I'd like to be able to cache those definitions between runs, but I need a task that only downloads the definitions to do that.

I don't really know anything about maven plugins, so this may be horribly wrong. If it is, let me know and I'll try to fix it!

### DIFF
--- a/dependency-check-core/src/main/java/org/owasp/dependencycheck/Engine.java
+++ b/dependency-check-core/src/main/java/org/owasp/dependencycheck/Engine.java
@@ -468,7 +468,7 @@ public class Engine {
     /**
      * Cycles through the cached web data sources and calls update on all of them.
      */
-    private void doUpdates() {
+    public void doUpdates() {
         LOGGER.info("Checking for updates");
         final UpdateService service = new UpdateService(serviceClassLoader);
         final Iterator<CachedWebDataSource> iterator = service.getDataSources();
@@ -533,7 +533,7 @@ public class Engine {
      * @throws NoDataException thrown if no data exists in the CPE Index
      * @throws DatabaseException thrown if there is an exception opening the database
      */
-    private void ensureDataExists() throws NoDataException, DatabaseException {
+    public void ensureDataExists() throws NoDataException, DatabaseException {
         final CveDB cve = new CveDB();
         try {
             cve.open();
--- a/dependency-check-maven/src/main/java/org/owasp/dependencycheck/maven/UpdateMojo.java
+++ b/dependency-check-maven/src/main/java/org/owasp/dependencycheck/maven/UpdateMojo.java
@@ -1,0 +1,126 @@
+/*
+ * This file is part of dependency-check-maven.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * Copyright (c) 2015 Maxime Bury. All Rights Reserved.
+ */
+package org.owasp.dependencycheck.maven;
+
+import java.util.Locale;
+import java.util.logging.Level;
+import java.util.logging.Logger;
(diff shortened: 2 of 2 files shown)


stevespringett:
For CI, the dataDirectory should be specified. This allows you to have multiple jobs (or a single repetitive job) that can access a common data directory used by Dependency-Check. The data directory contains the database and index that are created after a successful NVD update. The data directory should be located outside of your projects workspace so that it doesn't get wiped out every time. It will then only periodically update when necessary.

I'm not sure adding an update task is the best approach, especially since the combination of a dataDirectory and the autoUpdate property can be set.

Bowbaq:
The problem is not so much where the data ends up (we can tell CircleCI to cache arbitrary directories) but when it gets downloaded.

CircleCI has a dedicated step to download dependencies, test data and such. Caching happens at the end of this step (and the cache is persisted between runs). CircleCI then runs the tests, but nothing gets cached at this stage.

Right now the definitions get downloaded as part of running the tests,  and so the database doesn't get cached. I could run the analysis during the dependencies step, but that's inelegant. It seems preferable to have a dedicated task to update the definitions, and then have another one that's dedicated to running the checks.

Does that make sense?

stevespringett:
Have you tried specifying the Maven phase and goal you want Dependency-Check to run in?

```
<plugin>
    <groupId>org.owasp</groupId>
    <version>1.2.9</version>
    <configuration>
        <format>ALL</format>
        <autoUpdate>true</autoUpdate>
    </configuration>
    <executions>
        <execution>
            <phase>verify</phase>
            <goals>
                <goal>check</goal>
            </goals>
        </execution>
    </executions>
</plugin>
```

In this example, it's executing the plugin during the Maven verify phase which occurs after compile, test, and package. 

Note, disabling NVD updates can be controlled through setting the value of autoUpdate, which can also be a variable in the POM.

For more configuration options, refer to http://jeremylong.github.io/DependencyCheck/dependency-check-maven/configuration.html

Bowbaq:
I think you're saying I can skip the updates and run only the analysis. What I want is the opposite: download the updates and skip the analysis. Is that possible?

leRadwan:
A separate update task called independently of a scan would be very useful for integration scenarios like CI. That enables scenarios of setting a schedule for NVD updates independent of when scans are run using Dep Check. Ours was a CLI integration though, so would be great to have that exposed there as well.

stevespringett:
Ok, so the objective is to have a Maven goal that can keep a centralized data directory up-to-date. 

Depending on the size of your CI infrastructure, it may be beneficial to use something like NIST Data Mirror (https://github.com/stevespringett/nist-data-mirror). Many build servers are locked down and prohibited from accessing the Internet as part of controlling third-party dependency usage and preventing XBI vulnerabilities. Many orgs simply run a NIST data mirror that gets updated periodically, and tell Dependency-Check to use the mirror instead.

I'll let @jeremylong comment on the Maven goal.

stevespringett:
@jeremylong If we add an update goal to Maven, we need to do the same with the Ant task and CLI. If this is something we want to do, does it make sense to make it a two part update?
- Option 1 - Mirror NIST NVD data by performing the download and not processing it.
- Option 2 - Perform a regular DC update task

Option 1 would be good for hosting NIST NVD data on an intranet, accessible via HTTP(S), and available to any Dependency-Check instance running with network access to the mirror.

Option 2 would be good for keeping an up-do-date data directory for a smaller build infrastructure where the build machines all have access to the same file system.

jeremylong:
I don't see a problem with adding an update goal for Maven and an
updateOnly option for Ant and the CLI. However, I think we should keep the
nist-data-mirror separate. Also, documentation outlining how to use the
nist-data-mirror should be created.

Steve - I'll followup with you regarding the nist-data-mirror off list.

--jeremy

jeremylong:
This PR was manually re-created in order to implement this in the CLI and Ant task as well as a maven goal. The site documentation has been updated.

Bowbaq:
Thanks for implementing this feature. Is it automatically available on Maven, or does it need some kind of version bump?

jeremylong:
A new version will need to be released. I have one more change in flight and then I will push 1.2.10 (likely within a week).

Best Regards,

Jeremy

Bowbaq:
Hi Jeremy, do you know when you plan to tag a new release?

stevespringett:
This feature will been added to v1.2.11 of the Jenkins plugin as well.

jeremylong:
I would expect within 2 weeks. I have some documentation and a couple of bug fixes to complete before the next release.

--Jeremy

Issue #207, opened 2015-04-01
Add Python Analyzer


dwvisser:
Analyzer should gather evidence from the various downloadable artifacts from the [Cheese Shop](https://pypi.python.org/pypi):
- Eggs: .egg and .zip extensions
- Wheels: .whl extension
- Sdists: .tar.gz extension

Also gather evidence from a Python installation's site-packages (e.g., a system's installed packages, or a virtualenv's installed packages):
- .dist-info/MANIFEST
- .egg-info/PKG-INFO
- .egg/EGG-INFO/PKG-INFO
- actual Python package folders

I already have a prototype successfully parsing out MANIFEST files and PKG-INFO files from the file system and .whl archives. I need to finish that part, and add scanning of the actual Python packages (in file system and sdists).

P2exe- and Py2app-deployed code (i.e., compiled) is out of scope for this issue tracker. Certain kinds of applications that embed the Python interpreter may by scannable though.

jeremylong:
+1 - I know several people that have asked for Python.

In addition to the core Analyzer we likely need to enhance the suppression analyzer and base suppression XML file to limit some of the obvious false positives. One thought I had was to enable to suppression to be based off of the description within the CVE entry - specifically because most of the python related CVEs  appear to have a '.py' or 'python' listed in the description. So if those are not present the CVE may not be related to python library (this would need to be researched to ensure valid matches are not masked).

jeremylong:
see pull #217

PR #217, opened 2015-04-14
Add Analyzers for Python Distribution Files and Packages

dwvisser:
This is to address #207. The following has been added:
- PythonDistributionAnalyzer - Analyzes Python wheel and egg distribution formats, as well as their unbundled forms that can sit directly in your python environment.
- PythonPackageAnalyzer - Directly analyzes Python package source code for evidence of known Python libraries.
- Test data and unit tests for the new analyzer classes.
- Suppression of mapping any artifact found under a dist-packages/ or site-packages/ folder with cpe:/a:python:python*
- Command-line options for disabling the Python analyzers.


dwvisser:
I just added commit b90774c with an additional false positive suppression for the current version of Django.

dwvisser:
I just added dadf49c with suppression of additional non-Python-based application CPEs. I am _not_ adding suppression of packages that refer to JavaScript code. Generally, those packages actually include the code they mention, and are either a web application using it, or a plugin to some larger project. I.e., the inferred dependency has a chance of being correct.

The net effect of this change when scanning a test dataset was to remove 54 definite false vulnerable dependencies from a result set showing 118 vulnerable dependencies. The dataset had a total of 380 wheel files downloaded from PyPI or built from source distributions downloaded from same.

dwvisser:
afa9c84 is a refinement which takes away a small number of extra false positives in my dataset.

PR #231, opened 2015-05-08
I created a Dependency Check Gradle plugin

wmaintw:
Hi,

The Dependency Check is really useful, we can get lots of benefits from it, but I found that it doesn't have gradle plugin, although I can work around this, I think a dedicated gradle plugin is a better choice.

Hence, I create this Dependency Check gradle plugin, it utilise the functions in dependency-check-core module, it will verify project dependencies, update data feed, analyse each dependency and generate report.   

For now, this plugin is just a MVP, more development work is required to add more functionalities, but at least, it works for gradle project.


jeremylong:
Thank you for the PR!  I will absolutely be merging this - but it will likely not make it into the next release as I am doing final verification right now.

wmaintw:
My pleasure, I know there are lots of development work required, it's not possible to release it recently, I will keep work on this, make it configurable, performance improvement, error handling, etc.

Thanks a lot :)

wmaintw:
Hi Jeremy, since the 1.2.11 just got released, would you please approve this pull request?

Best Regards.

jeremylong:
Thanks! The PR has been merged. However, could you please go through and add the legalese(Apache 2.0), including the copyright Wei Mai or Thoughtworks...  I noticed in the pom.xml for the gradle plugin it still has my name in the copyright.

--Jeremy

wmaintw:
Sure, will do that, thank you very much.

PR #236, opened 2015-05-27
Fix table name in dbStatements

ercpe:
Changed table name to "cpeEntry" to avoid SQL errors on MySQL (case sensitive)


jeremylong:
Thanks!

PR #252, opened 2015-06-13
More dependency management

hansjoachim:
Moves more version numbers from the various modules (-cli, -maven, -utils) into the parent pom. That way, if multiple modules use the same dependency, they will have the same version. Also makes it easier to maintain by keeping the version numbers in a single place.

Upgarded maven-core, -plugin-api and -settings to latest version.

Added comments for dependencies where newer versions are available, but upgrading will introduce problems or warnings.

Verified `mvn clean install` still runs successfully.

### DIFF
--- a/dependency-check-cli/pom.xml
+++ b/dependency-check-cli/pom.xml
@@ -310,7 +310,6 @@ Copyright (c) 2012 - Jeremy Long. All Rights Reserved.
         <dependency>
             <groupId>commons-cli</groupId>
             <artifactId>commons-cli</artifactId>
-            <version>1.2</version>
         </dependency>
         <dependency>
             <groupId>org.owasp</groupId>
--- a/dependency-check-core/pom.xml
+++ b/dependency-check-core/pom.xml
@@ -391,7 +391,6 @@ Copyright (c) 2012 Jeremy Long. All Rights Reserved.
         <dependency>
             <groupId>commons-io</groupId>
             <artifactId>commons-io</artifactId>
-            <version>2.4</version>
         </dependency>
         <dependency>
             <groupId>commons-lang</groupId>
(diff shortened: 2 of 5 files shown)


hansjoachim:
Two comments:
The -core module remains. The long-term plan there is to move the "useful" dependencies which we wish to keep up to date into the root pom's dependencyManagement while leaving the "only-used-for-tests-and-we-need-ancient-versions-because-they-need-to-be-vulnerable-to-verify-the-vulnerabilities-are-caught" dependencies alone. Would need to split them into either of this categories to do it properly. I haven't looked closely at this yet, but the current patch should fix the other modules.

Slightly off-topic, I've mentioned this before, but I haven't found a good way to organize the reporting plugins version numbers in a good way. According to [this bug report](https://issues.apache.org/jira/browse/MSITE-443), but looking at [one of the comments](https://issues.apache.org/jira/browse/MSITE-443?focusedCommentId=14451563&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-14451563) and going by my personal experience I can't get it to work. I can't find that now, but I vaguely remember something about a related bug which affected Maven prior to 3.1.x, maybe that's it.

jeremylong:
We could move the old vulnerable test dependencies to an active profile to
really denote them as only test dependencies...  thoughts?

--jeremy

hansjoachim:
> We could move the old vulnerable test dependencies to an active profile to
> really denote them as only test dependencies...  thoughts?

I don't know, I hadn't heard about activeprofiles until you mentioned it now. From what I skimmed briefly, it looks like it might work though.

I guess if the dependencies are moved into a separate profile they wouldn't show up in the list where newer versions are available. Of course, users/developers shouldn't need to specify some particular profile when building things. Most who clone the repo or have prepared a patch will run `mvn clean install` to verify things are working.  Presumably this is the part where activeprofiles help so that people don't have to specify some particular profile when building.

I'm curious how the active profile affect other things though, and whether that would still include those dependencies when running `mvn versions:display-dependency-updates`. I guess we should attempt to move one or two into a profile, set it as active and see how it works.

As long as we separate the ones we wish to keep upgrading from those which should be frozen, it think it sounds ok as a solution.

Issue #239, opened 2015-06-02
Consider using native Ant logging instead of java.util.logging


awood:
I'm using the Dependency Check ant task, but using it from Ruby via the [Apache Buildr](http://buildr.apache.org) too.  I parse the generated XML to display the results to users and therefore want to suppress the output from the Ant task.  Since the ant task is using java.util.logging directly I don't really have a way to suppress the output.  If it were using [Ant's native logging functionality](http://javadoc.haefelinger.it/org.apache.ant/1.8.1/org/apache/tools/ant/Task.html#log%28java.lang.String%29) then I could access the Ant BuildListener and suppress or redirect the output to /dev/null.

If this sounds like something you would be interested in, I would be happy to write the code and provide a pull request.

jeremylong:
This is planned as part of the fix for issue #164.  I will try to include
this in the next major point release (1.3).

In addition to ant logging for the ant task I plan on also using maven
logging for the maven plugin, etc.

--jeremy

jeremylong:
While some tweaks may still occur - the base adapter for ant has been committed to 1.2.12-SNAPSHOT and will be included in the next release.

Issue #164, opened 2014-10-24
Log messages routed to home/java.log


jeremylong:
In some cases log messages are being routed to the default log file location for JUL. This may be an opportunity to switch to logback.

jeremylong:
After further investigation the log files going to the default log file location for JUL was caused during unit testing. This ticket is being changed from a bug to an enhancement request to change from JUL to logback (or other logging framework).

jaypatrickhoward:
In lieu of a all-out migration off of JUL, in the interim, could you just modify the owasp JUL logging so that it respects the global JUL configuration?  My logging.properties file is this:

handlers=java.util.logging.ConsoleHandler
.level=OFF

And I have stderr and stdout redirected to /dev/null.  When I run the dependency check plugin I still get "INFO" log statements from the following loggers written to ~/java0.log:

org.owasp.dependencycheck.Engine
org.owasp.dependencycheck.data.update.StandardUpdate
org.owasp.dependencycheck.data.update.task.DownloadTask
org.owasp.dependencycheck.data.update.task.ProcessTask

jeremylong:
JUL is no longer in use.

PR #253, opened 2015-06-15
Slf4j over ant

colezlaw:
Patch for jeremylong/DependencyCheck#164 and jeremylong/DependencyCheck#239. This switches everything to Log4J. The CLI still uses JULi as its implementation, but all the logging messages are going through SLF4J. The Maven plugin and the Ant plugin both use their own internal logging mechanisms rather than using SLF4J. Furthermore, adapters have been written for both so that SLF4J messages from Core are routed through the native logging interfaces for each.


dependency-check-cli/pom.xml
```diff
@@ -322,5 +322,12 @@ Copyright (c) 2012 - Jeremy Long. All Rights Reserved.
             <artifactId>dependency-check-utils</artifactId>
             <version>${project.parent.version}</version>
         </dependency>
+        <!-- Logging implementation. We may change this in the future -->
+        <dependency>
+            <groupId>org.slf4j</groupId>
+            <artifactId>slf4j-jdk14</artifactId>
+            <version>1.7.12</version>
```
hansjoachim, line 329: Please put version numbers for dependencies under dependencyManagement in the root pom. :) Makes it easier to keep track of them when they're in one place, and it's a single place to upgrade once a new release is out. (See also #252).

dependency-check-core/pom.xml
```diff
@@ -362,6 +362,26 @@ Copyright (c) 2012 Jeremy Long. All Rights Reserved.
     <dependencies>
         <!-- Note, to stay compatible with Jenkins installations only JARs compiled to 1.6 can be used -->
         <dependency>
+            <groupId>org.slf4j</groupId>
+            <artifactId>slf4j-api</artifactId>
+            <version>${slf4j.version}</version>
```
hansjoachim, line 367: I might be missing something here, but where is the slf4j.version property defined?

hansjoachim:
Hi :)

I'm just an occasional contributor, so I haven't looked at this in-depth, but I left a couple of comments.

colezlaw:
Closing this to rebase with merges and fix issues with versioning.

colezlaw:
@hansjoachim I closed the pull request to fix your questions plus another Jeremy had open. However, I've always been puzzled by the dependencyManagement thing - I know how it's SUPPOSED to work, but if I define a version for slf4j-api in the parent POM dependencyManagement section, then define the dependency without a version in child projects, I still get an error in the child that I didn't specify a version. However, things like commons-io don't have version numbers and those work.

hansjoachim:
Like you said, that sounds like it should work. Even stranger if it works for other dependencies. I don't know what would be causing that, but if you have an example, let me know and I can take a look at it. :)

colezlaw:
Sorry. I misspelled "org". That's fixed.

What all dependency versions do you want in dependencyManagement? All of
them? Or just ones that appear in multiple projects?

hansjoachim:
> What all dependency versions do you want in dependencyManagement? All of
> them? Or just ones that appear in multiple projects?

Please add all of them to dependencyManagement. For one thing they might be picked up by other modules in the future (or when new ones are introduced) and it means there will be a single file which needs updating when newer versions are released. :)

Btw, I don't know how obvious it is, but I've tried to keep the list in dependencyManagment in alphabetical order.

PR #254, opened 2015-06-16
Now switched to slf4j

colezlaw:
This should close out issue #239 - this is everything moved over to SLF4J including Maven and Ant - except SLF4J messages from Core will go through Ant's own logger or Maven's own logger.


jeremylong:
Thanks!

PR #256, opened 2015-06-17
Some minor dependency-check-core compiler warning fixes

dwvisser:
Isolated sources of deprecation warnings, and added warning suppression annotations. Also added a minor Enumeration -> Enumeration<JarEntry> "fix".


jeremylong:
Thanks for the patch!

PR #259, opened 2015-06-20
Moving all versions into the parent project

colezlaw:
Per @hansjoachim's recommendation to move all dependency versions into the parent dependencyManagement section. In core, I've kept the test dependencies which are used for guinnea pigs at their versions in the core pom - should we consider making a separate project for these dependencies?


hansjoachim:
I'm not sure what to do with the "test-dependencies". It has been mentioned a couple of times, but @jeremylong has a suggestion towards the end of #252 which might work.

PR #260, opened 2015-06-21
Reporting plugin version numbers

hansjoachim:
As mentioned earlier (in #252) I've fought a bit with the reporting plugins and tried to place their version numbers under pluginManagement like the others. Unfortunately, this doesn't seem to work in practice. The alternative solution I've landed on is to place the version numbers in global properties in the root pom and then make sure these are assigned each place the plugins are used. Perhaps not ideal, but makes things consistent at least.

Note that this is work in progress, i.e. not intended for merging yet. Feel free to review and comment though. I'll remove the WIP tag when I consider it fit for inclusion, probably once I've dealt with all the modules.

Several of the plugins have newer versions available, but I wish to land this first so that all modules are synced using the same version before attempting to upgrade any. I have noticed that some plugins (notably maven-project-info-reports-plugin, and I believe findbugs) have different versions in different modules. I've assumed they've simply become out of sync for some reason, but please let me know if this was somehow intentional.

I've also marked at least one place with TODO(code review) which should probably be addressed before this is closed.


hansjoachim:
It should be noted that I have some slight problems verifying that `mvn site` still works as expected. I get some weird stacktraces when generating the findbugs report. I believe this is due to [a bug in the findbugs-plugin](https://github.com/gleclaire/findbugs-maven-plugin/issues/9), and I also get it with latest master so it doesn't seem like something I've introduced. I'll experiment with openjdk7 and see whether that works better.

hansjoachim:
Did some sample checks with JDK-7 now and all pages seem to work. I consider this ready for review now.

jeremylong:
Thanks!

Issue #208, opened 2015-04-01
Add analyzer of C/C++ project source and build files


dwvisser:
Example scenario:
- Developer wants to use the “foo” library.
- However, to make it “easy to install” and not depend on the version (if any) installed on the system, she creates a “foo” subdirectory and _copies the entire foo library_ into that subdirectory.
- Further, the “foo” subdirectory has a “configure.in” file, and **that file** tells us the version of _foo_.
- Then, her top-level build system will cd into the foo library, and build, when needed.  (Typically with a recursive make).  The final executable will the embed the “.o” file in this “foo/” subdirectory.

The proposal here is to create an Analyzer that that is able to scan for some common ways of embedding version information in C/C++ build files. Further, items like the license and copyright comment headers can be mined for name and vendor information.

I'm planning on implementing this in a little while.

jeremylong:
I know several people have asked for C/C++ support. I think this area of research is definitely worth-while. However, I am concerned about the false positive rates an analyzer like this might produce.

--Jeremy

PR #270, opened 2015-07-06
CMake Analyzer implementation

dwvisser:
This submission is to partly address #208. The analyzer inspects CMake build files for project name and version information. Sample data is included that gets used by the unit tests.


jeremylong:
After looking at this a bit I think we need to make a change to some core functionality - specifically the supportsExtension(String) defined in the Analyzer Interface. As the CMakeAnalyzer seems to only care about a specific txt file name. Otherwise this could end up creating dependencies for all txt files that were scanned. My proposal would be to change supportsExtension(String) to supportsFileName(String name, String ext). So we would call this like Analyzer.supportsFileName("something.txt", ".txt");.

Thoughts?

jeremylong:
Because PR #275 was merged, this PR now has merge conflicts.

dwvisser:
Now that #275 is merged, I will be able to use a filename-based filter, and clean this code up some.

dwvisser:
Fixed merge conflicts, and updated to FileFilter mechanism, which nicely simplified the code.

jeremylong:
Dale,

I just realized that this analyzer would be active in Maven, Ant, and Groovy - I could see adding this to the Ant, but probably need to disable by default in the Maven and Groovy plugin? Thoughts?  I'll have to go back and look at the other C analyzers too.

--Jeremy

dwvisser:
Good point. I haven't thought about those users with this (and Autoconf and OpenSSL). I think it's sensible to disable by default in the Maven and Gradle tools.

dwvisser:
Do you want me to add changes that disable this and the other C analyzers in the Maven and Gradle tools?

jeremylong:
If you have time - please, otherwise I will get to it (hopefully) this
weekend).

--Jeremy

jeremylong:
This has been merged

PR #278, opened 2015-07-13
add setting for customize http method when querying timestamp

wmaintw:
Currently, the Downloader will use `HEAD` as HTTP request method to query timestamp, it works most of the time, but it will prevent user from updating if user is behind a strict proxy which only allow `GET` and `POST` method.

So I introduced a setting item:

`downloader.quick.query.timestamp=true`

The default value is `true`, so the Downloader will use `HEAD` in `getLastModified()`, and this is the current behavior, won't affect existing users. The user then could change the value to `false` and in this way, the Downloader will use `GET` as HTTP method to query timestamp.


wmaintw:
Hi Jeremy, do you have any comments about this change? would you please approve it?

jeremylong:
I plan on merging this - however, there are a lot of other tasks that need to be performed (updating the other interfaces (Maven, CLI, Ant) and updating the related documentation. I should be able to do this soon but I am in the middle of debugging a build issue that only shows up in some linux environments using JDK8.

--Jeremy

wmaintw:
Got it, thank you very much.

jeremylong:
This change has been merged.

PR #279, opened 2015-07-14
Eliminated duplicate key in dependencycheck.properties file

dwvisser:
Commented out first instance of cpe.url, and moved 2nd instance up. Assumption: the 2nd value was being used.


jeremylong:
Thanks for catching this...

PR #298, opened 2015-08-03
Embedded ID spec; with a fast, efficient analyzer

dwvisser:
- # 212: This pull request introduces `src/site/markdown/analyzers/embedded-id-spec.md`, which lays out the concept of ASCII Embedded Identifiers primarily intended for C and C++ code to reliably place the embedded identifiers into object files. A key feature of the embedded identifier itself, is that it is prefixed by a random hexadecimal sequence that is both unlikely to randomly occur, and is a fixed sequence that can be efficiently scanned for in any file, i.e., including source files for any language.
- # 211: A very fast and efficient analyzer has been implemented that combines the [Knuth-Morris-Pratt algorithm](https://en.wikipedia.org/wiki/Knuth%E2%80%93Morris%E2%80%93Pratt_algorithm) with a `java.io.RandomAccessFile`, using buffered reads and avoiding `seek()` calls. On my system, I didn't notice much, if any, overhead with the addition of scanning all files with this analyzer. Performance tests saw speeds approaching the memory bandwidth of my machine.

If accepted, a priority on my current funding would be to then go ahead and select some prominent open source projects that can be sensitive security-wise (e.g., those found at https://www.coreinfrastructure.org/programs/census-project), and submit embedded ID patches.


dwvisser:
Anticipating a potential question here...

I chose `java.io.RandomAccessFile` with buffered reads over `java.io.BufferedInputStream` or other `InputStream` implementations, because each file would only be hit once or twice at most. It has been shown by others that `RandomAccessFile` used like this, is fast from the start: http://mechanical-sympathy.blogspot.com/2011/12/java-sequential-io-performance.html

jeremylong:
Dale,

As I've stated previously - while it would be great to have something like this my belief from efforts like CPE and SWID that this type of standard would not be used. This adds overhead to releases as people would have to remember to go change the number which could easily lead to multiple releases with identical embedded IDs.

--Jeremy

dwvisser:
CPE and SWID have suffered poor adoption for various reasons, which I’ve attempted to mitigate here. Both are IDs registered/assigned using a central authority. While there are free e-mail- and web- based ways to obtain CPEs, one has to know to do it, and there is not necessarily an incentive before a CVE is created. SWIDs are worse. As with most things ISO, financial investment is needed to be able to implement and benefit from the standard. That is a non-starter for many FOSS projects. Here, I’ve attempted to define a voluntary opt-in open standard that is easy to implement. I added a CPE variant, for those projects that already have an assigned CPE.

With regards to the difficulty of remembering to update the version. Projects can and would use build tools to automate this at the time of release. Most projects that have been through multiple release cycles utilize build tools to “single source” the version number.

jeremylong:
I've discussed this with several people and we just don't think this is going to work. I understand the goal; but getting other projects to adopt this will be extremely tough. Additionally, even if we did get 1-2 projects on board it almost gives a false sense of security.

PR #300, opened 2015-08-05
Added homebrew installation instructions

colezlaw:


dwvisser:
I think you're supposed to modify `dependency-check-cli/src/site/markdown/index.md.vm` in the master branch, instead.

colezlaw:
Ah - I was looking for that - you can reject the patch, then.

PR #310, opened 2015-08-09
Node.js analyzer

dwvisser:
This new analyzer utilizes the [org.json parser](http://mvnrepository.com/artifact/org.json/json) to parse Node Package Manager [package description files](http://browsenpm.org/package.json).

It appears many node.js packages have CPEs where the vendor name is set as: `productname_project`, so the analyzer synthesizes this form as vendor evidence.


colezlaw:
I thought we were still redirecting folks to retire.js for Javascript (node and otherwise) support.

jeremylong:
I don't think retire.js handles node. I know I wasn't going to spend time
building one until more of the other tickets were closed... But I'm okay
adding this in.

--Jeremy

colezlaw:
If you're okay adding it in, that's fine. But retire.js has better node
support than UI type libraries. At least, the node plugin is very easy to
run, and the list of node packages on the naughty list is much longer than
the list of browser ones.

colezlaw:
That does bring up something else...I had been waiting to see if we had a
direction on which json processor to include. Since that decision has been
made, I can start working on PHP support.

dwvisser:
AFAIK, Retire.js maintains its own "vulnerable versions" database, which uses NVD as just one of the inputs.

I think there's no harm in adding this node.js analyzer to D-C. If merged, then on the site documentation for the analyzer, I would add something to this effect: that retire.js is a more comprehensive tool for node.js and client JavaScript, but that this analyzer has the advantage of always being completely up to date with the NVD if D-C is run regularly.

As for JSON processor, I picked the most popular json.org library which I have used on other projects. There are other choices, though, and if licensing concerns (json.org has a very permissive, yet oddball homegrown license) rule it out, there _are_ alternatives that also implement [JSR 353](https://jcp.org/aboutJava/communityprocess/final/jsr353/index.html).

jeremylong:
The latest org.json is compiled with Java 8 - as we are supporting a Jenkins plugin this is a problem. I believe it has to be 1.6 compatible.

dwvisser:
OK. I'll look for a JSR 353 alternative that doesn't hit the Jenkins compatibility issue.

dwvisser:
Switched to Glassfish reference implementation of Java JSON Processing API. Modified code to to run against the spec, rather than custom org.json calls that were being used before.

PR #312, opened 2015-08-12
Rubygems analyzer

dwvisser:
This analyzer utilizes some custom regular expressions to parse Ruby Gem specifications out of .gemspec and Rakefile files. The specifications are Ruby code, and so _could_ vary from the format I'm parsing. However, I've found in practice that .gemspec files downloaded from https://rubygems.org/ code the values directly using string and list literals in the specification block.


jeremylong:
Dale,

For Ruby - [bundler-audit](https://github.com/rubysec/bundler-audit) would be the tool of choice. Why would we want to include a Ruby analyzer in dependency-check?

--Jeremy

mprins:
With jruby gem files can become first class dependencies (just like jar files) eg. this [maven plugin](https://github.com/GeoDienstenCentrum/sass-maven-plugin) depends on and bundles gem files. I'm not sure the way this is implemented would catch a use-case like that, but it would be nice.

PR #316, opened 2015-08-14
Help for scanning ISO images on Linux, Mac OS X, and Windows

dwvisser:
Since it is not easily possible to examine an ISO from Java, how to leverage OS capabilities.

### DIFF
--- a/src/site/markdown/analyzers/index.md
+++ b/src/site/markdown/analyzers/index.md
@@ -5,7 +5,7 @@ to extract identification information from the files analyzed.
 
 | Analyzer | File Types Scanned | Analysis Method |
 | -------- | ------------------ | --------------- |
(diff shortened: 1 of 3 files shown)


jeremylong:
We need to add warnings to this around binaries (i.e. C/C++). This tool does not scan SO files, non-.NET DLL and EXE, and other binary formats. OWASP dependency-check will be able to scan file types listed in https://jeremylong.github.io/DependencyCheck/analyzers/index.html. As the documentation is written now it sounds like we could find any  dependency/library on the system that is vulnerable and that isn't true.

In addition, we should consider expanding the list of [file type analyzers](https://jeremylong.github.io/DependencyCheck/analyzers/index.html) to indicate file extensions to make this more clear.

dwvisser:
All good points. I will work on adding the clarifications.

dwvisser:
I added a statement referring to the analyzers page, and made the analyzers page much more descriptive, utilizing a 3-column table. I also made the side menu of analyzers alphabetical to match the page.

PR #330, opened 2015-08-30
Ruby bundler-audit Analyzer

dwvisser:
The [rubysec team](https://github.com/rubysec) does a great job with their [ruby-adisory-db](https://github.com/rubysec/ruby-advisory-db) and [bundler-audit](https://github.com/rubysec/bundler-audit) projects. Therefore, I have created an analyzer that invokes `bundle-audit` (if present) in directories containing `Gemfile.lock" files. It then uses the resulting output to populate D-C Dependency & Vulnerability objects for reporting.

To test this, you'll of course need to install [ruby](https://www.ruby-lang.org/en/downloads/), [gem](https://rubygems.org/pages/download), [bundler](http://bundler.io/#getting-started), and [bundler-audit](https://github.com/rubysec/bundler-audit#install). If you put the bundler-audit executable in your environment's path, then no additional Dependency-Check configuration is needed. For a `Gemfile.lock` referring to known vulnerable libraries, you can point the CLI at `dependency-check-core/src/test/resources/ruby`.


jeremylong:
Dale - does this disable the other ruby analyzer if it can be used? This was you might reduce the noise, yet still allow a multi-language project to collect the data via a single end reporting tool...

dwvisser:
This is complementary to the other analyzer, which doesn't depend upon the developer using Bundler.

dwvisser:
Oh, I think I understand the point you're making now. I could see doing that, but I would want to log a message during processing that the other analyzer was turned off, and would want a way to force it back on.

jeremylong:
One of my concerns is that this would duplicate the dependencies between the two analyzers. If both analysis options are turned on the dependencies identified by both analyzers should be reported as a single dependency with evidence/findings from both routes. My vote would be disable the original analyzer if the bundler file is found. Then add a forceRubyAudit option.

stevespringett:
In the Jenkins plugin. There is a global option that allows administrators to select which analyzers are enabled. By default, everything except Nexus is enabled by default.

screen shot 2015-09-10 at 9 36 17 am

From my understanding, this approach would not work for the Ruby analyzers. It's one or the other but not both? If this is true, is it possible to have a single Ruby analyzer that figures out which underlying analyzer to use?

dwvisser:
I wish I were more of a Ruby developer. I don't know what the population of rubygems-but-not-bundler users are. I think I would prefer it if, when the bundle-audit analyzer _successfully_ initializes, that it logs a warning that it may be desirable to disable the Gemspec analyzer.

jeremylong:
Dale,

That would be my preference - if bundled analyzer initialized successfully
the gemspec analyzer would get disabled.

Jeremy

dwvisser:
I've been trying the "successful bundler-audit analyzer initialization disables gemspec analyzer" approach. I've been unsuccessful thus far, but probably can ultimately get it to work. My difficulty stems from two factors:
- `Analyzer` instances are meant to be independent items invoked by the Engine
- `ServiceLoader` makes no guarantees about load order (which affects later invokation sequence)

I really think it would be better just to issue an informative warning upon successful bundler analyzer initialization that the user may wish to disable the gemspec analyzer in order to avoid redundant results.

jeremylong:
Can't you run the bundler analyzer in an earlier phase? It really isn't
analyzing...

dwvisser:
I'll try that. Thanks. :-)

dwvisser:
OK. That was much simpler. Now the first time a `Gemfile.lock` is about to be processed, the Gemspec analyzer is disabled. I added a new phase `PRE_INFORMATION_COLLECTION` to be distinct from the `INITIAL` phase.

jeremylong:
Dale,

Due to timing and life events this won't be in the 1.3.1 release; I need to spend a bit more time reviewing and possibly adding a couple of things to this. I truly appreciate the work on this.

--Jeremy

dwvisser:
I understand. Thanks for letting me know. 😊

dwvisser:
Just merged master into my pull request branch, and dealt with the merge conflicts.

dwvisser:
Latest commit: Explicity shortcut running of unit tests if setup fails, just as AssemblyAnalyzer does.

dwvisser:
I just brought this up-to-date with the master branch and re-tested the functionality.

PR #333, opened 2015-09-03
Updated plugins.

awhitford:
- I was able to update the site plugin to version 3.4
  - I double checked that the markdown pages still worked (by upgrading `doxia-module-markdown` to 1.6), and that the _file type analyzer_ pages work.
- I noticed that some plugins were not the latest version, so updated those.
- I noticed that some plugins versions were mismatched (surefire-report); so updated that.


jeremylong:
Thanks!

PR #334, opened 2015-09-03
Removed deprecated code.

awhitford:
- Component annotation has been deprecated; Parameter should be used.
- ArtifactMetadataSource has been deprecated; along with ArtifactMetadataRetrievalException -- since they were not used, easy to remove.


jeremylong:
Thanks for the PR!

PR #336, opened 2015-09-05
Added time measurements for key steps.

awhitford:
I would like to gain better insight as to what is taking time.

As a result, added some basic timing for some key steps like running an Analysis, Checking for Updates, and Downloading and Processing NVD files.


jeremylong:
The other thing you may want to time is the creation of the in memory lucene index - as this can also take a bit more time. I have updated org.owasp.dependencycheck.analyzer.CPEAnalyzer

PR #338, opened 2015-09-07
Pom cleanup

awhitford:
Moved a lot of the reporting plugins from each module to the parent pom.  This greatly decreases the level of repetition, making the project more consistent and easier to manage.

There were also some redundancies addressed -- like not needing to declare the `maven-compiler-plugin`, or no need to repeat elements like `license` which are already inherited from the _parent pom_.

The cobertura plugin was upgraded and made consistent between version build and reporting.  (Previously 2.7 was used for build, but then 2.6 was used for reporting.)

The `maven-plugin-plugin` was moved to the _maven module_ since it is only used there.  Properties introduced (or renamed) to ensure plugin version consistency between _build_ and _reporting_.

Note that some plugins, like `jxr`, have no impact to modules that have no java source, so it is safe to move to parent pom and non-java modules like gradle and jenkins will not be impacted.

The `maven-project-info-reports-plugin` mentioned in modules referenced an invalid `help` report, as well duplicate reports from the parent pom, so this cleanup yields fewer build warnings.

### DIFF
--- a/dependency-check-ant/pom.xml
+++ b/dependency-check-ant/pom.xml
@@ -267,20 +267,6 @@ Copyright (c) 2013 - Jeremy Long. All Rights Reserved.
         <plugins>
             <plugin>
                 <groupId>org.apache.maven.plugins</groupId>
-                <artifactId>maven-project-info-reports-plugin</artifactId>
-                <version>${reporting.project-info-reports-plugin.version}</version>
-                <reportSets>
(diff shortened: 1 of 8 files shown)


jeremylong:
Thanks for all of the PRs! - I truly appreciate the help.

awhitford:
Happy to help.

PR #343, opened 2015-09-08
Code tweaks

awhitford:
Minor code tweaks to suppress FindBugs, Checkstyle, PMD, and JavaDoc alerts.

### DIFF
--- a/dependency-check-cli/src/main/java/org/owasp/dependencycheck/App.java
+++ b/dependency-check-cli/src/main/java/org/owasp/dependencycheck/App.java
@@ -416,7 +416,7 @@ public class App {
     }
 
     /**
-     * Takes a path and resolves it to be a canonical & absolute path. The caveats are that this method will take an Ant style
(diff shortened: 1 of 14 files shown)


jeremylong:
Again, thanks for help cleaning up the code. I generally go through (most) of the checkstyle/pmd/findbugs issues just prior to a release - I should change this to after accepting PRs as some do not look at these reports prior to sending the pull request.

I did notice one bug that may exist elsewhere in the code - there is a `LOGGER.warn("test %s", arg);`. This uses the String.format syntax rather then the slf4j syntax of `"{}"` for the placement of the arguments. May have to write a rule to flag this in the future.

awhitford:
Yeah, I noticed that %s in the logger...  Glad to see that cleaned up.

colezlaw:
Those would probably be my fault. There were a handful of cases where with the old logger we used String.format, and I failed to change those over to SLF4J syntax.

PR #344, opened 2015-09-10
Db opts

awhitford:
Replaced `"".equals(str)` with `str.isEmpty()`.
Added missing `serialVersionUID`.
Added types for raw declarations (Iterator and HashMap).

### DIFF
--- a/dependency-check-core/src/main/java/org/owasp/dependencycheck/analyzer/AssemblyAnalyzer.java
+++ b/dependency-check-core/src/main/java/org/owasp/dependencycheck/analyzer/AssemblyAnalyzer.java
@@ -147,7 +147,7 @@ public class AssemblyAnalyzer extends AbstractFileTypeAnalyzer {
 
             // First, see if there was an error
             final String error = xpath.evaluate("/assembly/error", doc);
-            if (error != null && !"".equals(error)) {
+            if (error != null && !error.isEmpty()) {
                 throw new AnalysisException(error);
             }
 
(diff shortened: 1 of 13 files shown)


jeremylong:
Most of the changes are great. I'm concerned about the MERGE statement. Currently, we have users that have swapped out for Oracle and MySQL. I believe the merge will work fine in Oracle - but the syntax for MySQL is different (see [INSERT... ON DUPLICATE KEY UPDATE...](https://dev.mysql.com/doc/refman/5.0/en/insert-on-duplicate.html)). Currently, there is no built in mechanism to use a different resource bundle for a different SQL Dialect (this is planned, just not built yet).

awhitford:
I understand the apprehension on the `merge` statement.  I wondered about that.

I was mislead because I saw that `ADD_DICT_CPE` already exists and it uses `merge`.  And then I also noticed that [`merge`](https://en.wikipedia.org/wiki/Merge_%28SQL%29) was added to the ANSI SQL 2003 standard.

Because the code structure is different (one merge statement vs two statements for insert and update), I thought an ideal strategy is to query the [DatabaseMetaData](http://docs.oracle.com/javase/7/docs/api/java/sql/DatabaseMetaData.html) to see if the merge statement is available.  Alas, I don't see any such way to derive that information.

Let me think about this some more...  Worse case is that I will revert that so that you can accept the other minor clean up.

awhitford:
OK, I reverted the `merge` command.  Hopefully you can accept the rest.

jeremylong:
Sorry for the confusion because dead code uses the merge statement (I would have caught that prior to a release in testing). The `ADD_DICT_CPE` and supporting code was created to test a theory - turns out we didn't need the code. I left the code in place and forgot to mark it as unused/un-needed. (I didn't want to wholly delete the code so I could test things in the future)

PR #345, opened 2015-09-12
Upgrade jmockit to 1.19

hansjoachim:
https://github.com/jmockit/jmockit1/issues/196 has been resolved and the latest version of jmockit works fine with openJDK again. `mvn clean install` ran successfully.


jeremylong:
Was this tested with openJDK? when we tried to upgrade to 1.17 we got build errors (per the comment).

hansjoachim:
> Was this tested with openJDK?

Yes, built with openJDK8. I guess I could have specified that in the original comment.

PR #346, opened 2015-09-12
Move version number for maven-dependency-plugin out into a variable s…

hansjoachim:
…o we can also use it in the reporting section

Currently, when running for instance `mvn versions:display-dependency-updates`, I get several error messages saying that the plugin lacks a version number. Unfortunately it looks like plugins in the reporting section doesn't pick up version numbers from dependencyManagement, so I've added a property to keep it in one place. The warning went away, and we are able to keep the version number consistent, but I'm not too happy with this workaround, so if anyone has a better idea feel free to let me know.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -129,6 +129,7 @@ Copyright (c) 2012 - Jeremy Long
         <apache.lucene.version>4.7.2</apache.lucene.version>
         <slf4j.version>1.7.12</slf4j.version>
         <logback.version>1.1.3</logback.version>
+        <dependency-plugin.version>2.10</dependency-plugin.version>
         <reporting.checkstyle-plugin.version>2.16</reporting.checkstyle-plugin.version>
         <reporting.cobertura-plugin.version>2.7</reporting.cobertura-plugin.version>
         <reporting.pmd-plugin.version>3.5</reporting.pmd-plugin.version>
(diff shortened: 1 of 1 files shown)


awhitford:
My suggestion is to add a declaration of any reporting plugin to the `pluginManagement` section, then you can remove the repeated version from the reporting section.  [Maven 3 will resolve Reporting plugins the same way as build plugins.](http://maven.apache.org/plugins/maven-site-plugin/maven-3.html#Version_Resolution)

hansjoachim:
Ok, I went back and double-checked this now. 

It appears that the warnings I got were wrong and fixed in a later maven release (I was running 3.0.5). See the https://issues.apache.org/jira/browse/MSITE-443?focusedCommentId=14451732&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-14451732 and the next comment.

I experimented with reverting my changes and it indeed looks like it picks up the version number when running the maven commands, even though it seems to have problems adding the version number to the effective POM. In summary, I believe this pull request would only offer added complexity, so I no longer think it should be merged. 

Note though, the comment above the one I linked to, where someone had problems with getting the version number from pluginManagement to apply in the reporting section. I'm fairly sure I ran into similar issues when I first experimented, and that's why (initially most) some plugin version numbers are placed in properties. Things seems to have been cleaned up a bit since I last touched this part, so it might be easier ways to do this now.

Issue #153, opened 2014-09-12
Allow for filepath patterns when specifying scan directories


scheffelaar:
It would be very convenient if, when specifying which directories should be scanned for vulnerabilities, it would be possible to use glob-style filepath patterns instead of having to specify hard filepaths. As far as I can tell, the CLI version will take the first path specified by a pattern (ie. if you give the argument -s "/*" in a dir with multiple subdirectories it'll only search one of them) while the Jenkins plugin doesn't seem to be able to handle wildcards at all.

This would be useful for automated testing in an environment where filepaths change -- in my case, I'd like to specify filepaths that contain version numbers that would get changed every time someone tags a release.

stevespringett:
It is unclear if you want this enhancement made to the CLI or Jenkins plugin.

For the Jenkins plugin, you can specify valid Jenkins environment variables such as WORKSPACE and BUILD_NUMBER and the plugin will replace the vars with the actual values. The plugin currently does not allow for Ant-style wildcards, but this should be able to be easily added.

jeremylong:
Thank you for pointing out the bug in the CLI. Currently, the only wild
card support was added to support only scanning specific file types. So
that one could (in theory) specify \directory*.jar   and then only JAR
files in the specified directory would be scanned. However, if one
specified the scan directory as just  -s "directory" then all files in all
sub-directories would get scanned. The bug, in the current implementation,
is if you add a single asterisk at the end of the path the behavior is not
what was intended.

In your case, likely specifying the directory without a wild card may work.

Regardless, the single asterisk at the end of the path is generating
unexpected behavior and will be fixed. Additionally, documentation will be
updated.

--Jeremy

scheffelaar:
Steve: Ideally, both. I was mostly trying to point out the current functionality in both plugins as I understood them.

Also, probably worth noting that I was using the CLI plugin on Linux.

Jeremy: Whoops, probably a mediocre example on my part. The wildcard works the same even if it's in the middle of the file path. So, in this structure:

```
basedir -- dirA -- target -- some jars
        |- dirB -- target -- some jars
```

Providing the argument "-s basedir/*/target" will work but will only scan one of the target dirs, presumably which one is first in the resulting set.

stevespringett:
Ant style pattern sets have been added to the Jenkins plugin (1.2.5-SNAPSHOT). This will allow for using various wildcard patterns in the scan path.

jeremylong:
[...]

PR #347, opened 2015-09-12
Use Directoryscanner from the ant package

hansjoachim:
As discussed briefly in #153, this replaces the copy'n'pasted DirectoryScanner and related classes with the one from the ant jar. I didn't get a reply on some of the follow-up questions, so I figured I might present my experiment here and we might discuss it.

This replaces nearly 10.000 lines with less than 10 and seems to work just as fine. It does introduce a dependency for this module (ant is already used elsewhere in the project), so there might be arguments for keeping it the way it is.


awhitford:
Funny, I was looking at this too.  I basically did the same thing -- EXCEPT that I also leveraged the [maven-shade-plugin](https://maven.apache.org/plugins/maven-shade-plugin/shade-mojo.html#minimizeJar) to merge a subset of the `ant` dependency (to satisfy the concern of including the `ant` dependency -- which honestly shouldn't be a big concern if you ask me).

See my PR #354.  Personally I prefer yours since I am not concerned about including an Ant dependency.  But if that does bother you, the Shade plugin can help.

hansjoachim:
Cool :) 

I guess it is up to the maintainer to review whether this should be merged and in that case which approach.

Btw, do you have any suggestions for resources or small examples to look at for the shade plugin? I've read a bit about it, but I don't think I fully grasped it yet. I would especially like how to properly compare before and after to see how it affects the artifacts.

jeremylong:
Thanks! I went back and forth over just included ant as a dependency vs. what I had done. I agree this is cleaner.

awhitford:
The Shade Plugin can basically merge jar files into your jar.  This is often used to avoid packing around a third party jar, or perhaps even embedding a particular version to avoid some kind of conflict.  (It can rename the embedded jar too -- so `org.apache.ant` could be renamed `org.owasp.ant`, for example -- now it can coexist with `org.apache.ant`.)

Besides the [plugin documentation](http://maven.apache.org/plugins/maven-shade-plugin/), personally I inspect the resulting jar to understand if the configuration is doing what I want.  You can use the `jar` command like:
  `jar tf target/dependency-check-1.3.1-SNAPSHOT.jar`

Or since a `jar` is also really a `zip` file, you can use an archive utility like 7-Zip to inspect.

_Hope this helps!_

PR #349, opened 2015-09-12
Fixes various warnings

hansjoachim:
Fixes some warnings in the code base.

See the diff or individual commits for details.

Added a TODO when I found a commented out assertion. The test failed when I tried to enable it, but I am not sure whether it should pass or not...


jeremylong:
Thanks for helping out with some of this cleanup!

--Jeremy

hansjoachim:
No problem :)

Now that #347 has been merged, it should be easier to clean up things since I no longer need to "filter out" the copy-pasted parts. Expect more in the future. ;)

PR #351, opened 2015-09-12
Added optional merge property support via a database dialect.

awhitford:
Support for the `merge` command is dependent on the [database dialect](https://en.wikipedia.org/wiki/Merge_%28SQL%29).
As a result, we need to extend the Database commands depending on the database.

This adds a dbStatements file for H2.  The MERGE_PROPERTY is optionally specified; if so, it is used and the database is updated with the `merge` command (or theoretically an upsert -- if more dialects are supported).  Otherwise the traditional update/insert strategy is employed.

Note that saveProperties method was broken (insert was never executed) and, since it was also unused, I chose to remove it (rather than fix it).

_(I know that this may be a little crazy or controversial.  Curious if you'll find this approach acceptable.)_

### DIFF
--- a/dependency-check-core/src/main/java/org/owasp/dependencycheck/data/nvdcve/CveDB.java
+++ b/dependency-check-core/src/main/java/org/owasp/dependencycheck/data/nvdcve/CveDB.java
@@ -29,8 +29,10 @@ import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
+import java.util.Locale;
 import java.util.Map;
 import java.util.Map.Entry;
+import java.util.MissingResourceException;
 import java.util.Properties;
 import java.util.ResourceBundle;
 import java.util.Set;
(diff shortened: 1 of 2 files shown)


jeremylong:
I really like the way you implemented this - much cleaner then what I had planed. Before merging this I will add an Oracle and MySQL dialect. I do see one small problem in that the ON DUPLICATE UPDATE may take more parameters... I may have to build a stored proc for MySQL.

awhitford:
Would a trick like this help?  [Reusing an anonymous parameter in a prepared statement](http://stackoverflow.com/questions/2924942/reusing-an-anonymous-parameter-in-a-prepared-statement)

Alternatively [this](http://stackoverflow.com/questions/2309970/named-parameters-in-jdbc) points to [NamedParameterJdbcTemplate](http://docs.spring.io/spring/docs/2.5.x/api/org/springframework/jdbc/core/namedparam/NamedParameterJdbcTemplate.html).

PR #353, opened 2015-09-13
Replaced StringBuffer with more efficient StringBuilder.

awhitford:
`StringBuffer` is synchronized, while `StringBuilder` is not.
`StringBuilder` is designed to be a drop-in replacement for `StringBuffer`.
Since synchronization is not an issue, `StringBuilder` should be used.


hansjoachim:
LGTM :+1:

jeremylong:
Some of those are in the copied Ant code - maybe we should commit back to them ;)

--Jeremy

PR #354, opened 2015-09-13
Replace copied Ant source code with a shaded dependency

awhitford:
Copying the Ant code creates code management issues (as outlined in issue #153).  For example, it is hard to understand what precisely was changed and it is difficult to merge future changes.

Rather than copying the Ant source code, this simply declares the ant library as a dependency to the CLI module.  To satisfy the concern of including Ant as a dependency, the [maven-shade-plugin](https://maven.apache.org/plugins/maven-shade-plugin/shade-mojo.html#minimizeJar) is leveraged to merge necessary Ant classes into the final jar.

This is admittedly not as aggressive as the hand coding, but more manageable over time.


awhitford:
Oops...  It looks like I branched off another branch instead of master.  As a result the first 2 commits are really a separate pull request -- and very minor.  But it was looking closer at the Utils module that made me ponder this whole including Ant thing.  I see there is another PR #347 that is very similar (and simpler).  So, not sure which one you'd prefer.

jeremylong:
I went with 347. I did go back and remove the deprecated methods that you pointed out.

PR #357, opened 2015-09-15
Fixed Maven warning messages 

dwvisser:
…by providing a version for maven-dependency-plugin.


hansjoachim:
Which version of Maven are you running? I recently submitted a similar fix, but it looks like the version number from pluginManagement is actually used and that the warning is due to a bug in maven. See #346 for more info.

dwvisser:
Ahh.. I was unaware. I'll close this out then.

PR #359, opened 2015-09-21
Update to 1.3.1 and fix 'apply-plugin' name

jreckner:
I think the 1.3.1 update was missed in the gradle build file.

Also the readme had a small typo on one of the `apply-plugin:` lines 

@jeremylong

### DIFF
--- a/dependency-check-gradle/README.md
+++ b/dependency-check-gradle/README.md
@@ -30,7 +30,7 @@ buildscript {
     }
 }
 
-apply plugin: 'dependency.check'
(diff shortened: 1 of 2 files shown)


jeremylong:
Sorry about the delay in the gradle release. I am not actually handling this myself (similar to the Jenkins plugin) - sometimes there is a small gap between the updates for these two. I have merged Wei Ma's PR #38 which contained everything you had in 359 plus a few more things.

Thanks - and I'll try to keep the release in sync better.

PR #370, opened 2015-10-09
Do not need or want to call toString for a logger parameter.

awhitford:
Any logger parameter will be turned into a `String`, but you want that to happen by the `debug` method.  If _Debug logging_ is not enabled, then the overhead of calling `toString` is avoided.


colezlaw:
Arrrgghh!! When I moved everything over to SLF4J from JULI, I thought I was very careful to do that. I spent days tracking them all down.

Thanks for catching what I missed.

PR #390, opened 2015-10-22
Option to skip Nvd check (cveValidForHours)

awhitford:
In response to my [Runtime Performance Observations](https://groups.google.com/forum/#!topic/dependency-check/5FDw9imKGQ0) post, this PR adds an option to skip redundant, repetitive (and expensive) checks for NVD updates.

There is a new, optional setting called `cve.check.validforhours`.  This is exposed in the Maven plugin as `cveValidForHours`.  If it is not set, it is business as usual -- same behavior.  But, if the value is positive, then an NVD check done with the specified hours will be skip the NVD check and just proceed with the analysis.

### DIFF
--- a/dependency-check-core/src/main/java/org/owasp/dependencycheck/data/nvdcve/DatabaseProperties.java
+++ b/dependency-check-core/src/main/java/org/owasp/dependencycheck/data/nvdcve/DatabaseProperties.java
@@ -46,6 +46,10 @@ public class DatabaseProperties {
      */
     public static final String MODIFIED = "Modified";
     /**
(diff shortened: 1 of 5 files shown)


awhitford:
_This is huge!_  I am very excited to finally see this in action because the constant queries against NIST really add up to wasted time.

If you are changing the default value of 4 hours, is it being disabled/enabled during unit tests?  Because when _this_ project builds, it needs to exercise the NIST download (and ideally it should exercise the skip logic too).

jeremylong:
Thanks for pointing that out. I updated the test properties to use 0 hours.

PR #391, opened 2015-10-22
Change CpeUpdater to use CPE_MODIFIED_VALID_FOR_DAYS

awhitford:
Shouldn't CPE Updater use the `CPE_MODIFIED_VALID_FOR_DAYS` setting (not CVE)?


jeremylong:
Thanks for catching this.  Note, this is not actually used ATM; it has been left in incase we decide to use some of the data from the CPE dictionary to improve matching.

Issue #393, opened 2015-10-24
Execution broken with latest updates


awhitford:
I just pulled the latest code and now using it breaks.

Looks like the _deprecated_ `aggregate` parameter is _null_, then line 355 of `BaseDependencyCheckMojo` is attempting a boolean check:

```
    private void validateAggregate() throws MojoExecutionException {
        if (aggregate) {
            final String msg = "Aggregate configuration detected - as of dependency-check 1.2.8 this no longer supported. "
                    + "Please use the aggregate goal instead.";
            throw new MojoExecutionException(msg);
        }
    }
```

Why not simply remove the _deprecated_ `aggregate` parameter?

Alternatively, it can be defaulted to `false`.

jeremylong:
Sorry about that, I mised two coding tasks into a single commit and obviously failed to test the changes to aggregate - I added a null check. This was left in to avoid breaking builds when people upgraded to a new version of dependency-check.

PR #394, opened 2015-10-24
Deprecated aggregate is truly optional

awhitford:
If aggregate is truly optional, need to do a null check otherwise risk an NPE.
See issue #393 -- looks like a bug was introduced by PR #390.


jeremylong:
I added the null check in a previous commit.

PR #396, opened 2015-10-27
Upgraded plugins.

awhitford:


pom.xml
```diff
@@ -224,7 +224,7 @@ Copyright (c) 2012 - Jeremy Long
                 <plugin>
                     <groupId>org.apache.maven.plugins</groupId>
                     <artifactId>maven-surefire-plugin</artifactId>
-                    <version>2.18.1</version>
+                    <version>2.19</version>
```
jeremylong, line 227: I reverted this change previously because I was having trouble with debugging via my IDE.

--Jeremy

awhitford:
Oh, that's weird.  I was surprised to see 2.19 elsewhere and just this one line had 2.18.1...

What IDE are you using?
Did you have a particular error message or stack trace?

jeremylong:
I keep switching back and forth between Netbeans and Intellij.  I couldn't get Netbeans to attach when using 2.19.

awhitford:
I think there might be a bug with Maven and Profiles because `mvn versions:display-plugin-updates` lists some odd results for the `core` module:

```
[WARNING] The following plugins do not have their version specified:
[WARNING]   maven-failsafe-plugin ..................................... 2.18.1
[WARNING]   maven-surefire-plugin ...................... (from super-pom) 2.10
[WARNING]   maven-surefire-report-plugin ................................ 2.19
```

I can't really explain that 2.10 version.

colezlaw:
The Super POM is the POM that all POM's inherit from - so if a plugin in
your POM doesn't specify a version, or if it's a plugin that's "on by
default", its version is specified from the Super POM. I don't recall
precisely where it's located, but I _think_ it's in
lib/maven-model-builder-<version>.jar - the
org/apache/maven/model/pom-4.0.0.pom is the super pom. Either that pom or
one of its dependencies will specify maven-surefire-plugin and its version.
If you look at
https://maven.apache.org/guides/introduction/introduction-to-the-pom.html#Super_POM,
it has their example ones. If you've got an Eclipse-based IDE (Eclipse,
STS, etc.) and open your pom and then look at the effective pom tab, it'll
show what gets resolved as your pom, although it doesn't specify where
everything got declared to put it in that state.

awhitford:
The command `mvn help:effective-pom -Doutput=effpom.xml` will generate the _effective pom_.  Version 2.10 of Surefire is not mentioned anywhere, and the code is correctly specifying the version in `pluginManagement` by the project's top level pom -- that should be enough.

The 2.10 version is coming from the `profile` definitions (see `MySQL-IntegrationTest` and `Postgresql-IntegrationTest`).  I suspect that since the profiles are not activated, the snippets aren't reflecting the updated versions from `pluginManagement` (and thus shows the super pom).  If this is the case, then this isn't really a problem...

The real issue is:  what debugging problem is Jeremy experiencing with his IDE?

When I originally queried their JIRA, I didn't see any known issues.  But now, I see a bunch and it looks like a 2.19.1 release may be imminent:

https://issues.apache.org/jira/issues/?jql=project%20%3D%20SUREFIRE%20AND%20fixVersion%20%3D%202.19.1

jeremylong:
Merged #435 instead.

PR #398, opened 2015-10-28
Refactored settings code.

awhitford:
Added a utility function (`setSettingsString`) to avoid repetitive logic.


jeremylong:
Great idea - but this coding pattern exists in other locations as well. I think we should add a setStringIfNotNull(String), setIntIfNotNull(Integer), and setBooleanIfNotNull(Boolean) to the Settings class. 

Thoughts?

Jeremy

awhitford:
Sure...  Centralizing this logic into Settings sounds wise.

awhitford:
Please see PR #402.  This PR is now obsolete.

PR #407, opened 2015-11-07
Upgraded dependencies.

awhitford:


hansjoachim:
lgtm :)

When testing this locally, I also saw that slf4j has a newer version, so I've added that in #411.

I don't know if this is the right place, but a bit more general discussion:
Regarding upgrades, currently dependencies built for Java 7 or newer has been avoided since the Jenkins-plugin need to support Java 6. I noticed that the recent LTS release of Jenkins, 1.625.1 (http://jenkins-ci.org/changelog-stable), now requires Java 7. I don't expect this to change much in the short-term since the Dependency-Check plugin is also useful for older Jenkins releases and it makes sense to support them. However if it at some point in the future becomes relevant for the plugin to require a newer version of Jenkins (due to API or other features), it's nice to know. I don't think the Java 6 requirement is holding us back much though, all I can really remember is that Lucene was rolled back to an earlier version.

PR #408, opened 2015-11-07
Closes #363 - only verify artifacts used in main application #363

### LINKED ISSUE #363: Gradle plug-in: only verify artifacts used in main application
We currently use the Gradle dependencyCheck plug-in for an Android project. The problem is that the checker flags libraries with CVE's that are only used by unit test plug-ins, they don't end up in the actual application.

It would be nice if we could configure the dependency check plug-in so that it only verifies the artifacts that actually end up in the application.

willis7:
This enhancement has not restricted the analysis to only artefacts used by the main application as per the request. Instead it gives the user the option to specify what to analyse, whilst still defaulting to the current behaviour of analysing all. I have added integration tests to show this behaviour.

Additionally, I have added a Gradle wrapper to ensure all future contributors use the same version of Gradle for development.


jeremylong:
I was going down a completely different path when you posted this PR (and then I had a ton of travel and was unable to respond). I added a `skipTestGroups` options then when set to true would skip any group that started with 'test'.

I think there might be value in both solutions, yours that limits to a specific configuration and mine that removes test scoped groups. Any chance you could resolve the merge conflicts?

Thanks!

--Jeremy

willis7:
Ahh yes, that is a cool option to have. I will look at the conflicts and reply in due course.

willis7:
I see you have renamed a few files, which is making the conflict merging a little ugly. Given you have addressed the immediate concern with your change, shall we just reject and close this PR?

jeremylong:
Sure, we can close this.

Issue #419, opened 2015-12-14
Duplicate CPE Index Created message (possible leak)


awhitford:
I noticed this output from the Maven plugin:

```
[INFO] Checking for updates
[INFO] Skipping NVD check since last check was within 12 hours.
[INFO] Check for updates complete (1123 ms)
[INFO] Analysis Starting
[INFO] Creating the CPE Index
[INFO] CPE Index Created (1721 ms)
[INFO] Creating the CPE Index
[INFO] CPE Index Created (0 ms)
[INFO] Analysis Complete (40638 ms)
```

You will notice that the `CPE Index Created` message appears twice.  I think this is because the `CPEAnalyzer` class calls `open` twice -- once during `initialize` and then again during an `open`.

I'm going to guess that the `open` call from `initialize` may be removed.  If the call is truly redundant, resources may be leaked (like `cve`).

PR #420, opened 2015-12-14
Issue #419 - Avoiding a duplicate CPE Index Created message

awhitford:
And avoid a resource leak.  Removing `open` from `initialize` didn't work, but avoiding a double-open seems to be a logical solution.


jeremylong:
That is so obvious after looking at it - I can't believe I didn't have the conditional in my orginal code.

Thank you!

Issue #426, opened 2015-12-21
The EngineVersionCheck.shouldUpdate() deletes the value of the CURRENT_ENGINE_RELEASE


rrialq:
When the value of the CURRENT_ENGINE_RELEASE property is the same as the value of the updateToVersion the method updates the property with the empty string value.
This seems no sense, since this property is used to compare against the last release readed from web.
The problem is that the value is overwritten with an empty string, because when no new version is available the code is executed one of twice.
1. When properties does not match then CURRENT_ENGINE_RELEASE property is updated.
2. Next time the properties match, so the CURRENT_ENGINE_RELEASE property is overwritten with the empty string (I think has no sense).
3. Next time properties does not match, so CURRENT_ENGINE_RELEASE property is updated.

In OracleDB there is no difference between empty strings and null: it stores a null string. After that when the plugin read the records in properties table and stores its values in HashMap it fails with a NullPointerException.

jeremylong:
This should now be resolved in the current snapshot.

PR #427, opened 2015-12-21
Solves the ISSUE #426. No need to update the CURRENT_ENGINE_RELEASE p…

rrialq:
…roperty when it has value.

In current code when the CURRENT_ENGINE_RELEASE contains the same version as in remote web, it overwrites
the property with the empty string, but in OracleDB there is no difference between empty string and null,
and it stores a null value, so when the plugin read all properties and it tries to store them in a HashMap,
the plugin fails with a NullPointerException.

And the code executes one of twice for the same version, when there is no need to save it until it changes.


jeremylong:
While this would solve the NPE; I believe it would break the intended functionality. Instead, we would need to implement deleteProperty.

jeremylong:
I incorporated similiar changes via a recent commit. Hopefully, you will not run into NPE issues with the Oracle database dialect.

PR #432, opened 2015-12-28
Code tweaks 2015-12-28

awhitford:
Noticed a few minor opportunities to improve the code.

### DIFF
--- a/dependency-check-core/src/main/java/org/owasp/dependencycheck/data/update/nvd/NvdCve12Handler.java
+++ b/dependency-check-core/src/main/java/org/owasp/dependencycheck/data/update/nvd/NvdCve12Handler.java
@@ -99,7 +99,6 @@ public class NvdCve12Handler extends DefaultHandler {
                 software = null;
             }
         } else if (!skip && current.isProdNode()) {
-
             vendor = attributes.getValue("vendor");
             product = attributes.getValue("name");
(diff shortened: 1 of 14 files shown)


dependency-check-core/src/main/java/org/owasp/dependencycheck/data/update/nvd/NvdCve12Handler.java
```diff
@@ -112,15 +111,16 @@ public void startElement(String uri, String localName, String qName, Attributes
                 /*yes yes, this may not actually be an "a" - it could be an OS, etc. but for our
                  purposes this is good enough as we won't use this if we don't find a corresponding "a"
                  in the nvd cve 2.0. */
-                String cpe = "cpe:/a:" + vendor + ":" + product;
+                final StringBuilder cpe = new StringBuilder();
```
jeremylong: Consider changing this to:

new StringBuilder(8+vendor.length()+product.length()+(edition==null?0:1+edition.length()));

awhitford:
What about `num`?

Maybe just simplify to:  `20 + vendor.length() + product.length()` ?

Will `num` and `edition` plus the constant characters exceed 20 characters?

Note that it was the `String += String` pattern that I was finding objectionable and inefficient.

I committed another adjustment along the lines of your suggestion.

PR #448, opened 2016-02-06
Upgraded SLF4J to 1.7.14 and the maven-compiler-plugin to 3.5.

awhitford:


awhitford:
BTW...  Not sure what the apprehension is to upgrade `commons-lang3` to version **3.4** (from 3.3.2).  The comment says there is a Jenkins issue, but that sounds odd to me.

stevespringett:
@awhitford The apprehension is that the dependencies used in core and util modules cannot be newer than the dependencies in Jenkins analysis-core plugin. Refer to https://github.com/jenkinsci/analysis-core-plugin/blob/master/pom.xml for it's dependencies.

hansjoachim:
Looks good. [slf4j 1.7.16 is out now](http://www.slf4j.org/news.html), so might want to switch to that one before merging?

PR #473, opened 2016-03-20
Handle bundle-audit not available case and fix RubyBundleAuditAnalyzer test cases

biancajiang:
changes:
1. Fixed RubyBundleAuditAnalyzerTest by adding the missing Gemfile\* files.
2. Added error logging from bundle-audit in RubyBundleAuditAnalyzer.
3. Disable RubyBundleAuditAnalyzer if it failed to initialize, so that RubyGemspecAnalyzer can be kept enabled.
4. New test case for bundle-audit not available case.


jeremylong:
Thanks!

PR #490, opened 2016-04-25
Fixed CVSS for Ruby.

### LINKED ISSUE #485: Incorrect CVSS score in DependencyCheck generated report
Hi, we are experiencing an issue with incorrect cvss scores in the report that Dependency Check generates (both XML and HTML). We are running DependencyCheck on a very simple Sinatra app. Bare reproduction is possible with just the `Gemfile` and `Gemfile.lock`. DependencyCheck detects that our app is vulnerable to CVE-2015-3225. On the NVD website, this [CVE](https://web.nvd.nist.gov/view/vuln/detail?vulnId=CVE-2015-3225) is stated to have a CVSS v2 base score of `5.0`. However, the DependencyCheck generated report states that this CVE has a `cvssScore` of `-1.0`. What is the reason for this discrepancy? Are we utilizing DependencyCheck incorrectly?

Command used to run dependency check:

``` bash
./dependency-check/bin/dependency-check.sh --project ExampleApp --scan sinatra/ --out . --format XML
```

Reproduction with the following `Gemfile` and `Gemfile.lock` in the `sinatra` directory:

`Gemfile`:

```
# encoding: utf-8
source 'https://rubygems.org'

gem 'sinatra'
```

`Gemfile.lock`:

```
GEM
  remote: https://rubygems.org/
  specs:
    rack (1.5.2)
    rack-protection (1.5.2)
      rack
    sinatra (1.4.4)
      rack (~> 1.4)
      rack-protection (~> 1.4)
      tilt (~> 1.3, >= 1.3.4)
    tilt (1.4.1)

PLATFORMS
  ruby

DEPENDENCIES
  sinatra
```

Generated XML Report:
https://gist.github.com/RochesterinNYC/adfb8d2e7507ff823fbcdc759bc9487c

Running `bundle-audit update` does not change the generated cvss score.

geramirez:
this bug was discovered when scanning ruby applications and getting back
`-1` cvss. this turns out to be a problem with bundle-audit cve
database.

Our solution was to use the NVD database, which dependency check uses to
get the CVSS scores for Ruby only if the Criticality is missing from
bundle-audit output. 

Keep in mind there are compilation errors with the
commit atm.

Fixes #485

Signed-off-by: Gabriel Ramirez gabriel.e.ramirez@gmail.com

### DIFF
--- a/dependency-check-core/src/main/java/org/owasp/dependencycheck/analyzer/RubyBundleAuditAnalyzer.java
+++ b/dependency-check-core/src/main/java/org/owasp/dependencycheck/analyzer/RubyBundleAuditAnalyzer.java
@@ -20,6 +20,7 @@ package org.owasp.dependencycheck.analyzer;
 import org.apache.commons.io.FileUtils;
 import org.owasp.dependencycheck.Engine;
 import org.owasp.dependencycheck.analyzer.exception.AnalysisException;
+import org.owasp.dependencycheck.data.nvdcve.CveDB;
 import org.owasp.dependencycheck.dependency.Confidence;
 import org.owasp.dependencycheck.dependency.Dependency;
 import org.owasp.dependencycheck.dependency.Reference;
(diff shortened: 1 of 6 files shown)


jeremylong:
I fixed the compile issues and finished the code changes... However, I did notice one thing - see the TODO in the RubyBundleAuditAnalyzer. Should we just always look up the vulnerability from the NVD even if we have a "score" from bundle audit? Our final report would have a more accurate CVSS score wouldn't it?

jvshahid:
@jeremylong i think so. It wasn't clear why a decision was made to map the criticality text to a numerical value. This mapping can probably happen if the cve is missing from the db. thoughts ?

jvshahid:
@jeremylong any plans to cut a new release with this fix ?

jvshahid:
ping

jeremylong:
Soon..  life/work happend.  Finishing up another update and should release
within a week.

Jeremy

PR #491, opened 2016-04-27
MSSQL Support

mwieczorek:
Just a script for MSSQL Server.


jeremylong:
I'm assuming you've tested dependency-check on MS SQL Server...  My only question would be do you think a dialect SQL file should also be created?

https://github.com/jeremylong/DependencyCheck/tree/master/dependency-check-core/src/main/resources/data

--Jeremy

mwieczorek:
Yes, I run the script, then run dependency check against the DB:  first in 'update mode', then 'scan mode'. Everything looks good.

About the dialect file: I looked at the dbStatements.properties and I think every statement is ok for MSSQL. In other dialect files I see 'MERGE_PROPERTY'  which is calling some stored procedure. 
But I saw in code that even if it's missing, everything will work fine.

Let me know if I should add MERGE_PROPERTY.

jeremylong:
Thanks for the PR

PR #494, opened 2016-05-03
 Align documentation with current project name specification flag

erikerikson:
When using the "--app" flag, the following warning is produced:

 [WARN] The 'app' argument should no longer be used; use 'project' instead.

 This change updates the documentation from suggesting "--app" to "--project"


jeremylong:
Thanks!

Issue #476, opened 2016-03-21
RubyBundleAuditAnalyzer lost original file path


biancajiang:
Use RubyBundleAuditAnalyzer to analyze Gemfile.lock will lose the file's original path b/c a temp file is created for each vulnerability found and its path in the temp dir is used.  Eg. from RubyBundleAuditAnalyzerTest.testAnalysis(), the returned dependency looks like:

Dependency{ 
**fileName**='Gemfile-redcarpet678732797223096147.lock',
**actualFilePath**='/var/folders/s4/_lmbxxyd75s4jysxry9qnskr0000gn/T/dctemp/Gemfile-redcarpet678732797223096147.lock',
**filePath**='/var/folders/s4/_lmbxxyd75s4jysxry9qnskr0000gn/T/dctemp/Gemfile-redcarpet678732797223096147.lock'}

I can see why the temp file path is used for more desirable checksum etc.  But shouldn't the original Gemfile.lock file path and name also be preserved?  Like the ArchiveAnalyzer will preserve the logical path of elements inside compressed files with **fileName** and **filePath**, and use **actualFilePath** for the temporarily decompressed path in temp dir.  In this case, it's a gem found inside Gemfile.lock file.  But the real path is not available in any attributes of the dependency.

biancajiang:
Any preference using one path attribute over another?  I'm happy to provide a PR for this.

jeremylong:
Again, I agree with you. The actualFilePath should not be changed; it is not used in the report. But the others, as you have pointed out, can and should be updated to provide the additional information. If you have time to work on this great.

--Jeremt

Issue #471, opened 2016-03-17
RubyGemspecAnalyzer doesn't work for Rakefile?


biancajiang:
Even RubyGemspecAnalyzer.java has a FILTER to support both .gemspec and Rakefile:
private static final FileFilter FILTER
            = FileFilterBuilder.newInstance().addExtensions(GEMSPEC).addFilenames("Rakefile").build();

Its analyzeFileType() only uses patterns for .gemspec for all cases:

```
    final Matcher matcher = GEMSPEC_BLOCK_INIT.matcher(contents);  //line 118
    if (matcher.find()) {
    ...
```

So matcher.find() is always false for Rakefile.

Is this an overlook?

jeremylong:
I did not author this plugin - I believe including the Rakefile was in early testing but there are issues with identifying the exact version used. The gemspec file contains more accurate informattion. We likely need to exclude the rakefile from the FileFilter.

PR #497, opened 2016-05-06
Ruby dependency enhancement

### LINKED ISSUE #477: RubyBundleAuditAnalyzer only report dependencies with vulnerabilities
RubyBundleAuditAnalyzer works great to report vulnerabilities.  But it ignores other dependencies.

RubyGemspecAnalyzer, FileNameAnalyzer(?) and other analyzers tries to report on all dependencies it can find.  

So for Ruby, RubyBundleAuditAnalyzer and RubyGemspecAnalyzer compliments each other, rather than overlap.  RubyGemspecAnalyzer also provides Vendor evidences (that we use) while RubyBundleAuditAnalyzer does not.

So instead of disable RubyGemspecAnalyzer when bundle-audit is available, maybe the better thing to do is to make RubyBundleAuditAnalyzer precedes RubyGemspecAnalyzer.  So that when Evidence from both exist, RubyBundleAuditAnalyzer can take precedence.   Similar to Jar file of a maven package:  when multiple evidence from multiple analyzers exist, the order of sources is consistent - central always tops the list.

If this sounds reasonable, I'd be happy to provide a PR for it.

biancajiang:
Enhancement to Ruby dependencies, this should resolve #477, #476, #471.  Main changes:
1. new RubyBundlerAnalyzer with RubyBundlerAnalyzer:

This analyzer accepts the fully resolved .gemspec created by the Ruby bundler (http://bundler.io) during install for better evidence results. It also tries to resolve the dependency packagePath to where the gem is actually installed. Then during {@link AnalysisPhase.PRE_FINDING_ANALYSIS}  {@link DependencyBundlingAnalyzer} will merge two .gemspec (one original, one from bundle install) dependencies together if  <code>Dependency.getPackagePath()</code> are the same.

Rationale:
Ruby bundler creates new .gemspec files under a folder called "specifications" at deploy time, in addition to the original .gemspec files from source. The bundler generated  .gemspec files always contain fully resolved attributes thus provide more accurate  evidences, whereas the original .gemspec from source often contain variables for attributes that can't be used for evidences.
1. RubyBundlerAnalyzer will also be disabled if RubyBundleAuditAnalyzer is working.
2. new field Dependency.packagePath for the directory of a package.  By default this is the same as filePath.  But for cases of packagePath is different than the filePath, this is very useful.
3. Rakefile is not currently supported yet.  commented out with place holder for future support and test case.


biancajiang:
There are also a bunch of enhancements to Ruby gem evidences.

PR #502, opened 2016-05-25
Update initialize_mysql.sql

xthk:
lower cased "properties" in UPDATE statement

I get an error with MySQL 5.5.48-37.8 on executing this file:

```
ERROR 1146 (42S02) at line 57: Table 'dependencycheck.Properties' doesn't exist
```

It seems, this is just an upper/lower case issue, which is easy to change.


jeremylong:
thanks for reporting the issue and the PR.

PR #507, opened 2016-06-03
Use batch updates where possible

jabbrwcky:
Use batch update for references and vulnerable software if supported by DB.


jeremylong:
Thanks for the PR!

PR #516, opened 2016-06-14
Correctly apply weightings when searching for CPEs

msrb:
The signature of `searchCPE` method is:
`searchCPE(String vendor, String product, Set<String> vendorWeightings, Set<String> productWeightings)`

Therefore it seems that the parameters are in incorrect order now.


jeremylong:
Thank you for noticing that; truly appreciate the PR.

PR #526, opened 2016-07-06
Updating gradle dependencyCheck documentation

nicolastrres:
I was trying to execute the gradle example, but I was getting this error:
`Could not get unknown property 'ALL' for object of type org.owasp.dependencycheck.gradle.extension.CheckExtension.`

This error is because the quotes are missing on the example.


jeremylong:
Thank you for the pointing this out and the PR! Truly appreciate it.

--Jeremy

PR #528, opened 2016-07-08
Compatibility fixes for MariaDB JDBC driver

felfert:
This pull request adds compatibility for MariaDB's JDBC driver.

Background:
While most JDBC drivers silently accept a conn.prepareCall(sql) with sql **not** containing "? = CALL(...)" (and return a PreparedStatement instead), MariaDB's driver is a little bit more restrictive and throws an exception in this case.

MariaBD developers consider this not an error, but actually an improvement.
See: https://jira.mariadb.org/browse/CONJ-63

Fortunately, DependencyCheck does **not** need CallableStatements anywhere, because none of the called functions return any value. Thus using a CallableStatement is superfluous anyway.

Rule of Thumb:
Use CallableStatement cs = conn.prepareCall(...) **only** if the sql actually uses an OUT parameter.

### DIFF
--- a/dependency-check-core/src/main/java/org/owasp/dependencycheck/data/nvdcve/ConnectionFactory.java
+++ b/dependency-check-core/src/main/java/org/owasp/dependencycheck/data/nvdcve/ConnectionFactory.java
@@ -20,7 +20,7 @@ package org.owasp.dependencycheck.data.nvdcve;
 import java.io.File;
 import java.io.IOException;
 import java.io.InputStream;
-import java.sql.CallableStatement;
+import java.sql.PreparedStatement;
 import java.sql.Connection;
 import java.sql.Driver;
 import java.sql.DriverManager;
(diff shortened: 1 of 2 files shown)


felfert:
1. Reasoning for using MariaDB's JDBC instead of mysql's JDBC: Less restrictive License. LGPL(MariaDB) vs. GPL(mysql)
2. Example pom.xml using MariaDB's JDBC driver:

``` xml
<project>
    <modelVersion>4.0.0</modelVersion>
    <groupId>dummy</groupId>
    <artifactId>dummy</artifactId>
    <version>1.0-SNAPSHOT</version>
    <build>
        <plugins>
            <plugin>
                <groupId>org.owasp</groupId>
                <version>1.4.1-SNAPSHOT</version>
                <dependencies>
                    <dependency>
                        <groupId>org.mariadb.jdbc</groupId>
                        <artifactId>mariadb-java-client</artifactId>
                        <version>1.4.6</version>
                    </dependency>
                </dependencies>
                <configuration>
                    <databaseDriverName>org.mariadb.jdbc.Driver</databaseDriverName>
                    <connectionString>jdbc:mariadb://my.cvedb.host/cvedb</connectionString>
                    <databaseUser>depscan</databaseUser>
                    <databasePassword>NotReallyMyDbPassword</databasePassword>
                </configuration>
                <executions>
                    <execution>
                        <goals>
                            <goal>update-only</goal>
                        </goals>
                    </execution>
                </executions>
            </plugin>
        </plugins>
    </build>
</project>
```

PR #539, opened 2016-08-12
filter out version from jar filename for package name

biancajiang:
A small improvement to FileNameAnalyzer:  if there's version number identified in file name, filter that out for the package name to get more accurate names.  e.g.  "library-name-1.4.1r2-release.jar", version number was identified, use "library-name" as the package name instead of the full file name "library-name-1.4.1r2-release".

Changes:
1. added a new function DependencyVersionUtil#parsePreVersion() to parse the package name if there's version number already identified.
2. use the new function in FileNameAnalyzer#analyze() to set a more accurate name if available.


biancajiang:
Adding a small fix to JarAnalyzer.java to use value of specification-version as version from Manifest:
https://github.com/jeremylong/DependencyCheck/pull/539/commits/8cd377b99f857772277a452a11134227507a14c3#diff-39ea9cf3596adff34a7d5454bb2ab91aL688

PR #542, opened 2016-08-14
document skipConfigurations and scanConfigurations for gradle plugin

bodewig:
closes jeremylong/dependency-check-gradle/#12


jeremylong:
Thanks for the PR!

PR #543, opened 2016-08-15
Add Dockerfile for dependency check

ryrizo:
Thought you might be interested in this. I've been working on a docker container that scans other containers and had success getting dependency check working. This is a slimmed down version of the Dockerfile I use. 

**Usage:**
`git clone https://github.com/jeremylong/DependencyCheck`
`cd DependencyCheck`
`docker build . -t dependency-check`
`docker run dependency-check <options>`

There are several options for getting the code into the container, the easiest of which is mounting it as a volume.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,33 @@
+#
+# OpenJDK Java 7 JDK Dockerfile
+#
+# https://github.com/dockerfile/java
+# https://github.com/dockerfile/java/tree/master/openjdk-7-jdk
+#
+
+# Pull base image.
+FROM ubuntu:14.04
+
+# Install Java.
+RUN \
+  sed -i 's/# \(.*multiverse$\)/\1/g' /etc/apt/sources.list && \
+  apt-get update && \
+  apt-get -y upgrade && \
+  apt-get install -y build-essential && \
+  apt-get install -y software-properties-common && \
+  apt-get install -y byobu curl git htop man unzip vim wget && \
+  apt-get update && \
+  apt-get install -y openjdk-7-jdk && \
+  rm -rf /var/lib/apt/lists/*
+
+
+# Java Version and other ENV
+ENV JAVA_HOME=/usr/lib/jvm/java-7-openjdk-amd64 \
+    DOWNLOAD=https://bintray.com/jeremy-long/owasp/download_file?file_path=dependency-check-1.4.2-release.zip
+
+RUN curl -L ${DOWNLOAD} -o ./dependency-check.zip && \
+  unzip dependency-check.zip
+
+WORKDIR /dependency-check/bin
+
+ENTRYPOINT ["/bin/sh", "dependency-check.sh"]


RubieV:
Adding @wurstbrot in, as a java 8 based docker image seems better maintainable. 

https://github.com/wurstbrot/Docker-OWASP-Dependency-Check/blob/master/Dockerfile

wurstbrot (commented):
This Dockerfile will run the processe as root user.

wurstbrot:
As a Dockerfile-Manager I would like to download the current release by a link like https://bintray.com/jeremy-long/owasp/download_file?file_path=dependency-check-current-release.zip

In that case, I do not need to check for a new release and each docker user will be able to download the current release. After the automatic build processes in the night.

wurstbrot:
I changed my Dockerfile for public usage. I use it in Jenkins. 
Should I merge it and create a pull request?

In addition it would be nice to have the build under https://hub.docker.com/r/owasp/ but I don't know the procedure in OWASP.

wurstbrot (changes requested):
This Dockerfile will run the processe as root user.
Also I am not sure how to get the report.

ryrizo:
@wurstbrot Thanks for your feedback! It looks like the openjdk image will have longer support than the current base image (java:8). To get the report out of the container, I write it to a mounted volume (looks like you do as well). The report might not be written to a volume when running as non-root [see here](https://forums.docker.com/t/root-user-or-non-root-user-inside-container/966/2).

**Running as root within container**

`A root user within a LXC container cannot (in theory) escalate to be root on the host machine; but many people believe that it is possible to do so. It is certainly harder to do with Docker containers (thanks to the capability restrictions) but if security is a big concern, you should stack up multiple safety mechanisms.` [from here](https://groups.google.com/forum/#!msg/docker-user/e9RkC4y-21E/JOZF8H-PfYsJ)

Although I agree there is potential risk added by running as root within the container, the container is not persistent which reduces the likelihood. Perhaps you can convince me otherwise. 

Let me know your thoughts. I can update this PR to use the openjdk base image and update the dependency check version.

ryrizo:
Also a side note: I made a script to ship results from the container to AWS S3 and SumoLogic. That way I can schedule the job daily and blow away the jenkins workspace each time. So I haven't had to retrieve the report produced by the tool.

wurstbrot:
@ryrizo Here an example why you should use users in docker:
As an attacker I can use docker with running processes as root to place an evil bash script with suid bit in a shared folder (in this case, it doesn't matter how he gained access as we want to apply defence in depth).
Afterwards, an attacker gains (somehow) access to a shell as unprivileged user, so that he can run the script with superuser priviliges and execute any command.

The report is written by using 777 permissions on the mounted folder initaly. After the first run you can change permission to 700.

In your link, you cite the wrong line from my point of view. The line to be cited is:

> The best practice is to combine:
> - running your process as non-privileged user within the containers (docker lets you do that easily)
> - stripping the container from all the potentially dangerous system capabilities (docker does that automatically)
> - running an hardened Linux, with e.g. a grsec-enabled kernel, or with your distro's security module (SELinux, AppArmor...)

ryrizo:
@wurstbrot Good example. I can't argue against best practices and defense in depth. My perspective was more of a risk based decision based on my current mitigations - which isn't the best context for an open source tool. Would you want to submit a new PR and I'll close this one?

wurstbrot:
@RubieV please close this due to #577 .

RubieV:
@ryrizo is the starter of this PR

Issue #552, opened 2016-09-06
org.slf4j.impl.StaticLoggerBinder causes duplicate classes


gertjanal:
A dependency should be used to make LoggerFactoryBinder available instead of a copy of the class

[INFO] --- duplicate-finder-maven-plugin:1.2.1:check (default) @ common-util ---
[INFO] Checking compile classpath
[INFO] Checking runtime classpath
[INFO] Checking test classpath
[WARNING] Found duplicate and different classes in [org.owasp:dependency-check-maven:1.4.2, org.slf4j:slf4j-log4j12:1.7.10]:
[WARNING]   org.slf4j.impl.StaticLoggerBinder
[WARNING]   org.slf4j.impl.StaticLoggerBinder
[WARNING] Found duplicate classes/resources in runtime classpath.
[WARNING] Found duplicate and different classes in [org.owasp:dependency-check-maven:1.4.2, org.slf4j:slf4j-log4j12:1.7.10]:
[WARNING]   org.slf4j.impl.StaticLoggerBinder
[WARNING] Found duplicate classes/resources in test classpath.
[ERROR] Failed to execute goal org.basepom.maven:duplicate-finder-maven-plugin:1.2.1:check (default) on project X: Found duplicate classes/resources! -> [Help 1]

gertjanal:
working on it

gertjanal:
OK, so the StaticLoggerBinder is not the original StaticLoggerBinder. Why would we need a patched version and why is it in a package of a different party?

colezlaw:
`StaticLoggerBinder` is the class that you rewrite for SLF4J to be able to find the logging implementation being used. This is the black magic and witchcraft it uses to prevent other classloader issues. In other words, this class **must** be named this. We wrote our version of it so that the DC internals that log via SLF4J calls make it to the Maven logger.

Evidently, some other plugin you're using also uses SLF4J, so they did the same thing to make their own custom logger.

Is there an option to for Dependency Check to fork?

jeremylong:
Currently there is no option to fork dependency-check (within the plugin itself). The proposed PR is not the correct solution as libraries should not define a [binding](http://www.slf4j.org/codes.html#multiple_bindings). However, @colezlaw isn't that what we've done by defining a `StaticLoggerBinder`? Instead of adding a logging binding via the pom we've added one via code. That being said, I'm not sure how else one would route all logging from a library to the base Maven logging mechanism...

jeremylong:
[...]

PR #553, opened 2016-09-06
org.slf4j.impl.StaticLoggerBinder causes duplicate classes

gertjanal:
Removed org.slf4j.impl.StaticLoggerBinder and added proper dependency


gertjanal:
Resolves #552

gertjanal:
OK, so the StaticLoggerBinder is not the original StaticLoggerBinder. Why would we need a patched version and why is it in a package of a different party?

colezlaw:
We can't do this. `StaticLoggerBinder` is a class that **we** provide which tells SLF4J which factory to use for logging implementation. This class is necessary to find the binding to the Maven logger so that internal API's which use SLF4J ultimately make their way into the Maven log.

Does the duplicate class warning cause the build to fail? Or is this just a warning from a plugin that looks for duplicate classes? Can that tool be configured to ignore some duplicates?

colezlaw:
And I understand the duplicate now. There's an adapter for Maven so that internal calls to SLF4J make it to the Maven logger, and we provided an Ant adapter so that internal calls to SLF4J make it to the Ant logger. 

The class that this introduces would force all SLF4J calls to go to Log4J, and would end up being a duplicate at runtime (because the CLI uses a different SLF4J implementation).

jeremylong:
Per the discussion in issue #552 - this PR is not being accepted.

PR #589, opened 2016-10-06
Hardening

pierre-ernst:
- Prevented XXE security vulnerabilities when parsing malicious XML
- Prevented MITM attacks by enforcing https 

This is useful to protect security engineers against developers who want to play tricks against them by giving their security engineers malicious jar files to scan.
I call this scenario "**Revenge P0wn**"
 :-)

### DIFF
--- a/dependency-check-core/src/main/resources/dependencycheck.properties
+++ b/dependency-check-core/src/main/resources/dependencycheck.properties
@@ -4,7 +4,7 @@ autoupdate=true
 max.download.threads=3
 
 # the url to obtain the current engine version from
-engine.version.url=http://jeremylong.github.io/DependencyCheck/current.txt
+engine.version.url=https://jeremylong.github.io/DependencyCheck/current.txt
 
 #temp.directory defaults to System.getProperty("java.io.tmpdir")
 #temp.directory=[path to temp directory]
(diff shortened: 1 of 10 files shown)


dependency-check-core/src/main/resources/dependencycheck.properties
```diff
@@ -62,7 +62,7 @@ cve.url-2.0.base=https://nvd.nist.gov/feeds/xml/cve/nvdcve-2.0-%d.xml.gz
 cve.cpe.startswith.filter=cpe:/a:
 
 cpe.validfordays=30
-cpe.url=http://static.nvd.nist.gov/feeds/xml/cpe/dictionary/official-cpe-dictionary_v2.3.xml.gz
+cpe.url=https://static.nvd.nist.gov/feeds/xml/cpe/dictionary/official-cpe-dictionary_v2.3.xml.gz
```
jeremylong, line 65: While the update is fine - this is not currently used.

dependency-check-core/src/main/resources/dependencycheck.properties
```diff
@@ -4,7 +4,7 @@ autoupdate=true
 max.download.threads=3
 
 # the url to obtain the current engine version from
-engine.version.url=http://jeremylong.github.io/DependencyCheck/current.txt
+engine.version.url=https://jeremylong.github.io/DependencyCheck/current.txt
```
jeremylong, line 7: Thanks for pointing this one out - when I originally wrote this I don't think gh-pages supported HTTPS.  Thanks!

jeremylong (approved):
Thank you for the PR! I truly appreciate the assist!

PR #590, opened 2016-10-08
Cleanup

stefanneuhaus:
- Fix typos
- Provide proper error message in case the (default) property file is not available. Ran into this issue once in combination with my dev setup and the Gradle daemon.

### DIFF
--- a/dependency-check-core/src/main/java/org/owasp/dependencycheck/exception/ExceptionCollection.java
+++ b/dependency-check-core/src/main/java/org/owasp/dependencycheck/exception/ExceptionCollection.java
@@ -25,7 +25,7 @@ import java.util.List;
 /**
  * A collection of several exceptions.
  *
- * @author Jeremy Lomg
+ * @author Jeremy Long
  */
 public class ExceptionCollection extends Exception {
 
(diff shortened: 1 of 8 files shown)


dependency-check-core/src/main/java/org/owasp/dependencycheck/exception/ExceptionCollection.java
```diff
@@ -25,7 +25,7 @@
 /**
  * A collection of several exceptions.
  *
- * @author Jeremy Lomg
```
jeremylong, line 28: well that's embarrassing... I even got my own name wrong ;)

jeremylong (approved):
Thanks! I truly mean this; I really appreciate the time you've put into reviewing my code and providing these updates back to the project.

dependency-check-core/src/main/java/org/owasp/dependencycheck/exception/ExceptionCollection.java
stefanneuhaus: I thought you put it into the _ExceptionCollection_ class on purpose. ;-)

I like supporting projects that deserve it. By saying so I do not mean that your code is crap. ;-) It's rather the opposite: I try to improve things, if I observe an issue in the tools that I like and use myself.

Issue #591, opened 2016-10-09
Feature Request: support parallel execution of analyzers of the same type


stefanneuhaus:
I analyzed the quite bad runtimes for my Gradle project and observed that the CentralAnalyzer was the culprit: it performs a blocking HTTP request for each dependency. For non-trivial projects this easily sums up to a remarkable duration.

To improve one could perform the dependency analyzation tasks of the CentralAnalyzer in parallel. As a generalization one could do so for almost all analyzers (there are a couple of exceptions):
- sequentially process all analyzers
- for each analyzer process the analysis tasks in parallel

Running analyzers _of different types_ in parallel would be much harder, as there a dependencies between the analyzer types.

stefanneuhaus:
I provided a PR as a solution proposal: #592

I attached my sample Gradle build file. Runtimes on my machine using Gradle 3.1 and warmed up Gradle Daemon:
- `1.4.3`: ~9:00 min
- `1.4.4-SNAPSHOT`: 1:29 min (already contains the fixes for [Dependency deduplication](https://github.com/jeremylong/DependencyCheck/commit/44edcabe15ca98b60d98d0fc71d8296e87349436) and [improved test configuration detection](https://github.com/jeremylong/dependency-check-gradle/issues/17))
- `1.4.4-SNAPSHOT` + this PR: 0:22 min

[build.gradle.txt](https://github.com/jeremylong/DependencyCheck/files/518019/build.gradle.txt)

jeremylong:
This is actually something I've been thinking about implementing. I'll spend tomorrow morning reviewing and testing the implementation. While I did have to change the checksum calculations (earlier today) to use a _slightly_ less efficient algorithm - a PR like this one looks like it will make up for the minor performance hit I had to re-introduce.

stevespringett:
@stefanneuhaus Thats a substancial improvement!

jeremylong:
@stefanneuhaus thanks!  #592 was merged

PR #592, opened 2016-10-09
Support parallelism for analyzers of the same type

stefanneuhaus:
- Analyzer declares `boolean supportsParallelProcessing()`.
- If an Analyzer supports prallel processing, it will process its tasks with up to 4 \* #cores threads.
- Fixed some synchronization issues.
- minor changes


stevespringett:
Thanks for the commit. Out of curiosity, do you have any benchmarks that show the performance improvements?

jeremylong:
@stevespringett see issue #591 where initial performance stats have been posted.

PR #596, opened 2016-10-11
Don't reset the H2 DB when it's at a non-standard path

thomasgl-orange:
When using a H2 JDBC address with a custom file path (not the default "%s"), the DB is always erased.  

In the ConnectionFactory, such "jdbc:h2:file:" JDBC addresses are treated as a special case, to create the schema if the h2.db file is not found at the expected path (which is fine). The issue is that, when checking for existence of this file, its path is not extracted from the actual JDBC address, but assumed to be the default one, in the data directory.  Since there is no such file, the schema is re-created.

FYI, my use case for customizing this H2 JDBC address is to get a shared DB for Jenkins analysis.  I'm setting, in Jenkins global configuration, the H2 JDBC address to "jdbc:h2:file:~/something/blah;OPTIONS...", and this way I don't have to modify the data configuration in each job.

### DIFF
--- a/dependency-check-core/src/main/java/org/owasp/dependencycheck/data/nvdcve/ConnectionFactory.java
+++ b/dependency-check-core/src/main/java/org/owasp/dependencycheck/data/nvdcve/ConnectionFactory.java
@@ -134,7 +134,7 @@ public final class ConnectionFactory {
             boolean shouldCreateSchema = false;
             try {
                 if (connectionString.startsWith("jdbc:h2:file:")) { //H2
-                    shouldCreateSchema = !h2DataFileExists();
+                    shouldCreateSchema = !h2DataFileExists(connectionString);
                     LOGGER.debug("Need to create DB Structure: {}", shouldCreateSchema);
                 }
             } catch (IOException ioex) {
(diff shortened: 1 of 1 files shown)


stevespringett:
This commit will likely break Dependency-Track, and the Dependency-Check Jenkins plugin along with all other integrations that already use dependency-check in non-standard paths.

Why does the code not use Settings.getDataDirectory()? Many integrations specify the data directory this way.

thomasgl-orange:
I don't understand.  I'm sure you're rightly smelling some potential issues with this change that I will have overlooked, but I can't figure out what exactly.  Could you be more specific?

All I can see, so far, is an issue in the current ConnectionFactory code:
- on the one hand it is possible to set an arbitrary H2 URL, and thus to choose an arbitrary name and location for the .h2.db file,
- on the other hand, the decision to (re)initialize the H2 DB is made by checking existence of a .h2.db file specifically in the data directory, so basically assuming that the H2 URL has not been customized.

An other way to look at this issue is that, currently, the same logic to build the default H2 filepath (based on the data dir and db name) is duplicated in these two methods:
- `Settings.getConnectionString(String, String)`
- `ConnectionFactory.h2DataFileExists()`

...except that in `Settings.getConnectionString` it's only used if the URL happens to includes a `%s` place-holder, whereas in `ConnectionFactory.h2DataFileExists` it is used unconditionally.
My patch is an attempt at leaving `Settings.getConnectionString()` alone in charge of determining the actual JDBC URL, and making `ConnectionFactory.h2DataFileExists()` simply go with it and trust whatever filepath has been set or substituted in this URL.

As for my use case for hitting this issue, in case it was not clear, it is something like this:
- I'm in Jenkins, using the plugin in version 1.4.3
- I want to share my H2 database between different jobs, to save time/bandwidth (and obviously, I can configure the required locks / exclusions to make this viable)
- one option would be to keep the default H2 URL (`jdbc:h2:file:%s;SOMETHING`) in the Jenkins global configuration, and to set the same data directory in each job configuration. It works fine, and the DB file is kept with its full contents at the end of a job execution, and then reused.  But it is not convenient to setup, because it implies repeating the same specific configuration (the data dir path) in every single job.
- my other option would be to customize the H2 URL once for all in the global Jenkins configuration, with a fixed file path (for instance `jdbc:h2:file:/home/jenkins/depcheck;SOMETHING`), and voilà!  Except it doesn't work, because each time a JDBC connection is created, `h2DataFileExists()` decides the database is missing, so the schema gets re-created, and the next job has to reimport the complete NVD again.

stevespringett:
Thanks for clarifying the issues you're running into and the proposed solution. I think I understand the issue and agree that setting up the data directory for each job can be a pain if you have a lot of jobs. I'm running into the same situation personally.

With this fix, it seems like you've corrected part of the issue. The data directory will still be created in the projects workspace (by default) however. So you'll have duplicate lucene indexes for every job. 

Ideally, you'll want a global configuration in the Jenkins plugin to specify the data directory and then on a per-job basis have the option to override the global config if necessary. I've created a ticket for this and will include it in the next version of the Jenkins plugin.

https://issues.jenkins-ci.org/browse/JENKINS-38938

thomasgl-orange:
Thanks a lot Steve for JENKINS-38938. Now using a custom build of Jenkins dependency-check plugin 1.4.3 + your `globalDataDirectory` patch, and no more custom H2 JDBC URL. It works perfectly fine for my use-case.

I still think there is something to fix in dependency-check itself about `ConnectionFactory.h2DataFileExists()` making too much assumptions on the expected .h2.db file path, but I don't hit this bug anymore, thanks to your changes on the Jenkins plugin.

> So you'll have duplicate lucene indexes for every job. 

Actually, I've not yet seen any Lucene index files in the data directory...  On my Jenkins agents, in these directories, I've always only seen some .h2.db files (and sometimes the corresponding .trace.db files, in case of warning message), that's why I was assuming that me changing the H2 JDBC URL was enough to share the data between jobs.  But maybe it depends on some options or features I've not enabled?

stevespringett:
@thomasgl-orange Glad it's working for you. I've been meaning to add that feature anyway, so this ticket provided motivation to actually get it done.

I'll let @jeremylong work the rest of this pull request.

jeremylong:
@thomasgl-orange the lucene index is an in memory index that gets built during each execution.

As to the path in the connection string, yes I completely understand the concern, but if using the built in H2 database it is far better to use the [`--data`](http://jeremylong.github.io/DependencyCheck/dependency-check-cli/arguments.html) to move the database to another location.

PR #598, opened 2016-10-13
Security issue in documentation of security product is amusing.

ezarko:
Setting any file to be executable as well as writable by other users is a security concern.  Updated documentation.


jeremylong:
Thank you for pointing that out...  Can't believe I put 777. Regardless, that is no longer necessary as the .sh is already marked as executable when packaged. I just removed this from the documentation.

PR #606, opened 2016-11-02
NexusAnalyzer should not ignore analyzer.nexus.proxy

thelateperseus:
NexusSearch honours the analyzer.nexus.proxy setting to identify the jar via a SHA1 hash and receives a URL to download the POM.  NexusAnalyzer then attempts to download the POM ignoring the analyzer.nexus.proxy setting, which fails if Nexus is within the local network (and doesn't require a proxy)


jeremylong:
Thanks for pointing this out. I changed the implementation of the fix so I did not actually use your code - but thanks for pointing out precisely where the issue was!  This will be included in the next release.

thelateperseus:
Much appreciated - this is an excellent tool!

PR #614, opened 2016-11-13
Fix handling of numerical versions

stefanneuhaus:
The proposed fix is to compare parts of the version in an alphanumerical way, if the part contains a leading zero.

### DIFF
--- a/dependency-check-core/src/main/java/org/owasp/dependencycheck/dependency/VulnerableSoftware.java
+++ b/dependency-check-core/src/main/java/org/owasp/dependencycheck/dependency/VulnerableSoftware.java
@@ -226,18 +226,21 @@ public class VulnerableSoftware extends IndexEntry implements Serializable, Comp
 
     /**
      * Determines if the string passed in is a positive integer.
+     * To be counted as a positive integer, the string must only contain 0-9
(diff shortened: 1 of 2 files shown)


jeremylong:
Thanks for pointing this issue out.  I need to spend some time thinking about the proposed fix to consider if there might be consequences...

dependency-check-core/src/main/java/org/owasp/dependencycheck/dependency/VulnerableSoftware.java
```diff
@@ -234,6 +234,13 @@ private static boolean isPositiveInteger(final String str) {
         if (str == null || str.isEmpty()) {
             return false;
         }
+
+        // numbers/versions with leading zeros should not be treated as numbers
+        // (e.g. when comparing "01" <-> "1")
+        if (str.charAt(0) == '0') {
```
jeremylong: This would treat '0' as a string. We need to add a `&& str.length() > 1`

stefanneuhaus: Good catch! I fixed it as proposed.

PR #634, opened 2016-12-29
rework the enabled / disabled logic

hgschmie:
If an analyzer is disabled from the configuration, it should not be
initialized (because some of the may actually fail during that process
nor should the engine log in any way that those exist.

With these changes, it is possible for me to turn off unwanted
analyzers (e.g. Ruby analyzers for a java project) from the maven
plugin and not confuse my users with spurious misleading messages.


jeremylong:
Thanks for the PR.  I'll merge this in the next few days.

hgschmie:
Thanks, great to hear that.

PR #635, opened 2016-12-29
adds maven configuration switches for more analyzers

hgschmie:


jeremylong:
For the non-FileTypeAnanlyzers I do not believe configuration options should be made available to disable them. These have been added to the core engine for the purposes of allowing plugins to disable a specific analyzer and replace it with a new implementation.

However, as you have pointed out several of the FileTypeAnalyzers do not have configuration options. When this gets merged I plan on:
1. Removing the configuration options for the non-file type analyzers (unless someone can make a compelling argument for allowing these analyzers to be configured).
2. Validating that the configuration options exist in the CLI, gradle, and Ant plugins.
3. Updating the documentation

hgschmie:
Ok. I am less (not :-) interested in digging through the cli, gradle and ant code, so I may not do that. Where would I put a pull request for documentation?

jeremylong:
No worries about the other interfaces, I can take care of that.  If you don't mind updating the documentation it is in the maven site for each of the interfaces. For Maven see https://github.com/jeremylong/DependencyCheck/tree/master/dependency-check-maven/src/site/markdown

PR #636, opened 2016-12-29
adds a new flag 'failBuildOnAnyVulnerability'

hgschmie:
In our build system, we enable checkers based on boolean
values. Currently, the only way to enable failing the build on
vulnerabilities is by providing a numeric value (0-10) for another
property. This change adds a boolean switch that will fail the build
if any vulnerability is present (we have a strict "no vulnerabilities
in our builds" policy).

### DIFF
--- a/dependency-check-maven/src/main/java/org/owasp/dependencycheck/maven/BaseDependencyCheckMojo.java
+++ b/dependency-check-maven/src/main/java/org/owasp/dependencycheck/maven/BaseDependencyCheckMojo.java
@@ -159,6 +159,12 @@ public abstract class BaseDependencyCheckMojo extends AbstractMojo implements Ma
     @Parameter(property = "failBuildOnCVSS", defaultValue = "11", required = true)
     private float failBuildOnCVSS = 11;
     /**
+     * Fail the build if any dependency has a vulnerability listed.
+     */
+    @SuppressWarnings("CanBeFinal")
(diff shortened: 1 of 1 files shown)


jeremylong:
This would be equivalent to setting:
```xml
<failBuildOnCVSS>0</failBuildOnCVSS>`
```
As such, I don't see the need to add additional configuration options that have the same affect.

hgschmie:
Hey Jeremy,

thanks for looking at the pull request. I hope that I can change your mind. I am maintaining a little project to provide a basic setup when building software with Maven. The idea is to have scaffolding and "opinionated" configuration for a set of curated plugins and checkers.  This is at https://github.com/basepom/basepom/

I want to use this plugin in a similar way as I integrated checkstyle, jacoco, pmd, findbugs and a number of other plugins. For each of those, two properties exist:

https://github.com/basepom/basepom/blob/master/foundation/pom.xml#L77-L117

One, `basepom.check.skip-<xxx>` allows a plugin execution to be skipped, the other `basepom.check.fail-<xxx>` turns on whether an error flagged by the plugin is fatal or not. Each of those are booleans and they cascade into larger groups (such as `basepom.check.fail-basic`). 

I can do the former thing with your plugin. I can not do the latter because there is no boolean to control "failure/non-failure" based on the outcome.  Unfortunately, maven is pretty rigid and the general consensus with most plugins is to provide booleans that control skip and failure. 

I hope that you may reconsider applying this patch.

PR #644, opened 2017-01-18
Add troubling JAR file name to the exception when JAR reading errors occur

oosterholt:
Minor change: When JAR reading errors occur, at least add the file name to the exception. Without it, finding the troubling JAR is hard.


jeremylong:
Thanks!

Issue #650, opened 2017-01-31
DependencyMergingAnalyzer also merges all evidences to main 


biancajiang:
@jeremylong Thanks for #630!  I actually had some local changes very similar but yours is great!

The only side effect is that by merging all evidences to the main dependency, the boundary of dependencies from distinctive analyzers is blur now.  

#### Example
when both dependency by CocoaPodsAnalyzer and dependency by SwiftPackageManagerAnalyzer co-exist (which is not uncommon), it's great DependencyMergingAnalyzer will recognize it and make the CocoaPodsAnalyzer depd the main one, SwiftPackageManagerAnalyzer depd the related.  But by merging the evidence to the main dependency sort of "polluted" the evidence in SwiftPackageManagerAnalyzer depd, even though all evidence are still available from the related dependencies.  

Can the evidence be kept in their original dependency objects so the boundary are kept?  If any use case needs the complete list of evidence, they can still be collected from the bundled dependencies? 

I'd be happy to provide a PR if this sounds reasonable.

jeremylong:
I can't think of a use case for needing the evidence separated more then is already done by the "source" field? As these two package management systems are pointing at the same dependency/artifact - the evidence should be the same. Do you have a concrete reason why this would be needed?

biancajiang:
In addition to vulnerability, we also use the packages identification and their metadata (evidences) for other purposes, where consistent results for packages is key.  So when multiple artifacts co-exist, the distinction of the main vs. related dependencies is important. For example, when both Swift.package and .podspec exist for the same package, evidences from the main dependency source always has priority, unless other sources provide higher confidence evidence.  It worked well before v1.4.5.

Interestingly, it seems the related dependencies do not have evidences anymore, even after [the evidence merge to the main dependency](https://github.com/jeremylong/DependencyCheck/blob/master/dependency-check-core/src/main/java/org/owasp/dependencycheck/analyzer/DependencyMergingAnalyzer.java#L179-L181) is removed.  But I haven't found when the evidence removal was done.  

Any reasons for this change?

biancajiang:
@jeremylong PR #654 is available.

We heavily rely on package metadata for open source package and license for legal compliance purpose.  Hence we are a bit more sensitive to the consistency of package metadata.  The new test class should be able to cover the two cases here.

jeremylong:
@biancajiang one solution to meet your needs would be to disable the core dependency merging analyzer and replace it with your own. would that work?

[...]

PR #654, opened 2017-02-06
PR for #650 with test cases

biancajiang:
Changes:

1. new test AnalyzersIntegrationTest.java with two test cases:
- testSwiftDependencyBundling() shows the original pb in #650. 
- testNodeDependencyBundling() shows relatedDependencies from duplicates no longer have evidences.  

2. two main changes to DependencyMergingAnalyzer.java:
- skip merging of evidences to fix the original pb in #650 
- populate evidences back to duplicates.  This should be a better approach in performance than including duplicates for dependency analysis.
- handle more co-located package manager cases.

Package metadata consistency are important to us for open source package and license identification for legal compliance.

Please let me know if any questions.


biancajiang:
more change per codacy suggestions.  Feedback welcomed.

jeremylong:
One of the reasons for merging the evidence collected from different package management frameworks is to try to decrease noise/duplicates in the results and increase detection; thus, if independent analyzers (for different filetypes/package managers) happen to collect different pieces of evidence for the same dependency that when put together can assist in finding a correct match.  With regard to your test case and the reason that the node modules' related dependency has no evidence - that is an issue with the node analyzer that we'd have to look into.

Are you using the evidence from the HTML or XML report?  We could expand one or both to show the evidence on the related dependencies AND mark the evidence as having come from a non-primary source if it was from merging (i.e. it came from a related dependency).  That way, for you reporting purposes you could filter the non-primary (I'll likely come up with a better name) and then you would  also have the evidence attached to the original related dependency.

Thoughts?

dependency-check-core/src/test/java/org/owasp/dependencycheck/analyzer/AnalyzersIntegrationTest.java
```diff
@@ -0,0 +1,186 @@
...
+import org.owasp.dependencycheck.dependency.Evidence;
+import org.owasp.dependencycheck.dependency.EvidenceCollection;
+import org.owasp.dependencycheck.utils.Settings;
+
+/**
+ * Integration tests for testing certain analyzers working together.
+ * 
+ * @author Bianca Jiang (@biancajiang)
+ */
+public class AnalyzersIntegrationTest extends BaseTest {
```
jeremylong, line 42: This should extend `BaseDBTestCase` as it utilizes the engine and database.  There is a chance that the database would not exist when the test runs and using the BaseDBTestCase ensures that a minimum database is in place. (and yes, I should probably rename the BaseDBTestCase to BaseIntegrationTestCase or something like that).

PR #668, opened 2017-02-21
Misc minor performance tweaking and cleanup

stefanneuhaus:
My aim was to further speedup the CVE DB update of the OOTB setup using H2 by making CveDB thread-safe and processing the update files in parallel.

Though I tried several things the speedup is not very impressive. The mean runtime of `gradlew --no-daemon dependencyCheckPurge dependencyCheckUpdate` decreased by ~3 sec on my machine/internet connection.

Nevertheless the PR could be valuable:
* refactoring of CveDB eases further optimizations/parallelization
* small performance improvement
* minor cleanup

I want to share some observations which you might find interesting:
* Performance of the DB updates only plays an important role in the OOTB setup, as in this case the effect of the cached DB is restricted (lack of a central DB and varying build servers).
* Though CveDB is thread-safe now, increasing parallelism by means of `NvdCveUpdater.PROCESSING_THREAD_POOL_SIZE` even degrades performance at the moment. This is probably not very surprising, as CveDB maintains a single connection and thus uses synchronization.
* Using a connection pool is probably a good idea, as this would require far less synchronization. Advantages:
** Enterprise setup: parallel processing should scale very well
** H2: ~5-10% speedup (testrunning to estimate the speedup required some tweaks)
* Submitting DB requests is not the bottleneck in the fast H2 embedded mode as no network IO is involved.
* H2 connection string
** `FILE_LOCK=SERIALIZED`: omitting this brings some speedup but is probably not an option, as there could be builds running in parallel
** also using `LOCK_MODE=0` (isolation level Read Uncommitted): speedup only in conjunction with omitting `FILE_LOCK=SERIALIZED`. The same drawbacks as above apply.
* Enabling batch support for H2 did not bring any improvements. Tested with autocommit enabled and disabled.


stevespringett:
Changing file lock in the JDBC URL will have consequences in the Jenkins plugin. I would highly advise against this.

Also be careful of the threadpool. I'm not entirely sure if parallel Jenkins jobs would actually know about each others pool or not. This would require a some testing.

stefanneuhaus:
* "file lock in the JDBC URL": I fully agree, we should not trade performance for reliability.
* I meant "connection pool", not "thread pool" (I corrected it above). And yes, this still would require some testing.

If one wanted to further improve performance, introducing a connection pool would probably be the next step. I do not want to suggest this, as this would be at least a moderate effort/risk task and it would be questionable if it is worth the effort. I just wanted to document my findings.

dependency-check-core/src/main/java/org/owasp/dependencycheck/data/update/nvd/NvdCve20Handler.java
```diff
@@ -272,7 +272,9 @@ private void saveEntry(Vulnerability vuln) throws DatabaseException, CorruptInde
         final String cveName = vuln.getName();
         if (prevVersionVulnMap != null && prevVersionVulnMap.containsKey(cveName)) {
             final List<VulnerableSoftware> vulnSoftware = prevVersionVulnMap.get(cveName);
-            vuln.getVulnerableSoftware().addAll(vulnSoftware);
```
jeremylong: Okay, due to my horrible code - the change from a loop calling `updateVulnerableSoftware` to using `getVulnerableSoftware().addAll()` will cause problems and not work as expected.  This is likely a bug I need to fix; but consider the following two test cases;  `testAddAll()` fails:

```java
    @Test
    public void testAddAll() {
        List<VulnerableSoftware> list = new ArrayList<>();
        VulnerableSoftware vs = new VulnerableSoftware();
        vs.setCpe("cpe:/a:something:something:1.0.0");
        list.add(vs);
        
        vs = new VulnerableSoftware();
        vs.setCpe("cpe:/a:something:something:1.0.0");
        vs.setPreviousVersion("1");
        List<VulnerableSoftware> adds = new ArrayList<>();
        adds.add(vs);
        
        assertEquals(1,list.size());
        list.addAll(adds);
        assertEquals(1,list.size());
    }
    
    @Test
    public void testRemoveAdd() {
        List<VulnerableSoftware> list = new ArrayList<>();
        VulnerableSoftware vs = new VulnerableSoftware();
        vs.setCpe("cpe:/a:something:something:1.0.0");
        list.add(vs);
        
        vs = new VulnerableSoftware();
        vs.setCpe("cpe:/a:something:something:1.0.0");
        vs.setPreviousVersion("1");
        List<VulnerableSoftware> adds = new ArrayList<>();
        adds.add(vs);
        
        assertEquals(1,list.size());
        list.remove(vs);
        list.add(vs);
        assertEquals(1,list.size());
    }
```

stefanneuhaus: Correct. Fortunately this was covered by an integration test. To me it seems like `VulnerableSoftware` does not consistently implement `equals()` and `hashCode()`. I already had a quick look but did not find any issues.

One could also check if the integration test fails when removing the seemingly superfluous contains/remove stuff from `Vulnerability.updateVulnerableSoftware()`:
```
public boolean updateVulnerableSoftware(VulnerableSoftware vulnSoftware) {
    if (vulnerableSoftware.contains(vulnSoftware)) {
        vulnerableSoftware.remove(vulnSoftware);
    }
    return vulnerableSoftware.add(vulnSoftware);
}
```

I "fixed" the failing integration test in the first place by remigrating to the former implementation. But I agree, this is probably a different issue which is worth investigating.

stefanneuhaus:
I think the issue arises due to the fact that
* VulnerableSoftware does not take `previousVersion` into account for `hashCode()`/`equals()` (and also `compareTo()`).
* Vulnerability maintains a *Set* of VulnerableSoftware.
* Consumers of Vulnerability (like NvdCve20Handler) use *Lists* of VulnerableSoftware to update the Vulnerability.

So I think the test cases for VulnerableSoftwareTest should rather look like:
```
    @Test
    public void testAddAll() {
        Vulnerability obj = new Vulnerability();

        List<VulnerableSoftware> list = new ArrayList<>();
        VulnerableSoftware vs1 = new VulnerableSoftware();
        vs1.setCpe("cpe:/a:something:something:1.0.0");
        list.add(vs1);
        VulnerableSoftware vs2 = new VulnerableSoftware();
        vs2.setCpe("cpe:/a:something:something:1.0.0");
        vs2.setPreviousVersion("1");
        list.add(vs2);

        obj.getVulnerableSoftware().addAll(list);

        Set<VulnerableSoftware> vulnerableSoftwareSet = obj.getVulnerableSoftware();
        assertEquals(1, vulnerableSoftwareSet.size());
        VulnerableSoftware vulnerableSoftware = vulnerableSoftwareSet.iterator().next();
        assertEquals("1", vulnerableSoftware.getPreviousVersion());
    }

    @Test
    public void testRemoveAdd() {
        Vulnerability obj = new Vulnerability();

        List<VulnerableSoftware> list = new ArrayList<>();
        VulnerableSoftware vs1 = new VulnerableSoftware();
        vs1.setCpe("cpe:/a:something:something:1.0.0");
        list.add(vs1);
        VulnerableSoftware vs2 = new VulnerableSoftware();
        vs2.setCpe("cpe:/a:something:something:1.0.0");
        vs2.setPreviousVersion("1");
        list.add(vs2);

        for (VulnerableSoftware vs : list) {
            obj.updateVulnerableSoftware(vs);
        }

        Set<VulnerableSoftware> vulnerableSoftwareSet = obj.getVulnerableSoftware();
        assertEquals(1, vulnerableSoftwareSet.size());
        VulnerableSoftware vulnerableSoftware = vulnerableSoftwareSet.iterator().next();
        assertEquals("1", vulnerableSoftware.getPreviousVersion());
    }
```

But as they only test the semantics of equals()/hashCode() we do not need them: `testEquals()` and `testEquals2()` already cover this exact test case.


Rather we may want to fix Vulnerability s.t. the following test cases pass (`testAddAll()` fails):
```
    @Test
    public void testAddAll() {
        Vulnerability obj = new Vulnerability();

        List<VulnerableSoftware> list = new ArrayList<>();
        VulnerableSoftware vs1 = new VulnerableSoftware();
        vs1.setCpe("cpe:/a:something:something:1.0.0");
        list.add(vs1);
        VulnerableSoftware vs2 = new VulnerableSoftware();
        vs2.setCpe("cpe:/a:something:something:1.0.0");
        vs2.setPreviousVersion("1");
        list.add(vs2);

        obj.getVulnerableSoftware().addAll(list);

        Set<VulnerableSoftware> vulnerableSoftwareSet = obj.getVulnerableSoftware();
        assertEquals(1, vulnerableSoftwareSet.size());
        VulnerableSoftware vulnerableSoftware = vulnerableSoftwareSet.iterator().next();
        assertEquals("1", vulnerableSoftware.getPreviousVersion());
    }

    @Test
    public void testRemoveAdd() {
        Vulnerability obj = new Vulnerability();

        List<VulnerableSoftware> list = new ArrayList<>();
        VulnerableSoftware vs1 = new VulnerableSoftware();
        vs1.setCpe("cpe:/a:something:something:1.0.0");
        list.add(vs1);
        VulnerableSoftware vs2 = new VulnerableSoftware();
        vs2.setCpe("cpe:/a:something:something:1.0.0");
        vs2.setPreviousVersion("1");
        list.add(vs2);

        for (VulnerableSoftware vs : list) {
            obj.updateVulnerableSoftware(vs);
        }

        Set<VulnerableSoftware> vulnerableSoftwareSet = obj.getVulnerableSoftware();
        assertEquals(1, vulnerableSoftwareSet.size());
        VulnerableSoftware vulnerableSoftware = vulnerableSoftwareSet.iterator().next();
        assertEquals("1", vulnerableSoftware.getPreviousVersion());
    }
```

So basically the confusion comes from the fact, that we have two different notions of equality for VulnerableSoftware: sometimes we seem to care about the `previousVersion`, sometimes we don't. My guess is that this is not actually a bug, as the suprising logic like e.g. in `NvdCve20Handler.saveEntry()` is exactly for the reason to accomplish this.

Rather this seems like something that one may want to cleanup. As the code is a bit complex it is not easy for me to overlook the consequences of a refactoring. The complexity drivers that I observed are:
* different notions of equality
* VulnerableSoftware is not immutable, but has public modifiers like `setPreviousVersion()`, `setName()`, and even `parseName()`.
* VulnerableSoftware extends from IndexEntry, which is also not immutable.
* VulnerableSoftware implements Comparable: thus also `compareTo()` must be consistent with `equals()`/`hashCode()`. Some consumers actually use sorted collections like TreeSet, but it is not always easy to check if they actually rely on this. Not implementing Comparable and letting the consumers decide on the sorting they require could reduce complexity and may even be a performance gain.

## Conclusion
IMHO this issue probably does not indicate an actual bug. It is rather something that could need some cleanup.

jeremylong:
Completely agree - the VulnerableSoftware (and related) needs to be cleaned up. Your PR is what really highlighted this for me. The rest of the changes look great - thanks!  I've opened #677 so that I remember to cleanup the VulnerableSoftware code.

Issue #663, opened 2017-02-18
Restoring Maven 3.0 compatibility


aikebah:
Within my company we are working on a gradual migration from Maven 3.0 to Maven 3.3. We would like to do those in regular maintenance. Incompatibility of the DependencyCheck maven plugin with Maven 3.0 means that until we do that maintenance we can no longer run periodic scans on the codebase of the individual projects until we've made them Maven 3.1+ capable (there are more plugins out there with a 3.0/3.1 version brake that all need careful updating when migrating from 3.0 to 3.1+).

Having Dependency-check be equally capable of running on Maven 3.0 as well as 3.1+ would thus be a great benefit to us as it will enable us to continue scanning all our projects periodically for newly discovered vulnerabilities in our maven-based CI-environment. (We recently found out that our current scans no longer work due to the fix for the symptom mentioned in issue #562, but the fix for that is in 1.4.3 which already requires running Maven 3.1+). 

Looking at the basic functionality of DependencyCheck I suspect that the only reason for incompatibility with older versions of Maven is the dependency-resolution, which was changed to Eclipse Aether in August 2016. 

Is this dependency resolution indeed the only reason for the Maven 3.1+ requirement? 
If so I would propose to use the same version-agnostic depedency resolution that is used by the maven community themselves in e.g. the maven-dependency-plugin for its artifact resolution for dependency:source (using helper libraries from the org.apache.maven.shared groupId). I would be willing to invest time on getting BaseDependencyCheckMojo to resolve artifacts in a similar way and file a pull-request, but before taking the effort would like to ensure that this is indeed the only reason for the Maven 3.1+ requirement of the plugin.

jeremylong:
The issue with dependency resolution is the aggregate goal. At the time the
parent is executing the dependencies of the child modules have not been
resolved.

If you can propose something that will allow the resolution of dependencies
in child modules that will work with 3.0 and later - please let me know.

--jeremy

aikebah:
I want to experiment along the lines of the AbstractDependencyFilterMojo logic of resolving artifacts as present in getClassifierTranslatedDependencies (which gets called for the not-yet resolved source dependencies on the dependency:source goal). 

[...]

PR #669, opened 2017-02-21
Re-implemented dependency-resolution to work across maven 3.0 and 3.1+

aikebah:
Changed the dependency-resolution mechanism similar to how maven's dependency-plugin does missing source dependencies resolution in its dependency:source goal. This restores compatibility with maven 3.0 while retaining the aggregate goal fixes.

This commit would solve my issue #663


jeremylong:
The changes look great!  I'm going to do some testing this weekend before merging.

jeremylong:
Thanks again for this PR; truly appreciate the contribution to make this Maven 3.0 compatible.

aikebah:
YW, looking forward to the next release

Issue #665, opened 2017-02-20
[Feature Request] Adding pom version to projectinfo section of the dependency check report


Prakhash:
Hi all,

I would like to request a new feature from the dependency check. 

As shown below, Dependency check currently provides only the 3 options in the projectinfo, in addition to this, I wish to add the project version to the projectinfo section.

    <projectInfo>
        <name>WSO2 Carbon - UI</name>
        <reportDate>2017-02-20T20:23:53.695+0530</reportDate>
        <credits>This report contains data retrieved from the National Vulnerability Database: http://nvd.nist.gov</credits>
    </projectInfo>

Like this below :

     <projectInfo>
        <name>WSO2 Carbon - UI</name>
        <version>4.4.0</version>
        <reportDate>2017-02-20T20:23:53.695+0530</reportDate>
        <credits>This report contains data retrieved from the National Vulnerability Database: http://nvd.nist.gov</credits>
    </projectInfo>

 This is a really an essential requirement for people who are maintaining products with multiple versions and we create customized reports from the XML report generated by Dependency check. So it is not practical to always go and edit the bulk HTML reports

 **My Solution**

  For now, I have implemented this by forking the dependency check code by doing the following changes. If this an acceptable way of doing this, I can send a PR to you, or else please recommend a way to do this

1. Introduced a new constructor as below in [1] with the applicationVersion included

  ```
    public ReportGenerator(String applicationName,String applicationVersion, List<Dependency> dependencies, List<Analyzer> analyzers, DatabaseProperties properties) {

        this(applicationName,dependencies,analyzers,properties);
        context.put("pomVersion",applicationVersion);
    }
```

2. Changed the invoking maven method as below with the p.getVersion() attribute [2]

	`final ReportGenerator r = new ReportGenerator(p.getName(),p.getVersion(), engine.getDependencies(), engine.getAnalyzers(), prop);`

3. Changed the schema as in [3]

4. Changed  the XmlReport.vsl as in [4]

  [1] https://github.com/Prakhash/DependencyCheck/blob/master/dependency-check-core/src/main/java/org/owasp/dependencycheck/reporting/ReportGenerator.java#L128

  [2] https://github.com/Prakhash/DependencyCheck/blob/master/dependency-check-maven/src/main/java/org/owasp/dependencycheck/maven/BaseDependencyCheckMojo.java#L1049

[...]

PR #682, opened 2017-03-13
 changed the coordinates according to the pom file details

Prakhash:


Prakhash:
This part is done for maven in [1]

[1] https://github.com/jeremylong/DependencyCheck/issues/665

Issue #709, opened 2017-04-23
Create analyzer for Node Security Platform (nsp)


stevespringett:
nsp is a free source of information for published and non-published node.js vulnerabilities. nsp takes in a 'cleaned' version of package.json via a public REST API and returns an array of 0 or more advisories.

It is possible to use nsp as a supplemental datasource for package.json files. There are no requirements on storing datafiles locally, as nsp is purely a web service.

Due to the fact that the advisories are not located at the nvd and many of them do not have a published CVE associated with them, minor changes will need to be made in reporting to account for these differences.

This ticket is to track this enhancement.

stevespringett:
@jeremylong The nsp analyzer is a web service so the analysis is not actually part of ODC. The implementation of the nsp rest client IS part of ODC. What's your thought on @ Experimental for this? The analysis is solid and proven, but the ODC analyzer that interacts with nsp is not. I can think of arguments for both, but I'd like to get your take before I create the PR next week.

Issue #707, opened 2017-04-21
Enhanced support for new NVD feeds and CVSSv3


stevespringett:
The NVD will soon be releasing a beta of a new feed (both in XML and JSON formats) that support CVSSv3 scores and vectors. It will likely support CPE 2.3 as well.

This enhancement ticket is to ensure Dependency-Check can support:

- Downloading and processing of the new schema format
- Distinguish between CVSSv2 and CVSSv3 scores
- If possible, provide both sets of scores in dependency-check-report.xml (maybe html to)
- Can properly map severities for CVSSv2 and CVSSv3 scores (they are different) - refer to https://nvd.nist.gov/vuln-metrics/cvss

stevespringett:
Beta JSON feeds are now available from the NVD. The JSON feeds have CPE 2.3 syntax and include both CVSSv2 and CVSSv3 scores and impacts.

https://nvd.nist.gov/vuln/data-feeds#JSON_FEED

An updated XML feed with CVSSv3 and CPE 2.3 is not yet available.

stevespringett:
The JSON feed continues to not support impact and exploitability scores. Therefore, if these are necessary, it is now possible to perform calculations and retrieve vectors for CVSSv2 and CVSSv3 https://github.com/stevespringett/cvss-calculator

v6ak:
When elaborating with #733 and related, I've found another issue that might be fixed by this new import. It might be worth checking.

Description: I've seen some cases where OS (Windows/Linux/MacOS) CPEs are imported as application CPEs. For example, you can compare original [CVE-2010-1689](https://nvd.nist.gov/vuln/detail/CVE-2010-1689?cpeVersion=2.2#vulnConfigurationsArea
) to [imported data](https://gist.github.com/v6ak/d1002095cdd212ea64b73fdafed12fa6). In the original data, there is “cpe:/**o**:”, not “cpe:/**a**:”.

If desired, I can open a separate issue for it.

jeremylong:
This enhancement request has now bubbled to the top and is slated for implementation next.  See issue #1088 - ODC is not reporting on vulnerable libraries because the data in the older XML feeds does not include the necessary information for new vulnerabilities (2018 and some of the 2017 CVEs).

jeremylong:
As part of this enhancement the version matching needs to be re-worked. In most cases the file type analyzer itself can identify the correct version number - ODC should just use the version number identified for the CPE.

gbena:
[...]

PR #716, opened 2017-05-01
Added support to generate report in JSON format

ThomasGoeytil:
## Fixes Issue #
Added functionality

## Description of Change
Added support to generate JSON-report with -f JSON. 

no (Current integration test generates a JSON reports and fails if JSON is not valid).


stevespringett:
@ThomasGoeytil Thanks for the PR. We've been meaning to add JSON support for awhile now. One suggestion. We might want to version the JSON schema (similar to how we do the XML), because the schema will change when support for CVSSv3 is added. #707 

@jeremylong If you merge this in, we'll need to update the json template so that it supports #709  because as of now, it basically copies the xml version which currently doesn't account for the changes made in #714.

jeremylong:
Thanks for the PR!

PR #588, opened 2016-10-06
Add Dockerfile for dependency check

wurstbrot:
After a discussion on #577 I am creating a new pull request for the Dockerfile, with HTTPS source and fetching of current release.

After the pull request is accepted, I will create the repository on hub.docker.com.

PR #719, opened 2017-05-06
Improve dockerfile

janpapenbrock:
## Fixes Issue #

None.

## Description of Change

I've improved the Dockerfile to allow end users to use complete CLI functionality easier, without having to override `--entrypoint`.

Now it is possible to run 

```
docker run --rm owasp/dependency-check --version
```

which would before have been

```
docker run --rm --entrypoint "/usr/share/dependency-check/bin/dependency-check.sh" /owasp/dependency-check --version
```

### Breaking change

This is a breaking change. However, as the docker image is not published to Dockerhub at the moment, there is no breakage of officially documented behavior.

Not required.

## Additional notes

It would be great if the docker image could be published to Dockerhub as it is documented in README.

In lack of an official copy I created an automated build for our fork, here https://hub.docker.com/r/hochzehn/dependency-check/. This is not ideal, however, because we'd have to keep our fork in-sync to have an up-to-date image.

Therefore I hope that the official image can be published to Dockerhub, as mentioned in October in this comment: https://github.com/jeremylong/DependencyCheck/pull/588#issuecomment-252508904

CC @wurstbrot

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,6 +2,8 @@ FROM java:8
 
 MAINTAINER Timo Pagel <dependencycheckmaintainer@timo-pagel.de>
 
+ENV user=dockeruser
+
 RUN wget -O /tmp/current.txt http://jeremylong.github.io/DependencyCheck/current.txt && \
  current=$(cat /tmp/current.txt) && \
  wget https://dl.bintray.com/jeremy-long/owasp/dependency-check-$current-release.zip && \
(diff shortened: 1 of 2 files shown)


wurstbrot (approved):

wurstbrot:
Only @jeremylong can put this project on hub.docker.com in the OWASP category, as he is the owner in github.

jeremylong:
Thanks for the PR! I finally remembered to publish this to https://hub.docker.com/r/owasp/dependency-check/

janpapenbrock:
Awesome, thanks!

PR #720, opened 2017-05-06
Fix argument name for CLI.

janpapenbrock:
## Fixes Issue #

None.

## Description of Change

CLI documentation of how to enable experimental analyzers is currently wrong. The required argument is called `--enableExperimental`, whereas the documentation states `--experimental`.

Cf. https://jeremylong.github.io/DependencyCheck/dependency-check-cli/arguments.html
Cf. https://github.com/jeremylong/DependencyCheck/blob/master/dependency-check-cli/src/main/java/org/owasp/dependencycheck/CliParser.java#L1468-L1471

Not required.

### DIFF
--- a/dependency-check-cli/src/site/markdown/arguments.md
+++ b/dependency-check-cli/src/site/markdown/arguments.md
@@ -4,22 +4,22 @@ Command Line Arguments
 The following table lists the command line arguments:
 
 Short  | Argument&nbsp;Name&nbsp;&nbsp; | Parameter       | Description | Requirement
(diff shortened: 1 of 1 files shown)


janpapenbrock:
Sorry for the whitespace-bloat, had to adjust column sizes. Best viewed with whitespaces ignored: https://github.com/jeremylong/DependencyCheck/pull/720/files?w=1

jeremylong:
Thanks for the PR - that was likely a copy paste error.

PR #726, opened 2017-05-09
Suppress EntityFramework false positives.

### LINKED ISSUE #90: False positives in .NET
Dependency Check reports (with highest confidence) vulnerabilities CVE-1999-0201 and CVE-1999-0082 detected in Starksoft.Net.Ftp.dll which can be downloaded here: http://biko.codeplex.com/

One idea I had regarding some of these older CPE/CVE combos is just make a cut point of say 2001/2 and not include any earlier CVE data.

sethjackson:
## Fixes Issue

Fixes: https://github.com/jeremylong/DependencyCheck/issues/90#issuecomment-278581196 which I saw in my own scans.

## Description of Change

Suppress EntityFramework false positives.

No.


jeremylong:
Thanks!

PR #727, opened 2017-05-09
Update Windows development usage.

sethjackson:
So that the example invocation works in PowerShell and cmd.

## Fixes Issue

Invocation doesn't work OOB on `cmd`.

## Description of Change

Fix CLI invocation.

No


jeremylong:
thanks!

PR #744, opened 2017-05-24
Added some caching for CveDB in order to speedup some scans

v6ak:
## Fixes Issue #

Some items (e.g., Linux kernel) have many CVE entries, which takes much time to load it. When such CPE is matched for multiple dependencies, the scan takes much time and can easily hit the 10 minutes limit.

## Description of Change

Result of getVulnerabilities(String cpeStr) is cached under a soft reference. Other parts of CveDB are prepared for caching.

*no*

There is no new functionality, it is just enhancement of the old functionality. I've thought about tests, but it should just improve performance and I am not sure if it is appropriate to test it in unit/integration tests.

### DIFF
--- a/dependency-check-core/src/main/java/org/owasp/dependencycheck/data/nvdcve/CveDB.java
+++ b/dependency-check-core/src/main/java/org/owasp/dependencycheck/data/nvdcve/CveDB.java
@@ -36,6 +36,7 @@ import java.util.Properties;
 import java.util.ResourceBundle;
 import java.util.Set;
 import javax.annotation.concurrent.ThreadSafe;
+import org.apache.commons.collections.map.ReferenceMap;
 import org.owasp.dependencycheck.data.cwe.CweDB;
 import org.owasp.dependencycheck.dependency.Reference;
 import org.owasp.dependencycheck.dependency.Vulnerability;
@@ -48,6 +49,8 @@ import org.owasp.dependencycheck.utils.Settings;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import static org.apache.commons.collections.map.AbstractReferenceMap.HARD;
+import static org.apache.commons.collections.map.AbstractReferenceMap.SOFT;
 import static org.owasp.dependencycheck.data.nvdcve.CveDB.PreparedStatementCveDb.*;
 
 /**
@@ -90,6 +93,9 @@ public final class CveDB implements AutoCloseable {
      * The prepared statements.
      */
     private final EnumMap<PreparedStatementCveDb, PreparedStatement> preparedStatements = new EnumMap<>(PreparedStatementCveDb.class);
+
+    @SuppressWarnings("unchecked")
+    private final Map<String, List<Vulnerability>> vulnerabilitiesForCpeCache = new ReferenceMap(HARD, SOFT);
 
     /**
      * The enum value names must match the keys of the statements in the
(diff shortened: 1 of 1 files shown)


jeremylong:

dependency-check-core/src/main/java/org/owasp/dependencycheck/data/nvdcve/CveDB.java
```diff
@@ -91,6 +94,9 @@
      */
     private final EnumMap<PreparedStatementCveDb, PreparedStatement> preparedStatements = new EnumMap<>(PreparedStatementCveDb.class);
 
+    @SuppressWarnings("unchecked")
+    private final Map<String, List<Vulnerability>> vulnerabilitiesForCpeCache = new ReferenceMap(HARD, SOFT);
```
jeremylong: dependency-check, for the most part, is multi-threaded. While the `getVulnerabilities()` method is synchronized - it might be safer to use [`Collections.synchronizedMap(java.util.Map)`](http://docs.oracle.com/javase/7/docs/api/java/util/Collections.html?is-external=true#synchronizedMap(java.util.Map)).  Thoughts?

dependency-check-core/src/main/java/org/owasp/dependencycheck/data/nvdcve/CveDB.java
```diff
@@ -474,6 +481,7 @@ public synchronized Properties getProperties() {
      * @param value the property value
      */
     public synchronized void saveProperty(String key, String value) {
+        clearCache();
```
jeremylong, line 485: No need to clear the cache here. The saving of properties to the DB would have no affect on the vulnerabilities.

dependency-check-core/src/main/java/org/owasp/dependencycheck/data/nvdcve/CveDB.java
```diff
@@ -265,6 +271,7 @@ private synchronized void open() throws DatabaseException {
      */
     @Override
     public synchronized void close() {
+        clearCache();
```
jeremylong: For performance reasons, it would be better if this `clearCache()` was moved down 5 lines next to the `instance.usageCount = 0;`. As 'close()' may be called by multiple threads - another thread may still need to get vulnerabilities.

jeremylong (changes requested):
Overall I like the change! Couple of minor changes requested (see review comments) and the one question on the synchronization of the map.  All of the methods are synchronized, so that might be overkill - but I'm all for safety rather then creating weird un-reproducible bugs.

dependency-check-core/src/main/java/org/owasp/dependencycheck/data/nvdcve/CveDB.java
v6ak: Good point, I have missed that.

dependency-check-core/src/main/java/org/owasp/dependencycheck/data/nvdcve/CveDB.java
v6ak: IMHO it is not needed, but I'll change it if it helps to fit better in your culture.

dependency-check-core/src/main/java/org/owasp/dependencycheck/data/nvdcve/CveDB.java
v6ak: This is true, at least at this moment, but I believe:

1. This line helps maintainability. The logic behind is that one write clears everything from cache. We could break it in multiple parts and have, say, clearVulnerabilityCache and similar methods, but I see it as overkill at the moment. Also, one would have to be careful about cross-dependencies over multiple table (e.g., when adding a property that influences how are vulnerabilities read).
2. It does not hurt performance in typical usage.

jeremylong (approved):

jeremylong:
Thanks for the PR!

PR #745, opened 2017-05-24
Report Generation Cleanup

jeremylong:
## Description of Change

This PR modifies how the report generation should be handled. There was a lot of duplicate code in each of the plugin modules (ant, maven, etc., and even the CLI). The plugins should now use the engine's `writeReports` method instead of directly calling the `ReportGenerator`.

Existing tests were updated for the changed functionality.


jeremylong:

Issue #710, opened 2017-04-24
Exception caught: org.xml.sax.SAXParseException: DOCTYPE is disallowed when the feature "http://apache.org/xml/features/disallow-doctype-decl" set to true


vinod272:
Following is the entry from the log file.Please let me know if this issue is something to worry about, i think a particular jar file is not being analyzed.

2017-04-18 19:48:40,670 org.owasp.dependencycheck.data.central.CentralSearch:95
DEBUG - Searching Central url https://search.maven.org/solrsearch/select?q=1:"80b4ffe7c26ee97313bea2ddda5835fd38812ee4"&wt=xml
2017-04-18 19:48:40,673 org.owasp.dependencycheck.utils.SSLSocketFactoryEx:259
DEBUG - Available Protocols:
2017-04-18 19:48:40,673 org.owasp.dependencycheck.utils.SSLSocketFactoryEx:261
DEBUG - SSLv2Hello
2017-04-18 19:48:40,673 org.owasp.dependencycheck.utils.SSLSocketFactoryEx:261
DEBUG - SSLv3
2017-04-18 19:48:40,673 org.owasp.dependencycheck.utils.SSLSocketFactoryEx:261
DEBUG - TLSv1
2017-04-18 19:48:40,674 org.owasp.dependencycheck.utils.SSLSocketFactoryEx:261
DEBUG - TLSv1.1
2017-04-18 19:48:40,674 org.owasp.dependencycheck.utils.SSLSocketFactoryEx:261
DEBUG - TLSv1.2
2017-04-18 19:48:40,745 org.owasp.dependencycheck.utils.Downloader:183
DEBUG - Download of http://search.maven.org/remotecontent?filepath=org/eclipse/jetty/orbit/org.objectweb.asm/3.1.0.v200803061910/org.objectweb.asm-3.1.0.v200803061910.pom complete
2017-04-18 19:48:40,746 org.owasp.dependencycheck.xml.pom.PomParser:99
DEBUG - 
**org.xml.sax.SAXParseException: DOCTYPE is disallowed when the feature "http://apache.org/xml/features/disallow-doctype-decl" set to true**.
	at com.sun.org.apache.xerces.internal.parsers.AbstractSAXParser.parse(AbstractSAXParser.java:1239)
	at com.sun.org.apache.xerces.internal.jaxp.SAXParserImpl$JAXPSAXParser.parse(SAXParserImpl.java:648)
[...]

PR #750, opened 2017-05-30
710 pom parse error

jeremylong:
## Fixes Issue #710

When reading pom.xml files, ODC will now use an XML Reader that replaces common HTML named entities with the XML encoded unicode code point representation (i.e. `&oslash;` is converted into `&#216;`).

#### Have test cases been added to cover the new functionality?

yes


jeremylong:

PR #753, opened 2017-05-30
Fix #752 where skipping runtime-scoped maven artifacts also skipped c…

### LINKED ISSUE #752: Maven Compile-Time Dependencies Dropped When Runtime Dependencies Omitted
There is some bad logic in the maven plugin such that setting the `<skipRuntimeScope>` flag to `true` will also cause compile-time dependencies to be skipped.

Steps to reproduce:
- configure plugin with `<skipRuntimeScope>true</skipRuntimeScope>` argument
- include a compile-time dependency with an issue
- run mvn dependency-check:check
- observe no report is generated and the build does not fail (if configured to do so)

PR on the way to fix.  In the meantime, I've found that I can just remove the skip flag in the pom and it works (although it picks up runtime dependencies).

josh-cain:
## Fixes Issue 
#752 

## Description of Change
Isolates logic to determine which artifact scopes are excluded, and correct it such that runtime settings do not affect compile-time artifacts.

yes

### DIFF
--- a/dependency-check-maven/src/main/java/org/owasp/dependencycheck/maven/ArtifactScopeExcluded.java
+++ b/dependency-check-maven/src/main/java/org/owasp/dependencycheck/maven/ArtifactScopeExcluded.java
@@ -19,6 +19,8 @@ package org.owasp.dependencycheck.maven;
 
 import org.owasp.dependencycheck.utils.Filter;
 
+import static org.apache.maven.artifact.Artifact.SCOPE_RUNTIME_PLUS_SYSTEM;
+
 /**
  * Tests is the artifact should be included in the scan (i.e. is the
  * dependency in a scope that is being scanned).
@@ -53,6 +55,9 @@ public class ArtifactScopeExcluded extends Filter<String> {
 			return true;
 		}
 		if (skipRuntimeScope && org.apache.maven.artifact.Artifact.SCOPE_RUNTIME.equals(scope)) {
+			return true;
+		}
+		if (skipRuntimeScope && skipSystemScope && org.apache.maven.artifact.Artifact.SCOPE_COMPILE_PLUS_RUNTIME.equals(SCOPE_RUNTIME_PLUS_SYSTEM)) {
 			return true;
 		}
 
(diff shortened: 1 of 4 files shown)


dependency-check-maven/src/main/java/org/owasp/dependencycheck/maven/ArtifactScopeExcluded.java
```diff
@@ -0,0 +1,61 @@
...
+		if (skipProvidedScope && org.apache.maven.artifact.Artifact.SCOPE_PROVIDED.equals(scope)) {
+			return true;
+		}
+		if (skipSystemScope && org.apache.maven.artifact.Artifact.SCOPE_SYSTEM.equals(scope)) {
+			return true;
+		}
+		if (skipRuntimeScope && org.apache.maven.artifact.Artifact.SCOPE_RUNTIME.equals(scope)) {
+			return true;
+		}
+
```
jeremylong: Should this include:

```java
if (skipRuntimeScope && skipCompileScope && org.apache.maven.artifact.Artifact.SCOPE_COMPILE_PLUS_RUNTIME.equals(scope)) {
    return true;
}
if (skipRuntimeScope && skipSystemScope && org.apache.maven.artifact.Artifact.SCOPE_COMPILE_PLUS_RUNTIME.equals(SCOPE_RUNTIME_PLUS_SYSTEM)) {
    return true;
}
```

jeremylong (commented):
Thanks for the PR - just the one question on `SCOPE_COMPILE_PLUS_RUNTIME` and `SCOPE_RUNTIME_PLUS_SYSTEM`.

dependency-check-maven/src/main/java/org/owasp/dependencycheck/maven/ArtifactScopeExcluded.java
josh-cain: Up to you - didn't want to take liberties with that logic (I.E. `&&` vs `||`) when it comes to dual dependencies.  Wasn't certain whether one 'skip' vote would be enough, or if both would be required to skip said dependency entirely.

Will update PR.

josh-cain: Also, there is no 'skipCompileScope' option on the plugin.  Not really sure we'd want one either (can't really think of circumstances under which compile-time dependencies would *need* to be skipped).

What does that mean for SKIP_COMPILE_PLUS_RUNTIME?

jeremylong:
Thanks again for the PR!

PR #762, opened 2017-06-13
Bootclasspath fixes

anderruiz:
## Description of Change

*Current dependency check version does not work if it is loaded on bootclasspath. This patch fixes issues related with this by inserting an utility method in FileUtils that all classes call to load their resources. Additionally a minor trace fix is included*

*no*


jeremylong:

PR #787, opened 2017-07-05
Fixed the creation of the GrokAssembly.exe temp file and the cleanup of the temp config file.

vladt:
## Fixes Issue #

The GrokAssembly.exe temp file was not created anymore. Probably accidentally removed in https://github.com/jeremylong/DependencyCheck/commit/0075a7e1ce2eab3bf2e543f362e9303ea101f6e5.

## Description of Change

I added back the line that creates the GrokAssembly.exe temp file from resource and added asserts to tests to verify that the GrokAssembly executable and config temp files are created and cleaned up. I also added code to cleanup the temp config file.

Please note that this fixes all tests except one (AssemblyAnalyzerTest.testNonexistent), which still fails, but the failure doesn't seem related to the changes in this PR.

yes


jeremylong:

vladt:
Should I disable all tests (as it was before)? Should I disable only the test that fails now?

PR #789, opened 2017-07-06
Minor updates to PR#787

jeremylong:
See PR #787 - this PR contains a few minor enhancements around error handling and the test cases.


jeremylong:

dependency-check-core/src/test/java/org/owasp/dependencycheck/analyzer/AssemblyAnalyzerTest.java
```diff
@@ -223,11 +223,5 @@ public void testWithSettingMono() throws Exception {
     @After
     public void tearDown() throws Exception {
         analyzer.closeAnalyzer();
-        if (grokAssemblyExeFile != null) {
```
vladt: Without these asserts, there is nothing that verifies that the cleanup of the temp files is done correctly. Or is that something that doesn't matter?

PR #798, opened 2017-07-12
Engine execution modes

markrekveld:
## Fixes Issue #
This PR introduces a new feature, namely engine execution modes. 

## Description of Change
The change proposed introduces the concept of Execution Modes to the `Engine` with the following options:
1. Evidence Collection
2. Evidence Processing
3. Standalone (Default behaviour if no mode is specified)

### Evidence Collection
The `Engine` skips the database and only executes `Analyzers` of the following phases  `INITIAL`, `PRE_INFORMATION_COLLECTION`, `INFORMATION_COLLECTION`, `POST_INFORMATION_COLLECTION`.

### Evidence Processing
The `Engine` initialized the database and only executes `Analyzers` of the following phases  `PRE_IDENTIFIER_ANALYSIS`, `IDENTIFIER_ANALYSIS`, `POST_IDENTIFIER_ANALYSIS`, `PRE_FINDING_ANALYSIS`, `FINDING_ANALYSIS`, `POST_FINDING_ANALYSIS`, `FINAL`.

### Standalone
The `Engine` behaves just like it today on master, initializing the database, execution all `Analyzers` of all the phases.

*yes*

### DIFF
--- a/dependency-check-core/src/main/java/org/owasp/dependencycheck/Engine.java
+++ b/dependency-check-core/src/main/java/org/owasp/dependencycheck/Engine.java
@@ -583,7 +583,7 @@ public class Engine implements FileFilter, AutoCloseable {
                 }
             }
         }
-        for (AnalysisPhase phase : AnalysisPhase.values()) {
+        for (AnalysisPhase phase : mode.phases) {
             final List<Analyzer> analyzerList = analyzers.get(phase);
 
             for (Analyzer a : analyzerList) {
@@ -788,7 +788,7 @@ public class Engine implements FileFilter, AutoCloseable {
      */
     public List<Analyzer> getAnalyzers() {
         final List<Analyzer> ret = new ArrayList<>();
-        for (AnalysisPhase phase : AnalysisPhase.values()) {
+        for (AnalysisPhase phase : mode.phases) {
             final List<Analyzer> analyzerList = analyzers.get(phase);
             ret.addAll(analyzerList);
         }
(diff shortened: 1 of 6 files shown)


jeremylong:

dependency-check-core/src/main/java/org/owasp/dependencycheck/Engine.java
```diff
@@ -813,7 +880,9 @@ private void throwFatalExceptionCollection(String message, Throwable throwable,
      */
     public synchronized void writeReports(String applicationName, String groupId, String artifactId,
             String version, File outputDir, String format) throws ReportException {
-
+        if (mode == Mode.EVIDENCE_COLLECTION) {
```
jeremylong, line 883: Wouldn't we want to be able to generate a report containing the collected evidence (XML/JSON)? The velocity templates may need to be updated slightly so that the dependency objects can be fully reconstituted.  Thoughts?

jeremylong (approved):
Thanks - this is a great set of changes that will assist in using dependency-check in new ways. Only one question around the report generation if in evidence collection mode.

dependency-check-core/src/main/java/org/owasp/dependencycheck/Engine.java
markrekveld: > Wouldn't we want to be able to generate a report containing the collected evidence (XML/JSON)? The velocity templates may need to be updated slightly so that the dependency objects can be fully reconstituted. Thoughts?

Yeah for sure, we can use some lib like Jackson to output a JSON file with the dependencies when generating a report in evidence collection execution.

This would also open the door to import this JSON when doing evidence processing, maybe with an argument on the CLI.

Do you want me to add the report generation to this PR?

markrekveld:
@jeremylong can you give an update if we can expect this PR to be merged and released soon?

jeremylong:
@markrekveld at the latest the next release will be on July 23rd.

markrekveld:
@jeremylong cool thanks for excepting and merging this PR.

Issue #799, opened 2017-07-12
NPE in maven-plugin when system-scoped dependency is not resolved


aikebah:
While running DependencyCheck 2.0.1 against their project Arbi Sookazian encountered a NullPointerException in the BaseDependencyCheckMojo running mvn dependency-check:aggregate

[](https://groups.google.com/d/msg/dependency-check/II075ApylF8/9X8wuezTAgAJ)

[gist of the MVN log](https://gist.github.com/aikebah/2da08fc09b73de4a7243d08c2ff457b2)

PR #800, opened 2017-07-12
Fix issue #799 - Initialize exCol to prevent NPE

aikebah:
Fixes issue #799 

Initialize the exCol when still `null` prior to adding the DependencyNotFoundException for a system dependency

no


jeremylong:

jeremylong (approved):

Issue #793, opened 2017-07-10
java.util.MissingResourceException when trying to use owasp plugin with Postgres


evg1001:
When I try to use version 2.0.1 of the owasp dependency check plugin for maven it fails with error:

Can't find resource for bundle java.util.PropertyResourceBundle, key MERGE_PROPERTY -> [Help 1]

For full error log please check here:
https://gist.github.com/evg1001/e1a01a50ccb5602e7164fc1db6b5033a

jeremylong:
@evg1001 Any chance you could re-run the failed execution with the `-X` option and look for the entry in the log file: 

`Database product: ` 

Then post the line from the log file here?

stefanneuhaus:
I just observed the "Database product:" line in the reported error log: https://gist.github.com/evg1001/e1a01a50ccb5602e7164fc1db6b5033a#file-maven_owasp_dependency_plugin_error-txt-L305

PR #803, opened 2017-07-13
Fix MissingResourceException "MERGE_PROPERTY" for Postgres

stefanneuhaus:
This fixes issue #793.

Root cause is that the resource bundle for Postgres could not be resolved, as
* the database driver reported the database product to be "PostgreSQL"
* a custom Locale (used for defining the ResourceBundle) uses a lower-cased identifier/language
* you are using a case-sensitive file-system

I also did some minor refactorings in order to improve the logging in case of database connection issues.


jeremylong:

jeremylong:
@stefanneuhaus thanks again for this PR!

Issue #815, opened 2017-07-20
maven plugins failOnError does not work for aggregate goal


TorstenKruse:
I'm using maven plugin version 2.0.1.

The plugin supports failOnError to set exit code different from 0 in any case of error.

If I call "mvn dependency-check:check" with a broken suppressions file, this works fine and the result will be BUILD_FAILURE.

But if I call "mvn dependency-check:aggregate" it does not work and I get BUILD_SUCCESS.

Would be nice to see this changed. Thanks.

PR #819, opened 2017-07-22
Fix Error Reporting in Maven Aggregate Goal 

jeremylong:
## Fixes Issue #815

The maven aggregate goal does not fail the build (in all cases). This PR is to update the aggregate to match the error reporting found in the `check` goal.

yes - an integration test has been added


jeremylong:

Issue #773, opened 2017-06-23
Maven plugin ignores standard directory conventions


stevespringett:
The Maven plugin doesn't seem to support the scanning of directories like the CLI does. This causes problems when using standard Maven conventions like src/main/webapp and src/main/resources where scannable files may reside such as javascript, package.json, and .NET assemblies.

For webapps, scanning the resulting war after the fact is not an option as some of the evidence may have been altered or omitted during assembly (i.e. the minification of javascript or the omission of package.json).

IMO, the Maven plugin should be able to scan all directories defined through standard Maven conventions. Defaults should be provided and enabled by default and overridable through configuration.

jeremylong:
Completely agree. This should be added to the gradle plugin as well.

jeremylong:
When implementing consider using [FileSets](https://maven.apache.org/shared/file-management/fileset.html) to allow control over what is scanned.

stevespringett:
I've added this to the Maven plugin. I'll let someone more familiar with Groovy add it to the Gradle plugin.

PR #820, opened 2017-07-22
Adds Maven support for specifying multiple filesets in a 'scanSet'. #773

stevespringett:
Adds support for specifying multiple filesets in a a scanSet. #773. Defaults to standard Maven conventions of src/main/resources, src/main/filters, and src/main/webapp if not specified.

### DIFF
--- a/dependency-check-maven/src/main/java/org/owasp/dependencycheck/maven/BaseDependencyCheckMojo.java
+++ b/dependency-check-maven/src/main/java/org/owasp/dependencycheck/maven/BaseDependencyCheckMojo.java
@@ -47,6 +47,8 @@ import org.apache.maven.shared.artifact.resolve.ArtifactResolverException;
 import org.apache.maven.shared.dependency.graph.DependencyGraphBuilder;
 import org.apache.maven.shared.dependency.graph.DependencyGraphBuilderException;
 import org.apache.maven.shared.dependency.graph.DependencyNode;
+import org.apache.maven.shared.model.fileset.FileSet;
+import org.apache.maven.shared.model.fileset.util.FileSetManager;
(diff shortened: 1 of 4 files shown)


jeremylong:

dependency-check-maven/src/main/java/org/owasp/dependencycheck/maven/BaseDependencyCheckMojo.java
```diff
@@ -748,6 +760,36 @@ private ExceptionCollection collectDependencies(Engine engine, MavenProject proj
...
+
+        // Define the default FileSets
+        if (scanSet == null || scanSet.length == 0) {
+            FileSet resourcesSet = new FileSet();
+            FileSet filtersSet = new FileSet();
+            FileSet webappSet = new FileSet();
+            try {
+                resourcesSet.setDirectory(new File(project.getBasedir(), "src/main/resources").getCanonicalPath());
+                filtersSet.setDirectory(new File(project.getBasedir(), "src/main/filters").getCanonicalPath());
+                webappSet.setDirectory(new File(project.getBasedir(), "src/main/webapp").getCanonicalPath());
```
jeremylong, line 772: should dependency-check be looking at the src or target directory?  Only wondering as some build plugins will download and drop additional JS into the target. Scanning both may be okay as if the hash matches it will be merged to a single dependency (because the hashes match). However, scanning both could give rise to duplicates because I've seen other build plugins compress  and/or combine JS from the src before placing it in the target.

jeremylong (approved):
While I added a comment/question about the `src` vs. `target` for scanning - this is likely a build by build decision that teams would need to make. Leaving it as `src` is likely appropriate as long as this is clearly documented.

Issue #828, opened 2017-07-29
saveProperty() fails on Postgres


stefanneuhaus:
Saving properties via `CveDB.saveProperty()` fails on Postgres:
```
Unable to save property 'NVD CVE 2005' with a value of '1501345766000' to the database
MERGE_PROPERTY failed
org.postgresql.util.PSQLException: ERROR: syntax error at or near "CALL"
  Position: 1
        at org.postgresql.core.v3.QueryExecutorImpl.receiveErrorResponse(QueryExecutorImpl.java:2455)
        at org.postgresql.core.v3.QueryExecutorImpl.processResults(QueryExecutorImpl.java:2155)
        at org.postgresql.core.v3.QueryExecutorImpl.execute(QueryExecutorImpl.java:288)
        at org.postgresql.jdbc.PgStatement.executeInternal(PgStatement.java:430)
        at org.postgresql.jdbc.PgStatement.execute(PgStatement.java:356)
        at org.postgresql.jdbc.PgPreparedStatement.executeWithFlags(PgPreparedStatement.java:168)
        at org.postgresql.jdbc.PgPreparedStatement.executeUpdate(PgPreparedStatement.java:135)
        at org.owasp.dependencycheck.data.nvdcve.CveDB.saveProperty(CveDB.java:517)
```

I verified with a SQL client that the following statement fails with the same error:
```
CALL save_property('NVD CVE sns','1501345763000');
```

Tested with DependencyCheck 2.1.0 and Postgres 9.6.2.

PR #829, opened 2017-07-29
Fix save_properties on Postgres

stefanneuhaus:
Fixes issue #828.

Note that I had to call `execute()` instead of `executeUpdate()`on the PreparedStatement as otherwise an exception was thrown:
```
org.postgresql.util.PSQLException: A result was returned when none was expected.
        at org.postgresql.jdbc.PgPreparedStatement.executeUpdate(PgPreparedStatement.java:140)
        at org.owasp.dependencycheck.data.nvdcve.CveDB.saveProperty(CveDB.java:517)
```

I verified the fix with the latest Postgres release and with 9.2 (Postgres 9.0 and 9.1 were not tested as the definition of the stored function `save_property`in `initialize_postgres.sql` is not valid with these releases.).

I also verified that H2 (the only other DB that defines MERGE_PROPERTY) still works with the `execute()` call.


jeremylong:

Issue #835, opened 2017-08-08
NspAnalyzer fails to parse package.json


ktham:
`NspAnalyzer` throws a Json parsing exception when parsing `package.json`

Here is a snippet from one scan:

> WARN org.owasp.dependencycheck.AnalysisTask - An unexpected error occurred during analysis of '/root/Iterable/target/web/classes/main/META-INF/resources/webjars/iterable/1.0-SNAPSHOT/lib/bootstrap/package.json' (Node Security Platform Analyzer): org.glassfish.json.JsonObjectBuilderImpl$JsonObjectImpl cannot be cast to javax.json.JsonString
> WARN org.owasp.dependencycheck.AnalysisTask - An unexpected error occurred during analysis of '/root/Iterable/target/web/classes/main/META-INF/resources/webjars/iterable/1.0-SNAPSHOT/lib/moment-range/package.json' (Node Security Platform Analyzer): org.glassfish.json.JsonObjectBuilderImpl$JsonObjectImpl cannot be cast to javax.json.JsonString
> INFO org.owasp.dependencycheck.Engine - Finished Node Security Platform Analyzer (1 seconds)
> INFO org.owasp.dependencycheck.Engine - Finished Vulnerability Suppression Analyzer (0 seconds)
> INFO org.owasp.dependencycheck.Engine - Finished Dependency Bundling Analyzer (0 seconds)
> INFO org.owasp.dependencycheck.Engine - Analysis Complete (17 seconds)
>// ...
> org.glassfish.json.JsonObjectBuilderImpl$JsonObjectImpl cannot be cast to javax.json.JsonString
> org.glassfish.json.JsonObjectBuilderImpl$JsonObjectImpl cannot be cast to javax.json.JsonString
>   at org.owasp.dependencycheck.Engine.analyzeDependencies(Engine.java:649)

PR #836, opened 2017-08-08
Fix NspAnalyzer

ktham:
## Root Cause
`NspAnalyzer` fails to parse `package.json` files that contain the following keys:

```javascript

{
  // Should be parsed as a JsonArray but was parsed as JsonObject
  "bundledDependencies": [
    "node-pre-gyp"
  ],

  // Older libraries use license objects rather than JsonString
  {
  "license": {
    "type": "MIT",
    "url": "https://github.com/twbs/bootstrap/blob/master/LICENSE"
  }
}
```

## Changes

* `NspAnalyzer`: parse `"bundleDependencies"` as `JsonArray`
* `NspAnalyzer`: parse `"bundledDependencies"` as `JsonArray`
* `NspAnalyzer`: parse `"license"` as `JsonObject` and subelement `"type"` as `JsonString` in cases where older libraries used a [license object](https://docs.npmjs.com/files/package.json#license)

### DIFF
--- a/dependency-check-core/src/main/java/org/owasp/dependencycheck/analyzer/NspAnalyzer.java
+++ b/dependency-check-core/src/main/java/org/owasp/dependencycheck/analyzer/NspAnalyzer.java
@@ -43,6 +43,7 @@ import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import javax.json.Json;
+import javax.json.JsonArray;
 import javax.json.JsonException;
 import javax.json.JsonObject;
 import javax.json.JsonObjectBuilder;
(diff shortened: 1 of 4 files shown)


jeremylong:

stevespringett:
@ktham Thanks for the excellent PR - especially the additional (and much needed) tests. I've verified in my environment the changes. Are these the only changes necessary to fix #835?

ktham:
@stevespringett no problem, I started to use this library and needed this issue fixed 😄  It may not be exhaustive, but those are the only changes that is needed to fix #835 for the set of libraries that was scanned in my project

jeremylong:
@ktham Thank you for the PR! I'm trying to solve one more, non-related, issue and then I plan on pushing a new release (hopefully this weekend).

ktham:
@jeremylong Just wondering if you're close to cutting a release or if there are more things you're trying to squeeze in?

jeremylong:
Recently discovered a threading issue that needs to be dealt with - which is what was causing the delay. We will likely release a version this week and then spend the next dev cycle re-working some code to resolve the threading issues.

Issue #827, opened 2017-07-28
Error occurred connecting to the local database


giovaneliberato:
Hi everyone!

I'm running dependency check as a part of a CI pipeline in a Jenkins instance and suddenly this connection error started to happen. This is the second time we see this local database, but in the first time the job went back to normal after a few hours.

Here's the plugin configuration. Note that I added **`<connectionString>`** after trying to solve this error based on #410 

```
<plugin>
    <groupId>org.owasp</groupId>
    <artifactId>dependency-check-maven</artifactId>
    <version>2.0.1</version>
    <configuration>
        <failBuildOnCVSS>8</failBuildOnCVSS>
        <skipProvidedScope>true</skipProvidedScope>
        <outputDirectory>${project.build.directory}/reports/owasp</outputDirectory>
        <nuspecAnalyzerEnabled>false</nuspecAnalyzerEnabled>
        <assemblyAnalyzerEnabled>false</assemblyAnalyzerEnabled>
        <connectionString>jdbc:h2:file:%s;FILE_LOCK=FS;AUTOCOMMIT=ON;</connectionString>
    </configuration>
    <executions>
        <execution>
            <goals>
                <goal>check</goal>
            </goals>
        </execution>
    </executions>
</plugin>
```

Here's the stacktrace that jenkins is outputting 
```
[ERROR] Failed to execute goal org.owasp:dependency-check-maven:2.0.1:check (default-cli) on project default: An exception occurred connecting to the local database. Please see the log file for more details. Unable to connect to the database -> [Help 1]
org.apache.maven.lifecycle.LifecycleExecutionException: Failed to execute goal org.owasp:dependency-check-maven:2.0.1:check (default-cli) on project default: An exception occurred connecting to the local database. Please see the log file for more details.
	at org.apache.maven.lifecycle.internal.MojoExecutor.execute(MojoExecutor.java:216)
	at org.apache.maven.lifecycle.internal.MojoExecutor.execute(MojoExecutor.java:153)
	at org.apache.maven.lifecycle.internal.MojoExecutor.execute(MojoExecutor.java:145)
	at org.apache.maven.lifecycle.internal.LifecycleModuleBuilder.buildProject(LifecycleModuleBuilder.java:116)
	at org.apache.maven.lifecycle.internal.builder.multithreaded.MultiThreadedBuilder$1.call(MultiThreadedBuilder.java:188)
[...]

PR #839, opened 2017-08-13
H2 Upgrade

jeremylong:
## Fixes Issues

* https://github.com/jeremylong/dependency-check-gradle/issues/46
* https://github.com/jeremylong/DependencyCheck/issues/827

## Description of Change

In some cases dependency-check is forced to use a newer version of the H2 driver (Jenkins Pipelines, new versions of gradle, etc.) because the build engine is using a newer version of H2. As such, the previously used locking mechanism is not available. This PR changes the connection string so that the database does not perform any locking and allows dirty reads. The update process, executed by default each time ODC is run contains a locking mechanism so that only one instance of ODC can update the database at any given time.

Existing test cases cover these changes.


jeremylong:

PR #840, opened 2017-08-13
Improve Lucene Usage

jeremylong:
This PR updates the usage of Lucene - removing some unnecessary code and making the use a little more standard.  Additionally, the list of stop words was increased to included common entries in the NVD/CPE data that may not be listed in the dependencies evidence (project, inc, foundation, etc.).


jeremylong:

PR #848, opened 2017-08-18
Catch IOException parsing manifest

mealingr:
## Description of Change

Catch and log `IOException` when parsing manifest for two reasons:

1.  To prevent an `ExceptionCollection` exception being thrown by `engine.analyzeDependencies()` when analyzing unreadable jars/zips. This is useful because these errors could be common (x) and unnecessarily bloat any `ExceptionCollection`. Also it would make it easier to determine whether to continue or not especially for external uses of `engine.analyzeDependencies()`.

(x) when compressing files on macOS it may create mirror `._` files (same filename except with the `._` prefix), these files aren't actual copies they just contain file information, but if `engine.analyzeDependencies()`  sees `._some_file.jar` it will fail to parse its manifest throwing an `IOException` because it's unreadable.

2. To be consistent with other usages of `try (JarFile jar = new JarFile(dependency.getActualFilePath()))`

https://github.com/jeremylong/DependencyCheck/blob/fb2b3159e8dedaad06cb3b6f667975616fbd848f/dependency-check-core/src/main/java/org/owasp/dependencycheck/analyzer/JarAnalyzer.java#L313
https://github.com/jeremylong/DependencyCheck/blob/fb2b3159e8dedaad06cb3b6f667975616fbd848f/dependency-check-core/src/main/java/org/owasp/dependencycheck/analyzer/JarAnalyzer.java#L914

Edit: this has since been updated to incorporate suggestions - see https://github.com/jeremylong/DependencyCheck/pull/848#issuecomment-323534760
https://github.com/jeremylong/DependencyCheck/pull/848#issuecomment-323737697

*yes*


jeremylong:

jeremylong:
Completely understand the reason why you would want this. However, instead of just removing the exception it would be better to see if the file name, in this case `._some.jar`, starts with a `._` and if the dependency collection contains a `some.jar` - if that is the case skip and remove the `._some.jar`.  Another option, if we run into `._some.jar` and the magic bytes don't match a zip - skip and remove it?  Thoughts?

mealingr:
> Completely understand the reason why you would want this. However, instead of just removing the exception it would be better to see if the file name, in this case `._some.jar`, starts with a `._` and if the dependency collection contains a `some.jar` - if that is the case skip and remove the `._some.jar`. Another option, if we run into `._some.jar` and the magic bytes don't match a zip - skip and remove it? Thoughts?

These both sound like good ideas, I've made some changes in 5d87dc2 to show how this could work. Note especially the early return at https://github.com/mealingr/DependencyCheck/blob/5d87dc2942c9319eca3533d8a7b54a84be411d2d/dependency-check-core/src/main/java/org/owasp/dependencycheck/analyzer/JarAnalyzer.java#L255
which allows the `AnalysisException` to be avoided. Is there a reason it wasn't returning early before? If so maybe the early return should be restricted to just when macOS metadata files or non-zip files are encountered.

jeremylong:
Thanks for the PR and the quick response to the requested changes! I will be pushing a new release today or tomorrow.

Issue #842, opened 2017-08-14
Support skipping maven report generation


mattnelson:
When declaring the dependency check plugin in a parent pom, I would like to disable report generation for specific modules. Specifically in testing modules since I do not care about the vulnerability of the dependencies used for the tests and to help with build times.

jeremylong:
In your test modules add the following to your build:

```xml
<plugin>
  <groupId>org.owasp</groupId>
  <artifactId>dependency-check-maven</artifactId>
  <configuration>
    <skip>true</skip>
  </configuration>
</plugin>
```

AndrewJCarr:
The issue stems from having the report sets configured in the parent pom and being unable to override the report set in maven.

```
<reporting><plugins>
<plugin>
    <groupId>org.owasp</groupId>
    <artifactId>dependency-check-maven</artifactId>
    <reportSets>
        <reportSet>
            <reports>
                <report>aggregate</report>
            </reports>
        </reportSet>
    </reportSets>
</plugin>
</plugins></reporting>
```

The enabling of the skip option (as above) works for the build aspect but does not have any effect during site generation. It performs the download/analysis of the NVD database files during the site generation if they have not yet occured. This impacts the site generation times for the modules that have no need for OWASP scanning (mainly integration tests) if we leverage a common parent pom. 

It appears that if we place the following behavior in BaseDependencyCheckMojo#generate, the skip configuration would be honored during site generation. Would you be open to a pull request for this behavior?

```
if (skip) {
    getLog().info("Skipping " + getName(Locale.US) + " report generation");
}
```

jeremylong:
Yes, please submit a PR - we missed this when we originally added the skip. 

Thanks!

AndrewJCarr:
No problem. I'll put one up in a few days or so - need to carve out some time to get the project compiling, etc.

AndrewJCarr:
I opened PR #858 to add the skip functionality. Sorry it took so long to get a tiny fix submitted.

PR #858, opened 2017-08-25
#842 Honor skip configuration in reports

AndrewJCarr:
## Description of Change

In dependency-check-maven, checks the skip configuration option during report generation and honors the value.

No. I was unable to find a suitable location for a report test in the maven module. I did test it manually and locally. The report generation was successfully skipped by including the following into the `reporting` section of my child pom.

````
<plugin>
  <groupId>org.owasp</groupId>
  <version>2.1.1-SNAPSHOT</version>
  <configuration>
    <skip>true</skip>
  </configuration>
 </plugin>
````

I also tested `skip` being set to `false` as well as not setting it. Results in those cases were also as expected (and had the reports generate)


jeremylong:

jeremylong:
Thanks again for the PR.  I'm hoping the next release won't take that long but there are some threading issues that need to be resolved. Regardless, this will be included in the next release.

PR #862, opened 2017-08-29
don't stop SuppressionAnalyzer if one suppression file can't be load

stboissdev:
## Fixes Issue #
When you have multiple suppression files and a file can not be loaded, the following files will never be analyzed.

## Description of Change

when a file can't be load, no SuppressionParseException was throw , but a warning was logged with suppressionFilePath 

*yes*


jeremylong:

jeremylong:
@stboiss I've been considering this PR for a while now and I don't think it should be merged. Specifically, if a suppression file cannot be read that means the configuration is wrong and the build/execution should fail.  If you don't want the build to fail because the dependency-check-maven plugin (or gradle, ant, etc.) then you can configure them with `failOnError=false`.  Would that work?

stboissdev:
@jeremylong even with this option, if one file can not be loaded, the following files will not be loaded.

jeremylong:
Changes were made in master that will allow other suppression files to be loaded even though one of the files was invalid. Does that combined with `failOnError=false` satisfy your issue?

stboissdev:
OK rules are loaded but the analysisTask wil not execute. it would be necessary to differentiate fatal or non-fatal InitializationException.

jeremylong:
The concept of a fatal initialization exception has been added and will be included in the next release.

jeremylong:
This update was included in the 3.0.0 release.

Issue #870, opened 2017-09-04
Question about node


javixeneize:
Hi

Question about node packages. How is this analysed, and which option do you recommend?

Let me explain

If i send as scanpath just package.json, seems like it analyses only first level dependencies. In my scenario, no vulnerabilities detected

If i leave scanpath empty, it analyses all the dependencies in the project, and dependencies of the dependencies (it is, all the package.json in my workspace), and then, vulnerabilities are reported

Having said that... what is the right way to scan a node project?

I mean, if i have a dependency A, that uses a dependency B, that uses C, and C has a vulnerability, then, i understand that B has the same vulnerability as C, and also A has the same vulnerability, isnt it?

Thanks

jeremylong:
That is one of the interesting questions for software composition analysis.
Does a bug (regardless of security) from a dependency's transitive
dependency affect the application. Unfortunately, the answer is that it
would take a lot of research to figure that out for each vulnerability. In
your example, does A use B in a way that uses the code in B that uses the
vulnerable code in C? Or alternatively, is the vulnerable code in C
automatically exposed just due to it being some framework that auto-exposes
the vulnerable bits - regardless of the use by A and B.

This is one of the reasons that, on top of software composition analysis,
some companies are also looking at RASP technologies. I can't give a good
view/opinion into how RASP (or other tools) help protect against vulnerable
libraries.

Best advice, keep your libraries as current as you can (i.e. patch often -
not just for security), use layered protections, etc.  One of the reasons
you want to patch often is that when the critical security bug comes out -
very few patches are back ported, the resolution is almost always upgrade
to the safe version. If you've let your dependencies languish that upgrade
to a safe version could become much more time consuming.

--Jeremmy

stevespringett:
In addition to what Jeremy wrote, I also want to point out the multiple ways to scan a node project.

The node analyzer which relies solely on data in the NVD is typically not that useful since the Javascript community doesn't self-report (typically) and relies primarily on commits to track vulnerabilities. As a result of the ecosystem being relatively immature, the Node analyzer isn't that useful.

The NSP analyzer on the other hand, is very useful as it relies on a lot of manually curated data.

[...]

PR #875, opened 2017-09-10
Added support for retired analyzers. Retired NodePackageAnalyzer.

stevespringett:
Jeremy, The idea in retiring analyzers rather than simply removing them is to:

- allow users to continue to use dependency-check in rare scenarios where the analyzer may actually be useful
- prevent users from discovering an analyzer does not exist and spending time creating something we've previously deemed unsuccessful and removed

A @Retired annotation was created and the same mechanisms that provide the ability to enable/disable experimental analyzers have been applied to retired analyzers as well.

I envision the path from experimental analyzer to take one of two paths - they will either graduate to production, or be retired.


jeremylong:

stevespringett:
See also #870

jeremylong (approved):

brianf:
Fwiw, for our use case we still need the other node analyzer, but I understand the desire to retire these for the normal dep check use case. This is just an fyi.

Issue #860, opened 2017-08-28
UpdateException: Cause: Error updating 'CVE-2007-5199'


fpmosley:
### Error
Receiving the following error when executing a Dependency-Check scan, in a Jenkins job, using the Jenkins plugin.

```
[DependencyCheck] OWASP Dependency-Check Plugin v2.1.0
[DependencyCheck] Executing Dependency-Check with the following options:
[DependencyCheck]  -name = mobilecloud-owasp
[DependencyCheck]  -scanPath = /var/lib/jenkins/workspace/Mobile/Test-Jobs/mobilecloud-owasp
[DependencyCheck]  -outputDirectory = /var/lib/jenkins/workspace/Mobile/Test-Jobs/mobilecloud-owasp
[DependencyCheck]  -dataDirectory = /var/lib/jenkins/workspace/Mobile/Test-Jobs/mobilecloud-owasp/dependency-check-data
[DependencyCheck]  -dataMirroringType = none
[DependencyCheck]  -isQuickQueryTimestampEnabled = true
[DependencyCheck]  -jarAnalyzerEnabled = true
[DependencyCheck]  -nodeJsAnalyzerEnabled = false
[DependencyCheck]  -nspAnalyzerEnabled = true
[DependencyCheck]  -composerLockAnalyzerEnabled = false
[DependencyCheck]  -pythonDistributionAnalyzerEnabled = false
[DependencyCheck]  -pythonPackageAnalyzerEnabled = false
[DependencyCheck]  -rubyBundlerAuditAnalyzerEnabled = false
[DependencyCheck]  -rubyGemAnalyzerEnabled = false
[DependencyCheck]  -cocoaPodsAnalyzerEnabled = false
[DependencyCheck]  -swiftPackageManagerAnalyzerEnabled = false
[DependencyCheck]  -archiveAnalyzerEnabled = true
[DependencyCheck]  -assemblyAnalyzerEnabled = true
[DependencyCheck]  -centralAnalyzerEnabled = true
[DependencyCheck]  -nuspecAnalyzerEnabled = true
[DependencyCheck]  -nexusAnalyzerEnabled = false
[DependencyCheck]  -autoconfAnalyzerEnabled = false
[DependencyCheck]  -cmakeAnalyzerEnabled = false
[DependencyCheck]  -opensslAnalyzerEnabled = false
[DependencyCheck]  -showEvidence = true
[DependencyCheck]  -formats = XML HTML 
[...]

Issue #843, opened 2017-08-15
Error updating 'CVE-2003-1540'


giacgbj:
…and today the error is:

```
Aug 15, 2017 7:18:10 AM INFO org.owasp.dependencycheck.Engine doUpdates

Checking for updates

Aug 15, 2017 7:18:10 AM FINE org.owasp.dependencycheck.utils.Settings getDataFile

Settings.getDataFile() - file: '/home/jenkins/workspace/FooProject/dependency-check-data'

Aug 15, 2017 7:18:10 AM FINE org.owasp.dependencycheck.utils.Settings getConnectionString

Connection String: 'jdbc:h2:file:/home/jenkins/workspace/FooProject/dependency-check-data/dc;AUTOCOMMIT=ON;FILE_LOCK=SERIALIZED;'

Aug 15, 2017 7:18:10 AM FINE org.owasp.dependencycheck.utils.Settings getDataFile

Settings.getDataFile() - file: '/home/jenkins/workspace/FooProject/dependency-check-data'

Aug 15, 2017 7:18:10 AM FINE org.owasp.dependencycheck.data.update.NvdCveUpdater initializeExecutorServices

#download   threads: 12

Aug 15, 2017 7:18:10 AM FINE org.owasp.dependencycheck.data.update.NvdCveUpdater initializeExecutorServices

#processing threads: 8

Aug 15, 2017 7:18:10 AM INFO org.owasp.dependencycheck.data.update.NvdCveUpdater getUpdatesNeeded

starting getUpdatesNeeded() ...

Aug 15, 2017 7:18:10 AM FINE org.owasp.dependencycheck.data.update.NvdCveUpdater$TimestampRetriever call

Checking for updates from: http://urlWhichWorks/nistDataMirror/nvdcve-2.0-Modified.xml.gz

Aug 15, 2017 7:18:10 AM FINE org.owasp.dependencycheck.data.update.NvdCveUpdater$TimestampRetriever call

Checking for updates from: http://urlWhichWorks/nistDataMirror/nvdcve-2.0-2017.xml.gz

Aug 15, 2017 7:18:10 AM INFO org.owasp.dependencycheck.data.update.nvd.DownloadTask call

Download Started for NVD CVE - Modified

Aug 15, 2017 7:18:10 AM FINE org.owasp.dependencycheck.utils.Downloader fetchFile

Attempting download of http://urlWhichWorks/nistDataMirror/nvdcve-2.0-Modified.xml.gz

[...]

Issue #833, opened 2017-08-05
Unable to access the local database


giacgbj:
Logs: https://gist.github.com/giacgbj/84dd509c2c4916e18b7bfc9827b13d0a

- I've run the job many times before posting this issue
- I'm using an internal mirror which works
- the directory /home/jenkins/workspace/FooProject/dependency-check-data is writeable (775) by the user which launches Jenkins and it's removed each time the job is executed
- the dbs permissions are:
 -rw-rw-r-- 1 jenkins jenkins 2107392 Aug  5 22:19 dc.h2.db
 -rw-rw-r-- 1 jenkins jenkins     116 Aug  5 22:19 dc.lock.db
 -rw-rw-r-- 1 jenkins jenkins   43174 Aug  5 22:19 dc.trace.db

rompic:
sounds a little bit like https://github.com/jeremylong/DependencyCheck/issues/100 to me.
https://github.com/jeremylong/DependencyCheck/issues/100#issuecomment-38585732 -> "To recover from your current situation, you can either delete the whole data directory, or you might be able to just delete the .lock file in the data directory."

giacgbj:
If the .lock file is in "dependency-check-data" directory, that directory is removed before each build.

stevespringett:
Ensure you have auto update enabled in the job. If you're removing the project workspace for each build, autoupdate will need to be enabled so it can rebuild the database each time. Assuming that is done, do you have either of these plugins installed?

- https://wiki.jenkins.io/display/JENKINS/Pipeline+Maven+Plugin
- https://wiki.jenkins.io/display/JENKINS/Liquibase+Runner

These potentially could cause conflicts as they use beta versions of H2 instead of the stable 1.3 branch that Dependency-Check and the Jenkins H2 plugin uses.

rompic:
Having a separate job that updates the database regularly is a massive performance Improvement as far as I know.
See E.g. https://medium.com/@PrakhashS/checking-vulnerabilities-in-3rd-party-dependencies-using-owasp-dependency-check-plugin-in-jenkins-bedfe8de6ba8
https://blog.dominikschadow.de/2015/07/using-owasp-dependency-check-as-jenkins-plugin/

giacgbj:
@stevespringett Autoupdate is enabled (from the gist, i think it's the autoupdate='true' property).
"Unfortunately", I don't have those plugins installed.

[...]

Issue #802, opened 2017-07-13
Error reading CPE data


giacgbj:
I'm using the version 2.0.1.1.

The exception is:
```
[DependencyCheck] Analyzing Dependencies
[DependencyCheck] One or more exceptions were thrown while executing Dependency-Check
[DependencyCheck] Exception Caught: org.owasp.dependencycheck.data.update.exception.UpdateException
[DependencyCheck] Cause: Error updating 'CVE-1999-1185'
[DependencyCheck] Message: org.xml.sax.SAXException: Error updating 'CVE-1999-1185'
[DependencyCheck] org.owasp.dependencycheck.data.nvdcve.DatabaseException: Error updating 'CVE-1999-1185'
[DependencyCheck] Exception Caught: org.owasp.dependencycheck.exception.InitializationException
[DependencyCheck] Cause: org.owasp.dependencycheck.data.cpe.IndexException: Error reading CPE data
[DependencyCheck] Message: An exception occurred accessing the database
```

jeremylong:
Is this reproducible on your system (today)? Before testing again could you run the `dependencyCheckPurge` task and then run `dependencyCheckAnalyze`?
 
If it still fails, what version of Gradle is being used? OS?

giacgbj:
It seems the issue solved by itself. It could have simply been a network issue.

giacgbj:
...a network issue which happens quite often: every job Jenkins ran failed with a more generic error (probably due to version 2.1.0):
```
05:58:28 [DependencyCheck] Analyzing Dependencies
05:58:42 [DependencyCheck] One or more exceptions were thrown while executing Dependency-Check
05:58:42 [DependencyCheck] Exception Caught: org.owasp.dependencycheck.exception.InitializationException
05:58:42 [DependencyCheck] Cause: org.owasp.dependencycheck.data.cpe.IndexException: Error reading CPE data
05:58:42 [DependencyCheck] Message: An exception occurred accessing the database
```

I'm using Maven plugin on CentOS Linux 7.

Rerunning the jobs now, after about 3 hours the error disappears.

giacgbj:
I probably reproduced the issue. I'm using a mirror, so the files are always the same (until tomorow, when it will be updated).

On Jenkins I see:
```
00:00:49.007 [DependencyCheck] OWASP Dependency-Check Plugin v2.1.0
00:00:49.012 [DependencyCheck] Executing Dependency-Check with the following options:
[...]

PR #877, opened 2017-09-11
Thread Safety Updates

jeremylong:
## Fixes Issues
#860, #843, #833, #802, #690

## Description of Change

In Jenkins and other build systems - parallel builds were causing failures. This PR has refactored a lot of code to correct the threading issues. More detailed discussion of the changes can be found below:

The `Engine` object was updated so that when the list of dependencies is retrieved - it is an unmodifiable array of dependencies. If dependencies need to be added or removed methods exist on the Engine object itself to perform these functions.

`Settings` and `CveDB` singletons were refactored - analyzers can obtain references to these via the `Engine` object. 

An additional initialization step was added to the `Analyzer` interface. When an analyzer is instantiated by the engine the calls will be done in this order: `initialize(settings)`, `prepare(engine)`, `analyze(dependency, engine)`, and finally `close()`.

`EvidenceCollection` was completely refactored and for all intensive purposes any custom analyzers that previously used this object (obtained via the `Dependency` object for vendor, product, and version collections) must be updated to utilize the dependency object itself to update evidence. Example:

```java
/* old */
dependency.getVendorEvidence().addEvidence(evidence);
/* new */
dependency.addEvidence(EvidenceType.VENDOR, evidence);
```

The `Dependency` object is now thread safe. As such, any lists used (e.g. evidence, identifiers) are
unmodifiable copies of the internal list. Any modifications must be done via the dependency object itself (e.g. `dependency.removeIdentifier(i)`).

Tests were updated appropriately.


jeremylong:

stevespringett:
Fantastic job!!!

I noticed that the Travis build seems to always fail on Java7 - does the same on some of my projects. Know the issue?

Have 3.0.0-SNAPSHOT's been pushed yet, and if not, when do you think that will happen?

jeremylong:
@stevespringett regarding travis-ci - I believe their trusty image for java 7 no longer supports oraclejdk7; you have to change your .travis.yml to use openjdk7.

Regarding merging this PR into master - that will likely happen soon (which will cause the 3.0.0-snapshots to be pushed to artifactory). I'm doing some more local testing with mono and ruby to validate that I haven't broken anything ;)  I actually may merge this in a few hours... Tomorrow morning at the latest - unless I find a bug.

jeremylong:
While other bugs may cause the build to randomly fail (see #897 and there might be an issue in the Ant Tasks unrelated to #897 or it is the exact same issue) - this PR is considered complete and will be merged.

Issue #874, opened 2017-09-09
Provide an ability to suppress dependencies CVE listing and simply fail the build in maven


vrozov:
Such ability may be used in CI builds (public Travis and Jenkins).

jeremylong:
Unless I'm misunderstanding your request - these features already exist.

* [Suppressing False Positives](https://jeremylong.github.io/DependencyCheck/general/suppression.html)
* [Maven Configuration](https://jeremylong.github.io/DependencyCheck/dependency-check-maven/configuration.html) - see failBuildOnCVSS or failBuildOnAnyVulnerability

vrozov:
The request is to add an option to suppress adding list of CVEs that caused build to fail when failBuildOnCVSS or failBuildOnAnyVulnerability options are used and display generic error message without providing details on CVEs.

jeremylong:
Apparently I need to update the documentation as `showSummary` has been a boolean option for a very long time. To suppress the summary in the maven build use:

```xml
<plugin>
    <groupId>org.owasp</groupId>
    <artifactId>dependency-check-maven</artifactId>
    <version>2.1.1</version>
    <configuration>
        <showSummary>false</showSummary>
    </configuration>
    <executions>
        <execution>
            <goals>
                <goal>check</goal>
            </goals>
        </execution>
    </executions>
</plugin>
```

vrozov:
Even when `showSummary` is false, list of dependencies with CVEs are added to the error message. For example:
```
[ERROR] Failed to execute goal org.owasp:dependency-check-maven:2.1.1:check (default-cli) on project <foo>:
[ERROR]
[ERROR] One or more dependencies were identified with vulnerabilities that have a CVSS score greater than '5.0':
[ERROR]
[ERROR] commons-beanutils-1.7.0.jar: CVE-2014-0114
[ERROR] commons-collections-3.2.1.jar: CVE-2015-6420
[ERROR] commons-httpclient-3.1.jar: CVE-2014-3577
[ERROR] netty-3.7.0.Final.jar: CVE-2014-3488, CVE-2014-0193
[ERROR] curator-recipes-2.6.0.jar: CVE-2016-5017
[ERROR] hadoop-auth-2.6.0.jar: CVE-2017-3162, CVE-2016-5393
[ERROR] zookeeper-3.4.6.jar: CVE-2016-5017
[ERROR] jetty-6.1.26.jar: CVE-2011-4461
```
[...]

PR #880, opened 2017-09-14
Enhance csv report

elenoir:
## Fixes Issue #

#809 

## Description of Change

- Edit Velocity template
- Add utils method to fill CSV

## Have test cases been added to cover the new functionality?

*yes*

PR #881, opened 2017-09-14
Hide failure msg show summary

elenoir:
## Fixes Issue #

#874 

## Description of Change

- Add conditions on Ant/Maven/Core modules


jeremylong:

jeremylong (approved):
This appears to include the fix from PR #880 - totally fine as I'm planning on merging that change as well.  Quick review of this PR and it looks good. I'll merge this after I finish with threading issue (see #877).

PR #883, opened 2017-09-14
Fix the package.swift DisplayFileName to be something recognizable and unique in a report

brianf:
so the report doesn't list dozens of "package.swift" entries

## Description of Change

Taking a similar approach as used in other analyzers that always have the same filename. Include the parent name since it likely is more recognizable. Without this, scanning many swift packages ends up with a report loaded with "package.swift" entries repeated each time.

*yes*


jeremylong:

jeremylong (approved):

brianf:
After looking at some of the other analyzers, I'm going to take a standardized approach. Pulling this pull for now.

PR #884, opened 2017-09-15
Adding version to the composer.lock displayFileName

brianf:
## Description of Change
Added the version to the displayFileName
Added basic test for composer parsing, including the new version
Changed output to debug

*yes*


jeremylong:

jeremylong (approved):

brianf:
About the most recent changes... in subsequent testing I discovered that there was almost always an extra, empty dependency left around for the composer.lock file. Since this never gets evidence added to it, it was pointless to leave in the report. I made sure this only gets removed when there are dependencies added from parsing the lock file, so nothing is accidentally removed.

brianf:
The build simply ran past the allocated 15 minutes?

brianf:
After looking at some of the other analyzers, I'm going to take a standardized approach. Pulling this pull for now.

PR #905, opened 2017-10-01
Nuget name improvements

brianf:
## Description of Change
One additional commit which adds the nuget Analyzer tests and name updates.

This is on top of my other pr since it needs the dependency object changes.

Yes


jeremylong:

PR #906, opened 2017-10-02
Fix build warnings and add default goal 

ottlinger:
## Description of Change

* Maven warning about wrong usage of prerequisites for non-maven-plugin projects. Use enforcer plugin instead to enforce at least Maven 3.1.
* Add defaultGoal to allow easier local usage by just calling mvn instead of mvn clean install.

*no*, just a Maven change.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -150,9 +150,6 @@ Copyright (c) 2012 - Jeremy Long
             <url>https://jeremylong.github.io/DependencyCheck/</url>
         </site>
     </distributionManagement>
-    <prerequisites>
-        <maven>3.1</maven>
-    </prerequisites>
     <build>
     <defaultGoal>clean install</defaultGoal>
         <pluginManagement>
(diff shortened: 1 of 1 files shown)


jeremylong:

ottlinger:
@jeremylong what except for green builds can verify that the patch is okay? Thanks.

pom.xml
```diff
@@ -190,6 +188,21 @@ Copyright (c) 2012 - Jeremy Long
...
+                    <executions>
+                        <execution>
+                          <id>enforce-minimal-maven-version</id>
+                          <goals>
+                              <goal>enforce</goal>
+                          </goals>
+                          <configuration>
+                              <rules>
+                                  <requireMavenVersion>
+                                      <version>3.1</version>
```
jeremylong, line 200: This is actually defined in dependency-check-maven's [pom.xml](https://github.com/jeremylong/DependencyCheck/blob/master/dependency-check-maven/pom.xml#L92). However, as you are pointing out - that needs to be updated.

jeremylong (approved):
I'll update the PR to move the enforcer rule to just the maven plugin.  Thanks!

jeremylong:
@ottlinger the automated comment that appears to be from me asking "can one of the admins verify this patch?" is something GH adds.

PR #909, opened 2017-10-02
Improve centralanalyzer stability

stefanneuhaus:
## Description of Change

I observed sporadic connection errors when fetching the MavenArtifacts from MavenCentral (500). I implemented a retry (5 retries, initial delay 500 msec, double delay each time) for fetching these. Only if the last retry fails, the CentralAnalyzer gets disabled.

yes:
* Unit tests for the retry
* Manual integration test

### DIFF
--- a/dependency-check-core/src/main/java/org/owasp/dependencycheck/analyzer/CentralAnalyzer.java
+++ b/dependency-check-core/src/main/java/org/owasp/dependencycheck/analyzer/CentralAnalyzer.java
@@ -210,7 +210,7 @@ public class CentralAnalyzer extends AbstractFileTypeAnalyzer {
      * Performs the analysis.
      *
      * @param dependency the dependency to analyze
-     * @param engine the engine
+     * @param engine     the engine
      * @throws AnalysisException when there's an exception during analysis
      */
     @Override
--- a/dependency-check-core/src/main/java/org/owasp/dependencycheck/analyzer/CentralAnalyzer.java
+++ b/dependency-check-core/src/main/java/org/owasp/dependencycheck/analyzer/CentralAnalyzer.java
@@ -75,6 +75,13 @@ public class CentralAnalyzer extends AbstractFileTypeAnalyzer {
     private static final String SUPPORTED_EXTENSIONS = "jar";
 
     /**
+     * There may be temporary issues when connecting to MavenCentral.
+     * In order to compensate for 99% of the issues, we perform a retry
+     * before finally raising the {@link #errorFlag}.
+     */
+    private static final int NUMBER_OF_TRIES = 5;
+
+    /**
      * The analyzer should be disabled if there are errors, so this is a flag to
      * determine if such an error has occurred.
      */
@@ -83,7 +90,8 @@ public class CentralAnalyzer extends AbstractFileTypeAnalyzer {
     /**
      * The searcher itself.
      */
-    private CentralSearch searcher;
+    CentralSearch searcher;
+
     /**
      * Field indicating if the analyzer is enabled.
      */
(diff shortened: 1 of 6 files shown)


jeremylong:

stefanneuhaus:
Verified in 36 analyze runs that the retry effectively improved stability: I encountered 24 "Internal server error" (500), out of which 23 could be compensated with the next try, one time a second retry was needed. The result was stable:
```
  36 Found 21 vulnerabilities in project sample-project: CVE-2012-6153, CVE-2013-4499, CVE-2014-0107, CVE-2014-0114, CVE-2014-3577, CVE-2015-3192, CVE-2015-3253, CVE-2015-5211, CVE-2015-5262, CVE-2016-10006, CVE-2016-2510, CVE-2016-3093, CVE-2016-5007, CVE-2016-6497, CVE-2016-9878, CVE-2017-5662, CVE-2017-5662, CVE-2017-5662, CVE-2017-5664, CVE-2017-5664, CVE-2017-5929
```

stefanneuhaus:
Did even more testing. I consistently got consistent results in >1000 runs:
```
1017 Found 21 vulnerabilities in project sample-project: CVE-2012-6153, CVE-2013-4499, CVE-2014-0107, CVE-2014-0114, CVE-2014-3577, CVE-2015-3192, CVE-2015-3253, CVE-2015-5211, CVE-2015-5262, CVE-2016-10006, CVE-2016-2510, CVE-2016-3093, CVE-2016-5007, CVE-2016-6497, CVE-2016-9878, CVE-2017-5662, CVE-2017-5662, CVE-2017-5662, CVE-2017-5664, CVE-2017-5664, CVE-2017-5929
```

stefanneuhaus:
@jeremylong There are still three PMD rule violations in my PR. After a review my view on these ones is that I violated them on purpose. What is your preferred way to handle those three PMD rule violations?
- fix/workaround
- suppress via `@SuppressWarnings("PMD.<RULE_NAME>")` or some other way
- just keep them

stefanneuhaus:
@jeremylong I was wondering if letting the analyze task fail after the last unsuccessful retry would be even more appropriate. I would favor a failing build over unreliable results. What is your opinion on this?

jeremylong:
Regarding PMD - I'm okay with adding `@SuppressWarnings("PMD.<RULE_NAME>")` - I still have a few warnings in the core that I haven't cleaned up or suppressed.

Regarding failing vs. unreliable results - I agree a failure would be better.

dependency-check-core/src/main/java/org/owasp/dependencycheck/analyzer/CentralAnalyzer.java
```diff
@@ -256,8 +264,52 @@ public void analyzeDependency(Dependency dependency, Engine engine) throws Analy
...
+     * <p>
+     * As the connection to MavenCentral is known to be unreliable, we implement
+     * a simple retry logic in order to compensate for 99% of the issues.
+     *
+     * @param dependency the dependency to analyze
+     * @return the downloaded list of MavenArtifacts
+     * @throws FileNotFoundException if the specified artifact is not found
+     * @throws IOException           if connecting to MavenCentral finally failed
+     */
+    List<MavenArtifact> fetchMavenArtifacts(Dependency dependency) throws IOException {
```
jeremylong: Shouldn't this have a protected scope?

jeremylong (approved):

dependency-check-core/src/main/java/org/owasp/dependencycheck/analyzer/CentralAnalyzer.java
stefanneuhaus: I am fine with it, but it would not be my preferred way. Reasoning:
* private: If I did not want to unit test this method, this would be my favorite scope.
* package-private: This is my preferred scope for fields there are meant to be private but need to be accessible via unit tests. The rule in question is also declared as a [Controversial Rule](http://pmd.sourceforge.net/pmd-4.3.0/rules/controversial.html#DefaultPackage), which I would agree on.
* protected: compared to package-private this scope only makes sense if you want to access a field from a sub class which is located in another package. This is not the case in this example.

But I agree that too many of those SuppressWarning annotation clutter up the code, which is why I thinking making it protected is acceptable. I changed it accordingly.

stefanneuhaus:
I fixed the PMD rule violations and added the change to fail the analysis instead of disabling the CentralAnalyzer.

Hopefully there are not too many setups out there with permanent IOExceptions (for whatever reason, e.g. misconfigured `ANALYZER_CENTRAL_URL`) that have been silently ignored all the time and that now fail. But better to be aware of the failure. And at least there is still the option to simply disable the CentralAnalyzer.

jeremylong (approved):

PR #921, opened 2017-10-09
Fix typo: Engine.isDataBseRequired()

stefanneuhaus:
* Fix typo: In Engine isDataBseRequired() --> isDataBaseRequired()
* make method private

Not 100% sure, if you want this merged, as the method was part of the public API. To me it seems like the method should not be public anyway.


jeremylong:

Issue #937, opened 2017-10-20
v3.0.1 does not work with NodeJS


pdot86:
### Reproduce the issue by following these steps using both dependency-check@v2.0.1 and dependency-check@v3.0.1: 

- Set up an example NodeJS/NPM project and install a package with a known vulnerability
  - I used an old version of uglify-js, any version before 2.6.0 has vulnerabilities:
  - `mkdir test && cd test && npm init -y && npm i -D uglify-js@2.4.1`
- Run dependency-check on the project:
  - `dependency-check --enableExperimental --project test --log ./log.txt -f ALL -o ./ -s ./`

---

**v2.0.1 Scan Info and Results:**
- dependency-check version: 2.0.1
- Dependencies Scanned: 24 (8 unique)
- Vulnerable Dependencies: 1
- Vulnerabilities Found: 3
- Vulnerabilities Suppressed: 0
- Vulnerabilities:
  - cpe:/a:uglifyjs_project:uglifyjs:2.4.1
  - CVE-2015-8858
  - NSP-48
  - NSP-39

**v3.0.1Scan Info and Results:**
- dependency-check version: 3.0.1
- Dependencies Scanned: 9 (8 unique)
- Vulnerable Dependencies: 0
- Vulnerabilities Found: 0
- Vulnerabilities Suppressed: 0
- Vulnerabilities:
  - NONE

stevespringett:
The Node.js analyzer was retired in 3.0 due to large % of false positives and small % of true positives. If you want to use it, enable the retired analyzers.

All Node projects should use the NSP analyzer instead of the Node analyzer. The results are much more accurate.

jeremylong:
I just realized that while the property to re-enable the retired analyzers exists, it can currently only be passed in as an environment variable (`-Danalyzer.retired.enabled=true`).

The CLI, maven, gradle, etc. interfaces need to expose this as a configurable option.

stevespringett:
GitHub needs an `<oops> `emoji.

stefanneuhaus:
@stevespringett Maybe this one? :grimacing:

see [rxaviers/gist:7360908](https://gist.github.com/rxaviers/7360908) :octocat:

amenibouarada:
please , where is the solution , the error still exist (  Message: Could not perform NSP analysis. Invalid payload submitted to Node Security Platform.) thanks

ammy1999:
[...]

PR #939, opened 2017-10-22
Update documentation and configuration for Retired Analyzers

jeremylong:
## Fixes Issue #937 

The documentation has been updated to indicate that the Node.js analyzer was retired. The NSP analyzer should be used going forward. Additionally, configuration options were added to re-enable the retired analyzers.


jeremylong:

Issue #933, opened 2017-10-18
md5sum is not updated alongside of sha1sum by ComposerLockAnalyzer


bentmann:
- https://github.com/jeremylong/DependencyCheck/blob/e2a97738e176fd9bb7042f1f53a024c7c011460c/dependency-check-core/src/main/java/org/owasp/dependencycheck/analyzer/ComposerLockAnalyzer.java#L127
- https://github.com/jeremylong/DependencyCheck/blob/e2a97738e176fd9bb7042f1f53a024c7c011460c/dependency-check-core/src/main/java/org/owasp/dependencycheck/analyzer/CMakeAnalyzer.java#L228

These lines set the SHA1 on the dependency but nothing sets the MD5 correspondingly. The MD5 sum remains on the value it got when the `Dependency(File)` constructor called `determineHashes()` initially a few lines above those `setSha1sum()` calls.

This leads to funny data like below where the two dependencies erroneously share the same MD5 (21ec9e4212f01ba03fa6156e80872fbc) but properly have different SHA1 sums (22825656f245d881a8eb41c0aefc6fd13286dc8e vs da278cfddbad92988554c77026c42dad40750e31).

`````
[ {
  "vendors" : [ {
    "name" : "name",
    "source" : "file",
    "value" : "composer",
    "confidence" : "HIGH"
  } ],
  "vendorWeightings" : [ ],
  "products" : [ {
    "name" : "name",
    "source" : "file",
    "value" : "composer",
    "confidence" : "HIGH"
  } ],
  "productWeightings" : [ ],
  "versions" : [ ],
  "actualFilePath" : "M:\\temp\\dctemp2051907f-24cf-4bb0-9e17-df026a773f92\\check6817504487507768272tmp\\2\\drupal-8.3.7\\composer.lock",
  "filePath" : "M:\\downloads\\drupal-8.3.7.tar.gz\\drupal-8.3.7.tar\\drupal-8.3.7\\composer.lock",
  "fileName" : "drupal-8.3.7.tar.gz: drupal-8.3.7.tar: composer.lock",
  "packagePath" : "M:\\downloads\\drupal-8.3.7.tar.gz\\drupal-8.3.7.tar\\drupal-8.3.7\\composer.lock",
  "md5sum" : "21ec9e4212f01ba03fa6156e80872fbc",
  "sha1sum" : "22825656f245d881a8eb41c0aefc6fd13286dc8e",
  "identifiers" : [ ],
[...]

PR #940, opened 2017-10-22
Add Md5 checksum to the composer and cmake analyzers

jeremylong:
## Fixes Issue #933

Add Md5 checksum to dependencies analyzed by the composer and cmake analyzers.


jeremylong:

Issue #928, opened 2017-10-14
Consistency/comparison issues in domain classes


stefanneuhaus:
- VulnerableSoftware
  - compareTo() is inconsistent with equals(): though this is no MUST requirement, the javadoc indicates that it is strongly recommended and classes that violate this SHOULD clearly indicate it. As the compareTo() logic needs to be rather sophisticated, equals() could fallback to this like `compareTo() == 0`. Example of inconsistency:
```
    @Test
    public void testEqualsAndHashCode() {
        VulnerableSoftware instance1 = new VulnerableSoftware();
        instance1.setCpe("cpe:/a:mortbay:jetty:6.1");
        VulnerableSoftware instance2 = new VulnerableSoftware();
        instance2.setCpe("cpe:/a:mortbay:jetty:6-1");

        assertEquals(0, instance1.compareTo(instance2));
        assertEquals(instance1, instance2);
    }
```

- Dependency
  - equals()/hashCode(): contract violation as only equals() respects `packagePath`.
  - compareTo() is inconsistent with equals()
    - compareTo() merely respects `filePath`.
    - compareTo() compares case-insensitive, while equals() is case-sensitive.

- Identifier
  - compareTo() is inconsistent with equals() as only equals() respects `type`.
  - compareTo() SHOULD throw a NullPointerException when comparing with `null`.

- Evidence
  - compareTo() SHOULD throw a NullPointerException when comparing with `null`.

- DependencyVersion
  - if have not looked for an inconsistency example, but it looks at least fragile:  equals() and compareTo() have different sophisticated implementations, hashCode() just relies on `versionParts`
  - no contract violation, but a bit odd: compareTo() is not connected to the implementation of VulnerableSoftware. Not sure if this can cause unexpected results. See also issue #929.

To me it seems like not all classes need to implement Comparable. If this is actually the case, I would rather refrain from it to prevent these kind of issues. Especially as objects of these classes are not immutable while also being part of ordered collections, it is not easy to ensure consistency. In fact objects are being mutated in various places.

Also Dependency inherits from EvidenceCollection, which can complicate things when correctly implementing equals()/hashCode()/Comparable or maintaining thread-safety. Falling back to composition looks like an easy win here, as EvidenceCollection is package-private and not used anywhere else.

[...]

Issue #929, opened 2017-10-14
Incorrect implementation of Maven version order specification


stefanneuhaus:
Though the PR #925 improved things the [Version Order Specification](https://maven.apache.org/pom.html#Version_Order_Specification) is currently implemented in a rather heuristic manner, which may result in false positives or false negatives.

`org.apache.maven.artifact.versioning.ComparableVersion` ([javadoc](https://maven.apache.org/ref/current/maven-artifact/apidocs/org/apache/maven/artifact/versioning/ComparableVersion.html)) already implements it (presumably correctly). We could make use of that implementation.

Version parsing/ordering is implemented in `VulnerableSoftware` and also in `DependencyVersion`/`DependencyVersionUtil`. Maybe one can simplify things and merge these or otherwise get rid of duplicate code.

jeremylong:
Good points - I was actually planning on going back and looking at the ordering of RELEASE and SNAPSHOT builds. As I am about to publish the 3.0.0 we can include this in a following release.

PR #983, opened 2017-11-13
CompareTo() cleanup

jeremylong:
## Initial Fix Issue #928

Removed unnecessary TreeSets and removed or updated `compareTo()` in a few classes. There is more work to do for #928 and #929 - but this fix resolves some issues and speeds up the analysis.


jeremylong:

Issue #503, opened 2016-05-25
Error on importing CVE-2008-2079 from NVD XML files into MySQL database


xthk:
Hi all,

while running:

```
dependency-check.sh \
  --cveValidForHours 0 \
  --log dependency-check.log \
  --dbDriverName com.mysql.jdbc.Driver \
  --dbDriverPath mysql-connector-java-5.1.39/mysql-connector-java-5.1.39-bin.jar \
  --connectionString="jdbc:mysql://localhost:3306/dependencycheck" \
  --updateonly
```

I got these errors (in the logfile):

```
com.mysql.jdbc.exceptions.jdbc4.MySQLIntegrityConstraintViolationException: Duplicate entry '30517-47953' for key 'PRIMARY'
```

and

```
DEBUG - Unable to update details for org.owasp.dependencycheck.data.update.NvdCveUpdater org.owasp.dependencycheck.data.update.exception.UpdateException: org.xml.sax.SAXException: Error updating 'CVE-2008-2079'
```

Which on the console produces:

```
[WARN] Unable to update Cached Web DataSource, using local data instead. Results may not include recent vulnerabilities.
```

Return code of dependency-check.sh is "0", but that is another issue (#215) I think.

Logs (console and debug log) are attached. Please let me know if further information is needed.
- [output.txt](https://github.com/jeremylong/DependencyCheck/files/281869/output.txt)
- [dependency-check.log.txt](https://github.com/jeremylong/DependencyCheck/files/281867/dependency-check.log.txt)

jabbrwcky:
So I dug a bit into the problem which turns out to be some paradoxon apparently, as the TreeSet of Vulnerability contains two entries for CPE `cpe:/a:mysql:mysql:5.1.28a`: one with `previousVersion = null` and one with `previousVersion="1"`.

This is baffling as the previousVersion apparently is not used in the compareTo() implementation of VulnerableSoftware, which just compares the CPE strings. The CPE strings are completely the same.

I tried adding some unit tests with the data from CVE-2008-2079 but in the unit tests TreeSet behaves as expected.

It is somehow connected to the fact that TreeSet utilizes compareTo instead of equals to find duplicates, because the issue vanishes if I change the implementation in Vulnerability from SortedSet/TreeSet to Set/HashSet (which uses equals to prevent duplicates).

If sorting is not really crucial, the fastest fix would to use a HashSet for storing the vulnerabilities and sort it on the fly  whenever it is required.

[...]

PR #994, opened 2017-11-17
Added Batch insert for References and Vulnerabilities

alefq:
Applies batch inserts for reference and vulnerability tables,
solves the slow one-by-one insert process, for Vulnerabilities
with several references/vulnerabilities associated.

## Description of Change

Makes use of batch capabilities of PreparedStatement. Executes updates every 1000 ( MAX_BATCH_SIZE), slowing down time for vulnerabilities updates by a near 10x factor on external databases (Postgresql for instance). 
Tested on H2 (default), and Postgresql. 

no

Tests available already cover the process. All tests where executed before pushing code to forked repository.

### DIFF
--- a/dependency-check-core/src/main/java/org/owasp/dependencycheck/data/nvdcve/CveDB.java
+++ b/dependency-check-core/src/main/java/org/owasp/dependencycheck/data/nvdcve/CveDB.java
@@ -25,6 +25,7 @@ import java.sql.ResultSet;
 import java.sql.SQLException;
 import java.util.ArrayList;
 import java.util.Collections;
+import java.util.Date;
 import java.util.EnumMap;
 import java.util.HashMap;
 import java.util.HashSet;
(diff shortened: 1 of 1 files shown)


jeremylong:

jeremylong:
We previously had batching enabled and under MySQL ran into issue https://github.com/jeremylong/DependencyCheck/issues/503.

While other changes have been made to the `Vulnerability` and `VulnerableSoftware` objects - adding the batching support back in will require additional testing on MySQL (and my require additional updates to the `equals()` and `compareTo()` methods of `VulnerableSoftware` (see issue #928).

alefq:
Is there regression tests for those issues?
Sounds more like a Mysql jdbc driver problem rather a Dependency Check issue.
I can look a little deeper into the issue with some more information.
Is there a branch or tag with previous version you are mentioning?
I made the patch, and will keep it in my fork, because making millions of inserts, one by one (in nested loops), is really slow and unefficent for my postgresql setup.

jeremylong:
Unfortunately, no. I don't maintain a full suite of tests against each database. However, if we make batch support configurable I suppose anyone that runs into the issue they can disable batch support.  Look at the history of the CveDB  and you'll see that this was halfway implemented (i.e. it had a boolean option - but I do not believe I ever exposed it - and finally just removed it).

alefq:
I will add a flag to disable batch inserts, then create a new Pull Request.

Issue #993, opened 2017-11-17
NodeJS scan not recognizing dependencies in package.json files


TwilightMercy:
Hi,
I ran the NodeJS retired analyzer (with **--enableRetired** parameter) on a NodeJS Project using the latest version of the tool (3.0.2).
The HTML report  shows that the tool recognized as Dependencies only the **"package.json"** files. It doesn't scan their content and the open source dependencies.

Example:
![1](https://user-images.githubusercontent.com/29895680/32942361-7df0b388-cb91-11e7-904e-ac2fdfa441dc.png)

Any ideas?

stevespringett:
That is the way the NodeJS analyzer was designed. In my testing, I found that it produced a 100% false positive rate and a 100% false negative rate. This analyzer has been retired due to it not being useful. No further work is currently planned on improving it.

All Node projects should use the Dependency-Check NSP analyzer, which is enabled by default. The NSP analyzer utilizes the Node Security Platform API and will scan dependencies included in package.json.

TwilightMercy:
I understand, but if so - Even when using the default analyzers (just simply running the tool without **--enableRetired** or **enableExperimental**), I get the same results. only **package.json** files as Dependencies - and not their content.

jeremylong:
I completely understand the issue you are facing and I am working on revamping the Node.js and NSP analyzers.

TwilightMercy:
Ok, can you please update here when the problem is fixed?

TwilightMercy:
Is it possible to download the updated version now, because the version is still 3.0.2

jeremylong:
@barronen1 no.  the feature is not complete yet.  you are more then welcome to download the branch, compile it, and test it.

TwilightMercy:
@jeremylong I understand.

Can I ask if you can guide me in short how to do that? I downloaded the whole ZIP (with the comitted files).
And tried to to compile the files you changed using the command: **javac JavaFileName.java**, but it has a lot of errors.
Am I doing it wrong?

jeremylong:
This is a maven project.

```bash
git clone https://github.com/jeremylong/DependencyCheck.git
cd DependencyCheck
git checkout issue-993
mvn install
```

This should build version 3.0.3-SNAPSHOT.

TwilightMercy:
I guess that **mvn install** didn't work properly while building.

------------------------------------------------------------------------
[INFO] Reactor Summary:
[INFO]
[...]

Issue #991, opened 2017-11-16
CSV For Node Scan Doesn't Include Vulnerable Software Name.


jgamblin:
When you scan a package.json and look at the CSV output it doesn't include the name of the vulnerable software (tagged vulnerableSoftware in the json).

jeremylong:
This is being fixed the [issue-993](https://github.com/jeremylong/DependencyCheck/tree/issue-993) branch and will be included in the next release.

jaikishantulswani:
/GithubDependencyCheck.sh: line 5: jq: command not found
(23) Failed writing body
cat: 'results/*.csv': No such file or directory

PR #1006, opened 2017-11-26
Improve Node.js and NSP analyzers

jeremylong:
## Fixes Issue #993 and #991 

Removed the retired annotation from the Node Package Manager analyzer and updated it so that instead of reading package.json - it analyzes the package-lock.json or shrinkwrap.json. Only the package-lock.json or shrinkwrap.json from the root directory will be analyzed; if there are package-lock.json contained in the dependencies within the node_modules directory they will not be analyzed. 

Disabled the CPE Analyzer for Node dependencies; this is due to the incredibly high false positive rate. If one did want to re-enable the CPE Analyzer for Node dependencies simply remove "npm" from the setting `ecosystem.skip.cpeanalyzer`. This setting/environment variable is a comma seperated list of eco systems for which the CPE analyzer will be skipped. Currently, only "npm" is in the list but more may be added in the future.

Update the dependency-check.json report format. Specifically, the identifiers section was updated to better support Node. In addition, some minor changes were made to the HTML and CSV report formats.

Test cases were updated.


jeremylong:

stevespringett:
I'm taking a look. Will provide feedback Monday or Tuesday.

stevespringett (commented):
Upon initial review and execution, I noticed that the HTML and XML reports only contain vulnerable dependencies. Dependencies without vulnerabilities are not listed.

SHA1 and MD5 hashes should not be included in the report. The isVirtual should be set to true and the hashes should be null or empty.

On a typical node project that has multiple versions of the same dependency (such as jquery), you'll end up with the filename = ...package.json?jquery multiple times. What affect will this have? Will it only list jquery once? If so, that is not the behavior we want. We actually want jquery (in this case) being listed multiple times.

I have some more testing to do, but it's looking good so far. Will provide additional feedback Monday or Tuesday

jeremylong:
@stevespringett regarding "Dependencies without vulnerabilities are not listed." - do you have a shrinkwrap.json or package-lock.json? Those are analyzed by the Node Package Analyzer to obtain the list of dependent libraries used by the node application. Then the package.json is used by the NSP Analyzer to identify vulnerable dependencies.

If the package-lock.json only has multiple versions of the library then it will be listed multiple times. Depending on how things are setup they multiple versions of the same library will either result in "related dependencies" and/or additional entries in the "Referenced In Project/Scope:" section.

stevespringett:
In my experience, shrinkwrap json is typically only used when creating and publishing node modules. Modules that are published to the npm repo that include shrinkwrap.json will be guaranteed to include the specified version of each of the dependencies specified in the file.

When developing applications however, shrinkwrap json is rarely used. Applications are not published to the npm repo. I just checked, and none of the node apps that I've ever worked on have a shrinkwrap json. You can refer to https://github.com/OWASP/NodeGoat for an example of a Node application that doesn't have shrinkwrap.

In the case of NodeGoat, `npm ls` will result in a tree containing 1191 packages, whereas Dependency-Check only includes 4.

If `npm install` is run prior to Dependency-Check, Dependency-Check will collect evidence from node_modules and return a report consisting of 981 dependencies (706 unique).

jeremylong:
Regarding the npm-shrinkwrap.json vs the package-lock.json - the reason npm-shrinkwrap.json is being included is due to the [package-lock.json documentation](https://docs.npmjs.com/files/package-lock.json) that states:

> If both package-lock.json and npm-shrinkwrap.json are present in the root of a package, package-lock.json will be completely ignored.

Regarding the differences in the counts between `npm ls` and dependency-check if you look you will see a ton of duplicates listed in `npm ls`. On NodeGoat try running:

```bash
npm install
npm ls | sed 's/^[ └│┬├─└]*//' | sed 's/ deduped//' | grep -v ^$ | grep -v "owasp-nodejs-goat" | sort -fu | wc -l
```

You will get 860 unique dependencies. Previously you would have gotten the incorrect count from dependency-check as there was an error in parsing due to multiple licenses for some dependencies; I have just updated the PR so that when you run:

```bash
./dependency-check.sh -s /Users/someone/projects/samples/NodeGoat -l out.txt -o . --disableAssembly --project node
```

You will get 860 unique dependencies. We have to disable the Assembly Analyzer as one of the node.js dependencies contains an `exe`.

ammy1999:
hello, I add a stage dependencuCheck in jenkins file, as result : I have a lot of vulnerabilities and a html report , the problem now, I want add a suppress file for the false positive but when I click in suppress button, it doesn't work , So what can I do please !!

stevespringett:
try viewing the report outside of Jenkins - Jenkins has a built in CSP that prohibits the execution of inline javascript. Copy the file locally, and open it there.

PR #1007, opened 2017-11-27
make Error on personal Suppression File to not fatal

stboissdev:
## Fixes Issue #

if one personal suppression file fails to load, analyzer stops after prepare step and base suppression File Rules are not processed

## Description of Change
I change to make not fatal Error, if personal suppression file fails to load

?


jeremylong:

jeremylong:
thanks!

stboissdev:
@jeremylong, when do you plan to make the next release?

jeremylong:
My plan is next weekend. I want to go through the FP and likely add the ability to suppress NSP findings.

PR #1015, opened 2017-12-03
Version Upgrades

jeremylong:
## Description of Change

Moved several version numbers to properties and setup auto-incrementing the versions on every clean operation. Added a safety net via a [custom maven-enforcer rule](https://github.com/jeremylong/class-file-format-rule) that will break the build if we upgrade to a dependency that requires JRE v8+. We currently are supporting Java 7 until next June when Jenkins drops support for Java 7.

Existing tests


jeremylong:

stevespringett:
Jeremy. I think you might have misunderstood the email I sent. As of this June, Jenkins only support Java 8. They still have LTS releases that are still supported that can only run on Java 7, but all new releases since June require Java 8.

Issue #1018, opened 2017-12-04
reportOutputDirectory is undocumented in the Maven plugin


mackncheesiest:
`reportOutputDirectory` is currently undocumented in the [configuration page](https://jeremylong.github.io/DependencyCheck/dependency-check-maven/configuration.html) for the DependencyCheck Maven plugin.

Having the option documented would be beneficial.

jeremylong:
The `reportOutputDirectory` is not directly configured in the plugin; rather this is bound to the reporting `outputDirectory`. It is showing up in the plugin help due to the creation of a single mojo that can be used as both a build plugin or a site plugin. To configure the `reportOutputDirectory` you must configure the reporting section like this:

```xml
    <reporting>
        <outputDirectory>${basedir}/target/reports</outputDirectory>
        <plugins>
            <plugin>
                <groupId>org.owasp</groupId>
                <artifactId>dependency-check-maven</artifactId>
                <version>3.1.0-SNAPSHOT</version>
                <configuration>
                    <format>ALL</format>
                </configuration>
            </plugin>
        </plugins>
    </reporting>
```

I will update the javadoc.

mackncheesiest:
Ah, that makes sense. I wonder if there is then a way to mark it as `<editable>false</editable>` in the autogenerated `plugin.xml` so that it doesn't appear as a suggested parameter for the plugin during pom configuration.

I believe marking the parameter as `@Parameter(..., readonly = true)` in the same Java code would stop it from being suggested, but not sure it would necessarily play well with how the rest of the plugin behaves.

jeremylong:
@mackncheesiest thanks for the suggestion - marking the parameter as `readonly` does in fact remove it from the `help:describe`.  Thanks!

PR #1019, opened 2017-12-04
Added documentation entry for reportOutputDirectory configuration option

mackncheesiest:
## Fixes Issue #
#1018 

## Description of Change
Adds a documentation entry for the `reportOutputDirectory` Maven plugin configuration option

N/A


jeremylong:

mackncheesiest:
Actually, after experimenting with both the `outputDirectory` and `reportOutputDirectory` parameters with version 3.0.2 of the Maven plugin in both `<build>...</build>` and `<reporting>...</reporting>` configurations, I can't say that I even know what the `reportOutputDirectory` option even does. It never seems to affect behavior of either `mvn site` with an `aggregate` goal or `mvn package` with a `check` goal no matter what configuration I do. 

I was thinking it might affect `mvn site` as a corresponding pair to the `outputDirectory` parameter that is [ignored](https://jeremylong.github.io/DependencyCheck/dependency-check-maven/configuration.html) for site builds, but it doesn't seem to do that. The current Javadocs aren't [incredibly illustrative](https://github.com/jeremylong/DependencyCheck/blob/5a0e597124e6b76695c28c2dd71a4252e0023f5c/dependency-check-maven/src/main/java/org/owasp/dependencycheck/maven/BaseDependencyCheckMojo.java#L154-L159) about the distinction between the two, either.

Anyway, some input would be appreciated.

jeremylong:
See comments in issue #1018.

PR #1028, opened 2017-12-09
Resolve fp

jeremylong:
## False Positive Reduction

Resolves several of the reported false positives.  I took a different approach this time on the FP for database connectors/drivers; this round we suppressed the individual CVEs that were unrelated to the connector/driver. This leaves the match in place but will create possible FP in the future when new CVEs arise that are not related to the driver/connector.

This round of patching included both updates to the hint configuration and a minor change to the version matching algorithm. These two changes will reduce additional/non-reported FP.

Yes with regards to the changes to the version matching algorithm.


jeremylong:

Issue #1000, opened 2017-11-21
Feature Request: Emit message if lock file exists for updating NVD


aaronweaver:
If dependency checker is stopped before the update finishes the lock file remains in the data directory. Running dependency checker again "hangs" as it appears to wait for the lock file to be removed. Perhaps show a message that a lock file exists with the option to remove it.

jeremylong:
That is a really good idea - thanks for the suggestion. I suppose we could provide the instructions to remove the lock file and database...

ChristopherSchultz:
This would be helpful for https://github.com/jeremylong/DependencyCheck/issues/1027

Issue #1027, opened 2017-12-08
Quitting dependency-check during an update leaves stale lock file


ChristopherSchultz:
Launch dependency-check and, while it is performing a database update, press `CTRL-C` to quit the process. Immediately launch dependency-check again and the tool hangs after printing:

```
[dependency-check] SLF4J: Class path contains multiple SLF4J bindings.
[dependency-check] SLF4J: Found binding in [jar:file:/Users/chris/packages/dependency-check-ant/dependency-check-ant.jar!/org/slf4j/impl/StaticLoggerBinder.class]
[dependency-check] SLF4J: Found binding in [jar:file:/Users/chris/packages/dependency-check-ant/lib/dependency-check-ant-3.0.2.jar!/org/slf4j/impl/StaticLoggerBinder.class]
[dependency-check] SLF4J: See http://www.slf4j.org/codes.html#multiple_bindings for an explanation.
[dependency-check] SLF4J: Actual binding is of type [org.owasp.dependencycheck.ant.logging.AntLoggerFactory]
```

A bit of Googling led me to this issue: https://github.com/jeremylong/DependencyCheck/issues/1000

Deleting the lock file allowed me to resume my work.

It would be nice if dependency-check would clean-up after itself in case it gets interrupted. (If possible). Otherwise, perhaps https://github.com/jeremylong/DependencyCheck/issues/1000 is enough to mitigate this problem.

I happen to be using the Apache ant task, if that makes any difference.

(Thanks for the great tool!)

jeremylong:
I'll do some research on this one - some changes that need to be made are updating the error reporting per issue #1000 and updating the purge task (and related) to remove the lock file.

ChristopherSchultz:
I haven't dived into the code for dependency-check. If it's somewhat straightforward, I might be able to produce a PR.

jeremylong:
Thanks for reporting this issue - I got to learn something new regarding shutdown hooks. This makes ODC much more stable!

ChristopherSchultz:
Thanks for the quick response!

djjorjinho:
Hello,

Are we planning to handle these stale locks on KILL signals? Am I wrong to point out this solution only handles SIGINT?

PR #1030, opened 2017-12-12
Add Shutdown Hook to remove h2 lock file

jeremylong:
## Fixes Issue #1000 & #1027 

Updated console output if a lock file exists and added a shutdown hook to ensure that the custom h2 db lock file gets removed when the JVM exits.

### DIFF
--- a/dependency-check-core/src/main/java/org/owasp/dependencycheck/utils/H2DBCleanupHook.java
+++ b/dependency-check-core/src/main/java/org/owasp/dependencycheck/utils/H2DBCleanupHook.java
@@ -1,0 +1,34 @@
+/*
+ * Copyright 2017 OWASP.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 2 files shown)


jeremylong:

stefanneuhaus:
Is this also expected to work with the Gradle Daemon? I'm not 100% sure if the current approach will work when the Gradle daemon process survives the build.

jeremylong:
@stefanneuhaus I've tested the shutdown hook with gradle and it appears to work as expected (i.e. it gracefully shutdowns and removes the lock file).

stefanneuhaus:
@jeremylong I just also tested the shutdown hook. When not using the Gradle daemon the shutdown hook works as expected. But when using the daemon I am left with a lock file:

1. Remove H2 DB (in "~/.gradle/caches/dependency-check-data/3.0/") 
2. ./gradlew --daemon dCAn ("--daemon" is the default since quite some time)
3. Once the H2 update is in progress press CTRL+C
4. Observe that "dc.update.lock" exists

Run "jps" to get the PID of the "GradleDaemon" process and kill it. Observe that the lock file gets properly deleted.

Running builds with the Maven plugin or CLI client in this state should be no problem in most cases, as the Gradle build stores the DB in a different location per default.

Each time I started a new build my old daemon could not be reused and has been stopped prior to starting a fresh one (for whatever reason; I did not dig into this), thereby triggering the shutdown hook again. But of course this is not the default behavior. In fact there can be even multiple instances of the daemon running, each having different configurations.

Though I did not manage to run into issues, I would expect that this is at least possible. Nevertheless even if this solution is not 100% bullet proof, it is at least a big step forward and prevents issues in most cases.

PR #1034, opened 2017-12-17
updated lock file cleanup to be useable by other build plugins

jeremylong:
## Fixes Issue #1030 

Updated the shutdown hook mechanism so that other build plugins can utilize their own shutdown mechanism. See https://github.com/jeremylong/DependencyCheck/pull/1030#issuecomment-352187395

### DIFF
--- a/dependency-check-core/src/main/java/org/owasp/dependencycheck/utils/H2DBLock.java
+++ b/dependency-check-core/src/main/java/org/owasp/dependencycheck/utils/H2DBLock.java
@@ -30,6 +30,8 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 /**
+ * The H2 DB lock file implementation; creates a custom lock file so that only a
+ * single instance of dependency-check can update the embedded h2 database.
  *
  * @author Jeremy Long
  */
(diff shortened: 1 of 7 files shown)


jeremylong:

jeremylong:
@stefanneuhaus if you have time - could you take a look at this PR in addition to the update to the gradle plugin PR (https://github.com/jeremylong/dependency-check-gradle/pull/70). Does this resolve the cleanup issue?

stefanneuhaus:
@jeremylong I had a look at the PR: still the lock file remains when using the Gradle daemon. The reason for this is basically the same as before. ShutDownHookActionRegister also adds a shutdown hook to the JVM runtime:
```
    private ShutdownHookActionRegister() {
        Runtime.getRuntime().addShutdownHook(new Thread(new GradleShutdownHook(), "gradle-shutdown-hook"));
    }
```
Thus you need the Daemon to shutdown in order to get the lock released.

I like the idea of having configurable shutdown hooks for different clients. For Gradle we probably want to hook into Gradle's predefined lifecycle hooks. See e.g. [Gradle docs](https://docs.gradle.org/current/userguide/build_lifecycle.html#build_lifecycle_events), [Gradle beyond the basics](https://www.safaribooksonline.com/library/view/gradle-beyond-the/9781449373801/ch03.html) or [stackoverflow](https://stackoverflow.com/questions/24638691/when-gradles-hooks-are-added-in-build-lifecycle). I have not yet played with these. A task-based hook may be sufficient, but we have to test this with multi-project builds. Otherwise a project-based hook may be appropriate.

PR #1036, opened 2017-12-17
Suppress FP: io.vertx:vertx-config-kubernetes-configmap

msrb:
## Description of Change

Suppresses FP in io.vertx:vertx-config-kubernetes-configmap.

no


jeremylong:

jeremylong (approved):

Issue #632, opened 2016-12-16
Lucene Query Parse Error


brileyd:
I get an error when I run the dependency checker against my project.  Here is a link to the error:
https://gist.github.com/brileyd/a0a21d095a72520ab5d9ea0065edc659

The error eventually states this:
Caused by: org.apache.lucene.search.BooleanQuery$TooManyClauses: maxClauseCount is set to 1024

Would it be possible to set a property/option to extend the maxClauseCount?

Thank you!

jeremylong:
Any chance you can tell me what dependency caused this? I'd be interested in seeing the manifest as there is likely a better solution then increasing the maxClauseCount...  But I can also look into setting something up so that the maxClauseCount can be increased.

brileyd:
It is a fairly large project with 12 modules.  Is there a property or arg I can pass to the shell script to get it to tell me which dependency?  The reason I am using the command line option instead of the maven option is I am trying to get all the results on a single report rather than in each module.  I've updated gist with the full output of the script.  Thanks!

jeremylong:
To get all of the reports consolidated using Maven you can use the [aggregate goal](http://jeremylong.github.io/DependencyCheck/dependency-check-maven/configuration.html) in the parent pom (`mvn org.owasp:dependency-check-maven:1.4.4.1:aggregate`. 

It would be great if the library causing the reported issue could be isolated. Looking at the error it appears to have something to do with `portal-impl Newton portal ... Liferay Portal Enterprise Edition 6.2.10`.  So I'm guessing you have several Liferay related dependencies - I tried downloading liferay and scanning it but I did not see the error you've reported. However, central does not appear to have a 6.2.10 version - 6.2.5 is the highest version.

In your environment, using the CLI could you generate the log file (add `-l dc.log`.  The log file should help us figure out which dependency is causing the issue.

brileyd:
Attached is the log file and the dependency causing the error is the portal-impl.jar.  I tried to attach the manifest, but github will not let me upload it.  It does contain all the text seen in the error (all of the LPE-xxxx entries).
[dc.log.zip](https://github.com/jeremylong/DependencyCheck/files/666928/dc.log.zip)

Thanks for all the help!

jeremylong:
Can you email me the manifest file?  jeremy.long @ owasp.org

jeremylong:
[...]

PR #1044, opened 2017-12-21
Lucene query fix

jeremylong:
## Fixes Issue #632

Limited the size of the text being added to the query to 1000 characters per piece of evidence. While it still may be possible to hit the error in #632 it is unlikely.

Yes


jeremylong:

PR #1047, opened 2017-12-25
False negative on xmlsec

jeremylong:
## Fixes Issue 1043

Added a hard-coded CPE for xmlsec.  If we see `cpe:/a:apache:santuario_xml_security_for_java:` then we will add `cpe:/a:apache:xml_security_for_java`. Higher confidence data creating a match on the first is causing this one to be missed.


jeremylong:

PR #1048, opened 2017-12-27
Lucene upgrade, bug fixes, and general cleanup

jeremylong:
## Description of Change

Upgraded Apache Lucene to the newest version compiled with Java 7 (version 5.5.5), fixed tokenizing/filtering pipeline to correctly handle URLs (should create fewer FP on any CPEs that contain `http`), and general code cleanup.

yes


jeremylong:

PR #1101, opened 2018-02-11
Fix groovy script failure because of unescaped Windows File.separator.

JeneJasper:
[ERROR] gmavenplus-plugin:execute (add-dynamic-properties-clean) on project dependency-check-parent
              Error occurred while calling a method on a Groovy class from classpath.:
              InvocationTargetException: startup failed:
[ERROR] Script1.groovy: 2: unexpected char: '\' @ line 2, column 57.
[ERROR]              config = "file:///C:\dev\projects\DependencyCheck/src/main/config


jeremylong:

jeremylong (approved):

Issue #1105, opened 2018-02-14
dependency check 3.1.1 false negative for resteasy-client


borcherspm:
We were updating dependency check from version 2.1.0 to 3.1.1.  We use the gradle plugin.  With 2.1.1 there were multiple cve's listed for resteasy-client, but with 3.1.1 they are all gone.

### Output log
* https://gist.github.com/borcherspm/08aa1600c93c2ae3b27dd551629509fa

### False Negative
- http://search.maven.org/#artifactdetails%7Corg.jboss.resteasy%7Cresteasy-client%7C3.1.4.Final%7Cjar
- 2.1.0
  - cpe: cpe:/a:redhat:resteasy:3.1.4   Confidence:LOW
  - maven: org.jboss.resteasy:resteasy-client:3.1.4.Final   Confidence:HIGHEST
  - evidence has pom, file, manifest, gradle
  - cve: CVE-2016-6348, CVE-2016-6347, CVE-2016-6346, CVE-2016-6345
- 3.1.1
  - cpe: cpe:/a:rest-client_project:rest-client:3.1.4   Confidence:Low
  - maven: org.jboss.resteasy:resteasy-client:3.1.4.Final ✓   Confidence:Highest 
  - evidence has manifest, pom, central, file, gradle
  - cve: none

_Same thing is also happening for other resteasy jars._
* resteasy-jackson2-provider-3.1.4.Final.jar
* resteasy-jaxb-provider-3.1.4.Final.jar
* resteasy-jaxrs-3.1.4.Final.jar
* resteasy-multipart-provider-3.1.4.Final.jar
* resteasy-servlet-initializer-3.1.4.Final.jar
* resteasy-spring-3.1.4.Final.jar

jeremylong:
During the lucene upgrade a mistake was made that introduced this (and possibly other) false positives. I will be releasing a new version soon.

PR #1114, opened 2018-02-25
Lucene Updates

jeremylong:
## Fixes Issue #1105

The Lucene analysis was updated so that tokens were properly concatenated for accurate results. Additionally, general code cleanup, checkstyle, etc.

yes


jeremylong:

PR #1144, opened 2018-03-21
Fix issues with dotnet core 2.0 assemblies

thomasd-:
It turns out, running GrokAssembly with mono yields some errors with some dlls (not all). Running it with dotnet runtime fixes the problem.

## Description of Change

The main change is in the file core/src/main/java/org/owasp/dependencycheck/analyzer/AssemblyAnalyzer.java, where I copy a different version of GrokAssembly depending on the OS (I assumed that the original GrokAssembly works well on windows), then I runned the binary with dotnet instead of mono.

Changes have been made to the Dockerfile too, in order to install dotnet instead of mono.

The rest of the changes are cosmetic change made to reflect the change from mono to dotnet in the options and documentation.

I added a binary for GrokAssemblyCore (the version of GrokAssembly I made to run with .NET Core) in the resources. This binary can be obtained by compiling the sources here : https://github.com/thomasd-/GrokAssemblyCore

no, but some tests have changed


jeremylong:

jeremylong:
I am planning on incorporating the changes - however, I need to do some additional testing. Specifically, the original mono version appears to have issues reading .NET core assemblies.  Is the reverse true? In other words, for compatability do we need to maintain a mono grok assembly and a .net Core grok assembly?

thomasd-:
Good point,

```
 $ dotnet GrokAssemblyCore.dll GrokAssembly.exe
<?xml version="1.0" encoding="utf-8"?><assembly><error>Managed assembly cannot be loaded</error></assembly>
```

It looks that GrokAssemblyCore has issues with GrokAssembly, and probably with all applications compiled for .NET Framework.

I'd say we should run GrokAssembly with Framework binaries and GrokAssemblyCore with Core binaries. But I have no idea how to differentiate on type of binary from the other.

The simplest solution would be to run the Framework test and then the Core test if the first failed, but is it an acceptable solution ?

jeremylong:
Is there a way to read the file header (magic) to differentiate between Core and .NET Framework assemblies?  Possibly in the CLI headers section?  Otherwise, I'm okay with the simple solution of just trying both as long as it is silent to the end user - just put a TODO statement in to go back and re-evaluate if there is a better way to choose which assembly analyzer to use.

Thanks!

thomasd-:
Alright, I've spent some time on this, and the main problem here is that GrokAssembly is trying to load the binary to examine (that is, loading it in a way that it could be executed). This fails when the assembly has some dependency to .NET Core API that is not covered by .NET Framework.

I have changed the GrokAssembly to gather its informations in another way. 

I should close this PR and open a clean new one.

PR #1147, opened 2018-03-23
Update clone URL and script location

lllama:
## Description of Change

Updates the docs.

Git URL will only work for contributors.
Maven seems to build into a different dir, so have updated that as well.

Nope


jeremylong:

jeremylong:
Thanks for the PR!

Issue #1086, opened 2018-01-30
CLI only checks two things with RubyOnRails project: Gemfile.lock and Gemfile.lock:nokogiri and complains "Did not find org.owasp.dependencycheck.analyzer.RubyGemspecAnalyzer"


philCryoport:
NOTE: I'm not the dev here; I'm the DevOps guy whose previous experience is in large Java and PHP deployments, so I'm learning this stuff on the fly.

* Ruby is 2.3.6
* Rails is 5.0.1

Not sure how to securely send you the Dependency Checker CLI debug log -- obviously, I don't want to advertise which gems we're using in our RubyOnRails project...

Downloaded OWASP Dependency Checker 3.1.1 release and built on Ubuntu x64 with OpenJDK 8.

Ran CLI with:
```
% cd ~/DependencyCheck-3.1.1/dependency-check-cli/target/release/bin
 % ./dependency-check.sh --project "Foo" --scan "/home/bar/foo"
[INFO] Checking for updates
[INFO] Skipping NVD check since last check was within 4 hours.
[INFO] Check for updates complete (65 ms)
[INFO] Analysis Started
[INFO] Finished Archive Analyzer (0 seconds)
[INFO] Launching: [bundle-audit, check, --verbose] from /tmp/dctempc959a0b8-966e-482e-a9f8-acaf9f469e7f
[INFO] Ruby Bundle Audit Analyzer is enabled. It is necessary to manually run "bundle-audit update" occasionally to keep its database up to date.
[WARN] Did not find org.owasp.dependencycheck.analyzer.RubyGemspecAnalyzer.
[INFO] Launching: [bundle-audit, check, --verbose] from /home/pschlesinger/cryoportal-v2
[INFO] Finished Ruby Bundle Audit Analyzer (0 seconds)
[INFO] Finished File Name Analyzer (0 seconds)
[INFO] Finished Dependency Merging Analyzer (0 seconds)
[INFO] Finished Version Filter Analyzer (0 seconds)
[INFO] Finished Hint Analyzer (0 seconds)
[INFO] Created CPE Index (1 seconds)
[INFO] Skipping CPE Analysis for npm
[INFO] Finished CPE Analyzer (1 seconds)
[INFO] Finished False Positive Analyzer (0 seconds)
[INFO] Finished Cpe Suppression Analyzer (0 seconds)
[INFO] Finished NVD CVE Analyzer (0 seconds)
[...]

PR #1148, opened 2018-03-23
Support parallel bundler audit runs

lllama:
## Fixes Issue #

This will address the "Could not create temporary gem file" error mentioned in #1086.

## Description of Change
createTempFile will ensure that we have unique filenames for the gem
file, which will in turn ensure that we don't get an error when two
projects being checked at the same time have the same dependency.

No (sorry)


jeremylong:

jeremylong:
Thanks for the PR!

Issue #1145, opened 2018-03-22
Add 'suppress until' config to temporarily suppress a vulnerability


siladu:
In a situation where we know a dependency vulnerability fix is incoming, it would be nice to not have to remember to un-suppress it.

For example, CVE-2018-7489 is fixed: https://github.com/FasterXML/jackson-databind/issues/1931 
but awaiting release: https://github.com/FasterXML/jackson/wiki/Jackson-Release-2.9.5

#### Example
Proposed new config added to reenable warnings after specified date: `<until>2018-04-01</until>`
```xml
    <suppress>
        <notes><![CDATA[
   file name: jackson-databind-2.9.4.jar
   ]]></notes>
        <gav regex="true">^com\.fasterxml\.jackson\.core:jackson-databind:.*$</gav>
        <cve>CVE-2018-7489</cve>
        <until>2018-04-01</until>
    </suppress>
```

Similar to https://github.com/unruly/junit-rules/blob/master/README.md#ignore-tests-until-a-certain-date-or-datetime

jeremylong:
Interesting idea - thanks for the suggestion. It may take us a while to get to this - but PRs are always welcome.

aikebah:
@jeremylong I'm willing to give this a try.... I'll try to come up with a PR for this

jeremylong:
Note - we still need to update the documentation on this feature.  Regardless - Thanks for the PR!

PR #1149, opened 2018-03-25
Proposed fix for issue #1145

aikebah:
## Description of Change
- Implemented single-pass multi-xml-schema-validating parser similar to existing code for hints parsing
- Added new suppression xmlschema version with the 'until' for a suppression rule implemented as attribute of suppress (which seems more logical to me than a child element)
- updated code to support this temporary suppression of issues
- Added testcases for the proper parsing of the version 1.1 (existing) and 1.2 (new) suppressionfile xmlschema

yes (validating that suppressions with until 'in-the-past' do not appear in the suppression-rules and suppressions without until and with until 'in-the-future' do appear in the suppression-rules after parsing)

### DIFF
--- a/core/src/main/java/org/owasp/dependencycheck/xml/suppression/SuppressionHandler.java
+++ b/core/src/main/java/org/owasp/dependencycheck/xml/suppression/SuppressionHandler.java
@@ -18,8 +18,12 @@
 package org.owasp.dependencycheck.xml.suppression;
 
 import java.util.ArrayList;
+import java.util.Calendar;
 import java.util.List;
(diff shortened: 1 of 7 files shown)


aikebah:
@jeremylong Let me know if you'd rather have a PR from a rebased version to pass the travis-ci build (with a proper NVD datafeed configuration)

jeremylong:
@aikebah if you wouldn't mind rebasing this - I would appreciate it.

Thanks!

aikebah:
branch rebased to current master

Issue #740, opened 2017-05-22
aggregate goal seems to ignore reactor modules (always fetches from repo) in certain circumstances


jeff303:
There seems to be some issue with dependency-check-maven where, when using the `aggregate` type output, the plugin is downloading all dependencies from the repositories instead of using the local build versions via Reactor.  This started happening somewhere between version 1.4.0 and 1.4.5, since we did not experience this problem in 1.4.0.  And [this forum post](https://groups.google.com/forum/?fromgroups#!topic/dependency-check/88ctFUZBIns) suggests a large part of this was rewritten within that range.

Full build log available [here](https://www.dropbox.com/s/3lf0y5jworddkoy/build5.out?dl=0).  Relevant snippets pasted below.  Notice the `Downloading` part near the end.  This should not be happening, since this is a local module, so it should be built first, and its local artifacts used in the analysis.

```
Apache Maven 3.3.9 (bb52d8502b132ec0a5a3f4c09453c07478323dc5; 2015-11-10T10:41:47-06:00)
Maven home: /usr/local/Cellar/maven/3.3.9/libexec
Java version: 1.8.0_111, vendor: Oracle Corporation
Java home: /Library/Java/JavaVirtualMachines/jdk1.8.0_111.jdk/Contents/Home/jre
Default locale: en_US, platform encoding: UTF-8
OS name: "mac os x", version: "10.12.4", arch: "x86_64", family: "mac"
...<snip>...
[INFO] Reactor Build Order:
[INFO]
[INFO] StreamSets Data Collector RB Gen Maven Plugin
[INFO] StreamSets Data Collector
[INFO] StreamSets Data Collector Root Prototype POM
[INFO] StreamSets Data Collector Root POM
[INFO] StreamSets Testing Library
[INFO] StreamSets Data Collector Bootstrap
[INFO] StreamSets Utils
[INFO] StreamSets SSO
[INFO] StreamSets Data Collector Common
...<snip>...
...NOTE: this is where the undesired download happens.  It should be using the local Maven module...
[DEBUG] Using connector BasicRepositoryConnector with priority 0.0 for https://repo.streamsets.net/artifactory/libs-snapshot
Downloading: https://repo.streamsets.net/artifactory/libs-snapshot/com/streamsets/streamsets-datacollector-common/2.6.0.0-SNAPSHOT/streamsets-datacollector-common-2.6.0.0-20170501.145725-107.jar
```

[...]

PR #1165, opened 2018-03-28
Upgrade maven-dependency-tree version

aikebah:
## Description of Change
As a follow up to comments in issue #740
* Update maven shared library dependency-tree to the current version
* Adapt the DependencyCheckMojo where needed to restore compatibility with this newer lib

no new functionality; existing testcases pass and are assumed to cover dependency-resolution

### DIFF
--- a/maven/src/main/java/org/owasp/dependencycheck/maven/BaseDependencyCheckMojo.java
+++ b/maven/src/main/java/org/owasp/dependencycheck/maven/BaseDependencyCheckMojo.java
@@ -720,7 +720,7 @@ public abstract class BaseDependencyCheckMojo extends AbstractMojo implements Ma
      */
     protected ExceptionCollection scanArtifacts(MavenProject project, Engine engine) {
         try {
-            final DependencyNode dn = dependencyGraphBuilder.buildDependencyGraph(project, null, reactorProjects);
+            final DependencyNode dn = dependencyGraphBuilder.buildDependencyGraph(project.getProjectBuildingRequest(), null, reactorProjects);
             final ProjectBuildingRequest buildingRequest = newResolveArtifactProjectBuildingRequest();
             return collectDependencies(engine, project, dn.getChildren(), buildingRequest);
         } catch (DependencyGraphBuilderException ex) {
--- a/maven/src/main/java/org/owasp/dependencycheck/maven/BaseDependencyCheckMojo.java
+++ b/maven/src/main/java/org/owasp/dependencycheck/maven/BaseDependencyCheckMojo.java
@@ -26,7 +26,6 @@ import java.util.List;
 import java.util.Locale;
 import org.apache.maven.artifact.Artifact;
 import org.apache.maven.artifact.repository.ArtifactRepository;
-import org.apache.maven.artifact.resolver.filter.ArtifactFilter;
 import org.apache.maven.artifact.versioning.ArtifactVersion;
 import org.apache.maven.doxia.sink.Sink;
 import org.apache.maven.execution.MavenSession;
--- a/maven/src/main/java/org/owasp/dependencycheck/maven/BaseDependencyCheckMojo.java
+++ b/maven/src/main/java/org/owasp/dependencycheck/maven/BaseDependencyCheckMojo.java
@@ -26,6 +26,7 @@ import java.util.List;
 import java.util.Locale;
 import org.apache.maven.artifact.Artifact;
 import org.apache.maven.artifact.repository.ArtifactRepository;
+import org.apache.maven.artifact.resolver.filter.ArtifactFilter;
 import org.apache.maven.artifact.versioning.ArtifactVersion;
 import org.apache.maven.doxia.sink.Sink;
 import org.apache.maven.execution.MavenSession;
@@ -720,7 +721,7 @@ public abstract class BaseDependencyCheckMojo extends AbstractMojo implements Ma
      */
     protected ExceptionCollection scanArtifacts(MavenProject project, Engine engine) {
         try {
-            final DependencyNode dn = dependencyGraphBuilder.buildDependencyGraph(project.getProjectBuildingRequest(), null, reactorProjects);
+            final DependencyNode dn = dependencyGraphBuilder.buildDependencyGraph(newResolveArtifactProjectBuildingRequest(), null, reactorProjects);
             final ProjectBuildingRequest buildingRequest = newResolveArtifactProjectBuildingRequest();
             return collectDependencies(engine, project, dn.getChildren(), buildingRequest);
         } catch (DependencyGraphBuilderException ex) {
@@ -905,6 +906,7 @@ public abstract class BaseDependencyCheckMojo extends AbstractMojo implements Ma
     public ProjectBuildingRequest newResolveArtifactProjectBuildingRequest() {
         final ProjectBuildingRequest buildingRequest = new DefaultProjectBuildingRequest(session.getProjectBuildingRequest());
         buildingRequest.setRemoteRepositories(remoteRepositories);
+        buildingRequest.setProject(project);
         return buildingRequest;
     }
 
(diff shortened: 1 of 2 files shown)


aikebah:
@jeremylong Let me know if you'd rather have a PR from a rebased version to pass the travis-ci build (with a proper NVD datafeed configuration)

jeremylong:
@aikebah as with the other - would you mind rebasing this change?

Thanks!

aikebah:
Branch rebased to current master

jeremylong:
I upgraded maven-dependency-tree similar to your PR (which I was having issues with) and added a test case that I believe mimics the aggregate problem. This can be seen in the currently failing branch: https://github.com/jeremylong/DependencyCheck/tree/maven-tree-upgrade.   With your PR I kept getting blank dependency-check reports (i.e. zero dependencies).

I am beginning to think we may need to implement our own custom dependency resolution...

aikebah:
@jeremylong Looks like you overlooked one change that I discovered to be required by the new maven library: adding the MavenProject to the projectBuildingRequest, which in my PR is found at
https://github.com/jeremylong/DependencyCheck/pull/1165/commits/0815ed7a3645cd6a10d4cba1b06bb1b42de3101b#diff-34442e326d7be233d2d63917a6ee0ff6R909

In your branch after https://github.com/jeremylong/DependencyCheck/commit/ab174b4bb4379935ede44bab6f762d2f3c5c3341#diff-34442e326d7be233d2d63917a6ee0ff6R723
there should be a `buildingRequest.setProject(project);` to have the same effect

aikebah:
Created PR #1250 to your branch to have that omission fixed

aikebah:
> With your PR I kept getting blank dependency-check reports (i.e. zero dependencies)

With which test(s) does that occur? I rebased my branch (upgrade-maven-dependency-tree) locally to the current master to overcome the NSPAnalyzer test dependency-count issue and ran a `mvn clean install -DreleaseTesting`. I checked the reports inside maven/target/it and they appear to be OK to me.

If you'd like I can force-push to update this PR

jeremylong:
@aikebah thanks for pointing out my mistake.  So now everything seems to work - except the bug we need to fix ;)

I updated my branch, [maven-tree-upgrade](https://github.com/jeremylong/DependencyCheck/tree/maven-tree-upgrade), and when running `mvn install -DreleaseTesting=true` the invoker integration test, `740-aggregate`, fails with:

```
Caused by: org.apache.maven.plugin.MojoExecutionException: One or more exceptions occurred during dependency-check analysis
	at org.owasp.dependencycheck.maven.BaseDependencyCheckMojo.runCheck(BaseDependencyCheckMojo.java:954)
	at org.owasp.dependencycheck.maven.BaseDependencyCheckMojo.execute(BaseDependencyCheckMojo.java:590)
	at org.apache.maven.plugin.DefaultBuildPluginManager.executeMojo(DefaultBuildPluginManager.java:134)
	at org.apache.maven.lifecycle.internal.MojoExecutor.execute(MojoExecutor.java:208)
	... 11 more
Caused by: org.owasp.dependencycheck.exception.ExceptionCollection: One or more exceptions occurred during analysis:
	Could not find artifact org.owasp.test:fourth:jar:1.0.0-SNAPSHOT
	Could not find artifact org.owasp.test:fourth:jar:1.0.0-SNAPSHOT
	at org.owasp.dependencycheck.maven.BaseDependencyCheckMojo.collectDependencies(BaseDependencyCheckMojo.java:835)
	at org.owasp.dependencycheck.maven.BaseDependencyCheckMojo.scanArtifacts(BaseDependencyCheckMojo.java:726)
	at org.owasp.dependencycheck.maven.AggregateMojo.scanDependencies(AggregateMojo.java:69)
	at org.owasp.dependencycheck.maven.BaseDependencyCheckMojo.runCheck(BaseDependencyCheckMojo.java:922)
	... 14 more
```

Which in a lot of ways one should expect as when the aggregate goal runs the JAR, org.owasp.test:fourth:jar:1.0.0-SNAPSHOT, has not been built yet. I think the solution would be to narrow the exception handling for `ArtifactResolverException` to just wrap the call `artifactResolver.resolveArtifact()` on line 786. We then look to see if the specific artifact is in the reactor projects to be built. If it is, add an dependency with just the GAV info that doesn't point to the actual JAR...  Thoughts?

aikebah:
@jeremylong I agree that we  should handle the resolveArtifact differently, and your approach looks appropriate to me. But I think the modified logic should only apply for the aggregate goal (as I think that only in that case reactor projects (for a pristine version) would not yet yield a resolvable artifact)

I am wondering whether we should still make the engine scan the pom-file of the reactor project as a replacement of the logic lines 794-825 (not sure whether the information gained from these lines would lead to additional information in the aggregate report).

I'm offering to give it a try to implement based on your maven-tree-upgrade branch

jeremylong:
@aikebah After fixing test case 740 in my [maven-tree-upgrade](https://github.com/jeremylong/DependencyCheck/tree/maven-tree-upgrade) - the build and aggregate analysis works as expected. At this time I do not believe we need to do anything else to resolve issue #740.

Test case 740 does show an interesting result - the actual project itself is not included in the report. In 740 there are two JAR files that are not analyzed as they are not "dependencies" - but rather modules being built.  What are your thoughts on including modules in the report?

aikebah:
@jeremylong I thought your remark on the modules was regarding a change in behaviour, but as far as I can tell it describes the current behaviour of the aggregate report (At least I get a similar report with 3.1.2 if I go through the trouble of creating an initial build so that the artifacts are present in the local repository)

For my use-case not including the modules themselves if they are not a dependency of some other module would be fine. I would expect anyone responsible for building a project to know about their own vulnerabilities and only be interested in the vulnerabilities of **_external_** modules.
I would not expect a practice of somehow building/bundling some **_external_** module as a module of your own multimodule project tree if that external module was not also a dependency of one of your own modules (and thus would still be included in the analysis).
It also properly fits the brand of the tool, as the left-out modules are not dependencies. It might be useful for readers of the aggregate report to include a section (by default collapsed) to show which versions of which modules have had their dependencies scanned (rather than only showing the top-level project GAV coordinates as the module for which the aggregate report was created)


Regarding your testcase in its current form: it succeeds when other integration-tests have installed the org.owasp.tests:fourth project. For a proper test of issue #740 it needs artifacts that are not-yet present in the repository, so you should not reuse the artifact coordinates of other integrationtests. 

I was looking why the maven-plugin would build successfully locally with integration tests whereas a local mvn verify of just the it for 740 was failing. 

Turned out that in my case 690-threadsafety runs first and 740-aggregate later, so it finds the artifacts of 690 in the localRepo:
`[DEBUG] Skipped remote request for org.owasp.test:fourth:1.0.0-SNAPSHOT/maven-metadata.xml, locally installed metadata up-to-date.`
In Travis-CI your corrected test-case fails because 690-threadsafety runs later that 740-aggregate in the integration tests (so it did not yet install org.owasp.test:fourth:1.0.0-SNAPSHOT in the local repo). So there is still some work todo to get #740 fixed.

aikebah:
Locally fixed the testcase and trying implementation of the proposed fix. If that succeeds I'll create a PR against your [maven-tree-upgrade](/jeremylong/DependencyCheck/tree/maven-tree-upgrade) branch.

aikebah:
@jeremylong Maybe I did read your comment correct after all. I see that the child-module dependency is removed from the report when I apply the proposed fix to the code.

For my use-case I do not see a problem in this (as the module's dependencies are still properly scanned), but I can imagine other projects that build (some of) their dependencies from source including modules for (unmodified) open-source projects might want to have those modules scanned for known vulnerabilities as well.

Not sure if this warrants for an extension of the Engine to also support scanning of a bare POM-file as a dependency (writing one seems to be an easy task, basically it would be a stripped-down version of the setPomEvidence of the JarAnalyzer)

As I assume you've already prepared the proposed patch locally I'm creating a PR for just the test-case fix so that it will properly fail regardless of the sequence of integrationtests.

jeremylong:
See PR #1266.

aikebah:
@jeremylong From a quick-scan of the diff: looking forward to the next release including your changes. Looks good with the introduced virtual dependency.

PR #1170, opened 2018-03-29
Remove out-dated comment that the check tasks dependsOn dependencyChe…

stefanneuhaus:
The [Documentation](https://jeremylong.github.io/DependencyCheck/dependency-check-gradle/index.html#Step_2_Run_the_dependencyCheckAnalyze_task) of the Gradle plugin states:

> Once the dependency-check plugin is applied, if the Java plugin is being used dependency-check will automatically be added to the check task.

 This feature has been removed with jeremylong/dependency-check-gradle#26 . I updated the documentation accordingly.


jeremylong:

Issue #1177, opened 2018-04-02
Expected positive CVE id CVE-2017-8046


mjeanroy:
Hi,

A positive identification of CVE-2017-8046 is expected from dependency:

```xml
<dependency>
  <groupId>org.springframework.boot</groupId>
  <artifactId>spring-boot-starter-data-rest</artifactId>
  <version>1.5.2.RELEASE</version>
</dependency>
```

Note that the CVE is reported with a fixed version of `spring-boot-starter-data-rest`:

```xml
<dependency>
  <groupId>org.springframework.boot</groupId>
  <artifactId>spring-boot-starter-data-rest</artifactId>
  <version>2.0.0.RELEASE</version>
</dependency>
```

jeremylong:
This is related to #1088 with the base cause being that we need to convert from using the XML data feeds to the new JSON data feeds (issue #707).

jeremylong:
issue has been resolved.

PR #1178, opened 2018-04-02
added suppression spring boot and CVE-2017-8046

YoshiMan:
see https://pivotal.io/security/cve-2017-8046

## Description of Change

cve-2017-8046 is fixed in spring boot 2.0.0.Release. So it should be excluded. 

*no*, since this is a default config.

### DIFF
--- a/core/src/main/resources/dependencycheck-base-suppression.xml
+++ b/core/src/main/resources/dependencycheck-base-suppression.xml
@@ -2074,4 +2074,16 @@
         <gav regex="true">^io\.vertx:vertx-config-kubernetes-configmap:.*$</gav>
         <cpe>cpe:/a:kubernetes:kubernetes</cpe>
     </suppress>
+    <suppress>
+        <notes><![CDATA[
+        This suppresses the false positive
+        <dependency>
+            <groupId>org.springframework.boot</groupId>
+            <artifactId>.*</artifactId>
+            <version>2.0.0.RELEASE</version>
+        </dependency>
+        ]]></notes>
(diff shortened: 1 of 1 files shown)


jeremylong:

core/src/main/resources/dependencycheck-base-suppression.xml
```diff
@@ -2074,4 +2074,16 @@
...
+    <suppress>
+        <notes><![CDATA[
+        This suppresses the false positive
+        <dependency>
+            <groupId>org.springframework.boot</groupId>
+            <artifactId>.*</artifactId>
+            <version>2.0.0.RELEASE</version>
+        </dependency>
+        ]]></notes>
+        <gav regex="true">^org\.springframework\.boot:.*:2.0.0.RELEASE$</gav>
```
jeremylong, line 2086: Not sure I agree with the regex as I feel it is too broad (in that all artifacts are matched) and too specific in that only a single version is matched. See how this was done for a similar Jackson vulnerability: https://github.com/jeremylong/DependencyCheck/blob/master/core/src/main/resources/dependencycheck-base-suppression.xml#L534

I'd be okay if we suppressed this for everything but spring-data-rest using a negative lookahead.

jeremylong (changes requested):
See my comments about the regex.

jeremylong:
See issue #1177 - I believe this will be resolved once we upgrade from the XML data feeds to the JSON data feeds. The version information is more complete in the JSON data feeds.

jeremylong:
This will be included in the upgrade from XML to JSON data feeds.

PR #1184, opened 2018-04-03
build docker image in travis

stephengroat:
## Description of Change

builds docker image in travis
[can also run docker deploy from travis](https://docs.travis-ci.com/user/docker/#Branch-Based-Registry-Pushes)

N/A


jeremylong:

jeremylong:
The issue with this PR is that the call to docker occurs before the maven site is deployed. The docker build file grabs the latest version of dependency-check from the maven site hosted at: http://jeremylong.github.io/DependencyCheck/current.txt

So to make this work we would need to utilize a different mechanism to get the latest version. Additionally, the docker container should only be built on tagged versions. Lastly, to build the correctly tagged docker images we have been using the [publish_docker.sh](https://github.com/jeremylong/DependencyCheck/blob/master/publish_docker.sh)

.travis.yml
```diff
@@ -40,6 +41,12 @@ matrix:
         - JDK="JDK8"
```
jeremylong: Should the JDK8 section contain `stage: build`?

stephengroat:
understood

let me try throwing that change into the POM, that way it's using the built copy instead of a remote copy

.travis.yml
stephengroat: it's not the clearest thing in the world, but in https://docs.travis-ci.com/user/build-stages/#Build-Stages-and-Build-Matrix-Expansion, it shows that:

> Each job included in jobs.include inherits the first value of the array that defines a matrix dimension.

So, JDK8 inherits the build from the stage above. Not the clearest, but it seems to work

stephengroat:
@jeremylong let me know if this is more of what you're looking for. `dockerfile-maven` provides the ability to build and deploy docker images directly from maven. a little bit of authentication work would have to be done from within travis to get deploys working, but it's not impossible at all.

if that's the case, might want to delete `publish_docker.sh` as well and just use maven to deploy

jeremylong:
Thanks - this looks great.  On the next (non-emergency) release I will merge the PR.  I'll likely move the maven plugin to a profile and setup travis to publish when it processes a tagged version.

jeremylong (approved):

Issue #1120, opened 2018-02-28
Report files contain an exorbitant amount of whitespace characters


randomstability:
Java version: "1.8.0_162"

OS: Windows 7 Enterprise 64 bit
Service Pack: SP1
Patch level: Current through February 2018

DependencyCheck version: 3.1.1 CLI

This isn't critical but gives concerns over the robustness of the parsing occurring.

I don't feel I need to provide any example report.  But I used WebGoat as the file to scan (again, like my last report).  Almost half the report file size is from whitespace characters.  This is doubling the size of the output files.

There could easily be multiple sources of this whitespace bloat, but it seems like it could be a simple task to trim whitespace.  I'm no expert so I could be wrong.  Ideally locating the source calls causing this is best,  but that could be tedious.

This is fairly easy to assess using Notepad++ on Windows and removing all double space strings until they are all gone.  Then repeating this for empty lines and tab characters.  Saving out this reduced file can reveal the character count and size difference.

stevespringett:
Your observation is correct. The JSON and XML reports are designed to be both human and machine readable. 

Compressing the HTML report by default might be an acceptable thing to do. I'm not sure about the JSON or XML reports though. This might need to be an option as I find myself going into these reports on a regular basis to view them. Having to add whitespace to make them human readable again just creates a new problem.

But perhaps an option to compress the resulting report might be useful.

randomstability:
Thanks for the explanation.  At the least,  an option/flag would be nice, as I'm not dealing with small HTML reports here.

Also isn't whitespace irrelevant for modern file editors/viewers?  Don't they all format the data in a structured format?  I know a basic text editor won't but why is anyone using one?

stevespringett:
The use case I'm thinking of where whitespace is required is in the Jenkins plugin. In Jenkins, the XML report is used. The HTML report can optionally be generated but cannot be viewed from Jenkins because the report violates the default CSP settings, so most users don't bother. But viewing the XML report from the Jenkins workspace is something I always find myself doing.  So in these cases, the report would need to be human readable.

I recently scanned a Nexus repo with Dependency-Check which resulted in a report exceeding a few hundred megabytes, so I totally understand the need to compress as I could have benefited from that feature myself. :-)

jeremylong:
[...]

PR #1211, opened 2018-04-16
Remove Excess Whitespace

jeremylong:
## Fixes Issue #1120

Reduces the amount of whitespace in the generated reports. The exception is still the JSON report as it is still being pretty printed - this should be changed to an option.

Existing tests are in place.


jeremylong:

PR #1212, opened 2018-04-16
Reduce dependency on central

jeremylong:
## Fixes Issue - too many to reference...

- If a Java dependency already has analyzed a pom.xml the Central Analyzer will be skipped
- The Maven plugin no longer uses the central analyzer

existing


jeremylong:

PR #1220, opened 2018-04-21
CentralAnalyzer Configuration Updates

jeremylong:
This PR simply moves a few hard-coded configurations to be properties so that they could be updated without having to re-deploy code.


jeremylong:

PR #1228, opened 2018-04-23
Resolve FP on io.gitlab.arturbosch

reddyalready:
## Description of Change

Fixes a false positive encountered on `io.gitlab.arturbosch` namespace due to `io.gitlab` namespace (cpe) 

no


jeremylong:

Issue #1070, opened 2018-01-19
JenkinsPlugin: [DependencyCheck] Message: Unable to download the NVD CVE data.


ammy1999:
I am trying to run dependency check as a stage in jenkins file, but I have an error 

> [DependencyCheck] Message: Unable to download the NVD CVE data.
[DependencyCheck] org.owasp.dependencycheck.data.update.exception.UpdateException: Unable to download the NVD CVE data.
[DependencyCheck] 	at org.owasp.dependencycheck.data.update.NvdCveUpdater.update(NvdCveUpdater.java:130)

So how can I resolve this problem please !?

MenschNestor:
Same here. The problem has appeared overnight, the nightly build before still worked.

I had some success tracking the problem to SSLSocketFactoryEx in dependency-check-utils. It seems if I remove TLSv1 from the preferred protocols, the download works again. Unfortunately, the property `downloader.tls.protocols` is not exposed in the Maven plugin settings, so I cannot test if it would help setting it to `TLSv1.1,TLSv1.2,TLSv1.3`.

I tried the Oracle JDK and the BouncyCastle provider, but nothing worked so far.

Some excerpt from the logs (maven -X):
```
15:05:22 [DEBUG] #download   threads: 12
15:05:22 [DEBUG] #processing threads: 8
15:05:22 [INFO] starting getUpdatesNeeded() ...
15:05:22 [DEBUG] Checking for updates from: https://nvd.nist.gov/feeds/xml/cve/nvdcve-2.0-2002.xml.gz
15:05:22 [DEBUG] Checking for updates from: https://nvd.nist.gov/feeds/xml/cve/nvdcve-2.0-2015.xml.gz
15:05:22 [DEBUG] Checking for updates from: https://nvd.nist.gov/feeds/xml/cve/nvdcve-2.0-2006.xml.gz
15:05:22 [DEBUG] Checking for updates from: https://nvd.nist.gov/feeds/xml/cve/nvdcve-2.0-2007.xml.gz
15:05:22 [DEBUG] Checking for updates from: https://nvd.nist.gov/feeds/xml/cve/nvdcve-2.0-2004.xml.gz
15:05:22 [DEBUG] Checking for updates from: https://nvd.nist.gov/feeds/xml/cve/nvdcve-2.0-2005.xml.gz
[...]

PR #1237, opened 2018-04-24
removed tlsv1

jeremylong:
## Fixes Issue #1070

It appears having TLSv1 in the list of available protocols is causing issues in some environments. As such, as this should not be used anymore we are removing it from the default list.

See comments on #1070 and https://github.com/spotbugs/spotbugs-maven-plugin/pull/41.

Existing test cases cover the change.


jeremylong:

PR #1252, opened 2018-04-29
Migrate from deprecated commons-lang3 to supported commons-text

aikebah:
## Description of Change
Adapted the code to use the supported functions of the Apache Commons Text project in favor of the corresponding deprecated functions of Apache Commons Lang v3

no, it is assumed that the modified code is already covered by existing tests


stevespringett:
@jeremylong I validated this dependency does not have conflicts with Jenkins plugin.

PR #1260, opened 2018-05-02
Fix testcase 740-aggregate: give it unique GAV coordinates

aikebah:
## Description of Change
As referenced in #1165 comments a PR to update to 740-aggregate test so that it is guaranteed to fail regardless of the sequence in which the integrationtests are run.
The testcase will fail until #740 has been fixed.


jeremylong:

PR #1265, opened 2018-05-06
Fix Possible Path Traversal from Archive Files

jeremylong:
Resolves a possible security related issue where an archive (zip, jar, war, etc.) could contain a `name` with directory traversal sequences that would lead to a file being extracted to an unexpected location. An example archive can be found at  https://gist.github.com/grnd/61e3244bfff85cbe025f5bd4c60ba842. 

This PR adds the detection of the above scenario and if found an exception will be thrown.


jeremylong:

PR #1266, opened 2018-05-06
Fix dependency resolution

jeremylong:
## Fixes Issue #740

* Upgrades maven-dependency-tree (this PR supersedes #1165)
* When analyzing an aggregate build (in Maven) if a child module has not been built and as such, the JAR file does not exist, a warning will be printed and a dependency will be added that contains the evidence from the build - but does not point back to a specific file.
* Fixed bug that would cause pom.xml dependencies without an identifier to be incorrectly merged with any other JAR.

yes

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -49,6 +49,7 @@ after_success:
     fi;
 
 after_failure:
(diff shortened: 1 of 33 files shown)


jeremylong:

Issue #1271, opened 2018-05-11
A Way to Overwrite Skip Configuration


markdenihan:
As a user
I want to be able to overwrite a skip configuration in my pom.xml
So that I can specify in a mvn command when I want to skip/not skip dependency check

#### Example
pom.xml Dependency Check Plugin Config
```xml
<plugin>
    <groupId>org.owasp</groupId>
    <artifactId>dependency-check-maven</artifactId>
    <version>3.1.2</version>
    <configuration>
        <skip>true</skip>
    </configuration>
    <executions>
        <execution>
            <goals>
                <goal>aggregate</goal>
            </goals>
        </execution>
    </executions>
</plugin>
```
run maven from command line
```console
mark@mark-someMachine:~/someProject mvn clean verify -Ddependency-check.skip=false
```
Currently, this will result in dependency check being skipped. Is there any way to give the user property priority?

albuch:
You can use a property for the skipping value and override this on command line:

```
<properties>
    <dependency.check.skip>true</dependency.check.skip>
</properties>

...

<plugin>
    <groupId>org.owasp</groupId>
    <artifactId>dependency-check-maven</artifactId>
    <version>3.1.2</version>
    <configuration>
        <skip>${dependency.check.skip}</skip>
    </configuration>
    <executions>
        <execution>
            <goals>
                <goal>aggregate</goal>
            </goals>
        </execution>
    </executions>
</plugin>
```
and on command line:
``mvn clean verify -Ddependency.check.skip=false``

jeremylong:
Thanks for the suggestion. While one could define your own property - this is an easy enough fix.

PR #1283, opened 2018-05-18
Allow 'skip' to be used from an environment variable

jeremylong:
## Fixes Issue 1271

While there are many ways to accomplish #1271 (one could use a profile or other customer user property) in the pom.xml - this PR adds yet another way to control if dependency-check will be run as part of the build. One can configure dependency-check to be skipped by default (using `<skip>true</skip>`)in the Maven build. Then when you want to run dependency-check you can simply run:

```
 mvn verify -Ddependency-check.skip=false
```


jeremylong:

PR #1284, opened 2018-05-18
Resolve Several Reported False Positive

jeremylong:
This PR resolves many of the reported false positives.


jeremylong:

PR #1315, opened 2018-06-02
Cleanup/Java 7 Compatability

jeremylong:
GMaven Plus version information was configured in multiple places - consolidated these into a property and used version 1.5 to maintain Java 7 compatibility.


jeremylong:

Issue #866, opened 2017-08-30
FP in Springboot


javixeneize:
Hi

I am using a project that uses springboot in a non vulnerable version, and CVE 2016-9878 is being reported there

The analysis has found this in a component that is version 1.0.0 and uses springboot, but it is not springboot itself

See details below

Source	Name	Value
file	name	myname
file	version	1.0.0
Manifest	Implementation-Title	myname
Manifest	implementation-url	http://projects.spring.io/spring-boot/myname/
Manifest	Implementation-Vendor	Pivotal Software, Inc.
Manifest	Implementation-Vendor-Id	ID
Manifest	Implementation-Version	1.0.0-SNAPSHOT
Manifest	spring-boot-classes	BOOT-INF/classes/
Manifest	spring-boot-lib	BOOT-INF/lib/
pom	artifactid	myname
pom	description	Desc
pom	groupid	myname.myproject
pom	name	myname
pom	parent-artifactid	spring-boot-starter-parent
pom	parent-groupid	org.springframework.boot
pom	version	1.0.0-SNAPSHOT
Identifiers

maven: com.myname.myproject:1.0.0-SNAPSHOT   Confidence:HIGH
cpe: cpe:/a:pivotal_software:spring_framework:1.0.0   Confidence:LOW   suppress

Hope that helps

Thanks

jeremylong:
After reviewing this - resolving this class of FP will likely take a bit more effort.  In the meantime, I would suggest that users utilize a local suppression file.

ammy1999:
I don't unterstand How can I utilize a local suppression file !! Any help please

jeremylong:
@ammy1999 please see the answer at https://github.com/jeremylong/DependencyCheck/issues/981#issuecomment-347355230

ammy1999:
Thanks @jeremylong  , but please just a question why I should add the suppression file or this file exist in the code source !! it can remove the FP automatically or should update this file to remove successfully the FP

jeremylong:
Many users of dependency-check have a local suppression file (separate from the one included with dependency-check-core). This is done so you can more quickly suppress false positives, deal with false positives that are local to your environment (i.e. your organizations own libraries are being flagged incorrectly, etc.).

javixeneize:
Hi

Any chances of having this issue supressed in future releases?

Thanks

PR #1316, opened 2018-06-02
Resolve FP for spring-boot project perissue #866

jeremylong:
## Fixes Issue #866

Skips evidence from the spring-boot parent pom that causes mis-identification.


jeremylong:

PR #1332, opened 2018-06-15
StackOverflowException exception if two related dependencies reference each other

reyem:
## Fixes Issue #
fixes a stackoverflow exception if two related dependencies have each other in the <relatedDependencies> list. This caused a ping pong in checking the path again and again on the same files resulting in a StackOverflowException.fix stackoverflow exception.

## Description of Change
The change simply switches the arguments of hasSameBasePath to avoid the ping pong. So dependency1 is checked against each child of dependency2.

no


jeremylong:

jeremylong:
Thanks for the PR!

Issue #936, opened 2017-10-20
Analysis task timed out in Version 3.0.1


timothygruetzner:
Hi!

Since I updated to version 3.0.1 my analysis task times out. Everything runs fine until the jar analyzer finishes. Then the application waits for like 10 minutes, only to then fail with the message below:

[ERROR] null

Analysis task timed out.
[ERROR] One or more fatal errors occurred
[ERROR] null

The full log can be found here: https://gist.github.com/timothygruetzner/00f9a1b062fc7f68b4b5ab6d00c77879

(Sorry if this issue wasn't opened correctly, am very new to GitHub)

Thanks for your help!

timothygruetzner:
I was able to track down the issue a bit further: If I disable the central analyzer the task runs through perfectly.
Apparantly the connection to http://search.maven.org/ doesn't work from my network at the moment, from other networks everything works fine.

Still: I think the error handling on this one does not work that well! At least an error message could appear, that no connection to maven central was possible instead of silently waiting and then failing with a simple analyzer-timeout...

jeremylong:
While we know the central analyzer has increased the analysis time - this is the first report I've seen for a failure like this.  Could you re-run this with the `--log odc.log` and provide the log file?

In addition, we are looking at alternatives to the central analyzer.

TheNitek:
I think I ran into the same (or at least a similar issue). I ran " mvn clean package dependency-check:aggregate -Dmaven.test.skip=true -X" which resulted in the attached log file
[owasp.txt](https://github.com/jeremylong/DependencyCheck/files/1680636/owasp.txt)

I am on a proxied network, so network problems are highly likely (even though everything should be configured fine on my system)

jeremylong:
@TheNitek I'm curious how you ran dependency-check?  Was this just via the command line or via some other script that might have a timeout?  The reason I'm asking is that the error message was slightly incorrect - it should have been `Analysis task was cancelled` rather than `Analysis task timed out`.  

The actual call was to `Future.get()` - which in my research does not have a default timeout and will wait forever to complete.

TheNitek:
I ran it from command line as mentioned above: "mvn clean package dependency-check:aggregate -Dmaven.test.skip=true -X" (didn't change anything when removing clean, package and the maven.test.skip). Ran it on windows without any other magic.

v6ak:
[...]

PR #1333, opened 2018-06-17
Increase Timeout

jeremylong:
## Fixes Issue #936

Increases the default analysis timeout from 10 minutes to 20 minutes. In addition, the analysis timeout was made configurable by setting the system property odc.analysis.timeout` (note, the timeout is set in minutes).


jeremylong:

PR #1360, opened 2018-06-30
Bug fixes

jeremylong:
Fixes two bugs:

1) In some cases the Md5 checksum was not being calculated correctly and caused an error message when the report was generated
2) added the retry count to the Central Analyzer for purposes of downloading the pom.xml if it had not already been analyzed.


jeremylong:

Issue #1355, opened 2018-06-26
Fails to scan node packages in some cases


atomicpirate:
### Log
The log that was generated when I take the below steps with ODC 3.2.1 is posted [as a gist](https://gist.github.com/atomicpirate/79e4bb750a35783fa92de715f235bf24).

#### Steps to Reproduce
Create an empty directory and then add this package.json file:
```
{
  "name": "test",
  "dependencies": {
    "jquery": "latest"
  }
}
```
Then run:
```
npm install
dependency-check -s . -o . --project test
```

The expected behavior is that ODC would detect and scan the latest jquery from npm.

The actual behavior is that a report is generated indicating that zero dependencies were scanned: "Dependencies Scanned: 0 (0 unique)".

PR #1361, opened 2018-07-01
Fix NPM and NSP Analysis

jeremylong:
## Fixes Issue #1355

The NSP and NPM Analyzers had a conflict in that they would remove the necessary files the other analyzer needed or generated.  This PR resolves the issue.


jeremylong:

Issue #1016, opened 2017-12-04
unable to parse pom


PeterFromWiesbaden:
### Reporting Bugs/Errors
When parsing a project with the commandline Utility the job ends with the error "unable to parse pom". All reports seem to be generated though. But the jenkins job ends with an error.
We don't use the plugin because we were not able to integrate it in a Jenkins pipeline.

Regards,
Peter

[INFO] Analysis Started
[INFO] Finished Archive Analyzer (19 seconds)
[INFO] Finished File Name Analyzer (0 seconds)
[Fatal Error] :2:10: DOCTYPE is disallowed when the feature "http://apache.org/xml/features/disallow-doctype-decl" set to true.
[Fatal Error] :2:10: DOCTYPE is disallowed when the feature "http://apache.org/xml/features/disallow-doctype-decl" set to true.
[WARN] Unable to parse pom '/tmp/dctemp70cd02b5-58b6-4a24-b10e-370afc53f5a2/check3647187796509583746tmp/707/pom.xml'
[WARN] An error occurred while analyzing '/buildssd/workspace/10.1-owasp/server/taa/product/target/products/de.ruv.ruvconn.server.application.taa.product/win32/win32/x86/plugins/org.jslipc_0.2.0.jar' (Jar Analyzer).
[WARN] Unable to parse pom '/tmp/dctemp70cd02b5-58b6-4a24-b10e-370afc53f5a2/check3647187796509583746tmp/708/pom.xml'
[WARN] An error occurred while analyzing '/tmp/dctemp70cd02b5-58b6-4a24-b10e-370afc53f5a2/check8483265078113306401tmp/419/lib/jslipc-0.2.0.jar' (Jar Analyzer).
[INFO] Finished Jar Analyzer (2 seconds)
[INFO] Finished Central Analyzer (0 seconds)
[INFO] Finished Dependency Merging Analyzer (0 seconds)
[INFO] Finished Version Filter Analyzer (0 seconds)
[INFO] Finished Hint Analyzer (0 seconds)
[INFO] Created CPE Index (1 seconds)
[INFO] Finished CPE Analyzer (6 seconds)
[INFO] Finished False Positive Analyzer (0 seconds)
[INFO] Finished Cpe Suppression Analyzer (0 seconds)
[INFO] Finished NVD CVE Analyzer (0 seconds)
[INFO] Finished Vulnerability Suppression Analyzer (0 seconds)
[...]

PR #1372, opened 2018-07-06
Filter doctype from pom.xml

jeremylong:
## Fixes Issue #1016 

Doctypes defined in the pom.xml would cause parsing the xml to fail as doctypes are disallowed. The resolution was to just strip everything from the pom.xml up to the `<project>` tag.

yes


jeremylong:

PR #1373, opened 2018-07-06
Code cleanup, checkstyle suggestions, etc.

jeremylong:
General code cleanup.


jeremylong:

PR #1375, opened 2018-07-10
Only match '-' version to all versions if no other versions named

MrBerg:
## Fixes Issue #

For instance, https://nvd.nist.gov/vuln/detail/CVE-2018-8012 is currently assumed by dependency-check to affect **all** versions of Apache ZooKeeper, even though the list of vulnerable versions (<3.4.10, >=3.5.0, <= 3.5.3) does not include e.g. 3.4.11. This due to the first item in the list of vulnerable versions before 3.4.10 being `cpe:/a:apache:zookeeper:-` (`cpe:2.3:a:apache:zookeeper:-:*:*:*:*:*:*:*`)

Similar things happen with e.g. https://nvd.nist.gov/vuln/detail/CVE-2016-5017, https://nvd.nist.gov/vuln/detail/CVE-2018-0124 and https://nvd.nist.gov/vuln/detail/CVE-2018-0101.

The CPE 2.2 definition (https://cpe.mitre.org/files/cpe-specification_2.2.pdf), pp. 9-10 mentions that "
**The '-' term might also be used to describing an initial release** (bolding mine) if a platform does identify with a given component but no term is commonly used to describe it.  An example of this might be the initial release of an application before any updates have been released. [...]
It is important to note that when a CPE Name is being created for a specific platform type (i.e. a 
specific release of an application) then the hyphen should be used at all times (when a vendor 
term is not available) instead of a blank component.  The blank component is to be used only 
when a more general platform type is being enumerated.  (i.e. any version of an application)"

The CPE 2.3 naming specification (https://nvlpubs.nist.gov/nistpubs/Legacy/IR/nistir7695.pdf) P. 16 (printed page number is 10): "The logical value NA SHOULD be assigned when there is no legal or meaningful value for that attribute, or when that attribute is not used as part of the description. This includes the situation in which an attribute has an obtainable value that is null."
P. 24/18: "The logical values ANY and NA are defined as possible attribute values in WFNs. The logical value ANY SHALL bind to what [CPE22] calls a “blank” (i.e., an empty component, indicated by two 
sequential colons) in the URI. **The logical value NA SHALL bind to a single hyphen.** (bolding mine)"

The CPE dictionary (https://nvd.nist.gov/Products/CPE), marks initial versions with a hyphen and replaces it with the version number in subsequent versions. The CVE DB uses the same dictionary for listing affected versions. It does not always use the hyphen in the list of configurations e.g. in CVE 2018 `{
          "vulnerable" : true,
          "cpe22Uri" : "cpe:/a:apache:zookeeper",
          "cpe23Uri" : "cpe:2.3:a:apache:zookeeper:*:*:*:*:*:*:*:*",
          "versionEndExcluding" : "3.4.10"
        }`

So blank version numbers should match every version, and so do '-' versions *but only if there are no other named versions*.

## Description of Change

When finding matching software versions for a CVE in CVE DB, only assume that the special case (hyphen) matches all versions if there are no other version numbers listed.

Yes

### DIFF
--- a/core/src/main/java/org/owasp/dependencycheck/data/nvdcve/CveDB.java
+++ b/core/src/main/java/org/owasp/dependencycheck/data/nvdcve/CveDB.java
@@ -959,7 +959,7 @@ public final class CveDB implements AutoCloseable {
         String majorVersionMatch = null;
         for (Entry<String, Boolean> entry : vulnerableSoftware.entrySet()) {
             final DependencyVersion v = parseDependencyVersion(entry.getKey());
-            if (v == null || "-".equals(v.toString())) { //all versions
+            if (v == null || ("-".equals(v.toString()) && vulnerableSoftware.size() == 1)) { //all versions
                 return entry;
             }
             if (entry.getValue()) {
--- a/core/src/test/java/org/owasp/dependencycheck/data/nvdcve/CveDBIT.java
+++ b/core/src/test/java/org/owasp/dependencycheck/data/nvdcve/CveDBIT.java
@@ -32,6 +32,7 @@ import org.junit.Test;
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertNotNull;
 import static org.junit.Assert.assertNull;
+import static org.junit.Assert.assertFalse;
 import static org.junit.Assert.assertTrue;
 import static org.junit.Assert.fail;
 import org.junit.Before;
@@ -116,7 +117,7 @@ public class CveDBIT extends BaseDBTestCase {
         assertTrue("Expected " + expected + ", but was not identified", found);
 
         found = false;
-        expected = "CVE-2012-5370";
+        expected = "CVE-2012-5370"; // Affects all versions
         for (Vulnerability v : results) {
             if (expected.equals(v.getName())) {
                 found = true;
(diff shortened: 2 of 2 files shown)


jeremylong:

MrBerg:
Any progress on verifying this patch, @jeremylong et al?

jeremylong:
@MrBerg I apologize - I thought I had replied to this PR already (the comment from me above is just an automated comment added by github).  I completely agree with you that the handling of the `-` could use some improvements.  When this PR originally came in my plan was to incorporate the changes into the 4.0 release which will include a lot of other changes in this area.  However, due to life/work/etc. I have not been able to make progress yet on the 4.0 release (which the core is upgrading from the XML data feeds from the NVD to the JSON data feeds and that will move us from CPE 2.2 to CPE 2.3).

That being said, I will try and find some time to do some testing on the patch this weekend and include in a patch release shortly.

stevesuttonalertme:
Is anybody prepared to comment on the progress of this, please?  My team is awaiting this fix before progressing our use of the checker.  Thanks!

jeremylong:
Sorry for the delay - we are revamping the matching as we move from the XML data feeds to the JSON data feeds. Most of the matching will actually be implemented in a common CPE library: https://github.com/stevespringett/CPE-Parser/pull/1

I am currently verifying the version 2 of CPE-Parser to ensure it matches the specification. I will then implement the matching as specified in the CPE 2.3 spec.

jeremylong:
Sorry about the delay - this is covered in https://github.com/jeremylong/DependencyCheck/pull/1729.

PR #1377, opened 2018-07-11
General Code Cleanup and Bug Fixes

jeremylong:
Ensured properties were set correctly in multiple analyzers, updated the analysis of shaded jars to better display the correct information, code cleanup...


jeremylong:

PR #1385, opened 2018-07-18
Updates to support gradle android builds

jeremylong:
To better support android builds in the gradle plugin the default behavior of the `Dependency.getDisplayFileName()` was changed.

Additionally, the `DependencyMergingAnalyzer` was updated to merge related android dependencies (aar and the embedded classes.jar).

Other changes made were to the other file type analyzers to maintain correct behavior with the changes to the `getDisplayFileName()`.  Other cleanup was performed on some of the analyzers as part of this PR.


jeremylong:

bentmann:
> Other changes made were to the other file type analyzers to maintain correct behavior

The `CMakeAnalyzer` appears to have been missed, making its processed dependencies display "CMakeLists.txt" instead of the parsed name&version fields.

PR #1390, opened 2018-07-20
Change the GrokAssembly binary with one that works with dotnet Core.

thomasd-:
## Description of Change
GrokAssembly was loading the binaries to examine in order to identify it using Reflection.

The probleme was when the examined assembly was dependant of some .NET Core specific API, the assembly would refuse to load.

This new version of GrokAssembly lookup the information from the assembly in a way that does not fail when the Assembly is missing some API's.

You can check the source code of this binary here : https://github.com/thomasd-/GrokAssembly

*no*


jeremylong:

jeremylong:
Thank you for working on this! Truly - this is a much better solution then what I thought we were going to have to do.  However, and please don't take this personally, I'm not going to merge this PR ;)

I hope you understand the dangers of accepting a PR with an EXE in it. Instead - can you submit a PR against https://github.com/jeremylong/GrokAssembly?  I took a look at the [updates](https://github.com/thomasd-/GrokAssembly) you made and they look good.

TomJKlug:
Hi Jeremy, Noticed you accepted the merge on the GrokAssembly, we have been excited to utilize the docker container version in our CI/CD and was just wondering if there was more to check up on consuming the new GrokAssembly, or if we can expect an updated image soon. Thanks so much for your time and love the work you are doing on this!

jeremylong:
@TomJKlug While the GrokAssembly has been upgraded - I need to hold off till we release 4.0 to actually use the new dotnet core assembly in ODC. Switching from mono to dotnet core will be a breaking change for several users. That being said - 4.0.0 is already underway. There will likely be a release of 3.3.1 or 3.4.0 depending on what makes it in within a week. Then 4.0.0 will follow.

TomJKlug:
@jeremylong Thanks for the update, we will look forward to 4.0, thanks for the update!

thomasd-:
Hi, I see that you made the 4.0.0 release. Great !

However, this release does not seem to include the new GrokAssembly.exe ...

I believe there are people waiting for this change, when do you think it will be available ?

By the way, I don't understand why you wrote about switching from mono to dotnet-core being a huge change ? This new GrokAssembly.exe still runswith mono. Nothing has changed on this regard.

Thanks

jeremylong:
Yeah - version 4 was a somewhat unplanned release due to having vulnerable dependencies. To keep with semantic versioning I felt we need to bump the major version number - so the planned features I have said would be in 4.0.0 will be in 5.0.0.  Progress is being made on this but I do not have a firm timeline yet. I am hoping before the end of the year - a lot of things needed to change as we moved from the XML data feeds to the JSON data feeds.

thomasd-:
Thanks but I still don't understand why you need to do these changes before integrating this one ?

Correcting this issue is just a matter of changing a binary in the resource folder. Nothing in any part of the code needs to be changed.

jeremylong:
ACtually - it is a bit more then just swapping a binary.  In my testing we will need to change from mono to .net core. I also found a few bits of evidence that were not being extracted. The grok assembly code was updated but the changes on the ODC side have not been completed.

pms1969:
@jeremylong Any update on when this, or another solution, might be available.  We run the dependency check against a whole lot of dotnet core assemblies, and it's not optimal in it's current form.  I appreciate that it's not exactly simple, but it would be awesome if this could find it's way into a release soon.

EDIT:
I've just looked at your pull request.  @colezlaw doesn't look that active.  Is it best to just produce this patch as a fork in order to get it out there?

jeremylong:
See #1792.

Issue #1276, opened 2018-05-15
Implement Package URL into all standard reports


stevespringett:
### Abstract
There are inconsistencies in how identifiers are represented in the XML reports, both with regard to schema and the format of data.

### Observations
It appears that the identifier on relatedDependencies is only used for Maven projects which greatly reduces the usefulness of the element. It also appears the dependency-check schema is inconsistent between dependency and relatedDependency. One has a list of identifiers, the other only has a single identifier without any evidence. If a parser is using Jackson, or similar XML to object deserialization technique, both representations would need to be implemented on a single Dependency object resulting in fragile, less than ideal code.

In addition, the identifier on the dependency node has a slightly different format than the single identifier on the relatedDependency node.

### Recognition
Ideally, the solution would be to standardize both the schema and the format of data in the report. However, this may introduce breaking changes throughout the ecosystem.

### Proposed Solution
Leverage the Package URL specification to provide a structured, predictable, and supportable format to identify a components ecosystem and its relevant metadata for every dependency and relatedDependency.

The Package URL specification has already been implemented in Dependency-Track. Refer to https://docs.dependencytrack.org/datasources/routing/

The official Package URL Java implementation is located here: https://github.com/package-url/packageurl-java . This library will automatically create canonicalized Package URLs that adheres to the current specification.

The proposed solution should NOT modify the existing XML schema.

### Examples
```xml
<identifiers>
    <identifier type="maven" confidence="HIGH">
        <name>(org.apache.commons:io:1.3.4)</name>
    </identifier>
    <identifier type="package-url" confidence="HIGH">
        <name>Generated Package URL</name>
        <url>pkg:maven/org.apache.commons/io@1.3.4</url>
    </identifier>
</identifiers>
```
```xml
<identifiers>
    <identifier type="package-url" confidence="HIGH">
        <name>Generated Package URL</name>
        <url>pkg:npm/bootstrap@4.1.1</url>
    </identifier>
</identifiers>
```
```xml
<identifiers>
    <identifier type="package-url" confidence="HIGH">
        <name>Generated Package URL</name>
        <url>pkg:gem/jruby-launcher@1.1.2?platform=java</url>
    </identifier>
</identifiers>
```
```xml
<identifiers>
    <identifier type="package-url" confidence="HIGH">
        <name>Generated Package URL</name>
        <url>pkg:nuget/EnterpriseLibrary.Common@6.0.1304</url>
    </identifier>
</identifiers>
```

[...]

PR #1391, opened 2018-07-20
Sonatype OSS Index support

jdillon:
Adds support for analyzing dependencies using Sonatype OSS Index:

* https://ossindex.sonatype.org/


jeremylong:

jdillon:
Looking into why some of these automated checks have failed.

jdillon:
Appears the sync to central for the `ossindex-service-client` has not completed fully yet

core/src/main/java/org/owasp/dependencycheck/data/ossindex/detectors/MavenDetector.java
```diff
@@ -0,0 +1,97 @@
...
+                String[] parts = mid.getValue().split(":", 3);
+                if (parts.length == 3) {
+                    group = parts[0];
+                    name = parts[1];
+                    version = parts[2];
+                }
+            }
+        }
+
+        // TODO: may want to first consider just "central" and "nexus" sources, then guess from other evidence?
```
stevespringett, line 53: Do not do this! PackageURL contains standard types and 'maven' is a standard type. If you want to distinguish between maven central (the default when the maven type is specified) and another maven repo, use the 'repository_url' qualifier. Example: pkg:maven/org.apache.xmlgraphics/batik-anim@1.9.1?repository_url=repo.spring.io/release

Also, please consider using the official PackageURL implementation as standard types are included in it.

stevespringett:
If you're going to use the Sonatype implementation of PackageURL, be advised that it violates the current PackageURL specification by default. The generated purl will not be parsable by the official PackageURL implementations. So any system (such as Dependency-Track) will not be able to use the purl generated from Dependency-Check. It's also important to note the Sonatype docs https://ossindex.net/doc/coordinates state both an older version of the spec (starting with 'type') and the current version of the spec (starting with 'pkg:'), but the default RenderFlavor (as Sonatype calls it) is SCHEMELESS which violates the spec. I do not know if the other aspects of their implementation will generate non-compliant purls or not. Proceed with caution if you want to continue to use it.

I'd highly recommend using the official PackageURL implementation since that will always be kept up-to-date with the emerging spec and will be conformant by default.

The official implementation for Java is here: https://github.com/package-url/packageurl-java

It appears that when OSSIndex is enabled, #1276 may be satisfied, but when OSSIndex analyzer is disabled, support for PackageURL is removed. In order to achieve consistent support for PackageURL as defined in #1276, we may want to have the identifier added outside the OSSIndex code. That way, even if the OSSIndex analyzer is disabled, other analyzers (such as Nuget, Maven, or Npm) will still be able to generate valid PackageURL.

stevespringett:
https://github.com/sonatype/package-url-java/issues/4

core/src/main/java/org/owasp/dependencycheck/data/ossindex/detectors/MavenDetector.java
jdillon: I'm not quite sure what you are talking about.  This logic here doesn't have anything to do with the repository something comes from, but instead is suggesting to prefer the "central" and "nexus" evidence that DC has collected, which for Maven dependencies seem to have the best information about the coordinates.  Has nothing to do with PackageURL support for repository_url.

jdillon:
@stevespringett the PackageURL spec changed _after_ we implemented our implementation.  We were aware that there was some movement towards moving towards a 'pkg' scheme, but at the time we built our implementation this was not officially merged.

At the time, our implementation passed the testsuite, minus the issues with maven testsuite data, which I fixed here ( https://github.com/package-url/purl-spec/pull/37 )... but has just been gathering dust.  Since that request, the 'pkg' scheme change was made, which was about 2 months ago now.

We chose to make our impl support parsing both new + old, and support rendering new + old with a default.

Will continue this on the issue you created on the libraries git repo.

If DC wants to prefer the pkg scheme style rendering its pretty simple to change.

core/src/main/java/org/owasp/dependencycheck/data/ossindex/detectors/MavenDetector.java
stevespringett: Understood. I wasn't looking at the logic, only the TODO comment by itself. Just wanted to make sure it was understood.

core/src/main/java/org/owasp/dependencycheck/analyzer/OssIndexAnalyzer.java
```diff
@@ -0,0 +1,162 @@
...
+                }
+            }
+        }
+    }
+
+    private Vulnerability transform(final ComponentReport report, final ComponentReportVulnerability source) {
+        Vulnerability result = new Vulnerability();
+        result.setSource(Vulnerability.Source.OSSINDEX);
+
+        result.setName(String.format("%s: %s", source.getId(), source.getTitle()));
```
stevespringett: The name of the vuln is going to be too long to be useful and will negatively impact display in the Jenkins plugin, SonarQube plugin, OWASP Code Pulse, OWASP Dependency-Track, and potentially other systems. The name field is designed to be short (like CVE-2018-7269 for example). The current approach would likely only be acceptable for viewing in the resulting HTML report, but not external systems.

The title returned from OSSIndex consists of multiple fields which is designed to be human readable, not machine readable. Either we need to parse the title field returned from OSSIndex into something smaller and more meaningful, or we wait for https://github.com/sonatype/ossindex-public/issues/1 to be implemented which would essentially do that for us. 

Additionally, including a UUID (source.getId()) as the first part of the vulnerability name is likely not a good strategy. In the contexts of looking at the data in external systems, UUID's are not meaningful and difficult for people to parse and remember, whereas something like OSVDB-3092 or CVE-2018-7269 is much easier for people to digest.

As is, I will have to permanently disable OSSIndex and ignore the results in both the SonarQube plugin (for which I have no way to modify the visual presentation in their platform) and the Jenkins plugin (for which I have very limited means to modify the visual presentation).

jdillon: I think best I can do ATM with the data provided, is to drop the source.getId().

jdillon:
Addressed some concerns, fixed some issues when a failure does occur when fetching.  ATM depends on snapshots of the client, so not ready for merge yet.

core/src/main/java/org/owasp/dependencycheck/data/ossindex/detectors/MavenDetector.java
jdillon: okay, the problem is that the existing analyzers for "pom" evidence seem to mutate things for optimal lookup in the lucene database.   Its unfortunate that impl doesn't set new evidence for that purpose.  But it looks like the central and nexus analyzers leave the groupId and artifactId unmolested, which is what the comment is referring to.

jeremylong: I'm confused on what you mean by the Jar analyzer "mutates" data going into the evidence.  Evidence is just any textual data gathered for purposes of CPE identification. If you are looking for the actual groupid/artifactid/version - that should be in the identifiers collection - if the JarAnalyzer has the pom the GAV is added to both the evidence and the Identifiers. I am planning on splitting out the vulnerability identifiers and the repository (i.e. GAV) into separate collections in the Dependency object.

jdillon: @jeremylong there is logic that collects things like "groupid" that mutates the data:

```
[DEBUG]       Evidence{name=groupid, source=pom, value=apache.axis, confidence=HIGHEST}
```

Though the central and/nexus evidence appears to just use them asis:

```
[DEBUG]       Evidence{name=groupid, source=central, value=org.apache.axis, confidence=HIGH}
```

jdillon:
Client updated, pushed pom changes to bring it in.

p3pijn:
What work is still to be done to get this PR merged?

jeremylong:
@p3pijn I normally don't merge PRs that are failing the build (see travis-ci build above).

I know there are a few things in this PR that will be more easily accomplished in the 4.x branch (which I am actively working on). I was planning on figuring out why the PR was breaking the build after a make some foundational changes to ODC over the next 3-4 weeks.  4.0.0 with this PR should be released before the end of the year.

strongwinderic:
> @p3pijn I normally don't merge PRs that are failing the build (see travis-ci build above).
> 
> I know there are a few things in this PR that will be more easily accomplished in the 4.x branch (which I am actively working on). I was planning on figuring out why the PR was breaking the build after a make some foundational changes to ODC over the next 3-4 weeks. 4.0.0 with this PR should be released before the end of the year.

Hi! Just wondering if it is still on the roadmap, since 4.0.0 was launched, and the end of the year is passed. No pushing, just asking.

jeremylong:
The 4.x branch that was pushed was in response to a vulnerability.  I would not recommend using the 4.x branch as the FP rate went way up.  I am just running some final testing on the 5.0.0-M1 branch (see https://github.com/jeremylong/DependencyCheck/tree/jsonDataFeeds_M1).  I will likely release the branch - that is the branch that made significant changes that would more easily integrate with this PR.

If @jdillon doesn't get to it first - ensuring that this PR (or at least the intent and integration) will be included before the final 5.0.0 release.

jdillon:
looks like ODC 5.x simplifies this a bit as we no longer need to have package detectors and first-phase analysis to generate package-url.  So I think this can be simplified a lot now, will have a look in the next few days.

jdillon:
replaced by https://github.com/jeremylong/DependencyCheck/pull/1771

PR #1402, opened 2018-07-26
new list of mysql server cves

stephengroat:
## Description of Change

excludes new long list of mysql server cves

*yes/no*


jeremylong:

Issue #1394, opened 2018-07-23
Failed to initialize the RetireJS repo


ghost:
When the data directory is in a different path than the temporary directory, the retirejs analyzer fails with "Failed to initialize the RetireJS repo" exception. See [https://gist.github.com/bcartolo/d737e07be45a9eb3bfaa4bfcc6cdbdd0](https://gist.github.com/bcartolo/d737e07be45a9eb3bfaa4bfcc6cdbdd0)

waded:
One possible workaround is to disable the RetireJS analyzer for now, e.g. in the Maven plugin config:

    <configuration>
        <retireJsAnalyzerEnabled>false</retireJsAnalyzerEnabled>
    </configuration>

ghost:
Another one is to change the location of the data directory to the same drive of the temporary directory (Or the other way around).

stevespringett:
I also believe we need to remove the check to see if the analyzer is enabled or not prior to downloading. This behavior is inconsistent with other analyzers (like the NVD) which doesn't have this check. I had to 'enable' the retireJs analyzer in order to perform an updateOnly. Without the analyzer enabled, I received the same "Failed to initialize the RetireJS repo" exception.

jortkoopmans:
Also been hitting this problem, breaking my builds. I've rolled back to 3.2.1.

EDIT: Clarifying my setup: running Jenkins master in a docker container, on a CentOS7 host. Various build nodes around it, I've seen this problem occur on linux (CentOS7) based builds using the DependencyCheck.

stevespringett:
https://issues.jenkins-ci.org/browse/JENKINS-52792

javixeneize:
Hi. Is a Jenkins plugin version going to be released for this as well?

Thanks

stevespringett:
@javixeneize jenkins plugin 3.3.1 is already released and has this functionality.

javixeneize:
Hi. It is still failing for me on that version

[DependencyCheck] OWASP Dependency-Check Plugin v3.3.1

[DependencyCheck] Analyzing Dependencies
[DependencyCheck] One or more exceptions were thrown while executing Dependency-Check
[DependencyCheck] Exception Caught: org.owasp.dependencycheck.exception.InitializationException
[DependencyCheck] Cause: jsrepository.json (No such file or directory)
[DependencyCheck] Message: Failed to initialize the RetireJS repo
[...]

Issue #1404, opened 2018-07-27
Invalid cross-device link (DependencyCheck v3.3.0)


freecode1973:
I tried to switch to newest dependency-check-maven plugin, which results in following error:

[ERROR] Failed to execute goal org.owasp:dependency-check-maven:3.3.0:update-only (owasp-dependency-update) on project jenkins-maintenance: An exception occurred while downloading updates. Please see the log file for more details. Failed to initialize the RetireJS repo: /tmp/dctempfe32b35b-0b68-41f7-b514-c7c417e4b4fb/jsrepository.json -> /var/local/share/owasp-cache-3.3.0/jsrepository.json: Invalid cross-device link -> [Help 1]

* Directory /tmp is located on a fast, local hard disk (SSD). 
* /var/local/share is in this case an NFS share. I also tried to use local workspace for cve database directory or inside maven repository but same error occures. (With database I mean the directory set by dataDirectory property. Deleting this directory will not help)
* System is a RedHat linux, with Jenkins as build system

The problem does not occur if database and temporary directory are located on same hard disc partition. 

Is it possible to use soft links instead of hard links ? Alternatively...how to overwrite the temporary directory used by dependency-check-maven plugin, without affecting other maven modules (so I could set it to same disk where database is stored).

jeremylong:
Thanks, this is a duplicate of #1394 - although I had originally thought this was a windows only bug.  A fix will be published with the next release.

The only way to change the temp directory for a maven build would be to set the `temp.directory` system variable:
```
mvn -Dtemp.directory=/path/to/tmp verify
```

PR #1405, opened 2018-07-27
Retire JS Initialization Bug

jeremylong:
## Fixes Issue #1394 and #1404 

Instead of using move to copy the retire js database the code has been updated to perform a copy and delete.


jeremylong:

PR #1406, opened 2018-07-29
Added new Nuget packages.config analyzer (Nugetconf)

doshyt:
## Fixes Issue #
Adds support of .NET [packages.config](https://docs.microsoft.com/en-us/nuget/reference/packages-config) dependency format. This format is a popular way to declare dependencies for MSBuild projects without including them into *.csproj files or using *.nuspec files.

## Description of Change
Created a new analyzer "Nugetconf" using the existing Nuspec and MSBuild analyzers as inspirations; marked it as *experimental*. Updated settings, CLI options, various documentation files and added test cases.

Yes

### DIFF
--- a/ant/src/main/java/org/owasp/dependencycheck/taskdefs/Check.java
+++ b/ant/src/main/java/org/owasp/dependencycheck/taskdefs/Check.java
@@ -209,7 +209,7 @@ public class Check extends Update {
      */
     private Boolean nuspecAnalyzerEnabled;
     /**
-     * Whether or not the .NET Nuget config file Analyzer is enabled.
+     * Whether or not the .NET Nuget packages.config file Analyzer is enabled.
      */
     private Boolean nugetconfAnalyzerEnabled;
     /**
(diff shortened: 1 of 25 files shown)


jeremylong:

jeremylong:
@doshyt - this is one of the most complete and highest quality PRs we've seen. From clean, documented code to updating the user documentation. Thank you for adding the configuration to all of the implementations (maven, cli, ...). What really surprised me is that you even noticed that you needed to update the [dependencycheck-base-suppression.xml](https://github.com/jeremylong/DependencyCheck/pull/1406/files#diff-506caf42a5af07f2709b5eeb556a2aacL78) to filter out some obvious FP.

Thank you!

doshyt:
@jeremylong - thanks for the good words!

I also noticed that there is a lot of duplication in XPath parsing implementations between Nuspec, MSBuild and packages.config analyzers. Probably, it can be simplified and merged into the same parser that implements searches for these analyzers. I would love to work on it next.

PR #1411, opened 2018-07-30
Re-arrange the CSV format report to be more awk, etc, friendly.

### LINKED ISSUE #979: dependency-check.sh should have an option for plain-text results, too.
currently, i find the output of the shell-script it hard to quick-understand, as the xml/json are hard to read, and the html with all the JS even harder ( with cmdline tools like ´more´).

An option to have similar output like generated by the gradle-plugin or so would be nice.

aw-was-here:
## Fixes Issue #

fixes #979

## Description of Change

Re-arrange the CSV format report to be more awk, etc, friendly.
Additionally, use line continuation markers to make it more readable

no

### DIFF
--- a/core/src/main/resources/templates/csvReport.vsl
+++ b/core/src/main/resources/templates/csvReport.vsl
@@ -17,11 +17,48 @@ Copyright (c) 2017 Jeremy Long. All Rights Reserved.
 
 @author Jeremy Long <jeremy.long@owasp.org>
 @version 1 *###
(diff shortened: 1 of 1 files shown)


jeremylong:

jeremylong:
As this would be a breaking change for some users - I am going to hold off merging this until the 4.0.0 release (which is in the works).

aw-was-here:
I'm going to close this pull request for now.  After a significant amount of testing, it's very clear that the high false positive rate means that the confidence level is way more important than I originally thought.  

Also, it would appear that at some point in time the code actually supported providing a custom velocity macro. Restoring that functionality is a much better fix anyway.

Issue #1421, opened 2018-08-05
maven plugin should look at more than just java dependencies


aw-was-here:
It would appear that if a maven project has more than just Java bits, the dependency check plugin doesn't take those extra parts into consideration.

For example, Apache Hadoop has a [package.json](https://github.com/apache/hadoop/blob/bcfc9851f4c6d37d9a119c16ea7d4d253d3b9614/hadoop-yarn-project/hadoop-yarn/hadoop-yarn-ui/src/main/webapp/package.json) that doesn't seem to be getting found despite being in a directory that is part of the (default) scan set.

This is obviously bad for two reasons:

a) To get a complete audit, one is running both the maven plugin and the standalone version on a source tree, doubling the amount of time/resources required
b) Users may be led into a false sense of security because they may believe the maven plugin is actually checking everything.

stevespringett:
The Maven plugin defaults to using standard Maven conventions and scans files in:

- src/main/resources
- src/main/filters
- src/main/webapp

If package.json is located in one of these directories, it will be scanned. If the file is outside the directories defined by Maven conventions, you can use the `scanSet` parameter to specify the directories you want to scan.

Refer to https://jeremylong.github.io/DependencyCheck/dependency-check-maven/configuration.html

I did a quick scan of hadoop-yarn-ui using the following command: 
```bash
mvn org.owasp:dependency-check-maven:3.3.0:check
```
and it did find 4 vulnerable Node components using the built-in NSP Analyzer.

aw-was-here:
Hmm. I've been using :aggregate and it definitely does not report on any of the Node components.  Running :check does.  

Maybe I misunderstood what aggregate does. My goal is to get one report for the entirety of the project. If it isn't actually running through all of the modules, that's less than helpful. :(

jeremylong:
This is a bit if an oversight and I'm glad you reported it.  The intent is that the aggregate goal would scan the standard directories - but I can guarantee that right now it does not. As you have seen, the aggregate goal only scans dependencies defined in the pom.xml.

jeremylong:
@aw-was-here is there any chance the project is available on github? If not - can you tell us the project layout for the maven build and where the package.json exists?

aw-was-here:
Awesome, thanks!

Yes, it's definitely on github:  https://github.com/apache/hadoop

[...]

PR #1422, opened 2018-08-05
Fix aggregate scan set

jeremylong:
## Fixes Issue #1421 

The ScanSet implementation for the Maven plugin did not work for aggregate goals. This PR enables the standard Maven layout to be scanned (or alternatively one configured in the build).

Yes

### DIFF
--- a/maven/src/it/740-aggregate/pom.xml
+++ b/maven/src/it/740-aggregate/pom.xml
@@ -25,7 +25,7 @@ Copyright (c) 2018 Jeremy Long. All Rights Reserved.
     <modules>
         <module>first</module>
         <module>second</module>
-		<module>third</module>
+        <module>third</module>
     </modules>
     <build>
         <plugins>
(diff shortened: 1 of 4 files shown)


jeremylong:

jeremylong:
@stevespringett Could you take a look at this PR - specifically the bit of code that sets up the scan set: https://github.com/jeremylong/DependencyCheck/pull/1422/files#diff-34442e326d7be233d2d63917a6ee0ff6R1010

For the aggregate goal to work - I changed from using the scanSet field to a local variable and if it is an aggregate build with a ScanSet defined we now perform a deep copy and adjust the directory if it is not absolute to point to the child module.  Looks right in testing - but I would still like another set of eyes.

stevespringett:
@jeremylong I don't see anything that pops up as being incorrect. It appears the existing 'check' goal will continue to work as expected and the 'aggregate' goal should do what we think.

The only thing that caught my eye was:

`FileSet[] projectScan = scanSet;`

I was expecting .clone() here, but I went through the code and it doesn't appear there would be any anomalies from doing it the way its currently written.

jeremylong:
@stevespringett thanks - actually that assignment statement you pointed out should not have been there; it was never used and was just residual from an earlier, incorrect implementation.

jeremylong:
@stevespringett also - clone was not used as `FileSet` does not support clone and a clone of an array of objects only create a new array with references to the object. That doesn't work as I need to leave the original objects intact.

PR #1429, opened 2018-08-08
Ignore packages.config files in File Analyzer

doshyt:
## Fixes Issue #

As a side effect of introducing the new Nugetconf Analyzer that scans packages.config for dependencies, the old File Analyzer started picking up all analyzed packages.config files as dependencies and adding them to the dependencies list. This behavior creates a lot of noise and doesn't carry any valuable information.

## Description of Change

Added "packages.config" to the ignore list of the File Analyzer.

no


jeremylong:

jeremylong:
Thank you for the PR!

Drewster727:
Hi -- was this released? I still see packages.config files in my results in v4.0.2

Issue #1439, opened 2018-08-11
Error org.owasp.dependencycheck.xml.pom.PomParseException with ojdbc7.jar analysis


DmitriyStoyanov:
After upgrading OWASP Dependency Check CLI 3.2.0 -> 3.3.1 Faced with error:
org.owasp.dependencycheck.xml.pom.PomParseException: Unable to parse pom ...

after adding logs found, that issue with ojdbc7.jar file.
Just for test, I have created directory with only this jar file and started /d/tmp/dependency-check/bin/dependency-check.bat --project "OJDBC" --scan "." --format ALL --log odc.log
```
...
[INFO] Analysis Started
[INFO] Finished Archive Analyzer (1 seconds)
[INFO] Finished File Name Analyzer (0 seconds)
[INFO] Finished Jar Analyzer (0 seconds)
[WARN] Unable to parse pom 'C:\Users\User\AppData\Local\Temp\dctemp53e4902f-83d6-401c-883c-aaa181a26d9a\pom2726894874192483239.xml'
[WARN] An error occurred while analyzing 'D:\tmp\git\tmp\owasp-ojdbc-issue\ojdbc7.jar' (Central Analyzer).
[INFO] Finished Central Analyzer (388 seconds)
...
[INFO] Analysis Complete (392 seconds)
[ERROR] org.owasp.dependencycheck.xml.pom.PomParseException: Unable to parse pom 'C:\Users\User\AppData\Local\Temp\dctemp53e4902f-83d6-401c-883c-aaa181a26d9a\pom2726894874192483239.xml'
```

main issue that it cannot find needed lib in Central Repository:
```
DEBUG - Starting Central Analyzer
2018-08-11 15:42:33,522 org.owasp.dependencycheck.Engine:807
DEBUG - Parallel processing with up to 4 threads: Central Analyzer.
2018-08-11 15:42:33,522 org.owasp.dependencycheck.AnalysisTask:86
DEBUG - Begin Analysis of 'D:\tmp\git\tmp\owasp-ojdbc-issue\ojdbc7.jar' (Central Analyzer)
2018-08-11 15:42:33,523 org.owasp.dependencycheck.data.central.CentralSearch:126
DEBUG - Searching Central url https://search.maven.org/solrsearch/select?q=1:7c9b5984b2c1e32e7c8cf3331df77f31e89e24c2&wt=xml
[...]

PR #1443, opened 2018-08-14
Fix of issue #1439 

DmitriyStoyanov:
Error org.owasp.dependencycheck.xml.pom.PomParseException during analysis of not existed pom in Maven Central

## Fixes Issue #
#1439 
## Description of Change

*Added check that after all attempts to download we received needed file.
And added AnalysisException to catch cases when pom file cannot be parsed*

*no*


jeremylong:

DmitriyStoyanov:
I've checked it by myself on test project with only one ojdbc7.jar file and received just warning:
```
[INFO] Finished Jar Analyzer (0 seconds)
[WARN] Unable to download pom.xml for ojdbc7.jar from Central; this could result in undetected CPE/CVEs.
[INFO] Finished Central Analyzer (389 seconds) 
```
without exception as faced before

PR #1459, opened 2018-08-24
Fix broken table formatting in Command Line Arguments doc

doshyt:
## Description of Change

Added missing table separators

Checked in my local Markdown parser, looks better now!


jeremylong:

Issue #1479, opened 2018-09-07
False positive on aws-java-sdk-swf-libraries-1.11.22.jar CVE-2018-13525


stephengroat:
False positive on library aws-java-sdk-swf-libraries-1.11.22.jar - reported as cpe:/a:flow_project:flow:1.11.22
```xml
<dependency>
   <groupId>com.amazonaws</groupId>
   <artifactId>aws-java-sdk-simpleworkflow</artifactId>
   <version>1.11.399</version>
</dependency>
```

PR #1490, opened 2018-09-18
additional suppression rule for #1479

stephengroat:
## Fixes Issue #

#1479 

## Description of Change

additional suppression rule for flow project on simple workflow aws libs

no


jeremylong:

PR #1494, opened 2018-09-21
Fixing the test case failure behind proxy.

tyagiakhilesh:
## Description of Change

Though probably I could have changed that artifactory host URI protocol to `http` instead of `https` but I was not sure if you had any reason to choose that. Hence I just made sure, that particular test case runs in non proxy environment. Hence the change.

Not Applicable.


jeremylong:

PR #1495, opened 2018-09-21
Multiple proxy settings in maven settings breaks the plugin function.

tyagiakhilesh:
## Description of Change

In case we find multiple settings for proxy in maven settings, pick the first active one. Also, it might be very much possible that proxy definition is specified in maven settings file but they are not active. Hence throwing IllegalStateException is not really correct?! Hence letting null return in case none of the proxy settings are active.

I would love to. But not able to figure out where are the relevant test cases for this at the moment. Hence I going ahead and raising this PR. Let us see if any test case fails for this.

### DIFF
--- a/maven/src/main/java/org/owasp/dependencycheck/maven/BaseDependencyCheckMojo.java
+++ b/maven/src/main/java/org/owasp/dependencycheck/maven/BaseDependencyCheckMojo.java
@@ -1572,27 +1572,10 @@ public abstract class BaseDependencyCheckMojo extends AbstractMojo implements Ma
                         }
                     }
                 } else {
-                    Proxy httpsProxy = null;
-                    Proxy httpProxy = null;
-
                     for (Proxy aProxy: proxies) {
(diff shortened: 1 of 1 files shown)


tyagiakhilesh:
Tested the plugin with this in my settings.xml and things are working good. Can we please take this up for review? 

```
<proxies>
    <proxy>
      <id>http</id>
      <active>false</active>
      <protocol>http</protocol>
      <host>xxxxx.xx.xxxxxxx.net</host>
      <port>XXXX</port>
      <nonProxyHosts>local.net|some.host.com</nonProxyHosts>
    </proxy>
    <proxy>
      <id>https</id>
      <active>true</active>
      <protocol>https</protocol>
      <host>xxxxx.xx.xxxxxxx.net</host>
      <port>XXXX</port>
      <nonProxyHosts>local.net|some.host.com</nonProxyHosts>
    </proxy>
  </proxies>
```

jeremylong:

tyagiakhilesh:
Tested again with my latest changes. Working good.

Issue #1366, opened 2018-07-04
Migrate NSP Analyzer to Node Audit Analyzer


stevespringett:
According to https://blog.npmjs.org/post/175511531085/the-node-security-platform-service-is-shutting, NSP will be shutting down September 30, 2018.

What we know:
- Node Security Platform will be shutting down on September 30
- NPM AUDIT, the replacement for NSP CHECK, is available in NPM v6.0 and higher
- The current stable Node.js distribution still ships with NPM v5.6
- NPM AUDIT (as of v6.1 - current release) still relies heavily on Node Security Platform

After investigating the NPM AUDIT API, it is safe to assume that:
- Dependency-Check can safely migrate from using the NSP API to the NPM AUDIT API.
- The NPM AUDIT API provides nearly identical information about the advisories discovered from the package submitted.
- Vulnerability identification should continue to work as before

For organizations that rely on stable Node.js distributions, using Dependency-Check for vulnerability identification will be the only alternative.

Related: https://github.com/DependencyTrack/dependency-track/issues/173

stevespringett:
This will likely need to be a phased migration as the current Node Audit API still has dependencies on Node Security Platform resources.

1) Migrate the `nsp check` API to use the`npm audit` API. This will result in NSP advisories with links back to nodesecurity.io.
2) If/when nodesecurity.io is shut down, migrate the links to the advisories to their new home (they currently do not exist outside of nodesecurity.io)
3) Change the SOURCE from being NSP to NPMJS

stevespringett:
Dynamically creating package.json from fragments will no longer work.

The Audit API requires a valid package-lock.json to be submitted. The POST body can either be plaintext json or a gziped version of the lockfile.

https://docs.npmjs.com/cli/audit
https://docs.npmjs.com/files/package-locks
https://docs.npmjs.com/files/package-lock.json

The lockfile is partially scrubbed and the following fragment is added to it:
```json
"metadata": {
    "npm_version": "6.1.0",
    "node_version": "v10.5.0",
    "platform": "linux"
  }
```
In addition, the `requires` field in package-lock.json (a boolean) is replaced by an object containing, what appears to be, the dependencies defined in package.json. Overall, there seems to be a bit of over-engineering and dramatically increased complexity compared to the existing NSP check API.

Documentation for the Audit API does not exist yet. https://github.com/npm/registry/issues/355

The response from the Audit API contains 
- **actions** (we can ignore these)
[...]

PR #1521, opened 2018-10-08
npm audit update

jeremylong:
## Fixes Issue #1366

Updated ODC to use Node Audit instead of NSP.

yes


stevespringett (approved):
Overall, I think this is ready to go. I ran this against NodeGoat and came up with expected results. However, we may want to clarify somewhere what the expectations are. Running Dependency-Check against nodeGoat will result in 21 vulnerabilities. Running NPM Audit against NodeGoat will result in 163 vulnerabilities. The reason is that npm audit will count every single instance of the vulnerable component whereas ODC will count it only once. So setting the expectations in documentation somewhere may be necessary.

PR #1552, opened 2018-11-06
add option to verify <dependencyManagement/> section in POM

guidoschreuder:
## Description of Change

This PR adds the option to also scan the dependencyManagement section of POM's for vulnerablilities

Yes

## Note
consider this a rough draft or proof-of-concept, i'm not very familiar with the codebase, just putting this out there for review and commenting


guidoschreuder:
i've done some manual testing with both 'check' and 'aggregate' on project and all seems to work

for good measure it should probably need an actual test-case

jeremylong (approved):

jeremylong:
Truly appreciate the PR .- thanks!

PR #1560, opened 2018-11-14
Add Golang Dep Analyzer Support

yznima:


core/src/main/java/org/owasp/dependencycheck/analyzer/GolangDepAnalyzer.java
```diff
@@ -0,0 +1,142 @@
...
+      dep.setName(name);
+
+      System.out.println(name);
+      if (name != null && !name.isEmpty()) {
+        dep.addEvidence(EvidenceType.PRODUCT, GOPKG_LOCK, "name", name, Confidence.HIGHEST);
+      }
+
+      final String version = lock.getString("version");
+      System.out.println(version);
+      if (version != null && version.isEmpty()) {
```
trptcolin, line 121: You want `!version.isEmpty()` here, right?

core/src/main/java/org/owasp/dependencycheck/analyzer/GolangDepAnalyzer.java
```diff
@@ -0,0 +1,142 @@
...
+        dep.addEvidence(EvidenceType.PRODUCT, GOPKG_LOCK, "version", version, Confidence.HIGHEST);
+      }
+
+      final String revision = lock.getString("revision");
+      System.out.println(revision);
+      if (revision != null && revision.isEmpty()) {
```
trptcolin, line 127: `!revision.isEmpty()`?

trptcolin (commented):
Really glad someone's taking this work on! I'm new to Golang, but used to being able to do these scans on the JVM and would love to be able to get some more insight in Golang as well.

It seems like a good next step might be a test case (whether manual or automated) where there's a vulnerable dependency that we know for sure is in the h2 database. I don't know enough Golang history to know what'd be likely to appear, other than searching for `golang` in the h2 database's `cpeentry` table. 

In case this is useful - it took me awhile to sort out working access to h2 from the command line. Credentials are from a resources file in this OSS project, and this is non-private data, so I feel like it's fair game to paste them here in an issue:
```
rlwrap java -cp /Users/colin/.m2/repository/com/h2database/h2/1.4.196/h2-1.4.196.jar org.h2.tools.Shell
Welcome to H2 Shell 1.4.196 (2017-06-10)
Exit with Ctrl+C
[Enter]   jdbc:h2:~/test
URL       jdbc:h2:/Users/colin/Projects/oss/yznima-go-dep-check/cli/target/release/data/dc
[Enter]   org.h2.Driver
Driver
[Enter]
User      dcuser
[Enter]   Hide
Password  DC-Pass1337!
Connected
```

In trying to get a known-vulnerable test case together, I found that one of the few non-core-golang libraries I've found with a CVE, [x/crypto/ssh](https://nvd.nist.gov/vuln/detail/CVE-2017-3204) aka `golang:crypto` in the database, has no version numbers, just SHA-based dependencies. And in thinking a little harder about the Golang dependency management story yesterday, I was realizing that this *could* be a common thing (at least currently - I know [versioned modules](https://github.com/golang/go/wiki/Modules) will land at some point and hopefully change the ecosystem)?

But until then, it seems like this could be a pretty fundamental caveat for Golang scanning: for dependencies that use SHA-based versioning, I don't think there's a way to detect whether that version is affected - unless the CVE database had a complete list of affected SHAs somehow, which seems really unlikely to me. I'd be happy to be surprised though! I don't know enough about either the CVE database or this project to know whether there are workarounds.

hazcod:
Don't forget that since `go modules` are now supported officially, you can parse the `go.sum` and `go.mod` file.

Yesh-AJ-006:
Is the GoLang support live?

yznima:
> Is the GoLang support live?

Not yet, unfortunately.

jeremylong:
The current implementation can be found in the [yznima-go branch](https://github.com/jeremylong/DependencyCheck/tree/yznima-go).

We still need to add additional test cases and update the CPE Analyzer to allow for no version numbers on golang projects.

mvdbos:
For those who might be interested: I have just created a PR for go module dependency support here: #1999. It is different from this one, as it does not make use of Gopkg.lock as used by `dep`, but uses go.mod. This is the native way of working with modules in go, available since go 1.11 and default as of go 1.13.

jeremylong:
This has been merged with updates via a different PR.

PR #1601, opened 2018-12-08
Fix Issue 1600 (stale dependencies found for snapshot archive child modules when performing an aggregate build)

aikebah:
## Description of Change

In addition to the fix for #740 (use virtual dependency when a reactor module is not found during aggregate build) this fix makes DependencyCheck also use a virtual dependency when a found SNAPSHOT module is part of the reactor when an aggregate dependency-check is performed on a multimodule project.

This new default behaviour can be switched-off with a new configuration `setting: dependency-check.virtualSnapshotsFromReactor`

**Also bumped the version to the next snapshot (4.0.1-SNAPSHOT) as current master is still on 4.0.0** 

Yes, new integration testcase that demonstrates the issue was added in commit 4b01ad1176ec9609824d34e4cb9b9fa77cb002c2 (which breaks the build) and then the fix was coded to make the new test pass.

### DIFF
--- a/ant/pom.xml
+++ b/ant/pom.xml
@@ -20,7 +20,7 @@ Copyright (c) 2013 - Jeremy Long. All Rights Reserved.
     <parent>
         <groupId>org.owasp</groupId>
         <artifactId>dependency-check-parent</artifactId>
-        <version>4.0.0</version>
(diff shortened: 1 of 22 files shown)


jeremylong:
As always - I appreciate the high quality PR!  

Thanks for your contributions

PR #1604, opened 2018-12-10
Add more mysql client vs server suppressions

stephengroat:
## Description of Change

Add more mysql client vs server suppressions

no


jeremylong:
Thanks!

PR #1615, opened 2018-12-16
Changes to resolve FP increase

jeremylong:
## Description of Change

Externalized the minimum score for Lucene matches and increased the minimum score to 30 to resolve increase in FP after upgrading to Lucene 7.5.

### DIFF
--- a/core/src/main/java/org/owasp/dependencycheck/analyzer/CPEAnalyzer.java
+++ b/core/src/main/java/org/owasp/dependencycheck/analyzer/CPEAnalyzer.java
@@ -176,7 +176,7 @@ public class CPEAnalyzer extends AbstractAnalyzer {
             skipEcosystems = Arrays.asList(tmp);
         }
 
-        minLuceneScore = engine.getSettings().getFloat(Settings.KEYS.ECOSYSTEM_SKIP_CPEANALYZER, 30);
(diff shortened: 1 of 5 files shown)


stevespringett:
Is there any more work required to fix this (besides tests)?

Anything else necessary to merge and release 4.0.1?

Anything I can do to help?

I'm preparing to release Dependency-Track 3.4 later this week and I'd like to ship with these fixes in place, even if its a snapshot, as long as it works.

core/src/main/java/org/owasp/dependencycheck/analyzer/CPEAnalyzer.java
```diff
@@ -174,6 +176,8 @@ public void prepareAnalyzer(Engine engine) throws InitializationException {
             skipEcosystems = Arrays.asList(tmp);
         }
 
+        minLuceneScore = engine.getSettings().getFloat(Settings.KEYS.ECOSYSTEM_SKIP_CPEANALYZER, 30);
```
TobiX: Copy&Paste-Error?
```suggestion
        minLuceneScore = engine.getSettings().getFloat(Settings.KEYS.LUCENE_MIN_SCORE_FILTER, 30);
```

jeremylong:
@TobiX thank you for spotting the copy paste error.

PR #1616, opened 2018-12-17
exclude bower_components from NPM analyzers

thomasgl-orange:
## Fixes Issue #

The _Node Audit_ analyzer correctly ignores ``package-lock.json`` files found under a ``node_modules`` directory (same for ``npm-shrinkwrap.json`` files). This is a common behavior for analyzers based on ``AbstractNpmAnalyzer``, which is documented and implemented here:
https://github.com/jeremylong/DependencyCheck/blob/v4.0.0/core/src/main/java/org/owasp/dependencycheck/analyzer/AbstractNpmAnalyzer.java#L97

I think the same exclusion logic should be applied to modules found under a ``bower_components`` directory, for the same reasons (ie., avoiding scan of child modules).

## Description of Change

Auto-exclude files with path matching ``.../bower_components/...``, just like ``.../node_modules/...`` is already excluded.

*no*

Manual testing was done on a project which was still using Bower though.  And the code is simple enough...

### DIFF
--- a/core/src/main/java/org/owasp/dependencycheck/analyzer/AbstractNpmAnalyzer.java
+++ b/core/src/main/java/org/owasp/dependencycheck/analyzer/AbstractNpmAnalyzer.java
@@ -83,19 +83,22 @@ public abstract class AbstractNpmAnalyzer extends AbstractFileTypeAnalyzer {
     }
 
     /**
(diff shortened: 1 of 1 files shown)


jeremylong:
Can you point to a project or two containing bower components?

Issue #1622, opened 2018-12-18
False positive with jmh-core


coheigea:
There is a false positive with jmh-core:

jmh-core-1.21.jar (cpe:/a:sun:openjdk:1.21, org.openjdk.jmh:jmh-core:1.21, cpe:/a:oracle:openjdk:1.21) : CVE-2009-2689, CVE-2009-3879, CVE-2009-0733, CVE-2009-2475, CVE-2009-3883, CVE-2009-2476, CVE-2009-3884, CVE-2013-0169, CVE-2009-3880, CVE-2009-3881, CVE-2009-3882, CVE-2009-0581, CVE-2009-2690, CVE-2009-0723, CVE-2009-3728

PR #1645, opened 2018-12-26
suppression rule for #1622

stepio:
## Description of Change

Ignore false positive for `org.openjdk.jmh:jmh-core:*`.

*no*


jeremylong:
Thanks for the PR.  I did not see this and created the identical rule today.

stepio:
No problem, it happens - I just shared what I done for myself :)

Issue #1229, opened 2018-04-23
False positive on slf4j-api:1.8.0-beta2 


reddyalready:
False positive on library slf4j-api:1.8.0-beta2.jar - reported as cpe:/a:slf4j:slf4j:1.8.0.beta
```xml
<dependency>
   <groupId>org.slf4j</groupId>
   <artifactId>slf4j-api</artifactId>
   <version>1.8.0-beta2</version>
</dependency>
```

The issue (`CVE-2018-8088`) has been resolved in `1.8.0-beta2`, but dependency-check still flags the jar as vulnerable.

jeremylong:
Thanks - this is a somewhat known issue in that the update versions (alpha, beta, milestone, etc.) are not properly analyzed and generally marked as exploitable.

reddyalready:
Is it an issue with the cpe/regex matcher? Anything I can do to help?

jeremylong:
This is deep in the CPE identification (specifically the CPEAnalyzer and Dependency objects). However, changes to the version numbers to better support alpha, beta, snapshot, m1, etc. could impact other parts of dependency-check-core. I will try and add this to an upcoming change to switch from the NVD XML data feeds to the NVD JSON data feed.

dave-89:
According to the CVE entry and the Jira ticket, it seems that the vulnerable class is present in slf4j-ext, not in slf4j-api

eyecats:
Noticing what dave-89 noticed above - EventData is not present in slf4j-api, nor does slf4j-api pull in slf4j-ext which contains EventData. I verified slf4j-ext is not being brought into our project, however, OWASP dependency-check-maven vulnerability scan is flagging slf4j-api anyways. Anyone have any ideas? (note: I'm on slf4j-api-1.7.25.jar).

log4j-over-slf4j-1.7.25.jar is also being flagged, and, doesn't pull in slf4j-ext as well.

Aiko-Klostermann:
@reddyalready  "The issue (CVE-2018-8088) has been resolved in 1.8.0-beta2" 
We are currently having the same issue and want to upgrade to a newer version. We were searching for a source but couldn't find any changelog anywhere.

pmehtaupgrade:
[...]

PR #1672, opened 2019-01-08
suppression rule for #1229

stepio:
## Description of Change

Ignore false positive for `org.slf4j:slf4j-api:*`.

*no*

### DIFF
--- a/core/src/main/resources/dependencycheck-base-suppression.xml
+++ b/core/src/main/resources/dependencycheck-base-suppression.xml
@@ -310,6 +310,13 @@
     </suppress>
     <suppress base="true">
         <notes><![CDATA[
+        FP per #1229
+        ]]></notes>
+        <gav regex="true">^org\.slf4j:((?!slf4j-ext).)*:.*$</gav>
+        <cpe>cpe:/a:slf4j:slf4j</cpe>
+    </suppress>
+    <suppress base="true">
+        <notes><![CDATA[
         FP per #1027
         ]]></notes>
         <gav regex="true">^com\.github\.danielwegener:logback-kafka-appender:.*$</gav>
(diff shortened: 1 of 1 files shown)


stepio:
Ping @jeremylong.

Hello, any news?

core/src/main/resources/dependencycheck-base-suppression.xml
```diff
@@ -308,6 +308,13 @@
         <gav regex="true">^com.amazonaws:aws-java-sdk-swf-libraries:.*$</gav>
         <cpe>cpe:/a:flow_project:flow</cpe>
     </suppress>
+    <suppress base="true">
+        <notes><![CDATA[
+        FP per #1229
+        ]]></notes>
+        <gav regex="true">^org\.slf4j:((?!slf4j-ext).)*:.*$</gav>
+        <cpe>cpe:/a:slf4j:slf4j</cpe>
```
jeremylong: This really should be a very specific suppression otherwise we risk masking CVEs for slf4j.
```suggestion
        <cve>CVE-2018-8088</cve>
```

jeremylong (changes requested):
The suppression should be specific to the CVE in this case.

core/src/main/resources/dependencycheck-base-suppression.xml
stepio: I started with suppressing just `slf4j-api` as a constant value. But then I checked other comments and the original CVE and understood that although issue was raised only for `slf4j-ext`, most of the slf4j libs are detected.

So I decided that adding a single rule could be an improvement. But... you're the boss - `slf4j-api` is more than enough for my case.

I'll push the update today, thanks.

stepio: Sorry, ignore my previous note, you're completely right.

PR #1675, opened 2019-01-10
 Remove dependency to javax.xml.bind.DatatypeConverter that makes artifactory analyser crashing on Java 11

nhenneaux:
Remove dependency to javax.xml.bind.DatatypeConverter that makes artifactory analyser crashing on Java 11

### DIFF
--- a/core/src/main/java/org/owasp/dependencycheck/data/artifactory/ArtifactorySearch.java
+++ b/core/src/main/java/org/owasp/dependencycheck/data/artifactory/ArtifactorySearch.java
@@ -32,7 +32,6 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import javax.annotation.concurrent.ThreadSafe;
-import javax.xml.bind.DatatypeConverter;
 import java.io.FileNotFoundException;
 import java.io.IOException;
 import java.io.InputStreamReader;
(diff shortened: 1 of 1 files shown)


core/src/main/java/org/owasp/dependencycheck/data/artifactory/ArtifactorySearch.java
```diff
@@ -41,6 +40,7 @@
 import java.net.URL;
 import java.nio.charset.StandardCharsets;
 import java.util.ArrayList;
+import java.util.Base64;
```
Stephan202, line 43: The project [targets Java 8](https://github.com/jeremylong/DependencyCheck/blob/master/pom.xml#L426) so this should work :+1:.

Stephan202 (approved):

core/src/main/java/org/owasp/dependencycheck/data/artifactory/ArtifactorySearch.java
nhenneaux: Indeed but with this change it can also work with 11

Stephan202: I know ;). Just pointing out that this PR doesn't fix Java 11 compatibility at the expense of older still-supported releases. :)

Context for why this change is necessary is [JEP 320](https://openjdk.java.net/jeps/320).

jeremylong:
Thanks for the PR!

PR #1692, opened 2019-01-17
Correct doc on failOnError parameter

jrh3k5:
## Description of Change

Correct documentation on failOnError parameter - looks to be a copy/paste error from the dataFileName parameter

no


jeremylong:
Thanks for the PR!

PR #1699, opened 2019-01-24
Add exclusion filter to cuteoff some dependencies and their transitive

nfalco79:
This PR allow to define a typical maven artifacts exclusion filter (by GroupId:ArtifactId:Version) to exclude those dependencies and their transitive.

In my company we have a structured hierarchy of projects. Let's me explain better
- we have a framework
- technical modules developed on top our framework
- business modules developed on top of previous ones

Each modules is in a separate repository. With actual implementation of this plugin we commit a suppression file for the framework but each technical and business module have to replicate those suppression in their repository plus additional due their additionaldependencies.

The idea behind this PR is that, since each maven project ensure their dependencies are already checked we would to exclude internal dependencies and their transitive dependencies simply defining an exclusion like com.acme*:*
This allow each module technical/business to define only suppression for their additional dependencies.


nfalco79:
If for you it's a good idea I will spent time to provide test cases

Issue #1701, opened 2019-01-26
please add alternative constructor to Settings class


SingingBush:
please add another constructor to Settings class:

```
    public Settings(final Properties properties) {
        props = properties;
        logProperties("Properties loaded", props);
    }
```

This will make it a lot more convenient to use DependencyCheck from Java code

PR #1704, opened 2019-01-26
#1701: additional constructor on Settings class

SingingBush:
## Description of Change

simply add an additional constructor to the Settings class

*no* not really needed. it would be testing that Java works


jeremylong:
Thanks for the PR!

Issue #1702, opened 2019-01-26
FileUtils::getResourceAsStream only works if loading file in classpath


SingingBush:
it would be good if _getResourceAsStream_ made an additional attempt to create an input stream. I'll make a PR.

PR #1705, opened 2019-01-26
#1702: create FileInputStream if cannot get InputStream using classlo…

SingingBush:
…ader

## Description of Change

I found an issue using your project with a properties file that was not on the classpath. This change should resolve that.

*no*


jeremylong:
Thanks for the PR!

Issue #1703, opened 2019-01-26
improvement: make use of @Nullable and @Nonnull


SingingBush:
_javax.annotation_ is already available in the project but being used much. Using both `javax.annotation.Nonnull` and `javax.annotation.Nullable` would be helpful.

It's especially handy when your code is being called from Kotlin as the language will force the user to make a safe call with `?.`

kwin:
Please don't use those annotations. For a reasoning look at https://blog.codefx.org/java/jsr-305-java-9/. Other null annotations are  better suited (e.g. Eclipse or IntelliJ ones).

SingingBush:
ah cool, I was originally going to use the Jetbrains ones, went with javax.annotation because it was already in pom

PR #1706, opened 2019-01-26
#1703 use javax.annotation @Nullable and @Nonnull

SingingBush:
## Description of Change

as per title. I noticed that _javax.annotation_ is in the project and be used a little bit in places but there's a lot of places in which `@Nullable` and `@Nonnull` could be specified on methods. This PR updates some of the classes that I've been using today with those annotations.

*no*


SingingBush:
I'm going to close this and do a new PR for using Jetbrains annotations. see comments on #1703

PR #1731, opened 2019-02-17
Added CVSS v3 score to Ant and CLI build failure check

albuch:
5.0.0-M1 adds CVSS v3 score checking to the Maven Plugin but not for Ant and CLI.

This PR adds the CVSS 3.0 score checking to Ant and CLI.


jeremylong:
Thanks!

PR #1764, opened 2019-02-23
Issue-1751: Make Maven plugin aggregate goal use <repositories> secti…

srdo:
…ons of child modules when resolving dependencies

## Description of Change
Make the aggregate goal use the `<repositories>` of the child module when resolving dependencies for a child, rather than only using `<repositories>` from the parent POM.

Also noticed a test failure in the integration test for 1551, it was due to modification of an immutable collection. The fix is a one liner, and is included here (the wrapping of `collectorVisitor.getNodes()` in a new list).

Yes


jeremylong:
Thanks for the PR!

PR #1765, opened 2019-02-24
use Jetbrains annotations instead of findbugs #1703

SingingBush:
## Description of Change

When calling dependencycheck code from Kotlin it's handy to have the API make use of `@NotNull` and `@Nullable` annotations. I had previously made a PR to use javax.annotation as it was already pulled in by the findbugs dependency. This PR instead removes findbugs and uses:

```xml
    <dependency>
        <groupId>org.jetbrains</groupId>
        <artifactId>annotations</artifactId>
        <version>17.0.0</version>
    </dependency>
``` 

as per comment on #1703 using the Jetbrains equivalent is a better option.

*no*

### DIFF
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -187,7 +187,7 @@ Copyright (c) 2012 Jeremy Long. All Rights Reserved.
             <artifactId>joda-time</artifactId>
         </dependency>
         <dependency>
-            <groupId>com.google.code.findbugs</groupId>
+            <groupId>org.jetbrains</groupId>
             <artifactId>annotations</artifactId>
             <optional>true</optional>
         </dependency>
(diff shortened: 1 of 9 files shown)


jeremylong:
Thanks for the PR!

jdillon:
btw this seems a bit strange to me, why would you prefer a less standard set of annotations over those from javax.annotation that JSR 305 via findbugs provides?

SingingBush:
@jdillon there are arguments for and against using _javax.annotation_. As stated in description there was originally a PR for using the findbugs version but that was deleted in favour of this one. See the linked article in #1703 that gives some reasons not to use them, such as:

> FTR there were never any annotations published by JSR 305
> 
> As a result, if you redistribute an Oracle JRE with the “annotations claimed to be JSR 305 but never actually officially published from the JSR spec page” then you are violating the Oracle JVM redistribution terms.
> 
> Please stop referring to these as JSR 305 annotations… at best they are “presumed JSR 305 annotations”

I looked at https://jcp.org/en/jsr/detail?id=305 and it indeed confirms that _javax.annotations_ isn't any more "standard" than the package name Jetbrains use. So until a standard is published, projects should use anything they want.

jdillon:
@SingingBush thx for the reference.

Issue #1779, opened 2019-03-04
Invalid XML generated by 5.0.0-M1


whitfin:
**Describe the bug**

Invalid XML reports are generated when using the Maven Plugin v5.0.0-M1. Not familiar with the XML spec, but it's causing our sec tooling to fail validation of the XML and thus ignore the reports.

It's caused by this pattern being generated in the output report: 

```xml
<software "matched"="true" ...>
```

If you remove the quotes around `"matched"`, it becomes valid.

** Version of dependency-check used **
The problem occurs using version 5.0.0-M1 of the the Maven plugin.

**Expected behavior**

Valid XML should be produced, i.e. without the quotes.

PR #1780, opened 2019-03-04
Fix an XML validation issue in results

whitfin:
Fixes issue #1779 by removing quotes from invalid XML. I filed the issue but noticed that the fix is actually quite trivial.


jeremylong:
Thanks!

PR #1871, opened 2019-04-24
Fix project name required

tobmaster:
## Description of Change

Changed the project name parameter to required, as it is required by the cli tool

no


jeremylong:
The project name is not required in the 5.0.0-m1 or 5.0.0-m2 releases.  The documentation that is published is for the m2 release.  You can see here that the `project` argument is no longer required:

https://github.com/jeremylong/DependencyCheck/blob/0a7cc78c4b2a701bbe08a2b734112ffcdeda0dcb/cli/src/main/java/org/owasp/dependencycheck/CliParser.java#L1021-L1027

We hope to be code complete with the 5.0.0 branch soon. Sorry for the confusion.

Issue #1685, opened 2019-01-14
Failed to read results from the NPM Audit API


githubhs17:
Today from time to time I got following error:

15:26:59 [INFO] Finished Node.js Package Analyzer (0 seconds)
15:27:00 [INFO] Finished Dependency Merging Analyzer (0 seconds)
15:27:00 [INFO] Finished Version Filter Analyzer (0 seconds)
15:27:00 [INFO] Finished Hint Analyzer (0 seconds)
15:27:02 [INFO] Created CPE Index (1 seconds)
15:27:02 [INFO] Skipping CPE Analysis for npm
15:27:02 [INFO] Finished CPE Analyzer (2 seconds)
15:27:02 [INFO] Finished False Positive Analyzer (0 seconds)
15:27:02 [INFO] Finished NVD CVE Analyzer (0 seconds)
15:28:33 [WARN] An error occurred while analyzing '/src/package-lock.json' (Node Audit Analyzer).
15:28:34 [WARN] An error occurred while analyzing '/src/package-lock.json' (Node Audit Analyzer).
15:28:34 [INFO] Finished Node Audit Analyzer (92 seconds)
15:28:34 [INFO] Finished Vulnerability Suppression Analyzer (0 seconds)
15:28:43 [INFO] Finished Dependency Bundling Analyzer (8 seconds)
15:28:43 [INFO] Analysis Complete (103 seconds)
15:28:44 [ERROR] Failed to read results from the NPM Audit API (NodeAuditAnalyzer); the analyzer is being disabled and may result in false negatives.
15:28:44 [ERROR] Failed to read results from the NPM Audit API (NodeAuditAnalyzer); the analyzer is being disabled and may result in false negatives.

Can't find any informations about this error.
Does this mean there is an online service unavailable ?
If yes, what is the URL of this service ? (To check, before starting the scan)

Thanks

Additional information:
We are not using the OWASP Dependency-Check Plugin, we are working with the owasp/dependency-check docker image from https://hub.docker.com/r/owasp/dependency-check/

markdenihan:
@githubhs17 I had a similar problem to this. See #1679 

You can get more information from the Jenkins System Logs. 

The URL by default is http://registry.npmjs.org/-/npm/v1/security/audits

Jarl-dev:
I am getting the same errors since a few days.
But it does not occur every time. Are there any known issues with availability of the API in general?

markdenihan:
[...]

Issue #1679, opened 2019-01-11
Could not connect to Node Audit API: 503 No backends available


markdenihan:
Running dependency check v4.0.2 on a package-lock.json via Jenkins fails with the attached console logs / system logs

Jenkins slave/master are able to communicate with the registry.npmjs.org/-/npm/v1/security/audits API and get 200 responses.

Looks like the dependency check plugin just gets 503 errors. 

```
[DependencyCheck] OWASP Dependency-Check Plugin v4.0.2
[DependencyCheck] Executing Dependency-Check with the following options:
[DependencyCheck]  -name = dc_job
[DependencyCheck]  -scanPath = /home/ausernamehere/thingthatbuilds/workspace/dc_job/target
[DependencyCheck]  -outputDirectory = /home/ausernamehere/thingthatbuilds/workspace/dc_job
[DependencyCheck]  -dataDirectory = /home/ausernamehere/thingthatbuilds/dependencycheck-db
[DependencyCheck]  -suppressionFile = /home/ausernamehere/thingthatbuilds/workspace/dc_job/tests/security/dependency-check/suppressions.xml
[DependencyCheck]  -dataMirroringType = NIST CPE/CVE
[DependencyCheck]  -cveUrl12Modified = http://internal.mirror.int/downloads/nist/nvdcve-modified.xml.gz
[DependencyCheck]  -cveUrl20Modified = http://internal.mirror.int/downloads/nist/nvdcve-2.0-modified.xml.gz
[DependencyCheck]  -cveUrl12Base = http://internal.mirror.int/downloads/nist/nvdcve-%d.xml.gz
[DependencyCheck]  -cveUrl20Base = http://internal.mirror.int/downloads/nist/nvdcve-2.0-%d.xml.gz
[DependencyCheck]  -isQuickQueryTimestampEnabled = true
[DependencyCheck]  -jarAnalyzerEnabled = true
[DependencyCheck]  -nodePackageAnalyzerEnabled = true
[DependencyCheck]  -nodeAuditAnalyzerEnabled = true
[DependencyCheck]  -retireJsAnalyzerEnabled = true
[DependencyCheck]  -composerLockAnalyzerEnabled = true
[DependencyCheck]  -pythonDistributionAnalyzerEnabled = true
[DependencyCheck]  -pythonPackageAnalyzerEnabled = true
[DependencyCheck]  -rubyBundlerAuditAnalyzerEnabled = true
[...]

PR #1891, opened 2019-05-04
workaround for capacity problems of npmjs.org

RubyPaasche:
## Fixes Issue #
this is a workaround for npmjs.org audit-API capacity problems #1685 and #1679.

## Description of Change

Added retry counter for npmjs.org Audit-API requests.

### DIFF
--- a/core/src/main/java/org/owasp/dependencycheck/data/nodeaudit/NodeAuditSearch.java
+++ b/core/src/main/java/org/owasp/dependencycheck/data/nodeaudit/NodeAuditSearch.java
@@ -28,6 +28,8 @@ import java.net.URL;
 import java.nio.charset.StandardCharsets;
 import java.security.SecureRandom;
 import java.util.List;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.locks.LockSupport;
 import javax.annotation.concurrent.ThreadSafe;
 import org.json.JSONObject;
 import org.owasp.dependencycheck.utils.Settings;
@@ -99,7 +101,7 @@ public class NodeAuditSearch {
      * @throws IOException if it's unable to connect to Node Audit API
      */
     public List<Advisory> submitPackage(JsonObject packageJson) throws SearchException, IOException {
-       return submitPackage(packageJson,0);
+       return submitPackage(packageJson, 0);
     }
 
     /**
@@ -113,7 +115,7 @@ public class NodeAuditSearch {
      * package
      * @throws IOException if it's unable to connect to Node Audit API
      */
-    private  List<Advisory> submitPackage(JsonObject packageJson, int count) throws SearchException, IOException {
+    private List<Advisory> submitPackage(JsonObject packageJson, int count) throws SearchException, IOException {
         try {
             final byte[] packageDatabytes = packageJson.toString().getBytes(StandardCharsets.UTF_8);
             final URLConnectionFactory factory = new URLConnectionFactory(settings);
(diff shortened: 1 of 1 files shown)


core/src/main/java/org/owasp/dependencycheck/data/nodeaudit/NodeAuditSearch.java
```diff
@@ -127,6 +142,14 @@ public NodeAuditSearch(Settings settings) throws MalformedURLException {
                         final NpmAuditParser parser = new NpmAuditParser();
                         return parser.parse(jsonResponse);
                     }
+                case 503:
+                    LOGGER.debug("Invalid payload submitted to Node Audit API. Received response code: {} {}",
+                            conn.getResponseCode(), conn.getResponseMessage());
+                    if (count > 5){
+                        return submitPackage(packageJson, ++count);
```
jeremylong, line 152: should there be a sleep here?  If we are getting a 503 for capacity issues - shouldn't we be nice and pause a second before retrying?

core/src/main/java/org/owasp/dependencycheck/data/nodeaudit/NodeAuditSearch.java
```diff
@@ -127,6 +142,14 @@ public NodeAuditSearch(Settings settings) throws MalformedURLException {
                         final NpmAuditParser parser = new NpmAuditParser();
                         return parser.parse(jsonResponse);
                     }
+                case 503:
+                    LOGGER.debug("Invalid payload submitted to Node Audit API. Received response code: {} {}",
```
jeremylong: A 503 is not due to an invalid payload?

```suggestion
                    LOGGER.debug("Node Audit API returned `{} {}` - retrying request.",
```

core/src/main/java/org/owasp/dependencycheck/data/nodeaudit/NodeAuditSearch.java
```diff
@@ -127,6 +142,14 @@ public NodeAuditSearch(Settings settings) throws MalformedURLException {
...
+                            conn.getResponseCode(), conn.getResponseMessage());
+                    if (count > 5){
+                        return submitPackage(packageJson, ++count);
+                    }
+                    throw new SearchException("Could not perform Node Audit analysis. Invalid payload submitted to Node Audit API.");
```
jeremylong: Again - I don't think we should be calling this an invalid payload?

```suggestion
                    throw new SearchException("Could not perform Node Audit analysis - service returned a 503.");
```

jeremylong (changes requested):

core/src/main/java/org/owasp/dependencycheck/data/nodeaudit/NodeAuditSearch.java
RubyPaasche: Yeah sorry copy and paste mistake ..

core/src/main/java/org/owasp/dependencycheck/data/nodeaudit/NodeAuditSearch.java
RubyPaasche: Yeah sorry copy and paste mistake ..

core/src/main/java/org/owasp/dependencycheck/data/nodeaudit/NodeAuditSearch.java
RubyPaasche: I added an "LockSupport.parkNanos" it works the same like Thread.sleep but without Exceptions.
It is an dynamic sleep, it 500 ms multiplied with number of the retry.

thib3113:
Hello @rpaasche , After some research, it seems the node audit results are directly available in json on GitHub .

So, I think it will be really better, to download this vulnerabilities in the local database, and search on it, instead of calling npm audit for each package .

Here you can get it : https://github.com/nodejs/security-advisories
In the ecosystem folder, separate by folder for each dependency, and by vulnerabilities, with semver version vulnerable, and bonus, A cvsscore !!!

Just some more information on this repository, it's a work in progress, because as I read they wants to create a GitHub pages website, to search a package ... But not important in our case .

Other things, this repository, is an extract of this repository : https://github.com/nodejs/security-wg , but better separated ( not all the json in same folder, but separated by package for example ),  and As you can see in this issue : https://github.com/nodejs/security-wg/issues/494 they planned to migrate vulnerabilities to a separate repository. 

So, if you plan to use this, this link can be helpful : 
 - github api to get last master commit : `https://api.github.com/repos/nodejs/security-advisories/commits/master`
 - url to download the zip : `https://github.com/nodejs/security-advisories/archive/master.zip`

I'm not a Java developer, so I can't do a good java code for this (just little things), but if you need some testDatas, or informations, I can help

PR #1916, opened 2019-05-14
Remove multiple output format description

rouke-broersma:
As this is not true on current latest v4.0.2

## Description of Change
The docs say multiple output formats are supported by specifying the --format flag multiple times, but this is not the case in v4.0.2 which is the latest stable.

### DIFF
--- a/dependency-check-cli/arguments.html
+++ b/dependency-check-cli/arguments.html
@@ -122,7 +122,7 @@
 <td> -f   </td>
 <td> --format             </td>
 <td> &lt;format&gt;      </td>
-<td> The output format to write to (XML, HTML, CSV, JSON, JUNIT, ALL). Multiple formats can be specified by specifying the parameter multiple times. The default is HTML. </td>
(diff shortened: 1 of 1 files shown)


thib3113:
Not true on the latest release too .

jeremylong:
If you look at the actual published documentation: https://jeremylong.github.io/DependencyCheck/dependency-check-cli/arguments.html - you will see that 5.0.0-M3 does not allow multiple format parameters.  In addition, if you look at the documentation for the [4.0.2 version](https://github.com/jeremylong/DependencyCheck/blob/v4.0.2/cli/src/site/markdown/arguments.md) it does not say you can utilize multiple format parameters.

However, if you built the current master branch you can specify multiple format parameters.  I'm hoping that we can publish 5.0.0 within the week.

rouke-broersma:
For me it does say you can specify multiple format parameters on the latest published docs:

image

I had not noticed that the latest published docs are actually for the 5.0.0m3 preview build so I do see now that it doesn't say so for v4.0.2, thanks.

PR #1953, opened 2019-05-25
update MS SQL syntax

TimothyMorgan:
## Fixes Issue #
In the initialize_mssql.sql script the datatype BOOLEAN throws an error and should be BIT for MS SQL.
There is an extra ')' character in the create software table statement that cuts off the primary key constraint and causes a syntax error.

## Description of Change

Updating the syntax in the initialize_mssql.sql script so that schema can be initialized on MS SQL 2017

no


jeremylong (approved):

jeremylong:
Thanks! We haven't fully checked the scripts for the external databases.

Issue #1954, opened 2019-05-27
Temp directory not deleted after Dependency Track run


justin-tay:
**Describe the bug**
DependencyCheckScanAgent is not cleaning up "dctemp..." temp directories after each run.

** Version of dependency-check used **
dependency-check-core-4.0.2
dependency-track-embedded-3.5.0-SNAPSHOT

** Log file **
```
17:39:47.250 DEBUG [FileUtils] Failed to delete file: C:\Users\User\AppData\Local\Temp\dctemp78662722-e894-45c5-adef-574ead055076; attempting to delete on exit.
17:39:47.250 INFO [DependencyCheckTask] Processing Dependency-Check analysis results
```

**To Reproduce**
1. Upload BOM in Dependency Track

**Expected behavior**
Temp directories created should be cleaned up as it takes up quite a bit of space.

**Additional context**
This is because the following code attempts to delete the temp directory containing the h2 db via settings.cleanup(true) before the engine is closed.
```
settings.cleanup(true);
if (engine != null) {
    engine.close();
}
```

PR #1955, opened 2019-05-27
Fix for temp directory not being deleted after run

justin-tay:
## Fixes Issue #
https://github.com/jeremylong/DependencyCheck/issues/1954

## Description of Change
Clean up the temporary files only after the engine has been closed, otherwise the cleanup will fail to delete the files and will queue it for deletion on JVM exit.

No. Wasn't able to devise a test case as not only is settings private to the DependencyCheckScanAgent, the initial settings.cleanup(true) causes the tempDirectory in settings to be set to null so can't determine the actual tempDirectory used after execute is run. Manually tested by stepping through DependencyCheckScanAgentIT.testComponentMetadata() both before and after the fix.

### DIFF
--- a/core/src/main/java/org/owasp/dependencycheck/agent/DependencyCheckScanAgent.java
+++ b/core/src/main/java/org/owasp/dependencycheck/agent/DependencyCheckScanAgent.java
@@ -978,10 +978,10 @@ public class DependencyCheckScanAgent {
             }
             throw new ScanAgentException("One or more exceptions occurred during analysis; please see the debug log for more details.", ex);
         } finally {
-            settings.cleanup(true);
             if (engine != null) {
                 engine.close();
             }
+            settings.cleanup(true);
         }
         return engine;
     }


justin-tay:
1 test failure but unrelated to the change. Seems like CVE-2016-10750 published on 05/22/2019 specified "cpe23Uri" : "cpe:2.3: a:hazelcast:hazelcast:*:*:*:*:*:*:*:*" for the affected configuration and "cpe:2.3: a:hazelcast:hazelcast:2.5:*:*:*:*:*:*:*" looks like a correct identification in this case.

```
[INFO] Results:
[INFO] 
[ERROR] Failures: 
[ERROR]   CPEAnalyzerIT.testDetermineCPE_full:106->callDetermineCPE_full:156->lambda$callDetermineCPE_full$0:157 Unexpected match found: { dep:'hazelcast-2.5.jar', found:'cpe:2.3:a:hazelcast:hazelcast:2.5:*:*:*:*:*:*:*' }
[INFO] 
[ERROR] Tests run: 56, Failures: 1, Errors: 0, Skipped: 7
```

PR #2012, opened 2019-06-21
add authentication capabilities to OSS Index

jeremylong:
## Description of Change

Adds the ability to configure credentials for the OSS Index Analyzer.


jeremylong:
This PR also:
* fixes problems with the build when Central's APIs have an outage
* resolve issue with the CSV report output

PR #1999, opened 2019-06-14
Implement analyzer for go modules (>=1.11)

mvdbos:
## Fixes Issue #
Lack of support for Go modules. 

## Description of Change
This implementation supports the native `go mod` way of working with dependencies, which is enabled since go 1.11 and will be the default as of go 1.13.

## Have test cases been added to cover the new functionality?
Yes

PR #2018, opened 2019-06-24
Support go modules per #1999

jeremylong:
## Update to PR #1999

implements go modules support in dependency-check.

yes


mvdbos:
@jeremylong I wonder if this module needs to remain Experimental and for how long?

jeremylong:
The point around marking it as experimental is that we will likely see a higher false positive rate upon initial release.  When we get feedback on the usage and we can solve some of the FP issues then we can remove the experimental flag.  It is a way of telling people this is more in the beta release.

PR #2205, opened 2019-09-21
Bump commons-compress from 1.18 to 1.19 to fix CVE-2019-12402

albuch:
@jeremylong Could you check why @dependabot-bot didn't pick up this upgrade? Version 1.19 was released one month ago according to Sonatype repo.


jeremylong:
We've received bumps for commons-compress from @dependabot-bot before (see #1446 and #1318) - so that is a little curious. I just opened https://github.com/dependabot/feedback/issues/682.

PR #2238, opened 2019-10-04
Dependency Classifier Specification Fix

ddellspe:
## Description of Change

Added logic to check first the dependency specification, then the type specification for the details.  the Dependency can be null if not specified, but I wanted to make sure even if it was specified, but was an empty specification (which is allowed) that it would still provide the proper tracing.

*no*


jeremylong:
Thanks for the PR!

ddellspe:
NP Thanks for the great utility!

PR #2257, opened 2019-10-14
CMake variable resolution, some heuristics on product naming

gudaol:
This contribution was made possible during a 'LogMeIn Inc.' event.

## Description of Change
- Resolve $SOME_CMAKE_VARIABLE stuff, recursively
- heuristically apply ASDLIB -> LIBASD, while keeping original too in 'product' field(s)
- some fixes, extensions to the regexes that handle version information

no.
Tested manually with libpng


jeremylong:
@gudaol this change was reverted as it broke the build.  Due to environment setup issues I did not have time to figure out where the endless loop was occurring.  I had merged this as I thought the breaking build was due to something else - if you can fix the breaking build and submit a PR that passes the CI I will merge the change.

gudaol:
Will look into it soon, thanks!

PR #2268, opened 2019-10-17
Suppress FP for withdrawn CVE of issue 2247

aikebah:
## Description of Change
Suppress the withdrawn CVE for any version of antisamy

no, ran a local test to verify that suppression of this CVE was properly done


jeremylong:
While the correct solution to this problem was added with https://github.com/jeremylong/DependencyCheck/commit/f0d3f88a327f481f4e899a216e5d5118d687b66c - I merged this as well as the correct solution would only really solve the problem if users rebuilt their entire database.

PR #2274, opened 2019-10-19
Upgrade to NIST NVD JSON v1.1 URLs and JSON schemas

aikebah:
## Description of Change

Update DependencyCheck to use the new JSON v1.1 feeds of NVD

No, existing testcases cover the functionality. Test-case settings have been updated to use the correct NIST NVD datafeed URLs.


jeremylong:
@aikebah thank you for the PR! This would be a breaking change for many folks as they would need to update their internal mirrors.  As such, we should probably bump the major version - thoughts?

aikebah:
@jeremylong From the changes it's more a minor update, but I agree that it needs some attention with the release that people using a mirror are aware that they should make a change. 
[Nist Datamirror](/stevespringett/nist-data-mirror) is already updated, it's latest release (1.4.0) mirrors both JSON feeds.
People mirroring NIST NVD should in any case be aware that NIST willl cease publishing the 1.0 JSON-feed in the future. I  feared that what they said in the announcement would be in effect, but it's good to see that at least for some time they still publish both. Based on their XML deprecation path I would expect the 1.0 JSON to disappear in a few months time.

Bumping the major would also be fine with me, it would at least call attention to the fact that 'something has changed which deserves your attention and needs your action' for the people that mirror the NIST.
Main reason for making it only a minor would be that for people just using the tool as-is nothing needs to change except for upgrading the tool and a major might make them reluctant to just upgrade while for them that is a feasible scenario.

All in all making it a 5.3 release seems more logical to me than a 6.0 with respect to this change

aikebah:
After some more testing I think that a minor version is sufficient.

Running `mvn clean install -s settings.xml -DreleaseTesting -Danalyzer.node.package.enabled=false -Danalyzer.node.audit.enabled=false` with the original settings
```xml
<settings xmlns="http://maven.apache.org/SETTINGS/1.0.0"
          xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
          xsi:schemaLocation="http://maven.apache.org/SETTINGS/1.0.0
                          https://maven.apache.org/xsd/settings-1.0.0.xsd">
    <profiles>
        <profile>
            <id>default</id>
            <properties>
                <!-- 
                The following are used during testing - when testing locally some developers have a local mirror. 
                However, as travis-ci does not have a local mirror we need to set the original values.
                -->
                <odc.cve.url.base>https://nvd.nist.gov/feeds/json/cve/1.0/nvdcve-1.0-%d.json.gz</odc.cve.url.base>
                <odc.cve.url.modified>https://nvd.nist.gov/feeds/json/cve/1.0/nvdcve-1.0-modified.json.gz</odc.cve.url.modified>
            </properties>
        </profile>
    </profiles>
    <activeProfiles>
        <activeProfile>default</activeProfile>
    </activeProfiles>
</settings>
```

Yields a succesful build

Quite logical, because the main changes in between 1.0 and 1.1 are
 * support for CVSS 3.1 which from JSON structure perspective is 100% compatible with 3.0
 * removal of the affects section from the JSON (and making it optional in the JSON-schema), which is transparent for DependencyCheck as it does not use that section of the JSON feed

glsutter:
Jeremy - Pardon me for a couple of simple questions. 
1) When will this NVD JSON change get into the Docker Hub image?
2) How do we get the change included in our Maven dependency check? Does Maven depend on the Docker Hub image or on the Github repo? And which branch? 
Thanks

jeremylong:
1. When we cut the next release (probably a week or two).  Alternatively, it is open source - we try to keep the master branch stable so you can build it yourself.
2. See above for the release date.  No, dependency-check-maven does not rely on the docker image. You can either build the master branch yourself and use this feature now - or you can wait until the next release.

PR #2276, opened 2019-10-21
CMake variable resolution, some heuristics on product naming

gudaol:
This contribution was made possible during a 'LogMeIn Inc.' event.

## Description of Change
- Resolve $SOME_CMAKE_VARIABLE stuff, recursively
- heuristically apply ASDLIB -> LIBASD, while keeping original too in 'product' field(s)
- some fixes, extensions to the regexes that handle version information
- fixed infinite loop issue in #2257 

no.
Tested manually with libpng


jeremylong:
I apologize about how long this has taken me to get back to. However, it is still not passing the basic test cases.  This is because the [call to setVersion](https://github.com/jeremylong/DependencyCheck/pull/2276/files#diff-beab7592e120664755268331e167a5baR315) is passing in data that is not even remotely a version number.  Maybe we should be be filtering this via [parseVersion(version, true)](https://github.com/jeremylong/DependencyCheck/blob/master/core/src/main/java/org/owasp/dependencycheck/utils/DependencyVersionUtil.java#L93).

gudaol:
Thank you for getting back at this, I will look at it and try that way tomorrow!

jeremylong:
This PR was merged manually: https://github.com/jeremylong/DependencyCheck/commit/4ca7aa22ebbd11af42f01fbede5ee9c1378cffbc

PR #2287, opened 2019-10-27
Re-use mysql db queries for mariadb

aikebah:
## Description of Change
Use same query customization as mysql for mariadb

no


aikebah:
Manually tested locally using a dockerized [mariadb:latest](https://hub.docker.com/_/mariadb)

PR #2290, opened 2019-10-28
Add closing `fi` to `if` statement within *nix example

ddubson:
# Description of Change

Fixes unclosed if statement in Docker instructions for running dependency-check

Not needed


jeremylong:
Thanks for catching that.

Issue #2366, opened 2019-12-04
Debug logs are logging sensitive data when calling Settings.setString and Settings.getInt


jamesratzlaff:
**Describe the bug**
Pretty much the same as #2147. The fix/enhancement for this has been applied [Settings.logProperties](https://github.com/jeremylong/DependencyCheck/blob/c132d4d4601dd016834dcb25d4a287959d5f5cd8/utils/src/main/java/org/owasp/dependencycheck/utils/Settings.java#L761) and mitigates the issue on that method only.   [Settings.setString](https://github.com/jeremylong/DependencyCheck/blob/c132d4d4601dd016834dcb25d4a287959d5f5cd8/utils/src/main/java/org/owasp/dependencycheck/utils/Settings.java#L780) and [Settings.getInt](https://github.com/jeremylong/DependencyCheck/blob/c132d4d4601dd016834dcb25d4a287959d5f5cd8/utils/src/main/java/org/owasp/dependencycheck/utils/Settings.java#L1123) still log sensitive data.

 **Version of dependency-check used** 
dependency-check-utils-5.2.4

**Log file**
```[DEBUG][Settings#setString] - Setting: analyzer.artifactory.api.token='notMyActualToken'```

**To Reproduce**

1. Set your root log configuration to DEBUG

2. Do something similar to the following

```java 
Settings settings = new Settings();
settings.setString("analyzer.artifactory.api.token", "pwndViaDebugLog");
```

3. Look at the logs
4. Notice that the logs now contain something similar to

```
[DEBUG][Settings#setString] - Setting: analyzer.artifactory.api.token='pwndViaDebugLog'
```

**Expected behavior**
The setter and getter methods that log values should take into consider as to whether or not the key that has that value should mask that value

**Additional context**
I'll be submitting a pull request for this.

jeremylong:
Doh...  We completely missed the log statement in the setter - this should just be removed.

PR #2367, opened 2019-12-04
Added sensitive data masking when calling Settings.setString and Settings.getInt

jamesratzlaff:
## Fixes Issue #
#2366 

## Description of Change
Added masking functionality using ```getPrintableValue(String,String)``` for
```LOGGER.debug``` invocations in methods ```setString(String,String)``` and
```getInt(String,int)```

no


jeremylong:
Thanks for the PR!

Issue #2130, opened 2019-08-07
SanitizePackage throws NullPointerException on empty package-lock.json files


RexValkering:
**Describe the bug**

If `npm install` is run on a folder with no package, it generates the following empty `package-lock.json` file:

```json
{
  "lockfileVersion": 1
}
```
Running the CLI on a folder with such an empty file generates a NullPointerException in SanitizePackage:

```text
[WARN] An unexpected error occurred during analysis of 'E:\OwaspTest\package-lock.json' (Node Audit Analyzer): null
[ERROR] 
java.lang.NullPointerException: null
    at org.owasp.dependencycheck.data.nodeaudit.SanitizePackage.sanitize(SanitizePackage.java:70)
    at org.owasp.dependencycheck.analyzer.NodeAuditAnalyzer.analyzeDependency(NodeAuditAnalyzer.java:189)
    at org.owasp.dependencycheck.analyzer.AbstractAnalyzer.analyze(AbstractAnalyzer.java:131)
    at org.owasp.dependencycheck.AnalysisTask.call(AnalysisTask.java:88)
    at org.owasp.dependencycheck.AnalysisTask.call(AnalysisTask.java:37)
    at java.util.concurrent.FutureTask.run(Unknown Source)
    at java.util.concurrent.ThreadPoolExecutor.runWorker(Unknown Source)
    at java.util.concurrent.ThreadPoolExecutor$Worker.run(Unknown Source)
    at java.lang.Thread.run(Unknown Source)
```

This seems to be caused by these lines of code in [SanitizePackage.java](https://github.com/jeremylong/DependencyCheck/blob/0defa99ac4b40fc73715cee58ea92b1c78e8598d/core/src/main/java/org/owasp/dependencycheck/data/nodeaudit/SanitizePackage.java):

```java
final JsonObject dependencies = packageJson.getJsonObject("dependencies");
for (Entry<String, JsonValue> entry : dependencies.entrySet()) {
    ...
}
```
I presume this is because it does not check if `dependencies` is `null` before calling `entrySet()`.

**Version of dependency-check used**

The problem occurs using version 5.2.1 of the the cli. The cli is run on Windows Server R2 Datacenter 64-bit.

**Log file**

https://gist.github.com/RexValkering/8e0ff6c40b583a1e3856a359b80f644e

**To Reproduce**

```powershell
mkdir owasptest
cd owasptest
npm install .
dependency-check.bat -scan .
```

**Expected behavior**

[...]

PR #2430, opened 2020-01-14
Resolve problem with missing dependency in package-lock.json #2130

dzablo:
## Fixes Issue #
https://github.com/jeremylong/DependencyCheck/issues/2130

## Description of Change

I add a check if dependencies key exist in the package.json or package-lock.json.
These changes resolve the problem with throwing `java.lang.NullPointerException` 

yes


dzablo:
I didn't see there was a fix for it.

jeremylong:
Sorry - the team will try to do better about releasing patch versions more frequently.

PR #2446, opened 2020-01-24
Use pecoff4j to read dll/exe version information metadata on windows

pscamodio:
## Fixes Issue #
Dll and exe on windows that are not .NET assembly are only analyzed by the filename.
This is often not good enough because the filename can contain other numbers (x86, x64, ...) other than the version.

## Description of Change
To improve the situation I've reduced the confidence of the filename parsed version and created a new analyzer
The FileVersionAnalyzer use the pecoff4j library to extract, if possible, the version from the file metadata

no

### DIFF
--- a/utils/src/main/java/org/owasp/dependencycheck/utils/Settings.java
+++ b/utils/src/main/java/org/owasp/dependencycheck/utils/Settings.java
@@ -551,6 +551,10 @@ public final class Settings {
          */
         public static final String ANALYZER_FILE_NAME_ENABLED = "analyzer.filename.enabled";
         /**
+         * The key to determine if the File Version analyzer is enabled.
+         */
+        public static final String ANALYZER_FILE_VERSION_ENABLED = "analyzer.fileveraion.enabled";
+        /**
          * The key to determine if the Hint analyzer is enabled.
          */
         public static final String ANALYZER_HINT_ENABLED = "analyzer.hint.enabled";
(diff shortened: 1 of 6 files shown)


utils/src/main/java/org/owasp/dependencycheck/utils/Settings.java
```diff
@@ -550,6 +550,10 @@
          * The key to determine if the File Name analyzer is enabled.
          */
         public static final String ANALYZER_FILE_NAME_ENABLED = "analyzer.filename.enabled";
+        /**
+         * The key to determine if the File Version analyzer is enabled.
+         */
+        public static final String ANALYZER_FILE_VERSION_ENABLED = "analyzer.fileveraion.enabled";
```
HEllRZA: fileveraion typo

pscamodio: Typo fixed

pscamodio:
About the testing, it would require to add a .dll or .exe file with some metadata to the repo. I could reuse a publicy available dll or forge a minimal one, Do you want me to add this to have this merged?

jeremylong:
@pscamodio thank you for this PR!  As you can see I've expanded on it with PR #2448.  For the test case I just used one of the existing DLLs that is used in testing of the AssemblyAnalyzer.  Let's move the conversation and any additional updates to the new PR.

Issue #2485, opened 2020-02-20
Avoid recomputing file hashes unnecessarily


dbevacqua:
**Is your feature request related to a problem? Please describe.**
When a `Dependency` is created for a given file, the file's checksums are computed, regardless of whether this has been done before. In a multi-module Maven project with many common dependencies, this can slow checks considerably. For example, for the project I'm currently working on, ~50k checksums are computed for ~9k files. The scan takes nearly two minutes.

**Describe the solution you'd like**
Cache the checksum for a given hash function and file

**Describe alternatives you've considered**
I experimented with an implementation of `Checksums` which used Guava hashing and this appeared to speed up checks by ~15%. It is also quite concise, e.g.

```
HashCode hashCode = Files.hash(file, Hashing.md5());
```

**Additional context**
Profiling suggests that the vast majority of CPU time is spent computing file checksums, so this is definitely a hotspot

jeremylong:
Thanks for the PR!

PR #2486, opened 2020-02-20
Cache dependency file checksums

dbevacqua:
## Fixes Issue #

https://github.com/jeremylong/DependencyCheck/issues/2485

## Description of Change

Naive solution to avoid recomputing file checksums. Open to suggestions about a better place for the cache to sit. Possibly `Checksums`?

No


jeremylong (changes requested):
The cache should be in the Checksum class as the checksum class is used in other locations that might benefit from the enhancement.

dbevacqua:
Done. If/when you're happy with the change @jeremylong, I can rebase out the first two commits.
EDIT: realised my formatter settings were wrong, so have fixed and squashed it all down.

jeremylong (commented):
Looks good - but this will need to be rebased to include the now merged changes from #2488

Issue #2487, opened 2020-02-20
Use Guava Hashing


dbevacqua:
**Is your feature request related to a problem? Please describe.**

Profiling suggests that, once the vulnerability database has been created, the  majority of CPU time for a scan is spent computing file checksums

**Describe the solution you'd like**

The Guava implementations of MD5, SHA256 and SHA1 are faster than those using `MessageDigest` in `Checksum`. I experimented with an implementation of Checksums which used Guava hashing and this appeared to speed up checks by ~15%. It is also quite concise, e.g.

```
HashCode hashCode = Files.hash(file, Hashing.md5());
```

So I propose using Guava implementations instead

PR #2488, opened 2020-02-20
Use Guava Hashing to compute file checksums

dbevacqua:
## Fixes Issue #

https://github.com/jeremylong/DependencyCheck/issues/2487

## Description of Change

Use Guava `Hashing` for faster computation of file checksums

No

### DIFF
--- a/utils/src/main/java/org/owasp/dependencycheck/utils/Checksum.java
+++ b/utils/src/main/java/org/owasp/dependencycheck/utils/Checksum.java
@@ -18,10 +18,7 @@
 package org.owasp.dependencycheck.utils;
 
 import java.io.File;
-import java.io.FileInputStream;
 import java.io.IOException;
-import java.nio.ByteBuffer;
-import java.nio.channels.FileChannel;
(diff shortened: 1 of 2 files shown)


utils/src/main/java/org/owasp/dependencycheck/utils/Checksum.java
```diff
@@ -79,21 +81,25 @@ private Checksum() {
...
-                final byte[] bytes = new byte[b];
-                buf.get(bytes);
-                md.update(bytes, 0, b);
-                buf.clear();
-                b = ch.read(buf);
-            }
-            return md.digest();
+        HashFunction hashFunction = null;
+
+        switch (algorithm) {
```
jeremylong: To ensure we don't run into casing issues should this be:

```suggestion
        switch (algorithm.toUpperCase()) {
```

jeremylong (changes requested):
Thanks for the PR! I think it just needs one very minor update.

jeremylong (approved):

PR #2493, opened 2020-02-22
Fix broken checkstyle config after checkstyle-plugin update to v3.1.1

aikebah:
Update Checkstyle config for the breaking changes between v8.19 (maven plugin v3.1.0) and v8.29 (maven plugin v3.1.1)

## Description of Change
Updated Checkstyle config to the updated Checkstyle version used by plugin v3.1.1

Decided to just remove `allowUndeclaredRTE` and `allowThrowsTagsForSubclasses` and not introduce anything for that as per CheckStyle's own recommendation in their ticket.

no, not applicable; validated with `mvn -s settings.xml install site site:stage -DreleaseTesting -Danalyzer.node.package.enabled=false -Danalyzer.node.audit.enabled=false` that checkstyle checks properly pass now

### DIFF
--- a/src/main/config/checkstyle-checks.xml
+++ b/src/main/config/checkstyle-checks.xml
@@ -52,9 +52,15 @@
         <property name="maximum" value="0"/>
     </module>
 
+    <module name="LineLength">
+        <property name="ignorePattern" value="^ *\* *[^ ]+$"/>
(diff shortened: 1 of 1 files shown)


jeremylong:
Thanks!

Issue #2255, opened 2019-10-13
Performance improvement: Bootstrapping the H2 with CSV import (with PoC project)


skjolber:
**Is your feature request related to a problem? Please describe.**
Bootstrapping the H2 database is slow.

**Describe the solution you'd like**
Speed up processing by preprocessing JSON files into CSV files, then importing them.

See [this project](https://github.com/skjolber/dependency-check-bootstrap) for a working sample / proof of concept.

The performance enhancement is approximately 2.5x, complexity 'reasonable'. But it only works for H2 (as far as I know). Could this be an optimization worth including in the project?

**Describe alternatives you've considered**
More multithreaded / connections in CveDB. Didn't help much.

**Additional context**
See example project.

skjolber:
@jeremylong is this interesting? I guess some other databases will support CSV as well, like PostgreSQL

jeremylong:
My concern is setting up something like this that increases the maintenance tasks...

skjolber:
Hmm my impression was that this will not increase the maintenance tasks; it will only run from a fresh bootstrap. How / when do you think it will, and should I add the maintenance run to the example for comparison? @jeremylong

skjolber:
Or did you mean the developer workload for (software) maintenance? I should be able to unit-test more or less all use-cases. The parser methods (determineEcosystem(..) etc) should be extracted from the existing code for reduced duplication.

jeremylong:
Yes - I was referring to maintaining dependency-check. If the NVD updates their schema and we have to adjust this adds another location in the code base that would need to be updated.  I'm also looking at adding another datasource to update the database: https://developer.github.com/v4/object/securityadvisory/

Adding the github data may require schema changes and we would have to look at how that impact both the standard import and the csv import.

Now - I think there are other things we can do more standardly to improve the initial data load across all databases. Possibly disabling constraints and restructuring things so we are not deleting prior to inserting on the initial data load.

skjolber:
As far as I remember, the key here is to be able to preprocess the data files while downloading them, as this was a CPU-intensive operation that scales well (i.e. on all cores). I'll try to back that up with numbers.

So the optimal performance path might be to always have a preprocessing step, then two alternative ways of inserting - statement by statement (current behaviour), or bulk (like in my PoC). 

[...]

PR #2534, opened 2020-03-17
Detect ecosystems using AhoCorasickDoubleArrayTrie

skjolber:
## Partly fixes Issue #2255 
Combined performance and functional change.

## Description of Change
Detecting on ecosystems is done partly via searching the issue textual description for hints. This PR replaces a `String.indexOf(..)` approach with use of [AhoCorasickDoubleArrayTrie](https://github.com/hankcs/AhoCorasickDoubleArrayTrie). Search hits are now used to score the ecosystems (a simple count) for cases where a single issue could belong to multiple ecosystems. In addition, filtering is somewhat more advanced.

Yes. But some discussion is still needed:

 * add more / better keyword search? Perhaps even product searches, i.e. like from apache.
 * are unit test examples acceptable (in .txt files)?

The AhoCorasickDoubleArrayTrie project does not get top score on its code patterns in my book, but seems to do the job at an acceptable level.

## Results
Code for performance-testing and diff has been added to the [ecosystem](https://github.com/skjolber/datasets-json-benchmark/tree/ecosystem) branch in a seperate project.

 * single-thread performance: from about 8 seconds to 0.5 seconds. Effective improvement depends on the number of cores. 
 * diff attached in zip file.


skjolber:
[ecosystem.zip](https://github.com/jeremylong/DependencyCheck/files/4340892/ecosystem.zip)

Summary:
Composer added 13
Composer removed 51
Composer kept 18545
Java added 776
Java removed 37
Java kept 784
python added 0
python removed 2
python kept 412
null added 69
null removed 849
null kept 71152
CMAKE added 46
CMAKE removed 9
CMAKE kept 14526
npm added 57
npm removed 0
npm kept 142
perl added 1
perl removed 9
perl kept 269
ruby added 0
ruby removed 5
ruby kept 190

jeremylong:
This might take a bit of research as I'm comparing the differences between the old algorithm and the new.  So this is going to get merged - but may not make the next release.

skjolber:
I guess the database must be flushed when this change is released, at least I have not added any  migration support. 

Another discussion point: What is the ideal order when score are tied? Or would it make sense to map a single cve to multiple ecosystems?

jeremylong:
This is not going to make the 5.3.2 release. In reviewing this I've found other areas that need to be cleaned up with regard to the ecosystem mechanism. Current working branch is https://github.com/jeremylong/DependencyCheck/tree/updateEcosystemLogic - which includes this PR.

skjolber:
@jeremylong sounds good, lets move the discussion to the new PR when ready.

jeremylong:
I finally finished the updates for this PR. A new PR was submitted: #2610.

PR #2536, opened 2020-03-17
Correct JavaScript escaping on HTML reports

### LINKED ISSUE #2535: Markup injection in HTML report template
**Describe the bug**
The default HTML report template is vulnerable to markup injection due to reliance on `$enc.javascript` without `$enc.html` in nested contexts.

Dependency-Check Core version 5.3.1 (cli)

**Log file**
n/a

**To Reproduce**
1. Rename the `core/src/test/resources/javascript/ember.js` file to `"><iframe src=javascript:alert(1)>.js`
2. Run Dependency-Check against the renamed file
3. Open the generated HTML report in a browser
4. Observe a JavaScript `alert` dialog opening

**Expected behavior**
Markup injection should not be possible. All untrusted JavaScript strings should be encoded using both `$enc.javascript` *and* `$enc.html`.

**Additional context**
PR on the way.

jupenur:
## Fixes Issue #
Fixes #2535 

## Description of Change
Added `$enc.html` to several places in the HTML report template.

no


jeremylong:
I truly appreciate the identification of the issue and the PR. However, instead of fixing the nested context with double encoding we are just going to remove the onclick event attributes. See #2543.

PR #2544, opened 2020-03-19
Add mix audit analyzer for Elixir projects

defsprite:
## Fixes Issue #

There is no issue for Elixir support yet, but a stale PR #1340

## Description of Change

Adds support for Elixir dependencies found in `mix.lock` files by the fairly new [mix_audit](https://github.com/mirego/mix_audit) tool.

The implementation is kept fairly close to that of the `bundle-audit` analyzer with the following differences:

- bundle audit creates bogus temporary files, so that hashes on the dependency created will differ. This implementation manually creates the hashes derived from the package name.
- We try to find the vulnerability from the `cveDb`, as the result of mix audit contains that id. In case it cannot be found, a fallback vulnerability with limited information (and unknown severity) is built.

As I'm not too familiar with the code base and the project, this is an inital stab at the implementation and some rework is probably needed. Any feedback is greatly appreciated.
 
*yes*

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 language: java
 #group: deprecated-2017Q4
-dist: trusty
+dist: bionic
 
 env:
   global:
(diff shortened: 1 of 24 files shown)


jeremylong (changes requested):
Awesome!  thanks for the PR.  there is just one small problem - you need to add an entry in `src/main/resources/META-INF/services/org.owasp.dependencycheck.analyzer.Analyzer` for your new analyzer. Otherwise the analyzer will not be loaded by the ServiceLoader.

jeremylong (approved):

PR #2569, opened 2020-04-01
Add Maven wrapper

binkley:
See https://github.com/takari/maven-wrapper.

I ran:
```
mvn -N io.takari:maven:0.7.7:wrapper
```

## Fixes Issue #

N/A

## Description of Change

Provide Maven wrapper, akin to Gradle wrapper, for two reasons:

1. Easier onboarding of new devs (no need to install Maven separately)
2. Better reproducible build, with the project configuration controlling the version of Maven used

The Maven build itself is not tested.

## Documentation

I'll follow up later this week or weekend with documentation updates describing use of Maven wrapper.


mprins:
Not sure how fast Apache Maven moves, but the maven-wrapper as-is is now a legacy product as the functionality is to be integrated in the next minor version of Maven

jeremylong:
@mprins thanks! I didn't release that the wrapper was going to be part of Maven 3.7 (see notes on https://github.com/takari/maven-wrapper).

@binkley thanks for the PR - but I think we should wait for this to be part of core maven.

PR #2583, opened 2020-04-09
fix #2455 #2585

### LINKED ISSUE #2455: Version 5.3.0 java.lang.ClassCastException: org.glassfish.json.JsonStringImpl cannot be cast to javax.json.JsonObject
Run scan against React project (Node).

Version used 5.3.0. CLI

**Log file**
[WARN] An unexpected error occurred during analysis of 'package.json' (Node.js Package Analyzer): org.glassfish.json.JsonStringImpl cannot be cast to javax.json.JsonObject
[ERROR] 
java.lang.ClassCastException: org.glassfish.json.JsonStringImpl cannot be cast to javax.json.JsonObject
	at org.owasp.dependencycheck.analyzer.NodePackageAnalyzer.processDependencies(NodePackageAnalyzer.java:287)
	at org.owasp.dependencycheck.analyzer.NodePackageAnalyzer.analyzeDependency(NodePackageAnalyzer.java:262)
	at java.util.concurrent.FutureTask.run(FutureTask.java:266)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
	at java.lang.Thread.run(Thread.java:748)

### LINKED ISSUE #2585: NodeAnalyzer doesn't support local dependency or alias
**Describe the bug**

If I use local dependency or alias, It seems I get an error about the payload send to NPM .
After some search, it seems that the npm audit doesn't support local or alias modules .

Example to install a local package :
`npm install ./local_package`

Example to install an alias (not supported by npm)
`yarn add react-dom@npm:@hot-loader/react-dom`

CLI 5.3.2

**To Reproduce**
1. analyze node package.json / package-lock containing an alias or a local package

**Expected behavior**
Skip this dependencies .

I think, local package need to be analyzed separately (if repo with multi projects, dependency-check will already scan the others package) .

About the alias, I think the expected behavior is to skip it too .
To analyze it, install it at a dependency (without really saving it, just for the CI), or as a devDependency, or an optionnalDependency (but I think dependency-check doesn't check optionnalDependency)

**Additional context**
This issue is just to note bug I try to resolve with #2583

thib3113:
allow processDependencies with package.json dependencies format ( key = name, value = version )

And skip "not npm-audit compatible" npm dependencies

## Fixes Issue #

fix #2455
fix #2585 

## Description of Change

Actually, `processDependencies` can be called with the content of package.json, or lock file (npm-shrinkwrap.json, package-lock.json) .

But in package.json, dependency is like : 
```
{
   "braces": "1.8.5",
}
```
and in lock files 
```
{
"braces": {
      "version": "1.8.5",
      "resolved": "http://registry.citopia-dev.lan:4873/braces/-/braces-1.8.5.tgz",
      "integrity": "sha1-uneWLhLf+WnWt2cR6RS3N4V79qc=",
      "requires": {
        "expand-range": "1.8.2",
        "preserve": "0.2.0",
        "repeat-element": "1.1.3"
      }
    }
}
```

The other change, with to skip yarn alias, or local dependencies, because they are not handled by npm audit (npm audit crash if encoutering it)

Yes

I also revert https://github.com/jeremylong/DependencyCheck/commit/8f1f5663e48d57cb77dd9af722a8ab62f6c808c6 , because this doesn't solve the error, and add quotes in the string .


mprins:
I think having a lock file is the only way that transitive dependencies can be checked, which is exaclty the reason for using this tool. Not checking the bulk of the (transitive) dependencies leaves the potential of many vulnerable dependencies.

thib3113:
@mprins yes, I agree ... This PR is just to resolve the exceptions, not to add/remove features .

Else, I can "understand" the "need" of a minimal scan, only with package.json . If you are using yarn for example, you can't have a package-lock .... you will get a yarn.lock (following a format I didn't known, and the only way I found to parse it is to use the [yarnlock parser](https://www.npmjs.com/package/@yarnpkg/lockfile), but need a node project) .

The other way I found, is to use some libraries like https://www.npmjs.com/package/synp ... But, the library will a generate a different package-lock, because it will add alias, and so, break the npm tree checker . 

And if we try to just run `npm install` on a project managed by yarn, it will not respect the lock file ... so can install different version of the dependency

thib3113:
Bugs for the moment : 
 - RetireJS try to scan file that doesn't exist, maybe need to skip dependencies too . I'll check

thib3113:
The bug with RetireJS seems not related to this update (but, because I run the command in a CI folder, I think a CI job reset the node_modules when retireJS try to read js files) .

So, I'm waiting your return about this PR .

And I have some questions : 
 - what about optional dependencies ? I think, it can be interessting to analyze optional dependencies, if they are installed . (and so, you can use an alias, and the real dependency in optional, and just installing optionnal when running dependency-check), or things like that ? (it seems optionnal dependencies are installed like dependencies, but doesn't fail if install crash ... so maybe not the best for an alias, but can be interessting in some cases)


 - did we have a way to centralize node test data ? all in one folder, with subfolders ? So we can generate some tests, without duplicate them .

 - it seems that I have lot of "not commited files", generated by the install and tests : 
```
        ant/data/
        archetype/src/main/resources/archetype-resources/target/
        core/src/test/resources/pom/target/
        dependency-check-report.html
        maven/src/it/1527-aggregate-updated-dependencies/new/ear/target/
        maven/src/it/1527-aggregate-updated-dependencies/new/lib/target/
        maven/src/it/1527-aggregate-updated-dependencies/old/ear/target/
        maven/src/it/1527-aggregate-updated-dependencies/old/lib/target/
```
Do we need to add them to gitignore ?

- readme windows command seems outdated : 
```
> mvn -s settings.xml install
> .\dependency-check-cli\target\release\bin\dependency-check.bat -h
> .\dependency-check-cli\target\release\bin\dependency-check.bat --project Testing --out . --scan ./src/test/resources
```
`.\dependency-check-cli\target\release\bin\dependency-check.bat` seems to be `.\cli\target\release\bin\dependency-check.bat` .

jeremylong:
I am definitely planning on merging this (pending review, etc.) However, it might take me a little bit to get back to this as I am deep in #2534 - and have a few more updates to improve the speed an accuracy of the update process as it relates to the ecosystem.

Regarding centralizing the test cases - absolutely. We likely should centralize the test resources when possible and a node folder with subdirectories would work fine.

Regarding the not-committed files I'm not sure what is going on there as these files are not generated when I build (or are already excluded by `*/target/**`).

Finally, in what I can respond to today you are correct the README.md needs to be updated.  I'll update this now.

thib3113:
@jeremylong No problem . For ce moment, I'm using my build in my CI (so can debug if I see bugs) . So I can wait .

About the centralizing test cases, I didn't know how it's work .
And because nodejs add looooot of little files, duplacting them are not a good idea . And a common folder can be used for retireJs and npm and nodeaudit, with some specific modules .

---

For the moment, I see some tests on 
`src/test/resources/nodejs/package.json`, and on `core/src/test/resources/nodejs/package.json`

and I don't how (for example), in `core/src/test/java/org/owasp/dependencycheck/data/nodeaudit/NpmPayloadBuilderTest.java`
 I can reuse 
`src/test/resources/nodejs/package.json`
(or maybe with a long relative path, but maybe not the best) . 

If someone can help me about centralized the files, I can try to add more tests cases

---

I'll try remove each of them, and retry ... ( I'm not a java developper, so I've done lot of tries to compile it, and the default profile that jetbrains use, is not the correct one ). 
Just, the `dependency-check-report.html` is the one generated by the cli in the readme .

jeremylong:
We can centralize the test resources for npm.  The actual `*Test.java` files should not be moved though - these are in the packages (folder structure) as their corresponding Java class.  When we move any test resources we may have to update other test cases to correctly reference the new location.

jeremylong:
This was merged via #2622

PR #2595, opened 2020-04-15
fix #2594

### LINKED ISSUE #2594: False Positive on ch.qos.logback.contrib.logback*
#### Example
False positive on libraries ```ch.qos.logback.contrib.logback-json-core-0.1.5.jar```, ```ch.qos.logback.contrib.logback-jackson-0.1.5.jar```, ```ch.qos.logback.contrib.logback-json-classic-0.1.5.jar``` - reported as ```cpe:2.3:a:logback:logback:0.1.5:*:*:*:*:*:*:*``` which is the main logback project instead of logback-contrib
```sbt
    "ch.qos.logback.contrib"             %  "logback-json-classic"               % "0.1.5",
    "ch.qos.logback.contrib"             %  "logback-json-core"                  % "0.1.5",
    "ch.qos.logback.contrib"             %  "logback-jackson"                    % "0.1.5",
```

```ch.qos.logback.contrib``` should not be matched with ```cpe:2.3:a:logback:logback:0.1.5:*:*:*:*:*:*:*```

### PR DESCRIPTION
**Pull request by tzch96**
## Fixes Issue #
#2594 
## Description of Change

```ch.qos.logback.contrib*``` is no longer matched with ```cpe:2.3:a:logback:logback:0.1.5:*:*:*:*:*:*:*```

no

### DIFF
--- a/core/src/main/resources/dependencycheck-base-suppression.xml
+++ b/core/src/main/resources/dependencycheck-base-suppression.xml
@@ -4,21 +4,7 @@
         <notes><![CDATA[
         FP per #2594
         ]]></notes>
-        <packageUrl regex="true">^pkg:maven/ch\.qos\.logback\.contrib/logback\-json\-core@.*$</packageUrl>
-        <cpe>cpe:/a:logback:logback</cpe>
-    </suppress>
-    <suppress base="true">
-        <notes><![CDATA[
-        FP per #2594
-        ]]></notes>
-        <packageUrl regex="true">^pkg:maven/ch\.qos\.logback\.contrib/logback\-jackson@.*$</packageUrl>
-        <cpe>cpe:/a:logback:logback</cpe>
-    </suppress>
-    <suppress>
-        <notes><![CDATA[
-        FP per #2594
-        ]]></notes>
-        <packageUrl regex="true">^pkg:maven/ch\.qos\.logback\.contrib/logback\-json\-classic@.*$</packageUrl>
+        <packageUrl regex="true">^pkg:maven/ch\.qos\.logback\.contrib/.*$</packageUrl>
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: core/src/main/resources/dependencycheck-base-suppression.xml [hunk 1]
```diff
@@ -1,5 +1,26 @@
...
+        <packageUrl regex="true">^pkg:maven/ch\.qos\.logback\.contrib/logback\-jackson@.*$</packageUrl>
+        <cpe>cpe:/a:logback:logback</cpe>
+    </suppress>
+    <suppress>
+        <notes><![CDATA[
+        FP per #2594
+        ]]></notes>
+        <packageUrl regex="true">^pkg:maven/ch\.qos\.logback\.contrib/logback\-json\-classic@.*$</packageUrl>
+        <cpe>cpe:/a:logback:logback</cpe>
+    </suppress>
```
jeremylong: These three rules could probably be converted to a single rule that would be more future proof (i.e. adding additional contributions).
```suggestion
    <suppress base="true">
        <notes><![CDATA[
        FP per #2594
        ]]></notes>
        <packageUrl regex="true">^pkg:maven/ch\.qos\.logback\.contrib/.*$</packageUrl>
        <cpe>cpe:/a:logback:logback</cpe>
    </suppress>
```

jeremylong (changes requested):
Thank you for the PR - just one minor suggestion to make this a little more lenient in case additional "contrib" libraries are added in the future.

core/src/main/resources/dependencycheck-base-suppression.xml
tzch96: Agreed, this also takes care of other existing components which may be part of logback-contrib and potential future ones

Issue #2608, opened 2020-04-23
erratic relatedDependencies behaviour


valentijnscholten:
Hi,

I am seeing erratic behaviour when scanning the same set of artifacts over and over again.
This is when there are multiple artifacts with the same (vulnerable) (bundles) dependencies.

Situation:

app_a.war contains handlebars-2.0.0.js
app_b.war contains handlebars-2.0.0.js

Both wars are components in the same application, let's say a customer facing war and a operator facing war both using a common backend rest api (left out for simplicity).

Scanning with dependency check delivers XML report along the following lines:

```
<dependency isVirtual="false">
	<fileName>app_a.war: handlebars-2.0.0.js</fileName>
	<filePath>/var/lib/jenkins/workspace/middleware_develop/target/app_a.war/apidocs/lib/handlebars-2.0.0.js</filePath>
	<md5>501c421a9bc201f50c76c8d28af0cb36</md5>
	<sha1>f25d39dc72774e392d55d98dd1d1285b1e213809</sha1>
	<sha256>7cb481a09730ac4f570ec37702f2fa70ce197bec81100565c6817eb13666a796</sha256>
	<relatedDependencies>
		<relatedDependency>
			<filePath>/var/lib/jenkins/workspace/middleware_develop/anotherfolder/app_b.war/apidocs/lib/handlebars-2.0.0.js</filePath>
			<sha256>7cb481a09730ac4f570ec37702f2fa70ce197bec81100565c6817eb13666a796</sha256>
			<sha1>f25d39dc72774e392d55d98dd1d1285b1e213809</sha1>
			<md5>501c421a9bc201f50c76c8d28af0cb36</md5>
		</relatedDependency>
	</relatedDependencies>
	
```

When using vulnerability aggregators or reporting tools, this hides app_b.war from the results most of the time as these
tools do not usually look at relatedDependencies.

Problem is 2 that sometimes dependency check flips the order around:

```
<dependency isVirtual="false">
	<fileName>app_b.war: handlebars-2.0.0.js</fileName>
	<filePath>/var/lib/jenkins/workspace/middleware_develop/target/anotherfolder/app_b.war/apidocs/lib/handlebars-2.0.0.js</filePath>
	<md5>501c421a9bc201f50c76c8d28af0cb36</md5>
	<sha1>f25d39dc72774e392d55d98dd1d1285b1e213809</sha1>
[...]

PR #2609, opened 2020-04-24
add filename for relateddependencies to xml/json reports

valentijnscholten:
## Fixes Issue #
Implements item 3 from #2608: add filename in xml/json reports for related dependencies

testcases have been updated


jeremylong (changes requested):
Thanks for the PR! While a very limited number of JAR files are included as test resources - the preference is to add them as optional test dependencies in the `pom.xml`.  Mind updating this to remove the JAR file and include it via the pom?

core/src/main/resources/schema/dependency-check.2.4.xsd
```diff
@@ -0,0 +1,247 @@
...
+            <xs:element name="id" type="xs:string" minOccurs="1" maxOccurs="1"/>
+            <xs:element name="url" type="xs:string" minOccurs="0" maxOccurs="1"/>
+            <xs:element name="description" type="xs:string" minOccurs="0" maxOccurs="1"/>
+            <xs:element name="notes" type="xs:string" minOccurs="0" maxOccurs="1"/>
+        </xs:sequence>
+        <xs:attribute name="confidence" type="xs:string" use="optional"/>
+    </xs:complexType>
+    <xs:complexType name="relatedDependency">
+        <xs:sequence>
+            <xs:element name="fileName" type="xs:string" minOccurs="1" maxOccurs="1"/>
```
jeremylong, line 63: To make the schema backwards compatible I think this should be an optional (e.g. minOccurs="0":

```suggestion
            <xs:element name="fileName" type="xs:string" minOccurs="0" maxOccurs="1"/>
```

valentijnscholten: Isn't the idea that xml files have an xsd declaration in them that points to the schema version? If we need minOccurs=0, we could never add a "mandatory" field somewhere?

valentijnscholten:
> Thanks for the PR! While a very limited number of JAR files are included as test resources - the preference is to add them as optional test dependencies in the `pom.xml`. Mind updating this to remove the JAR file and include it via the pom?

I could see if there is an example already that does this? In the PR I used the existing mechanism in that test, but it feels a bit weird to includ jars indeed.

jeremylong:
@valentijnscholten Most of the test dependencies are defined in the core project pom.xml:

https://github.com/jeremylong/DependencyCheck/blob/73df62ee0e1407d99f1ef38b52785feef3d6ca98/core/pom.xml#L426-L432

https://github.com/jeremylong/DependencyCheck/blob/a503ca0eb95a7162a13fd1ff239706978909275a/core/src/test/java/org/owasp/dependencycheck/analyzer/JarAnalyzerTest.java#L174-L182

There are several JARs in the resources directory that are also used by `JarAnalyzerTest.java` - however, in some of these cases the JAR was slightly modified or does not exist in Central.

valentijnscholten:
Commit:
- use maven for test dependency 
- use v2.4 everywhere for generating report and in targetnamespace

Would you let me know if you want that `minOccurs` above to be 0 or 1?

core/src/main/resources/schema/dependency-check.2.4.xsd
jeremylong: As we are likely going to make the next version a major release (i.e. breaking changes) I'm okay with leaving this as 1 now.

jeremylong (approved):

valentijnscholten:
Not sure if a milestone label is needed, but I believe this will be in the upcoming 6.0.0

PR #2724, opened 2020-07-16
Inroducing perl cpanfile dependency analyzer

harj-the-dev:
## Fixes Issue #
No. This is for a new feature I required and wanted to share.

## Description of Change

Added perl scanner using dependencies listed in the cpanfile. Feature is hidden behind the `--enableExperimental` flag.
Example usage:
```
java -classpath "/path/*" -Dapp.name="dependency-check" -Dapp.repo="/hob" -Dapp.home="/path/" -Dbasedir="/path/" org.owasp.dependencycheck.App -s . --enableExperimental
```

*yes*

### DIFF
--- a/core/src/main/java/org/owasp/dependencycheck/analyzer/PerlCpanfileAnalyzer.java
+++ b/core/src/main/java/org/owasp/dependencycheck/analyzer/PerlCpanfileAnalyzer.java
@@ -55,6 +55,8 @@ import org.owasp.dependencycheck.utils.Checksum;
 import org.owasp.dependencycheck.utils.InvalidSettingException;
 import org.owasp.dependencycheck.dependency.Vulnerability;
 import org.owasp.dependencycheck.data.nvdcve.CveDB;
+import org.owasp.dependencycheck.data.nvdcve.ICveDBWrapper;
+import org.owasp.dependencycheck.data.nvdcve.CveDBWrapper;
 import org.owasp.dependencycheck.models.VulnerabilityDTO;
 
 import java.sql.ResultSet;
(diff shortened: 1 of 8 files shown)


harj-the-dev:
Hiya, Is there anything I can do to help get this feature included?

jeremylong:
The design of this analyzer does not follow how dependency-check is intended to work.  A `FileTypeAnalyzer` job is to create the dependency objects - in this case based off of the cpan file. The original cpan dependency should like be removed (see how the [PIP analyzer works](https://github.com/jeremylong/DependencyCheck/blob/bc907dfcc1d4f377dcf25c6d5b3ceaf3f25e8984/core/src/main/java/org/owasp/dependencycheck/analyzer/PipAnalyzer.java#L129-L131)); you then create [virtual dependencies](https://github.com/jeremylong/DependencyCheck/blob/bc907dfcc1d4f377dcf25c6d5b3ceaf3f25e8984/core/src/main/java/org/owasp/dependencycheck/analyzer/PipAnalyzer.java#L146) from the 'required', 'recommended'. etc. entries from the CPAN file.  Each dependency that gets created would have as much evidence as can be extracted (at least one entry in each 'vendor', 'product', and 'version' bucket - in addition one would want to add a [PURL identifier](https://github.com/package-url/purl-spec) (again see how the [PIP analyzer works](https://github.com/jeremylong/DependencyCheck/blob/bc907dfcc1d4f377dcf25c6d5b3ceaf3f25e8984/core/src/main/java/org/owasp/dependencycheck/analyzer/PipAnalyzer.java#L150-L151)).

Once this is done the other analyzers in the pipeline (CPE Analyzer, NVD CVE Analyzer, OSS Index Analyzer, etc.) will populate the vulnerability information.

The only analyzers that actually add vulnerabilities directly to the dependency object are those that are actually running some form of SCA vulnerability analysis such as Ruby Bundle Audit or `npm audit`.

jeremylong:
Thanks for the base code. I have updated the code and created a new PR based on your original PR (see #3378).

PR #2732, opened 2020-07-28
Update site links to GitHub main branch

sainaen:
## Description of Change

Fix the links from documentation site to GitHub that point at `master` instead of the `main` branch, causing 404-s.

Note, that the SCM urls in the various `pom.xml` were not updated (I’m not sure if that would break something or not).

*no*


jeremylong:
Thanks!

Issue #2796, opened 2020-09-07
npm modules that contain the name of a CPE are falsely identified in v6.0.0


aarongoldenthal:
Numerous npm modules that contain the name of a package with a valid CPE are falsely identified as that CPE in v6.0.0 (cli).  This did not occur with v5.3.2.  See some examples below:

```
|Vulnerability IDs                          |Package                                                                                                                                                      |Highest Severity                                   |CVE Count|Confidence|Evidence Count|
|-------------------------------------------|-------------------------------------------------------------------------------------------------------------------------------------------------------------|---------------------------------------------------|---------|----------|--------------|
|@babel/plugin-syntax-json-strings:7.8.3    |cpe:2.3:a:json_project:json:7.8.3:*:*:*:*:*:*:*                                                                                                              |pkg:npm/%40babel%2Fplugin-syntax-json-strings@7.8.3|HIGH     |          |Highest       |
|@types/json-schema:7.0.5                   |cpe:2.3:a:json_project:json:7.0.5:*:*:*:*:*:*:*                                                                                                              |pkg:npm/%40types%2Fjson-schema@7.0.5               |HIGH     |          |Highest       |
|@types/parse-json:4.0.0                    |cpe:2.3:a:json_project:json:4.0.0:*:*:*:*:*:*:* cpe:2.3:a:parsejson_project:parsejson:4.0.0:*:*:*:*:*:*:*                                                    |pkg:npm/%40types%2Fparse-json@4.0.0                |HIGH     |          |Highest       |
|decamelize-keys:1.1.0                      |cpe:2.3:a:decamelize_project:decamelize:1.1.0:*:*:*:*:*:*:*                                                                                                  |pkg:npm/decamelize-keys@1.1.0                      |HIGH     |          |Highest       |
|deep-extend:0.5.1                          |cpe:2.3:a:deep_extend_project:deep_extend:0.5.1:*:*:*:*:*:*:* cpe:2.3:a:extend_project:extend:0.5.1:*:*:*:*:*:*:*                                            |pkg:npm/deep-extend@0.5.1                          |CRITICAL |          |Highest       |
|deep-extend:0.6.0                          |cpe:2.3:a:deep_extend_project:deep_extend:0.6.0:*:*:*:*:*:*:* cpe:2.3:a:extend_project:extend:0.6.0:*:*:*:*:*:*:*                                            |pkg:npm/deep-extend@0.6.0                          |CRITICAL |          |Highest       |
[...]

PR #2878, opened 2020-10-11
Node.js False Positives

jeremylong:
## Fixes Issue #2796

Filter npm vulnerabilities when CPE doesn't match the product. Additionally, if the CPE contains a target_sw - ensure it is valid for Node.js.


jeremylong:
Note - some of this could be expanded to create more precise matching for other technologies.

PR #2899, opened 2020-10-18
Implementation of the new database schema for Oracle database

aikebah:
## Description of Change
Updated the Oracle initialization script and custom queries to match up with the new database layout of v6

Optimized querying Oracle by making larger fetches than by default to more closely match the other DBMSes that fetch *ALL* results directly into the ResultSet.

No, not feasible as there is no Oracle DB to run the testcases against.
Some manual tests run with various projects to compare the vulnerabilities output between them and a manual comparison of post-initial load counts for all tables between PostgreSQL and Oracle to make sure both import the same number of records for each.


jeremylong:
@aikebah thank you for the PR!

PR #2914, opened 2020-10-21
Add missing cleanup to UpdateMojo

aikebah:
## Description of Change

Add the missing cleanup routing to the UpdateMojo to remove the temporary folder at the end of an update-only run

no, ran update-only goal locally and verified that now the dctemp folder in the temp directory is removed


jeremylong:
Thanks for identifying this.  In some ways I almost want to move the cleanup into `engine.close()`...

PR #2916, opened 2020-10-23
Add completion for `dependency-check.sh`

jeremylong:
Add script for completion for dependency-check.sh so we can more easily select the correct arguments. After installing `completion-for-dependency-check.sh` (using `source completion-for-dependency-check.sh`) we can list the arguments using `<tab><tab>`. The following example lists all of the `--disable*` arguments:

```bash
$ ./dependency-check.sh --disable<tab><tab>
--disableArchive                      --disableCocoapodsAnalyzer            --disableNodeAuditCache               --disablePip
--disableAssembly                     --disableComposer                     --disableNodeJS                       --disablePipfile
--disableAutoconf                     --disableGolangDep                    --disableNugetconf                    --disablePyDist
--disableBundleAudit                  --disableGolangMod                    --disableNuspec                       --disablePyPkg
--disableCentral                      --disableJar                          --disableOpenSSL                      --disableRetireJS
--disableCentralCache                 --disableMixAudit                     --disableOssIndex                     --disableRubygems
--disableCmake                        --disableNodeAudit                    --disableOssIndexCache                --disableSwiftPackageManagerAnalyzer
```

Consider adding `source <path>/completion-for-dependency-check.sh` to your .bashrc if you use ODC a lot.

### DIFF
--- a/cli/src/main/resources/completion-for-dependency-check.sh
+++ b/cli/src/main/resources/completion-for-dependency-check.sh
@@ -111,7 +111,7 @@ _odc_completions()
 
 
     case "${prev}" in
-        -o|--out|-d|--data|--bundleAudit|--bundleAuditWorkingDirectory|--dbDriverPath|--dotnet|--go|-P|--propertyfile|--suppression|--hint|-l|--log)
+        -s|--scan|-o|--out|-d|--data|--bundleAudit|--bundleAuditWorkingDirectory|--dbDriverPath|--dotnet|--go|-P|--propertyfile|--suppression|--hint|-l|--log)
             COMPREPLY=( $(compgen -f -o default -- ${cur}) )
             return 0
             ;;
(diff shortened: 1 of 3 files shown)


cli/src/main/resources/completion-for-dependency-check.sh
```diff
@@ -0,0 +1,135 @@
...
+            COMPREPLY=( $(compgen -W "true false" -- ${cur}) )
+            return 0
+            ;;
+        -f|--format)
+            COMPREPLY=( $(compgen -W "HTML XML CSV JSON JUNIT ALL" ${cur}) )
+            return 0
+            ;;
+    esac
+    if [[ "$cur" == -* ]] ; then
+        # COMPREPLY=( $(compgen -W "$options" -- "$cur") )
```
jeremylong:
```suggestion
```

Issue #2949, opened 2020-11-11
Dockerfile doesn't function with OpenShift restricted SSC


namloc2001:
**Is your feature request related to a problem? Please describe.**
The current Dockerfile creates a container that cannot be started in OpenShift by a serviceAccount with the restricted SCC permissions. Instead it has to be created by a serviceAccount with the `anyuid` SCC or `nonroot` SCC.

This is because, as per https://docs.openshift.com/enterprise/3.2/creating_images/guidelines.html, 

By default, OpenShift Enterprise runs containers using an arbitrarily assigned user ID. This provides additional security against processes escaping the container due to a container engine vulnerability and thereby achieving escalated permissions on the host node.

For an image to support running as an arbitrary user, directories and files that may be written to by processes in the image should be owned by the root group and be read/writable by that group. Files to be executed should also have group execute permissions.

Adding the following to your Dockerfile sets the directory and file permissions to allow users in the root group to access them in the built image.

Because the container user is always a member of the root group, the container user can read and write these files. The root group does not have any special permissions (unlike the root user) so there are no security concerns with this arrangement.

**Describe the solution you'd like**
Solution is to amend/add the following code into the Dockerfile:
```
chown -R ${user}:0 /usr/share/dependency-check && \
chown -R ${user}:0 /report && \
chmod -R g=u /usr/share/dependency-check && \
chmod -R g=u /report
```

**Describe alternatives you've considered**
Alternatively, a new "OpenShift Dockerfile" could be created using the original as a base and applying these changes:
```
FROM owasp/dependency-check:6.0.2

ENV user=dependencycheck
ENV JAVA_HOME=/opt/jdk
ENV JAVA_OPTS="-Danalyzer.assembly.dotnet.path=/usr/bin/dotnet -Danalyzer.bundle.audit.path=/usr/bin/bundle-audit -Danalyzer.golang.path=/usr/local/go/bin/go"

USER 0

RUN chown -R ${user}:0 /usr/share/dependency-check && \
        chown -R ${user}:0 /report && \
	chmod -R g=u /usr/share/dependency-check && \
	chmod -R g=u /report

### remove any suid sgid - we don't need them
RUN for i in `find / -perm +6000 -type f`; do chmod a-s $i; done
USER ${user}

#VOLUME ["/src", "/report"]

WORKDIR /src

CMD ["--help"]
ENTRYPOINT ["/usr/share/dependency-check/bin/dependency-check.sh"]
```

[...]

PR #2950, opened 2020-11-11
Chown root group (0) for OpenShift

namloc2001:
## Fixes Issue #
#2949 

## Description of Change
As per https://docs.openshift.com/enterprise/3.2/creating_images/guidelines.html, to support Arbitrary User IDs. 

For an image to support running as an arbitrary user, directories and files that may be written to by processes in the image should be owned by the root group and be read/writable by that group. Files to be executed should also have group execute permissions.

Because the container user is always a member of the root group, the container user can read and write these files. The root group does not have any special permissions (unlike the root user) so there are no security concerns with this arrangement.

*no*


jeremylong:
Thanks for the PR!

Issue #2960, opened 2020-11-18
Maven Plugin: Proxy basic authentication fails with encrypted password, setting jdk.http.auth.tunneling.disabledSchemes="" comes too late


stefan-prange:
**Describe the bug**

I found (and solved) a problem with proxy authentication when downloading the NVD CVE vulnerability data from the internet over a proxy which requires basic authentication and uses http.

**My environment**
- Windows 10
- OpenJDK 11.0.8
- Maven 3.6.3
- Dependency-Check Maven plugin version 6.0.3
- Internet access is only possible over a proxy which requires basic authentication and which uses the http protocol on port 8080.
- Proxy access data is present as a \<proxies>\<proxy> element in my file \<HOME>/.m2/settings.xml 
- The element \<proxies>\<proxy>\<password> of that file contains an encrypted value, as described on https://maven.apache.org/guides/mini/guide-encryption.html

**Version of dependency-check used**

I'm using the Maven plugin with the following coordinates:
```xml
<plugin>
    <groupId>org.owasp</groupId>
    <artifactId>dependency-check-maven</artifactId>
    <version>6.0.3</version>
</plugin>
```

**Log file**

When running the command "mvn dependency-check:check" on my project, I get the following error information (log snippet):
```txt
Caused by: java.io.IOException: Unable to tunnel through proxy. Proxy returns "HTTP/1.1 407 Proxy Authentication Required"
    at sun.net.www.protocol.http.HttpURLConnection.doTunneling (HttpURLConnection.java:2177)
    at sun.net.www.protocol.https.AbstractDelegateHttpsURLConnection.connect (AbstractDelegateHttpsURLConnection.java:183)
    at sun.net.www.protocol.https.HttpsURLConnectionImpl.connect (HttpsURLConnectionImpl.java:168)
    at org.owasp.dependencycheck.utils.HttpResourceConnection.obtainConnection (HttpResourceConnection.java:206)
    at org.owasp.dependencycheck.utils.HttpResourceConnection.fetch (HttpResourceConnection.java:163)
```

**To Reproduce**
1. Set up an environment like mine, as described above, especially with an internet proxy that requires basic authentication and works with http.
2. Run `mvn dependency-check:check` or `mvn dependency-check:update-only` on an arbitrary Maven project in whose POM in its build/plugins or build/pluginManagement/plugins part you added the \<plugin> snippet shown above.

**My solution attempts:**

[...]

PR #2966, opened 2020-11-23
[issue 2960] Added support for encrypted proxy passwords in Maven set…

stefan-prange:
…tings

Also:
- Refactored handling of SecDispatcherException in class org.owasp.dependencycheck.maven.BaseDependencyCheckMojo.
- Added setting the system property "jdk.http.auth.tunneling.disabledSchemes" to an empty list already in
  org.owasp.dependencycheck.maven.BaseDependencyCheckMojo (if demanded), because setting it only
  in org.owasp.dependencycheck.utils.URLConnectionFactory comes too late to take effect when using DepCheck
  as a Maven plugin.

## Description of Change

Please see the commit message above and the code diff.
And https://github.com/jeremylong/DependencyCheck/issues/2960 for the background of this change.

*no*


jeremylong (approved):

jeremylong:
Thanks for the PR!

PR #2980, opened 2020-11-26
Add suppression for org.infinipan.protostream

ghost:
## Description of Change

Add suppression for org.infinispan.protostream

n/a


core/src/main/resources/dependencycheck-base-suppression.xml
```diff
@@ -3963,4 +3963,11 @@
         <packageUrl regex="true">^pkg:maven/com\.jcraft/jsch\.agentproxy.*$</packageUrl>
         <cpe>cpe:/a:jcraft:jsch</cpe>
     </suppress>
+    <suppress>
```
jeremylong: We generally add the base="true" attribute for core suppression rules. This prevents them from showing up in the "suppressed" section of the report.
```suggestion
    <suppress base="true">
```

ghost: Ah, yes. Oops :) Added. Thanks!

PR #2987, opened 2020-12-01
golang mod fix

irisdingbj:
## Fixes Issue #

-  Json parsing issues where I have `{ ` inside  the `go list -m -json`  result, eg: 

```
{         "Path": "github.com/Microsoft/hcsshim",
        "Version": "v0.8.7",
        "Replace": {
                "Path": "github.com/Microsoft/hcsshim",
                "Version": "v0.8.8-0.20200421182805-c3e488f0d815",
                "Time": "2020-04-21T18:28:05Z",
                "GoMod": "/xxxxxx/go/pkg/mod/cache/download/github.com/!microsoft/hcsshim/@v/v0.8.8-0.20200421182805-c3e488f0d815.mod"
        },         "Indirect": true,
        "GoMod": "/xxxxxx/pkg/mod/cache/download/github.com/!microsoft/hcsshim/@v/v0.8.8-0.20200421182805-c3e488f0d815.mod" }

``` 

- Add `declared-license` into the dependencies.

- Correct the `FilePath` , 'FileName`,'ActualFilePath` for the found go modules from the `Dir`

- Respect `analyzer.golang.path` in `dependencycheck.properties`  as 1st priority and backward to default if we have no such setting. 

.

### DIFF
--- a/core/src/main/java/org/owasp/dependencycheck/data/golang/GoModDependency.java
+++ b/core/src/main/java/org/owasp/dependencycheck/data/golang/GoModDependency.java
@@ -19,6 +19,14 @@ package org.owasp.dependencycheck.data.golang;
 
 import com.github.packageurl.MalformedPackageURLException;
 import com.github.packageurl.PackageURLBuilder;
+import com.google.common.base.Strings;
+import java.io.File;
+import java.io.IOException;
+import java.nio.charset.Charset;
+import java.util.Arrays;
+import java.util.HashSet;
+import java.util.Set;
+import org.apache.commons.io.FileUtils;
 import org.apache.commons.lang3.StringUtils;
 import org.owasp.dependencycheck.dependency.Confidence;
 import org.owasp.dependencycheck.dependency.Dependency;
@@ -45,6 +53,11 @@ public class GoModDependency {
      */
     private static final Logger LOGGER = LoggerFactory.getLogger(GoModDependency.class);
     /**
+     * A list of license files we recognize.
+     */
+    private static final Set<String> LICENSE_FILES = new HashSet<String>(Arrays.asList("LICENSE", "LICENCE", "LICENSE.TXT",
+            "LICENSE.MD", "LICENCE.MD", "LICENSE.CODE", "LICENCE.CODE", "COPYING"));
+    /**
      * The module path.
      */
     private final String modulePath;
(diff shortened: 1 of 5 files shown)


core/src/main/java/org/owasp/dependencycheck/data/golang/GoModDependency.java
```diff
@@ -155,9 +174,33 @@ private Dependency createDependency(Dependency parentDependency, String name, St
...
     }
 
+	private void extractLicense(Dependency dependency, File file) {
+
+		File[] files = file.listFiles();
+		if (files != null ) {
+			for (File f : files) {
+				if (supportedLicenseFilenames.contains(f.getName())) {
+					String license = f.getAbsolutePath();
+					dependency.setLicense(license);
```
jeremylong: Did you mean the license to be just the path to the file? Or should we read the content of the file into the license text in the dependency object?

core/src/main/java/org/owasp/dependencycheck/analyzer/GolangModAnalyzer.java
```diff
@@ -132,22 +133,18 @@ protected FileFilter getFileFilter() {
...
-                        );
-                        goPath = "go";
-                    }
-                }
+                    LOGGER.warn("Provided path to `go` executable is invalid. Trying default location. "
+                            + "If you do want to set it, please set the `{}` property",
+                            Settings.KEYS.ANALYZER_GOLANG_PATH
+                    );
+                    goPath = "go";
+               }
```
jeremylong: I don't believe this change is needed as this is contained in a synchronized block. We only need to set the goPath once.

jeremylong (commented):
Thanks for the PR - some good things in here. However, I had a PR in flight that changed some of what you were modifying. I already have a change in flight to incorporate some of you suggestions in the PR; as such there is no need to rebase and update this PR. However, I do have a question around your intent with the license file - see comment/question in the code...

Regarding the fix for the `Replace {}` issue, which does correctly have a trailing comma, I have modified the JsonArrayFixingInputStream to correctly handle a closing brace with a trailing comma (I'll push the change shortly).

core/src/main/java/org/owasp/dependencycheck/analyzer/GolangModAnalyzer.java
irisdingbj: we have `private static String goPath = "go"; `  Does this mean the setting from `dependencycheck.properties` will not take effect at all?

core/src/main/java/org/owasp/dependencycheck/data/golang/GoModDependency.java
irisdingbj: Yes, I am hesitant for this part as well..  Now I just set   it as filepath to the license file.    How do we deal with this for other analyzer? What is the suggested approach for this?

irisdingbj:
> Thanks for the PR - some good things in here. However, I had a PR in flight that changed some of what you were modifying. I already have a change in flight to incorporate some of you suggestions in the PR; as such there is no need to rebase and update this PR. However, I do have a question around your intent with the license file - see comment/question in the code...
> 
> Regarding the fix for the `Replace {}` issue, which does correctly have a trailing comma, I have modified the JsonArrayFixingInputStream to correctly handle a closing brace with a trailing comma (I'll push the change shortly).

Thanks @jeremylong  for the review! Can you please put a link for the fix you are working on? Thanks!

irisdingbj:
@jeremylong  Can we get this PR in? Thanks!

Issue #3026, opened 2020-12-17
False Positive on hawtio-wildfly-2.12.0.war


ghost:
When reporting a false positive please include:
- The location of the dependency (Maven GAV, URL to download the dependency, etc.)
- The CPE that is believed to be false positive
  - Please report the CPE not the CVE

#### Example
False positive on library hawtio-wildfly-2.12.0.war - reported as:
- `cpe:2.3:a:hawt:hawtio:2.12.0:*:*:*:*:*:*:*`
- `cpe:2.3:a:hawt.io:hawtio:2.12.0:*:*:*:*:*:*:*`
- `cpe:2.3:a:redhat:wildfly:2.12.0:*:*:*:*:*:*:*`

```xml
            <dependency>
                <groupId>io.hawt</groupId>
                <artifactId>hawtio-wildfly</artifactId>
                <version>2.12.0</version>
                <type>war</type>
            </dependency>
```

https://ossindex.sonatype.org/component/pkg:maven/io/hawtio/hawtio-wildfly@2.12.0

The vulnerabilities shown are for (ancient) versions of Wildfly.

jeremylong:
Thanks for the PR! This will be included in 6.0.4.

PR #3027, opened 2020-12-18
false positive on io.hawtio.hawtio-wildfly #3026

ghost:
## Description of Change

add base suppressions for identifiers that reference 'wildfly' part of dependency

n/a

### DIFF
--- a/core/src/main/resources/dependencycheck-base-suppression.xml
+++ b/core/src/main/resources/dependencycheck-base-suppression.xml
@@ -4071,4 +4071,25 @@
         <packageUrl regex="true">^pkg:maven/org\.infinispan\.protostream/protostream@.*$</packageUrl>
         <cpe>cpe:/a:infinispan:infinispan</cpe>
     </suppress>
+    <suppress base="true">
+        <notes><![CDATA[
+        false positive per #3026 on io.hawtio.hawtio-wildfly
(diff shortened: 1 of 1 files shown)


core/src/main/resources/dependencycheck-base-suppression.xml
```diff
@@ -4071,4 +4071,25 @@
...
+        ]]></notes>
+        <packageUrl regex="true">^pkg:maven/io\.hawt/hawtio\-wildfly@.*$</packageUrl>
+        <cpe>cpe:/a:hawt.io:hawtio</cpe>
+    </suppress>
+    <suppress base="true">
+        <notes><![CDATA[
+        false positive per #3026 on io.hawtio.hawtio-wildfly
+        ]]></notes>
+        <packageUrl regex="true">^pkg:maven/io\.hawt/hawtio\-wildfly@.*$</packageUrl>
+        <cpe>cpe:/a:redhat:wildfly</cpe>
```
jeremylong: It is possible to collapse the CPE nodes into a single suppression rule:

```suggestion
        <cpe>cpe:/a:hawt:hawtio</cpe>
        <cpe>cpe:/a:hawt.io:hawtio</cpe>
        <cpe>cpe:/a:redhat:wildfly</cpe>
```

Issue #3057, opened 2021-01-05
False Positive on taglibs-standard-impl


Marx314:
False positive on library taglibs-standard-impl-1.2.5.jar - reported as ```cpe:2.3:a:tag_project:tag:*:*:*:*:*:*:*:*```
https://mvnrepository.com/artifact/org.apache.taglibs/taglibs-standard-impl/1.2.5

```xml
<!-- https://mvnrepository.com/artifact/org.apache.taglibs/taglibs-standard-impl -->
<dependency>
    <groupId>org.apache.taglibs</groupId>
    <artifactId>taglibs-standard-impl</artifactId>
    <version>1.2.5</version>
</dependency>
```

bretik:
The same FP is being reported on:
- tagsoup-1.2.1.jar
- jstl-api-1.2.jar

PR #3058, opened 2021-01-05
fp per #3057

Marx314:
## Description of Change

False positive per #3057

no, sorry ^^


jeremylong:
Thanks for the PR!

PR #3063, opened 2021-01-08
 feat: add Yarn Audit analyzer

### LINKED ISSUE #2641: [ERROR] Could not perform Node Audit analysis. Invalid payload submitted to Node Audit API.
**Describe the bug**
I am getting this error message:
[ERROR] Could not perform Node Audit analysis. Invalid payload submitted to Node Audit API.

The problem occurs using version X.X.X of the ____ (cli, gradle plugin, maven plugin, etc.)
I am using Jenkins plugin for dependency check version 5.3.2
I am passing the following arguments to the dependency check:
--disableAssembly  true
--disableNodeJS  true

**Log file**

**To Reproduce**

**Expected behavior**

**Additional context**

### LINKED ISSUE #2814: Yarn install compatibility
Hi, I'm scanning a NodeJS project that has several package.json files in different folders. The developers use yarn for installing dependencies instead of npm so after running yarn command the package-lock files are not created (only node_modules folder).
When I run Dependency Check I get an error message for each package.json in the project because it does not find the corresponding package-lock (that should be created after running npm).

Is there a way of not getting these exceptions in my report? I believe I'm still finding all the dependencies in the project. Yarn creates the node_modules folder and that's where D-Check finds them.

Please tell me if I'm correct. Thanks!

### LINKED ISSUE #2842: Scanning with Yarn Package Manager
Hi,

I'm new to ODC and would like understand the the correct command line parameters I should be using.

Version: Command Line / 6.0.2

The project is built using Yarn package manager. I read that ODC requires a `package-lock.json` file. The commands I'm running are:

`npm install --package-lock-only --production`
`dependency-check.sh -s "[root folder]" --disableNodeJS --disableRetireJS --nodeAuditSkipDevDependencies`

This only reports scanning 27 dependencies when the `package.json` has 52 dependencies. And the report doesn't make sense in respect to the vulnerable packages listed are not in the `package.json`

So, my question is, what commands should I be running in order to scan correctly?

Many thanks

ssthom:
## Fixes Issue #

Fixes: https://github.com/jeremylong/DependencyCheck/issues/2842
Fixes: https://github.com/jeremylong/DependencyCheck/issues/2814
Fixes: https://github.com/jeremylong/DependencyCheck/issues/2641

## Description of Change

Updated the NodeAuditAnalyzer to be able to process the yarn.lock file. It uses `yarn audit --verbose` to get the request it sends to the Node Audit API and process that as if it was from the NPM lock file. It also uses the yarn audit `groups` param to be able to handle DependencyCheck's `--nodeAuditSkipDevDependencies` param

**yes** and also tested against my local yarn project


ssthom:
@jeremylong First time developing in this repo. Let me know if this is the right approach for this?

jeremylong:
Thanks for the PR - and there are things in here that will be used but I would like to see this re-structured a bit.  When we add a reliance on an external tool we should put this into an entirely separate analyzer; so in this case like `YarnAuditAnalyzer` and possibly extracting out some of the duplicate code into a parent class for both yarn and node audit analyzers.

One of the main reasons is so we can test to see if the tool is installed and working in the `prepareFileTypeAnalyzer()` method and output an appropriate warning if it is not, such as with https://github.com/jeremylong/DependencyCheck/blob/48a6bdad7d63dd6ffb39718fd4c58c500b68dd1a/core/src/main/java/org/owasp/dependencycheck/analyzer/GolangModAnalyzer.java#L270

In this case, if no yarn.lock is in the scan then the `prepareFileTypeAnalyzer()` would not be called. If a yarn.lock existed and yarn was not installed we can throw the error during setup with sufficient instructions for an end user to know they need to install yarn to be able to complete the scan.

ssthom:
@jeremylong How about this? I separated it into it's own Analyzer and moved some of the common code into `AbstractNpmAnalyzer`

It now prints the following to the user if Yarn isn't installed
```
[WARN] The Yarn Audit Analyzer has been disabled. Yarn executable was not found.
```

jeremylong:
@ssthom initial review - this looks great. really appreciate such a complete PR.  I will do some testing over the weekend, merge the PR, and then publish a release.  You don't have to worry about this as I'll handle it (just writing it down so I have notes); update to include:
* Standardize process output https://github.com/jeremylong/DependencyCheck/pull/2979
* Update the README.md section on [development pre-reqs](https://github.com/jeremylong/DependencyCheck#development-prerequisites) to include yarn.

josefschabasser:
Hi!
Sorry to interrupt you, but will this work with the docker image (owasp/dependency-check)?
Best regards!

jeremylong:
Current branch can be found here: https://github.com/jeremylong/DependencyCheck/tree/ssthom-yarnAudit

It looks like we also need to add `yarnPath` - as such, it might take a little bit longer.  Also, we likely need to install yarn in the docker image as well.

jeremylong:
@ssthom testing is finding more issues...  As I'm using a newer version of node then you were in testing I'm getting the following from `yarn audit --offline`:

> error owasp-nodejs-goat@1.3.0: The engine "node" is incompatible with this module. Expected version "12.x.x". Got "15.6.0"
error Found incompatible module.

Issue #3073, opened 2021-01-12
False Positive on java-cfenv-boot-pivotal-scs


Marx314:
When reporting a false positive please include:
- The location of the dependency (Maven GAV, URL to download the dependency, etc.) 
  - pkg:maven/io.pivotal.cfenv/java-cfenv-boot-pivotal-scs@2.2.1.RELEASE  (Confidence:High)
- The CPE that is believed to be false positive
  - cpe:/a:pivotal_software:spring_framework
  - cpe:/a:pivotal_software:spring_boot

#### Example
False positive on library java-cfenv-boot-pivotal-scs-2.2.1.RELEASE.jar - reported as cpe:2.3:a:pivotal_software:spring_framework:2.2.1:release:*:*:*:*:*:*

PR #3074, opened 2021-01-12
Fix False positive on java-cfenv-boot-pivotal-scs

Marx314:
java-cfenv-boot-pivotal-scs is being identified as 
- spring_boot 
- spring_framework 

leading to an interesting number of false positive

## Fixes Issue #
#3073

## Description of Change

Added false positive on java-cfenv-pivotal-scs in dependencycheck-base-suppression.xml for the greater good!

*no* / cry in unit test


jeremylong:
Thanks for the PR!

PR #3089, opened 2021-01-18
yarn audit

jeremylong:


Toilal:
This breaks scanning when yarn is not available on the system.

```
[ERROR] Failed to execute goal org.owasp:dependency-check-maven:6.1.0:check (default) on project estamp: One or more exceptions occurred during dependency-check analysis: One or more exceptions occurred during analysis:
[ERROR] 	InitializationException: Unable to read yarn audit output.
[ERROR] 		caused by IOException: Cannot run program "yarn": error=2, Aucun fichier ou dossier de ce type
[ERROR] 		caused by IOException: error=2, Aucun fichier ou dossier de ce type
```

jeremylong:
@Toilal then consider disabling the yarn analyzer.

mprins:
> Then consider disabling the yarn analyzer.

isn't disabling yarn only available as of (the unreleased) v. 6.1.1 for the maven plugin? (#3105 / 478237d9d0d4dcebb7957eb067bc25a0b2593ab9 )

Toilal:
@mprins yes it was not possible in v6.1.0, at least with maven. https://github.com/jeremylong/DependencyCheck/issues/3105

dpfrederick:
@jeremylong Howdy sir. Thanks for all of your hard work on this project.

I've been looking for an explanation for the "(YarnAuditAnalyzer); the analyzer is being disabled and may result in false negatives." message that we see sometimes. It looks like this is the pr where that message was added but I don't see any links to an issue or anything like that.

Can you point me somewhere with some more info? Should we expect the yarn audit analyzer to be deprecated at some point? Does this mean that the dependency checker won't be able to handle yarn.lock files anymore?

PR #3117, opened 2021-01-30
Simplify Golang analyzer

rofafor:
## Description of Change

Simplify Golang analyzer by getting rid of golang list modules fallback.

Go modules was introduced in version 1.11, so we don't have to support any older version. The "readonly" output seems to be working for both simple and complex repositories.

Simple golang repository without vendor directory:
```
git clone https://github.com/golang/example.git
cd example
for GOVERSION in 1.11 1.12 1.13 1.14 1.15; do echo "${GOVERSION}:"; docker run -it --rm -v $(pwd):/src -w /src golang:${GOVERSION} go list -json -m -mod=readonly all; done
```

Complex golang repository with vendor directory:
```
https://github.com/kubernetes/kubernetes.git
cd kubernetes
for GOVERSION in 1.11 1.12 1.13 1.14 1.15; do echo "${GOVERSION}:"; docker run -it --rm -v $(pwd):/src -w /src golang:${GOVERSION} go list -json -m -mod=readonly all; done
```

All output looks file.

no

### DIFF
--- a/core/src/main/java/org/owasp/dependencycheck/analyzer/GolangModAnalyzer.java
+++ b/core/src/main/java/org/owasp/dependencycheck/analyzer/GolangModAnalyzer.java
@@ -175,36 +175,6 @@ public class GolangModAnalyzer extends AbstractFileTypeAnalyzer {
     }
 
     /**
-     * Launches `go list -json -m all` in the given folder.
-     *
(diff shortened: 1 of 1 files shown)


jeremylong:
Completely agree - thanks for the PR!

PR #3119, opened 2021-01-30
Fix wrong formatting strings.

wilx:
## Description of Change

The are three logging statements which clearly want to print something but fail to specify that correctly.

no


jeremylong:
Thanks for the PR!

Issue #3241, opened 2021-03-30
False positives on Spring Cloud under io.awspring namespace


settesoft:
Some of the Spring Cloud libraries have been moved under a new namespace, and hence previously defined exclusions are not taking effect for these.

One such example is:
spring-cloud-aws-autoconfigure-2.3.0.jar
Reported as: cpe:/a:pivotal_software:spring_framework:2.3.0

```xml
<dependency>
    <groupId>io.awspring.cloud</groupId>
    <artifactId>spring-cloud-aws-autoconfigure</artifactId>
    <version>2.3.0</version>
</dependency>
```

I will raise a PR similar to the below snippet, for consideration (following the pattern [here](https://github.com/jeremylong/DependencyCheck/blob/df6384c30c832e1149555e68b8615e5c00f9d528/core/src/main/resources/dependencycheck-base-suppression.xml#L2446-L2454)):
```xml
<suppress>
<notes><![CDATA[
    False positive per issue #tbc
    ]]></notes>
    <gav regex="true">^io\.awspring\.cloud:spring-cloud-.*$</gav>
    <cpe>cpe:/a:pivotal_software:spring_framework</cpe>
    <cpe>cpe:/a:pivotal_software:spring_framework</cpe>
    <cpe>cpe:/a:context_project:context</cpe>
</suppress>
```

PR #3242, opened 2021-03-30
Add suppressions for io.awspring, ref #3241

settesoft:
## Description of Change

Suppress false positives on the io.awspring libraries, since they were moved from org.springframework.

Similar to the existing suppression for org.springframework [here](https://github.com/jeremylong/DependencyCheck/blob/df6384c30c832e1149555e68b8615e5c00f9d528/core/src/main/resources/dependencycheck-base-suppression.xml#L2446-L2454), this covers those libraries that have now been moved into a new package, io.awspring

No

### DIFF
--- a/core/src/main/resources/dependencycheck-base-suppression.xml
+++ b/core/src/main/resources/dependencycheck-base-suppression.xml
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <suppressions xmlns="https://jeremylong.github.io/DependencyCheck/dependency-suppression.1.3.xsd">
-   <suppress>
+   <suppress base="true">
         <notes><![CDATA[
 	    FP per #3241
 	    ]]></notes>
--- a/core/src/main/resources/dependencycheck-base-suppression.xml
+++ b/core/src/main/resources/dependencycheck-base-suppression.xml
@@ -1,5 +1,14 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <suppressions xmlns="https://jeremylong.github.io/DependencyCheck/dependency-suppression.1.3.xsd">
+   <suppress>
+        <notes><![CDATA[
+	    FP per #3241
+	    ]]></notes>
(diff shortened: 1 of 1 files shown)


settesoft:
I don't think the build failure is due to the change I made, but let me know if there is anything I need to do to fix.

Message from "Build with Maven" stage:
`Error:  Failed to execute goal org.apache.maven.plugins:maven-invoker-plugin:3.2.2:install (integration-test) on project dependency-check-maven: Execution integration-test of goal org.apache.maven.plugins:maven-invoker-plugin:3.2.2:install failed: Plugin org.apache.maven.plugins:maven-invoker-plugin:3.2.2 or one of its dependencies could not be resolved: Failed to collect dependencies at org.apache.maven.plugins:maven-invoker-plugin:jar:3.2.2 -> org.apache.maven.shared:maven-invoker:jar:3.1.0: Failed to read artifact descriptor for org.apache.maven.shared:maven-invoker:jar:3.1.0: Could not transfer artifact org.apache.maven.shared:maven-invoker:pom:3.1.0 from/to central (https://repo.maven.apache.org/maven2): Transfer failed for https://repo.maven.apache.org/maven2/org/apache/maven/shared/maven-invoker/3.1.0/maven-invoker-3.1.0.pom: Connection timed out (Read failed) -> [Help 1]`

core/src/main/resources/dependencycheck-base-suppression.xml
```diff
@@ -1,5 +1,14 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <suppressions xmlns="https://jeremylong.github.io/DependencyCheck/dependency-suppression.1.3.xsd">
+   <suppress>
```
jeremylong: Everything in the base suppression list should have the base attribute set to true. Without this these would still show up in the suppressed section of the report.
```suggestion
   <suppress base="true">
```

jeremylong:
Thanks for the PR!

Issue #2726, opened 2020-07-17
Telerik.Web.UI.dll


roxton001:
Hello,

we are trying to use your plugin to check any 3rd party libraries for vulnerabilities.

However, we tested it on Telerik.Web.UI 2013.2.717.40 library and no vulnerability was identified despite the fact the vulnerability for this library really [exists](https://nvd.nist.gov/products/cpe/detail/663834?keyword=cpe:2.3:a:telerik:ui_for_asp.net_ajax:2013.2.717:*:*:*:*:*:*:*&status=FINAL,DEPRECATED&orderBy=CPEURI&namingFormat=2.3)

![image](https://user-images.githubusercontent.com/9051831/87769679-45922080-c81e-11ea-9717-3a13f057506f.png)

Your tool works perfectly with a newer version of the Telerik library:
![image](https://user-images.githubusercontent.com/9051831/87770258-16c87a00-c81f-11ea-888c-0fce0929a50e.png)

Can you please somebody help us with this issue? 

Thanks

jeremylong:
For the scan that does not detect Telerik correctly:
1. what version of ODC are you using?
2. Please provide the log file (from the cli just add `--log=odc.log`
3. If possible, can you scan just the Telerik control and provide the XML version of the report?
4. What version of dotnet core are you using?

roxton001:
1. Dependency-Check Core version 5.3.2
2. [log.txt](https://github.com/jeremylong/DependencyCheck/files/4937565/log.txt)
3. [dependency-check-report.zip](https://github.com/jeremylong/DependencyCheck/files/4937572/dependency-check-report.zip)
4. dotnet --version 3.1.302

jeremylong:
I do not use the Telerik controls - so I'm not sure about this. Are there separate DLLs for the base Web UI and AJAX? Or are the ajax bits embedded in the Web UI DLL?

jeremylong:
Also - could you provide the XML report for the new DLL that was identified?

roxton001:
> Also - could you provide the XML report for the new DLL that was identified?

I just realized that the tool has not found any vulnerabilities for the newer version, it just generates CPE for the file. Sorry for misleading



> I do not use the Telerik controls - so I'm not sure about this. Are there separate DLLs for the base Web UI and AJAX? Or are the ajax bits embedded in the Web UI DLL?

Will have to check and let you know

roxton001:
[...]

PR #3244, opened 2021-03-30
added additional vendor and product for telerik

itsdrewmiller:
Fixes issue #2726 

I confirmed that this matches CVE-2019-18935 now - the original code had a typo on the type=product name=vendor, but the bigger issue is that there is no CPE that matches Progress/ASP.NET Ajax.  It looks like the overall list of CPEs that should match this dll is kind of long unfortunately, but this is the critical CVE so I think most important to find.


jeremylong:
Thanks for the PR!

itsdrewmiller:
Thanks for maintaining a popular open source package!  I know it can be a thankless pain in the ass sometimes.  😅

Issue #3262, opened 2021-04-06
NPE in NvdCveUpdater.performUpdate


wilx:
**Describe the bug**
NPE in NvdCveUpdater.performUpdate.

**Version of dependency-check used**
The problem occurs using version 6.1.5 of the Maven plugin

**Log file**

```
[ERROR] Failed to execute goal org.owasp:dependency-check-maven:6.1.5:aggregate (default-cli) on project com.ca.apm.supertop: Execution default-cli of goal org.owasp:dependency-check-maven:6.1.5:aggregate failed. NullPointerException -> [Help 1]
org.apache.maven.lifecycle.LifecycleExecutionException: Failed to execute goal org.owasp:dependency-check-maven:6.1.5:aggregate (default-cli) on project com.ca.apm.supertop: Execution default-cli of goal org.owasp:dependency-check-maven:6.1.5:aggregate failed.
        at org.apache.maven.lifecycle.internal.MojoExecutor.execute(MojoExecutor.java:224)
        at org.apache.maven.lifecycle.internal.MojoExecutor.execute(MojoExecutor.java:153)
        at org.apache.maven.lifecycle.internal.MojoExecutor.execute(MojoExecutor.java:145)
        at org.apache.maven.lifecycle.internal.LifecycleModuleBuilder.buildProject(LifecycleModuleBuilder.java:108)
        at org.apache.maven.lifecycle.internal.LifecycleModuleBuilder.buildProject(LifecycleModuleBuilder.java:76)
        at org.apache.maven.lifecycle.internal.builder.singlethreaded.SingleThreadedBuilder.build(SingleThreadedBuilder.java:51)
        at org.apache.maven.lifecycle.internal.LifecycleStarter.execute(LifecycleStarter.java:116)
        at org.apache.maven.DefaultMaven.doExecute(DefaultMaven.java:361)
        at org.apache.maven.DefaultMaven.execute(DefaultMaven.java:155)
        at org.apache.maven.cli.MavenCli.execute(MavenCli.java:584)
        at org.apache.maven.cli.MavenCli.doMain(MavenCli.java:213)
        at org.apache.maven.cli.MavenCli.main(MavenCli.java:157)
        at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
        at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
[...]

PR #3263, opened 2021-04-06
Only access/await CVE Modified processing task when it was actually submitted

aikebah:
CVE Modified processing task should only be accessed/awaited when CVE Modified download task actually submitted it. Fixes issue #3262

## Description of Change
Perform a non-null check on the result of the download task to ensure that an actual ProcessingTask was submitted by the DownloadTask. Otherwise skip awaiting/checking the results of the ProcessingTask.

No, think the condition is too hard to isolate in a concise testcase (needs a failing downloadtask for CVE Modified)

### DIFF
--- a/core/src/main/java/org/owasp/dependencycheck/data/update/NvdCveUpdater.java
+++ b/core/src/main/java/org/owasp/dependencycheck/data/update/NvdCveUpdater.java
@@ -304,9 +304,11 @@ public class NvdCveUpdater implements CachedWebDataSource {
             final Future<ProcessTask> task;
             try {
                 task = modified.get();
-                final ProcessTask last = task.get();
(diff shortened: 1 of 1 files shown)


aikebah:
@jeremylong do you agree to not include a unit test? Or should we extend the unit tests to more thoroughly cover the NvdCveUpdater logic from now on.
Feels to me like quite a task to accomplish compared to the change, but a thorough coverage should've surfaced the NPE before it hit our users.

jeremylong:
I do agree the test coverage needs to improve - but I'd rather get this obvious fix merged sooner rather then hold this up...  (which is why the test case coverage is where it is today)...

jeremylong (approved):

Issue #3307, opened 2021-04-21
NullPointerException on updateVulnerabilityInsertReferences(CveDB.java:1173)


mark-senne:
Hello everyone,

I'm using gradle plugin org.owasp:dependency-check-gradle:6.1.5 
and since 23:30 CEST I always get this NPE. An hour before with the same Version it was all good.

I didn't cut the long stacktrace, hope you don't mind. 
and I hope it helps you. 

EDIT: Sometimes it needs a little bit more time for "Checking for updates and analyzing dependencies for vulnerabilities"
but the rest is all the same

[...]
00:13:24  Verifying dependencies for project briefankündigung verwaltung
00:13:24  Checking for updates and analyzing dependencies for vulnerabilities
00:13:36  The execution of the download was interrupted
00:13:36  org.owasp.dependencycheck.data.update.exception.UpdateException: The execution of the download was interrupted
[...]

EDIT2:  "./gradlew dependencyCheckPurge" did not help in this matter. Get the some result.

Wish you alle the best 
       Mark Senne

23:45:03  + ./gradlew dependencyCheckAnalyze
23:45:04  To honour the JVM settings for this build a single-use Daemon process will be forked. See https://docs.gradle.org/6.8.3/userguide/gradle_daemon.html#sec:disabling_the_daemon.
23:45:05  Daemon will be stopped at the end of the build 
23:45:11  
23:45:11  > Task :dependencyCheckAnalyze
23:45:11  Verifying dependencies for project briefankündigung verwaltung
23:45:11  Checking for updates and analyzing dependencies for vulnerabilities
23:45:13  The execution of the download was interrupted
23:45:13  org.owasp.dependencycheck.data.update.exception.UpdateException: The execution of the download was interrupted
23:45:13  	at org.owasp.dependencycheck.data.update.NvdCveUpdater.performUpdate(NvdCveUpdater.java:317)
23:45:13  	at org.owasp.dependencycheck.data.update.NvdCveUpdater.update(NvdCveUpdater.java:125)
23:45:13  	at org.owasp.dependencycheck.Engine.doUpdates(Engine.java:860)
23:45:13  	at org.owasp.dependencycheck.Engine.initializeAndUpdateDatabase(Engine.java:667)
23:45:13  	at org.owasp.dependencycheck.Engine.analyzeDependencies(Engine.java:593)
23:45:13  	at org.owasp.dependencycheck.Engine$analyzeDependencies$0.call(Unknown Source)
[...]

PR #3311, opened 2021-04-22
fixing #3307 and workaround for stevespringett/CPE-Parser#76

rpaasche:
## Fixes Issue #
some cve have a cpe23URI of null, which cause a null pointer

## Description of Change
ignore null cpe23Uri

Add workaroung for  stevespringett/CPE-Parser#76 / ignoring parser exception for now

### DIFF
--- a/core/src/main/java/org/owasp/dependencycheck/data/nvdcve/CveDB.java
+++ b/core/src/main/java/org/owasp/dependencycheck/data/nvdcve/CveDB.java
@@ -1215,7 +1215,13 @@ public final class CveDB implements AutoCloseable {
 
         try {
             cpeEntries.forEach(entry -> {
-                builder.cpe(parseCpe(entry, cve.getCve().getCVEDataMeta().getId()))
+                // TODO may remove early out after https://github.com/stevespringett/CPE-Parser/issues/76 was fixed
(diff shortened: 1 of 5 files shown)


core/src/main/java/org/owasp/dependencycheck/data/nvdcve/CveDB.java
```diff
@@ -1252,7 +1264,12 @@ private Cpe parseCpe(DefCpeMatch cpe, String cveId) throws DatabaseException {
                     throw new DatabaseException("Unable to parse CPE: " + cpe.getCpe23Uri(), ex);
                 }
             } else {
-                throw new DatabaseException("Unable to parse CPE: " + cpe.getCpe23Uri(), ex);
+                String msg = String.format("Unable to parse CPE: %s", cpe.getCpe23Uri());
```
jeremylong, line 1267: the impact of not failing here is bigger than people might think. If the NVD data is corrupted again - a near silent fail like this will corrupt the database and can only be fixed by re-building the database after the data feeds are fixed.

jeremylong (commented):
While I truly appreciate the PR - I do not plan on merging this.  The NPE issues have been resolved with other PRs. However, I disagree with, for most users, will end up being a silent failure.

If one wants to create a more resilient pipeline - one should mirror the NVD data feed (and keep backups).

Issue #3291, opened 2021-04-18
Repeated scans result in different reports


valentijnscholten:
**Describe the bug**
I have a project which consists of 12 war files, and a shared war. I scan these using the jenkins plugin using `--scan **/*.war`.

The report that is outputted changes from time to time, even when the wars are binary identical to the previous runs.

Dependency check seems to be trying to group things together, or mark some wars as shade by other wars.

What I see is that sometimes war A is reported a dependency with war B as related dependencies. But sometimes it's different or the other way around.

The differences are so big that it's hard see what exactly is going on, but my suspicion is this is caused by having no predictable ordering. (a point mentioned in #2608 - now closed).

**Version of dependency-check used**
The problem occurs using version 6.1.3 of the jenkins cli

**Log file**
I can provide private links to logs offline

**To Reproduce**
Have a project with multiple wars, generated independently of eachother
Have a job copying all the wars into a folder
Scan the wars using the (jenkins) cli with --scan '**/*.war`

**Expected behavior**
When the wars are binary identical, the reports generated by dependency check should be identical as well.

valentijnscholten:
Did some more testing. Without knowing the intention behind the "shading" it's hard to understand what is happening. It looks like when I have a project that has:

A.war, which contains shared.jar
B.war, which contains shared.jar
C.war, which contains shared.jar
shared.jar

If I scan these 4 wars (or **/*.war), it looks like shared.jargets either "grouped" (or "shaded"?) under A.war xor B.war xor C.war. Could depend on filesystem ordering or maybe last update timestamp, haven't found a pattern yet.
But depending on what gets chosen you either get all the vulnerabilities from shared.jarunder only A.war or only B.war or only C.war. And separate from that you get them also under shared.jar. 

Not sure if this is intended, but I would expect all the vulnerabilities to be reported 4 times, once under A, once under B, once under C, once under shared. Does that make sense?
The current behavior only has 2 vulnerabilities in the report which hides the 2 others. And it's unpredictable which 2 get hidden. So if you have a tool like Defect Dojo trying to ingest the findings and keep track of new/mitigated findings things go wrong as it will start flipping findings from open to closed based on the unpredictability.

~~When I remove the shared.jarfrom the scan, it all works as expected. Each vulnerability in shared.jargets outputten 3 times. Once for each of the A/B/C.wars.~~

valentijnscholten:
Just did another run, with different result again using this setup:
A.war, which contains shared.jar
[...]

PR #3317, opened 2021-04-22
use predictable ordering as much as possible

valentijnscholten:
## Fixes Issue #
#3291 

## Description of Change
As described in #3291, every run of dependency check will/could generate a different report. As long as you scan 1 artifact, this differences are limited to ordering of vulnerabilities etc. Once you start scanning multiple (2, 3, 4. ... 10) artifacts, the differences start to affect which dependencies are considered "top level" dependencies, and which dependencies are seen as "related dependencies". In the end all dependencies and vulnerabilities should still be there, but differently ordered/grouped. However in #3291 I have seen that there are bigger differences in the reports. There can as much 25% in size difference in the xml report. And when importing the report into vulnerability management tools as OWASP Defect Dojo, shows difference total vulnerability count. Analyzing these 8-10MB xml reports was hard, so in search of finding the cause of this, I started to make some changes to have a predictable ordering in the reports (and internal data structures). 
This PR realizes this, and now it turns out the reports are consistently the same across runs of the scanner. So no more difference in vulnerability count or filesize or otherwise.

## Implementation ##
The PR replaces some `HashSets` with `TreeSets` to ensure natural ordering. A comparator is added to `Dependency.java` to allow ordering. It's based on the same comparator used elsewhere outside the class (based on `getActualFilePath()`) 

## TODO ## 
The `vulnerableSoftwareIdentifiers` property was also converted to a `TreeSet`, but still appears in different order across runs. Not where this is caused. But for now I'd like to get feedback before continuing on this path.

*no*, not sure if they are needed.

### DIFF
--- a/cli/src/main/java/org/owasp/dependencycheck/App.java
+++ b/cli/src/main/java/org/owasp/dependencycheck/App.java
@@ -21,9 +21,9 @@ import java.io.File;
 import java.io.FileNotFoundException;
 import java.io.IOException;
 import java.util.ArrayList;
-import java.util.HashSet;
 import java.util.List;
 import java.util.Set;
+import java.util.TreeSet;
 
 import org.apache.commons.cli.ParseException;
 import org.apache.tools.ant.DirectoryScanner;
@@ -316,7 +316,7 @@ public class App {
      * @return returns the set of identified files
      */
     private Set<File> scanAntStylePaths(List<String> antStylePaths, int symLinkDepth, String[] excludes) {
-        final Set<File> paths = new HashSet<>();
+        final Set<File> paths = new TreeSet<>();
         for (String file : antStylePaths) {
             LOGGER.debug("Scanning {}", file);
             final DirectoryScanner scanner = new DirectoryScanner();
--- a/core/src/main/java/org/owasp/dependencycheck/Engine.java
+++ b/core/src/main/java/org/owasp/dependencycheck/Engine.java
@@ -544,6 +544,7 @@ public class Engine implements FileFilter, AutoCloseable {
                     for (Dependency existing : dependencies) {
                         if (sha1.equals(existing.getSha1sum())) {
                             if (existing.getFileName().contains(": ") || dependency.getFileName().contains(": ")) {
+                                LOGGER.warn("continue for {} and {}", dependency.getFileName(), existing.getFileName());
                                 //TODO this won't be quite right 100% of the time. Its possible that the ": " would get added later
                                 continue;
                             }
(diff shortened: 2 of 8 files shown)


valentijnscholten:
Not sure if that NullPoint while downloading NVD database is related to the changes in this PR.

valentijnscholten:
Just went looking some more in the codebase and found out that in the past there have been PRs that actually removed the TreeSets and replaced them with HashSets. And removing / changing comparators to no longer ensure ordering. 
https://github.com/jeremylong/DependencyCheck/pull/983/files

jeremylong:
I believe some of these collections were previously ordered (in addition to others). The ordering was removed due to performance issues (on these and other collections).  Doing some test runs show that given `Dependencies Scanned: 6464 (4644 unique)` using the un-ordered version I got an average runtime of 117 seconds. Using your ordered collections I got an average runtime of 124 seconds.  Given that is a large number of dependencies - and most scans are not that large there is definitely a performance hit.

I wonder if all of these need to be tree sets. My guess is that the evidence, projectReferences,  relatedDependencies, vulnerabilities, and suppressed vulnerabilities do not need to be sorted. The ordering on the paths (https://github.com/jeremylong/DependencyCheck/pull/3317/files#diff-1593f793158c7b1dc3a7bd83329962ad1226667dc5d14ca1e68eef1bf39e89bbR319) is definitely needed.

valentijnscholten:
We could probably write a book on all the different Sets and their properties :-) Even with your change and the TreeSets, I still see the reports differing between scans. Sometimes there's a bunch of extra dependencies reported (most of the time they are not reported, but they are present). Without the ordering it was impossible to see what was happening, unless doing in memory comparisons between the sets and lists between runs. 

Even if you prefer not to use the TreeSets, wouldn't you agree it would be very helpful if the report was ordered predictably? Maybe it would be OK to do a one-time ordering after the analysis or when writing the report? Make a optional parameter for it for those who want to shave of those ms from their runs?

What would be the best way to provide more info on the differences in dependencies between runs? I haven't fully grasped the model and logic of DC in my mind yet to know what's going on. These microservices have lots of identical dependencies, so I wonder if by using the file hashes or looking at parts of the file path is causing them to be lost sometimes.

valentijnscholten:
Did some quick tests and with both your and my changes together (i.e. this PR), the reports are different between runs. Already before the line with the new sorting from your change, the dependencies are different. 
Once I disable parallel processing, everything is consistent between runs (but slower of course).

jeremylong:
@valentijnscholten do you have any specific test cases you could share? I went down a slightly different path then ordering everything (yes, we can look at sorting for the reports later - but for now I'd at least like to get the related dependency problem sorted).  Take a look at #3343.

valentijnscholten:
I just tested #3343, but get different reports, different sizes:
image
Without ordering it's not possible (for me) to make a good comparison between the reports. I will send you some wars by e-mail to reproduce.

jeremylong:
I just updated my branch. Take a look at #3343. I am now getting the same output from multiple executions (with the exception of the timestamp).

valentijnscholten:
superseded by #3343 which seems to fix #3291 and ensures predictable ordering in reports :-)

PR #3324, opened 2021-04-23
Fix two Null Pointer Exceptions

sfuhrm:
## Description of Change

This introduces a validation filter for DefCpeMatch since they seem to be incomplete from time to time.

*yes*


sfuhrm:
I'm prefering the more simple PR at the moment

Issue #2859, opened 2020-10-02
False Positives on jira-rest-java-client-api-5.2.1.jar


vilvo:
When reporting a false positive please include:
- jira-rest-java-client-api-5.2.1.jar (the latest) - https://mvnrepository.com/artifact/com.atlassian.jira/jira-rest-java-client-core/5.2.1

#### Example
False positives on library jira-rest-java-client-api-5.2.1.jar
```
<suppress>
--
  |   |   | <notes><![CDATA[
  |   |   | file name: site-activity-service.jar: jira-rest-java-client-core-5.2.1.jar
  |   |   | These CVEs are related to Jira server
  |   |   | ]]></notes>
  |   |   | <packageUrl regex="true">^pkg:maven/com\.atlassian\.jira/jira\-rest\-java\-client\-core@.*$</packageUrl>
  |   |   | <cve>CVE-2020-14164</cve>
  |   |   | <cve>CVE-2020-14165</cve>
  |   |   | <cve>CVE-2020-14167</cve>
  |   |   | <cve>CVE-2020-14168</cve>
  |   |   | <cve>CVE-2020-14169</cve>
  |   |   | <cve>CVE-2020-14172</cve>
  |   |   | <cve>CVE-2020-14173</cve>
  |   |   | <cve>CVE-2020-14174</cve>
  |   |   | <cve>CVE-2020-14178</cve>
  |   |   | <cve>CVE-2020-14181</cve>
  |   |   | <cve>CVE-2020-4021</cve>
  |   |   | <cve>CVE-2020-4022</cve>
  |   |   | <cve>CVE-2020-4025</cve>
  |   |   | <cve>CVE-2020-4024</cve>
  |   |   | <cve>CVE-2020-4028</cve>
  |   |   | <cve>CVE-2020-4029</cve>
  |   |   | </suppress>
```
In addition, a lot (68) of Critical CVEs between 2014 and 2019 which are expected to be fixed in the latest client library (if relevant).

PR #3354, opened 2021-05-05
Add suppression for jira-rest-java-client-app

Xaseron:
## Fixes Issue #
Fixes the same FP as in #2859 but for the  `jira-rest-java-client-app` dependency

## Description of Change
The `jira-rest-java-client-app` bundles the `jira-rest-java-client-core` which is already in the suppression file and some runtime dependencies. https://mvnrepository.com/artifact/com.atlassian.jira/jira-rest-java-client-app/5.2.2

*yes*


jeremylong:
Thanks for the PR!

PR #3365, opened 2021-05-08
Break out of loop looking for high enough vulnerabilities as soon as …

wilx:
…possible.

## Description of Change

Single line change, added `break` into a loop.

No.


jeremylong:
Thanks for the PR!

PR #3487, opened 2021-07-01
VERSIONENDEXCLUDING database field length (50) is too short to fit value

### LINKED ISSUE #3483: CVE_CPE_STARTS_WITH_FILTER set to 'cpe:2.3:' fails at CVE-2017-3223
**Describe the bug**
I want to not only import the application (a) related CVE data from the NVD, but also the OS (o) and hardware (h) related ones.  
When the Setting `Settings.KEYS.CVE_CPE_STARTS_WITH_FILTER` is set to `cpe:2.3:`, the OS, hardware, and application related CVEs should be imported, according to the JavaDoc of the Settings Key:
```
/**
* The starts with filter used to exclude CVE entries from the database.
* By default this is set to 'cpe:2.3:a:' which limits the CVEs imported
* to just those that are related to applications. If this were set to
* just 'cpe:2.3:' the OS, hardware, and application related CVEs would
* be imported.
*/
```
But what actually happens is that the updating process fails due to the [`CVE-2017-3223`](https://nvd.nist.gov/vuln/detail/CVE-2017-3223) having a Known Affected Software Configuration that is longer than 50 characters: `dh_ipc-ack-themis_eng_p_v2.400.0000.14.r.20170713.bin`, 53 characters to be precise.  
Which means that either the 53 character long string has to be cropped to a length of 50 characters, or the maximum character length of the field `VERSIONENDEXCLUDING` has to be extended to fit longer identifiers.

Using one of the two options `cpe:2.3:a`/`cpe:2.3:h` individually works fine. Only the option `cpe:2.3:o` causes this issue to appear.

The problem occurs using version `6.1.6` of the maven plugin.

**Log file**
[The error log](https://gist.github.com/Skyball2000/6f2657b2cc80775d2b5421344f5e6d7f)

**To Reproduce**
```
Settings settings = new Settings();
settings.setString(Settings.KEYS.CVE_CPE_STARTS_WITH_FILTER, "cpe:2.3:");

engine.openDatabase();

NvdCveUpdater updater = new NvdCveUpdater();
updater.update(engine);
```

**Expected behavior**
This should mirror the NVD database with type a/o/h cpes onto the local device and not crash on a value that is too long.

**Additional context**
This is my first issue here, let me know if I missed anything!

**Edit**
This seems to happen only once with the very specific [`CVE-2017-3223`](https://nvd.nist.gov/vuln/detail/CVE-2017-3223), no other cpe/cve causes this to happen, so the best solution for now might be to simply crop the `VERSIONENDEXCLUDING` to 50 characters if it should happen to be longer.

YanWittmann:
# Fixes Issue #
https://github.com/jeremylong/DependencyCheck/issues/3483, NVD Updater would crash on Setting CVE_CPE_STARTS_WITH_FILTER set to 'cpe:2.3:'
closes #3483 

## Description of Change

Since the field `VERSIONENDEXCLUDING` is only 50 characters long, the updater fails since it attempts to insert the string `dh_ipc-ack-themis_eng_p_v2.400.0000.14.r.20170713.bin` (more details see issue #3483).

Since this only happens once for the entire database, I made it so that the string gets cropped to 50 characters if it happens to be longer than 50 characters.

I tried this out and now the updater runs fine without errors when updating with the setting `CVE_CPE_STARTS_WITH_FILTER` set to `cpe:2.3:` or more specifically `cpe:2.3:o`.

no

### DIFF
--- a/core/src/main/java/org/owasp/dependencycheck/data/nvdcve/CveDB.java
+++ b/core/src/main/java/org/owasp/dependencycheck/data/nvdcve/CveDB.java
@@ -1048,10 +1048,10 @@ public final class CveDB implements AutoCloseable {
                         cveItemConverter.extractEcosystem(baseEcosystem, parsedCpe));
 
                 addNullableStringParameter(insertSoftware, 13, ecosystem);
-                addNullableStringParameter(insertSoftware, 14, parsedCpe.getVersionEndExcluding());
-                addNullableStringParameter(insertSoftware, 15, parsedCpe.getVersionEndIncluding());
-                addNullableStringParameter(insertSoftware, 16, parsedCpe.getVersionStartExcluding());
-                addNullableStringParameter(insertSoftware, 17, parsedCpe.getVersionStartIncluding());
+                addNullableStringParameter(insertSoftware, 14, parsedCpe.getVersionEndExcluding() != null && parsedCpe.getVersionEndExcluding().length() > 50 ? parsedCpe.getVersionEndExcluding().substring(0, 50) : parsedCpe.getVersionEndExcluding());
+                addNullableStringParameter(insertSoftware, 15, parsedCpe.getVersionEndIncluding() != null && parsedCpe.getVersionEndIncluding().length() > 50 ? parsedCpe.getVersionEndIncluding().substring(0, 50) : parsedCpe.getVersionEndIncluding());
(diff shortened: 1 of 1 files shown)


jeremylong:
A better solution would be to increase the size of the column. Truncating data tends to make it invalid...

YanWittmann:
You are right, I thought that for only one affected cve this would be acceptable but in a long term your solution is better.

YanWittmann:
I'll close the PR then.

jeremylong:
It would be great if you could submit a PR - it would be something like this (https://github.com/jeremylong/DependencyCheck/commit/1627a036443c932822e9c4882d63767cae426da1) only instead of inserts you would change the column width and increment the minor schema version number to 5.2.

PR #3488, opened 2021-07-01
Set length of versionEndExcluding, versionEndIncluding, versionStartExcluding and versionStartIncluding to 60

### LINKED ISSUE #3483: CVE_CPE_STARTS_WITH_FILTER set to 'cpe:2.3:' fails at CVE-2017-3223

YanWittmann:
## Fixes Issue #
See issue https://github.com/jeremylong/DependencyCheck/issues/3483
closes #3483 

## Description of Change

Set length of columns `versionEndExcluding`, `versionEndIncluding`, `versionStartExcluding` and `versionStartIncluding` to `60`
Set schema version to `5.2`

no

### DIFF
--- a/core/src/main/resources/data/upgrade_5.0.sql
+++ b/core/src/main/resources/data/upgrade_5.0.sql
@@ -2,4 +2,4 @@ UPDATE cpeecosystemcache SET ecosystem='MULTIPLE' WHERE vendor='tensorflow' AND 
 
 UPDATE cpeecosystemcache SET ecosystem='MULTIPLE' WHERE vendor='scikit-learn' AND product='scikit-learn';
 
-UPDATE Properties SET value='5.2' WHERE ID='version';
+UPDATE Properties SET value='5.1' WHERE ID='version';
--- a/core/src/main/resources/data/upgrade_5.0.sql
+++ b/core/src/main/resources/data/upgrade_5.0.sql
@@ -2,4 +2,4 @@ UPDATE cpeecosystemcache SET ecosystem='MULTIPLE' WHERE vendor='tensorflow' AND 
 
 UPDATE cpeecosystemcache SET ecosystem='MULTIPLE' WHERE vendor='scikit-learn' AND product='scikit-learn';
 
-UPDATE Properties SET value='5.1' WHERE ID='version';
+UPDATE Properties SET value='5.2' WHERE ID='version';
--- a/core/src/main/resources/data/initialize.sql
+++ b/core/src/main/resources/data/initialize.sql
@@ -28,8 +28,8 @@ CREATE TABLE cpeEntry (id INT auto_increment PRIMARY KEY, part CHAR(1), vendor V
 version VARCHAR(255), update_version VARCHAR(255), edition VARCHAR(255), lang VARCHAR(20), sw_edition VARCHAR(255), 
 target_sw VARCHAR(255), target_hw VARCHAR(255), other VARCHAR(255), ecosystem VARCHAR(255));
 
-CREATE TABLE software (cveid INT, cpeEntryId INT, versionEndExcluding VARCHAR(50), versionEndIncluding VARCHAR(50), 
-                       versionStartExcluding VARCHAR(50), versionStartIncluding VARCHAR(50), vulnerable BOOLEAN
+CREATE TABLE software (cveid INT, cpeEntryId INT, versionEndExcluding VARCHAR(60), versionEndIncluding VARCHAR(60),
+                       versionStartExcluding VARCHAR(60), versionStartIncluding VARCHAR(60), vulnerable BOOLEAN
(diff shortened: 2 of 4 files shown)


core/src/main/resources/data/upgrade_5.0.sql
```diff
@@ -2,4 +2,4 @@ UPDATE cpeecosystemcache SET ecosystem='MULTIPLE' WHERE vendor='tensorflow' AND
 
 UPDATE cpeecosystemcache SET ecosystem='MULTIPLE' WHERE vendor='scikit-learn' AND product='scikit-learn';
 
-UPDATE Properties SET value='5.1' WHERE ID='version';
\ No newline at end of file
+UPDATE Properties SET value='5.2' WHERE ID='version';
```
jeremylong: The upgrade_5.0.sql should not be modified. Instead, a new file `upgrade_5.1.sql` should be created that contains the upgrade steps for a schema to go from 5.1 to 5.2.

For this PR we should revert the change:
```suggestion
UPDATE Properties SET value='5.1' WHERE ID='version';
```

Then create the new file with something like (note, I haven't tested this SQL yet): 
```sql
ALTER TABLE software ALTER COLUMN versionEndExcluding SET DATA TYPE VARCHAR(60);
ALTER TABLE software ALTER COLUMN versionEndIncluding SET DATA TYPE VARCHAR(60);
ALTER TABLE software ALTER COLUMN versionStartExcluding SET DATA TYPE VARCHAR(60);
ALTER TABLE software ALTER COLUMN versionStartIncluding SET DATA TYPE VARCHAR(60);

UPDATE Properties SET value='5.2' WHERE ID='version'; 
```

jeremylong (changes requested):

core/src/main/resources/data/upgrade_5.0.sql
YanWittmann: That makes sense. I reverted the change to `upgrade_5.0.sql` and created the new file `upgrade_5.1.sql` according to your code.

YanWittmann:
Should I also revert the other file change then? I previously also changed `core/src/main/resources/data/initialize.sql` to a lenght of `60`:
```
CREATE TABLE software (cveid INT, cpeEntryId INT, versionEndExcluding VARCHAR(60), versionEndIncluding VARCHAR(60),
versionStartExcluding VARCHAR(60), versionStartIncluding VARCHAR(60), vulnerable BOOLEAN
```
But if the table gets upgraded by the sql instructions in `upgrade_5.1.sql` later on, shouldn't this be `50`?

jeremylong (approved):
Thanks for updating the PR! This will be merged before the next release.

jeremylong (approved):

Issue #3506, opened 2021-07-12
I would like the Node Package Analyzer to be able to skip dev dependencies. 


Aeyaa:
In a project I work on, we frequently fail security scans due to dev only dependencies.  

**Describe the solution you'd like**
I am aware of the --nodeAuditSkipDevDependencies flag, and I would like a similar option to be available for the Node Package analyzer. 

**Additional context**
Sometimes the node package analyzer seems to pick up CVEs that the Node Audit analyzer does not, which makes me uncomfortable disabling the Node Package analyzer all together. Instead, I would just like to be able to skip dev dependencies in the Node Package analyzer.

jeremylong:
The node package analyzer is planned for retirement as it is today. Some of its functionality will be folded into the node audit analyzer.

tmuldoon:
@jeremylong Can you expand on your statement that the node package analyzer is planned for retirement? We have built a framework that depends upon the CVE id format and we'd prefer to use this analyzer over the node audit analyzer (which, as you know, uses a very different id format). So, why retire the n.p. analyzer?

PS: Thank you!! The tool is exceptional.

jeremylong:
Some of the code from the node package analyzer will be incorporated into the node audit analyzer. So in the future the CVE data will be included with the info from Node Audit. Note, by not using the node audit analyzer you are likely going to have false negatives as the database used by `npm audit` has non-CVE related vulnerabilities.

PR #3507, opened 2021-07-12
Add --nodePackageSkipDevDependencies CLI option

Aeyaa:
Add an option '--nodePackageSkipDevDependencies' to skip dev
dependencies in the Node Package analyzer.

## Fixes Issue #

#3506 

## Description of Change

Add an option '--nodePackageSkipDevDependencies' to skip dev
dependencies in the Node Package analyzer.

When analyzing the the package-lock.json in the Node Package analyzer, if the dependency has the field `{..., "dev": true} ` and '--nodePackageSkipDevDependencies' option was used, then skip the analysis for this dependency. 

no


core/src/main/java/org/owasp/dependencycheck/analyzer/NodePackageAnalyzer.java
```diff
@@ -337,6 +337,12 @@ private void processDependencies(JsonObject json, File baseDir, File rootFile,
                 final String name = entry.getKey();
                 final String version;
                 boolean optional = false;
+                boolean skipDev = getSettings().getBoolean(Settings.KEYS.ANALYZER_NODE_PACKAGE_SKIPDEV, false);
+                boolean isDev = entry.getValue().asJsonObject().getBoolean("dev", false);
+
+                if (isDev && skipDev) {
+                    return;
```
jeremylong: Shouldn't this be continue instead of return?

```suggestion
                    continue;
```

jeremylong (changes requested):
While I'm okay adding this for now - the longer term solution is going to be folding the node package analyzer and node audit analyzer into a single analyzer. Just spotted one possible issue with the PR though.

core/src/main/java/org/owasp/dependencycheck/analyzer/NodePackageAnalyzer.java
Aeyaa: You are absolutely correct. It should be continue 🙂

jeremylong (approved):

jeremylong:
This was merged manually.

PR #3515, opened 2021-07-14
Issue 3494

aikebah:
## Description of Change

Fix the condition for the hints for spring-security so that it actually fires for file-based scans.

no


aikebah:
Converted to draft as further issue comments hint that we need more fixes for other Spring libraries that also do  not get the vmware vendor added in.

aikebah:
Abandoning this attempt as I've identified a root-cause that is somewhere else

Issue #3521, opened 2021-07-19
Additional OssIndexAnalyzer config options to prevent HTTP 429 responses on large scans.


RasmusOlesen:
We are currently scanning +10k dependencies (Dependencies Scanned: 11447 (11438 unique))
When the oss_cache directory is cleared or expires, we always run into "To Many Requests" errors from the OssIndexAnalyzer.
We then need to run the scan multiple times, getting the same error, until the oss_cache is fully populated again.

I have a couple of ideas on how to fix this:

**Add a batchSize config option: analyzer.ossindex.batchsize**
The default batchsize in org.sonatype.ossindex.service.client.OssindexClientConfiguration is set to 128.
Being able to increase this value might fix the problem, since there will be fewer requests to https://ossindex.sonatype.org/, and there will be longer intervals between them (depending on how the amount of requests over time is calculated).

**Configurable request delay: analyzer.ossindex.request.delay**
A full scan of this size already takes +20 minutes. If adding a 3s or 5s delay between each requst to https://ossindex.sonatype.org/ can prevent the scan from terminating, that would be a great option in my opinion.

**Configurable request retry count: analyzer.ossindex.request.retry.count**
I don't like this options since it just "brute forces" the necessary requests.

jeremylong:
PRs accepted.

PR #3528, opened 2021-07-20
#3521 New OssIndexAnalyzer config options

RasmusOlesen:
## Description of Change
Added two new config options  and request delay implementation to the OssIndexAnalyzer.

analyzer.ossindex.batch.size
Which maps to the corresponding config option in org.sonatype.ossindex.service.client.OssindexClientConfiguration

analyzer.ossindex.request.delay
Which delayes each request to the OssIndex Rest API by the configured amount of seconds.
If the configuration is present.

no

### DIFF
--- a/core/src/main/java/org/owasp/dependencycheck/analyzer/OssIndexAnalyzer.java
+++ b/core/src/main/java/org/owasp/dependencycheck/analyzer/OssIndexAnalyzer.java
@@ -50,6 +50,7 @@ import java.util.Arrays;
 import java.util.Collections;
 import java.util.List;
 import java.util.Map;
+import java.util.concurrent.TimeUnit;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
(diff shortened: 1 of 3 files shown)


jeremylong:
Have you tested these changes in your environment and did they solve your issue?  I'm asking because ODC spins up several copies of the same analyzer and runs them in parallel (this can be disabled per analyzer in code). As such, the "batch" size unless I'm mistaken will only be 1 or 2 (depending on the number of package URLs the dependency has).

The whole analyzer could be re-worked to not run in parallel - collect all the package URLs for all the dependencies and then submit them in batches.

RasmusOlesen:
I am working on linux so i never managed to get the project to build properly.
Kept getting GO and .NET related errors, when running maven.

But when debugging the behavior on the current Maven Dependency Check version (https://mvnrepository.com/artifact/org.owasp/dependency-check-maven/6.2.2),
it got the impression that the default 128 batch size was used,
and that the synchronized (FETCH_MUTIX) block (OssIndexAnalyzer line 136)
was preventing the requests from running parallel.

Which is why i suggested these changes in the first place.

jeremylong (approved):

jeremylong (approved):

PR #3532, opened 2021-07-21
Allow outputDirectory to be set from a user-property on the commandline

djBo:
## Fixes Issue #
1686

## Description of Change
Allow outputDirectory to be set from a user-property on the commandline using -DoutputDirectory

*no*


jeremylong (approved):

aikebah:
@jeremylong Not sure whether we should use this MR unchanged.
It has at least (potentially more cases) a propertyname clash with a default maven plugin [DependencyPlugin](https://maven.apache.org/plugins/maven-dependency-plugin/copy-dependencies-mojo.html#outputDirectory) which uses it to redirect the downloaded artifacts to a specific folder, which means it might trigger unwanted side-effects.
As a general rule I would say that a userProperty of a plugin should use a prefix to avoid such clashes. (e.g. dependencycheck.outputDirectory for this case)

PR #3540, opened 2021-07-25
Spring vendor fixes

aikebah:
## Description of Change

Fix JarAnalyzer to make existing base-hints match
Add base-hint to make gradle plugin match for Spring-projects that use gradle for the build
Update suppressions to avoid false positives with the improved vendor-assignment for Spring.io binaries
Fix an erroneous base-hint (Spring-security is not part of Spring-framework)

no

### DIFF
--- a/core/src/main/java/org/owasp/dependencycheck/analyzer/JarAnalyzer.java
+++ b/core/src/main/java/org/owasp/dependencycheck/analyzer/JarAnalyzer.java
@@ -606,9 +606,6 @@ public class JarAnalyzer extends AbstractFileTypeAnalyzer {
         }
 
         final String originalGroupID = groupid;
-        if (groupid != null && (groupid.startsWith("org.") || groupid.startsWith("com."))) {
-            groupid = groupid.substring(4);
-        }
 
         if ((artifactid == null || artifactid.isEmpty()) && parentArtifactId != null && !parentArtifactId.isEmpty()) {
             artifactid = parentArtifactId;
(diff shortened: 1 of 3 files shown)


aikebah:
Appears to me that the 629-jackson-dataformat failure is a failure that should've occurred for all PRs since 20 Jan 2021 as per the NIST NVD modification [Modified Analysis 1/19/2021 10:41:56 AM](https://nvd.nist.gov/vuln/detail/CVE-2018-1000873#VulnChangeHistorySection)

CVE-2018-1000873 changed from `cpe:2.3:a:fasterxml:jackson-databind` to `cpe:2.3:a:fasterxml:jackson-modules-java8`

aikebah:
Same holds for the 1512, which is using the same CVE for validation

jeremylong (dismissed):

aikebah:
Base hints contained a mixture of 'proper' full pom groupIds (that would be added as evidences by the MavenAnalyzer) and truncated pom groupIds (that would be added by the unmodified JarAnalyzer.
One of these was for jackson-core adding the java8 product as evidence which triggers the detection of CVE-2018-1000873 (as per the hint added in aa2ed52ec76a9c879cafc3d1fda790d094e437a2). Looking at the details that appears inappropriate, for which I'll file a separate issue to discuss it.
For this PR I'll put the appropriate com/org prefixes in the hints where they were left out to make them properly match again with the evidences gathered by Maven Plugin, Central-/Nexus- and Jar-Analyzer.

jeremylong (approved):

jeremylong:
Build passes on my machine. We may need to disable the node audit analyzer tests for a more stable build - possibly setup a new set of integration tests specifically to cover when ODC calls out to 3rd party services.

PR #3466, opened 2021-06-23
Treeset performance improvements

mfoo:
## Fixes Issue #
https://github.com/jeremylong/DependencyCheck/issues/3454, branched off of the `v6.2.2` tag.

These changes cut just under 20% off of my analysis time.

## Description of Change

Some of the instance variables in the `Dependency` class are already `TreeSet<Identifier>` objects. In a few places we return `Collections.immutableSet(new TreeSet(thingThatIsAlreadyATreeSet))`. The `new TreeSet(...)` call causes the contents of one `TreeSet` to be read into another one at a time, evaluating the `compareTo` method of the `Identifier` subclass to make sure the new `TreeSet` is in order.

The tests all pass locally but I am not certain if this behaviour is intended and that we want to return a copy of the set rather than an immutable wrapper of the set so please examine carefully.

## Have test cases been added to cover the new functionality?

No - open to suggestions, but the current test doesn't enforce that the collection is immutable or that it's a copy so I'm unsure of the intended behaviour.

PR #3561, opened 2021-08-04
Fix ConcurrentModificationException

jeremylong:
Initial fix for ConcurrentModificationException introduced by #3466.

### DIFF
--- a/core/src/main/java/org/owasp/dependencycheck/analyzer/FalsePositiveAnalyzer.java
+++ b/core/src/main/java/org/owasp/dependencycheck/analyzer/FalsePositiveAnalyzer.java
@@ -271,7 +271,7 @@ public class FalsePositiveAnalyzer extends AbstractAnalyzer {
      */
     protected void removeBadMatches(Dependency dependency) {
 
-        Set<Identifier> toRemove = new HashSet<>();
+        final Set<Identifier> toRemove = new HashSet<>();
         /* TODO - can we utilize the pom's groupid and artifactId to filter??? most of
          * these are due to low quality data.  Other idea would be to say any CPE
          * found based on LOW confidence evidence should have a different CPE type? (this
--- a/core/src/main/java/org/owasp/dependencycheck/analyzer/FalsePositiveAnalyzer.java
+++ b/core/src/main/java/org/owasp/dependencycheck/analyzer/FalsePositiveAnalyzer.java
@@ -170,7 +170,7 @@ public class FalsePositiveAnalyzer extends AbstractAnalyzer {
                     removalSet.add(i);
                 }
             }
-            removalSet.forEach((i) -> dependency.removeVulnerableSoftwareIdentifier(i));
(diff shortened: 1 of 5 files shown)


core/src/main/java/org/owasp/dependencycheck/dependency/Dependency.java
```diff
@@ -746,6 +746,13 @@ public synchronized void removeVulnerability(Vulnerability v) {
         return Collections.unmodifiableSet(relatedDependencies);
     }
 
+    /**
+     * Clears the {@link #relatedDependencies}.      *
```
jeremylong:
```suggestion
     * Clears the {@link #relatedDependencies}.
```

aikebah:
Same fix is needed for DependencyMergingAnalyzer at lines [147-149](https://github.com/jeremylong/DependencyCheck/blob/6832c6434cf60df146f0ff6704c4c082cee13519/core/src/main/java/org/owasp/dependencycheck/analyzer/DependencyMergingAnalyzer.java#L147-L149).

aikebah:
Sorry... overlooked that file in the changelog... It's already in there. So that leaves only FalsePostiveAnalyzer.removeBadMatches which removes items from vulnerableSoftwareIdentifiers while while iterating over them

```
exception: java.util.ConcurrentModificationException

java.util.TreeMap$PrivateEntryIterator.nextEntry(TreeMap.java:1211)
java.util.TreeMap$KeyIterator.next(TreeMap.java:1265)
java.util.Collections$UnmodifiableCollection$1.next(Collections.java:1044)
org.owasp.dependencycheck.analyzer.FalsePositiveAnalyzer.removeBadMatches(FalsePositiveAnalyzer.java:281)
```

aikebah (changes requested):
These look good to me, and also get rid of a [potentially dangerous state-modifying peek](https://4comprehension.com/idiomatic-peeking/) that was signalled by SonarLint.

When a [FalsePositiveAnalyzer#removeBadMatches](https://github.com/jeremylong/DependencyCheck/blob/6832c6434cf60df146f0ff6704c4c082cee13519/core/src/main/java/org/owasp/dependencycheck/analyzer/FalsePositiveAnalyzer.java#L281) fix gets added as far as I can see no further ConcurrentModifcationExceptions should happen (all other uses of the Set-getters appear to be read-only usage, already employ a defensive copy or imlement a 'gather' phase followed by an 'update' phase in the code)

aikebah (dismissed):
LGTM

core/src/main/java/org/owasp/dependencycheck/analyzer/FalsePositiveAnalyzer.java
```diff
@@ -271,6 +271,7 @@ private void removeJreEntries(Dependency dependency) {
      */
     protected void removeBadMatches(Dependency dependency) {
 
+        Set<Identifier> toRemove = new HashSet<>();
```
jeremylong:
```suggestion
        final Set<Identifier> toRemove = new HashSet<>();
```

PR #3582, opened 2021-08-11
Add engine.removeAll(Collection<Dependency>)

jeremylong:
Minor update to use removeAll when removing dependencies from the engine's collection.


jeremylong:
We cannot change to using removeAll - as in some cases if the same dependency was added twice and we only want to remove it once `removeAll` will remove both instances.

PR #3630, opened 2021-09-05
3620 improve doc on experimental analyzers

### LINKED ISSUE #3620: analyzer.swift.package.resolved.enabled=true is missing from conf
Noticed that analyzer.swift.package.resolved.enabled=true is missing from dependencycheck.properties but doc says it is enabled. Is this a mistake?

https://github.com/jeremylong/DependencyCheck/blob/main/core/src/main/resources/dependencycheck.properties
https://github.com/jeremylong/DependencyCheck/blob/125751267e5b8775c08451862fe1595c15493e3c/ant/src/site/markdown/configuration.md

PS
Took the project down, added this config and compiled the project but have some run time issues so not able to test my self. Access to https://nvd.nist.gov/feeds/json/cve/1.1/nvdcve-1.1-modified.meta fails with some cert issues

nhumblot:
## Description of Change

Fix #3620 

The goal of this PR is to :
* Make it more explicit that experimental analyzers property must be enabled for an experimental analyzer to run
* Enforce analyzers keys are all listed in the `dependencycheck.properties` file to prevent any misunderstanding.

I did not updated the cli documentation because the options are disabling analyzers instead of enabling them. I thought the misunderstanding was not present in this case. Another reason is also that I couldn't come with a short but clear & explicit sentence like the other tools.

Any suggestion is welcomed. :slightly_smiling_face: 

*yes*

### DIFF
--- a/core/src/test/java/org/owasp/dependencycheck/analyzer/DependencyCheckPropertiesTest.java
+++ b/core/src/test/java/org/owasp/dependencycheck/analyzer/DependencyCheckPropertiesTest.java
@@ -1,6 +1,5 @@
 package org.owasp.dependencycheck.analyzer;
 
-import org.jetbrains.annotations.NotNull;
 import org.junit.Assert;
 import org.junit.Test;
 
@@ -130,7 +129,6 @@ public class DependencyCheckPropertiesTest {
         return Collections.emptySet();
     }
 
-    @NotNull
     private Set<Class<?>> tryGetClasses(String packageName, BufferedReader reader) {
         return reader.lines()
                 .filter(line -> line.endsWith(".class"))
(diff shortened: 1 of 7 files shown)


core/src/test/java/org/owasp/dependencycheck/analyzer/DependencyCheckPropertiesTest.java
```diff
@@ -0,0 +1,153 @@
...
+
+        Class<?> superclass = clazz.getSuperclass();
+        while (superclass != Object.class) {
+            if (extendsAbstractAnalyzer(superclass)) {
+                return true;
+            }
+            superclass = superclass.getSuperclass();
+        }
+
+        return false;
```
aikebah: Can be coded more concise as

return AbstractAnalyzer.class.isAssignableFrom(clazz)

nhumblot: Thank you! I updated the code. :slightly_smiling_face:

nhumblot: Looks like one job got stuck and cancelled. Is it possible to relaunch it please? :pray:

jeremylong (approved):

PR #3655, opened 2021-09-17
Include external references from OSS Index API response in the report

kapseliboi:
## Fixes Issue #
DependencyCheck report is missing references that are visible on the OSS Index vulnerability page

## Description of Change

Add externalReferences reported by OSS Index to the DependencyCheck report as well.
Relevant links:
https://github.com/OSSIndex/vulns/issues/193
https://javadoc.io/doc/org.sonatype.ossindex/ossindex-service-api/latest/index.html

*no*

### DIFF
--- a/core/src/main/java/org/owasp/dependencycheck/analyzer/OssIndexAnalyzer.java
+++ b/core/src/main/java/org/owasp/dependencycheck/analyzer/OssIndexAnalyzer.java
@@ -339,7 +339,7 @@ public class OssIndexAnalyzer extends AbstractAnalyzer {
 
         // generate references to other references reported by OSS Index
         for (final String externalReference : source.getExternalReferences()) {
-            result.addReference("MISC", null, externalReference);
+            result.addReference("OSSIndex", externalReference, externalReference);
         }
 
         // attach vulnerable software details as best we can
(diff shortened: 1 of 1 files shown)


core/src/main/java/org/owasp/dependencycheck/analyzer/OssIndexAnalyzer.java
```diff
@@ -337,6 +337,11 @@ private Vulnerability transform(final ComponentReport report, final ComponentRep
         // generate a reference to the vulnerability details on OSS Index
         result.addReference(REFERENCE_TYPE, source.getTitle(), source.getReference().toString());
 
+        // generate references to other references reported by OSS Index
+        for (final String externalReference : source.getExternalReferences()) {
+            result.addReference("MISC", null, externalReference);
```
kapseliboi: I have no clue whether or not "MISC", null is a good combo.

jeremylong: From ODC's standpoint the source of the external reference would be the OSS Index. We can't use `null` for the name as that would impact the HTML report - so the best option would like be:
```suggestion
            result.addReference("OSSIndex", externalReference, externalReference);
```

core/src/main/java/org/owasp/dependencycheck/analyzer/OssIndexAnalyzer.java
```diff
@@ -337,6 +337,11 @@ private Vulnerability transform(final ComponentReport report, final ComponentRep
         // generate a reference to the vulnerability details on OSS Index
         result.addReference(REFERENCE_TYPE, source.getTitle(), source.getReference().toString());
 
+        // generate references to other references reported by OSS Index
+        for (final String externalReference : source.getExternalReferences()) {
+            result.addReference("OSSIndex", externalReference, externalReference);
```
jeremylong:
```suggestion
            result.addReference("OSSIndex", externalReference.toString(), externalReference.toString());
```

PR #3677, opened 2021-09-29
Indicate whether any updates were actually made 

Jonpez2:
## Description of Change
Adds an indication as to whether updates actually happened

no - where would be best to add it?


Jonpez2:
@jeremylong - Any interest in this patch?  Can I do some work to make it palatable to you please?
Thanks!

Jonpez2:
Nice thank you!
What’s the release process like? Can I possibly do something helpful there?

Thanks so much for your help - it’s a great product you’ve built here!

Issue #1416, opened 2018-08-03
feature request: support pnpm and yarn projects


DanielRuf:
pnpm uses a shrinkwrap.yaml file and yarn uses yarn.lock.

Both are different than the lockfile of npm, so it might be good to support them.

jeremylong:
The issue is that in a lot of ways - the security analysis of npm is just a wrapper around `nsp`. If one was strictly using node, I would recommend just using `nsp` or `npm audit`. ODC is useful in the case of multi-language projects or for organizations that wanted to have a single format that could be imported more easily into dependency-track or a vulnerability management platform.

DanielRuf:
> The issue is that in a lot of ways - the security analysis of npm is just a wrapper around nsp

Not anymore as `nsp` is deprecated and they use https://github.com/nodejs/security-wg and other resources now.

DanielRuf:
See https://www.npmjs.com/package/nsp

npmjs bought nsp.

jeremylong:
We know - and yet today, the dependency-check analysis is using `nsp` which will be around through Sept, 2018.

DanielRuf:
https://github.com/yarnpkg/yarn/issues/5808#issuecomment-409468136
https://github.com/yarnpkg/yarn/issues/5808#issuecomment-410171673
https://github.com/npm/npm-audit-report/

DanielRuf:
> We know - and yet today, the dependency-check analysis is using nsp which will be around through Sept, 2018.

Which is not so far away ;-)
Is there a PR for the migration to the audit API or other endpoints?

jeremylong:
@stevespringett any progress on the changing the endpoints for the NspAnalyzer?

stevespringett:
@jeremylong I'll be creating a branch for this soon as the commit is going to be decent size (20+ files modified already). I should have something posted later this week. After we replace NSP with NPM Audit however, we'll have to continue to make changes since NPM Audit, to this day, still has a requirement on NSP. So its a bit of a moving target that we'll have to stay on top of.

[...]

PR #3700, opened 2021-10-05
Add Pnpm AuditAnalyzer

iagotomas:
## Fixes Issue #

feature request: support pnpm and yarn projects #1416

## Description of Change

*Add a new AuditAnalayzer compatible with Pnpm projects*

*yes*


iagotomas (commented):
@jeremylong Could you have a look at the PR? It follows Yarn implementation, hope I didn't miss anything. Will be glad to hear your feedback :)

jeremylong:
Looks good - any chance you would be willing to update the site documentation as well?

https://github.com/jeremylong/DependencyCheck/blob/main/ant/src/site/markdown/configuration.md
https://github.com/jeremylong/DependencyCheck/blob/main/maven/src/site/markdown/configuration.md
https://github.com/jeremylong/DependencyCheck/blob/main/cli/src/site/markdown/arguments.md

I'll get the arguments updated in the gradle plugin with the next release.

iagotomas:
Sure, didn't notice these, thanks for pointing.

iagotomas:
Hi @jeremylong, I fixed the doc as you suggested and conducted some more tests which raised an issue with ProcessReader, not sure if you found it before. The inputstream provided by ProcessReader is topped at 64k and the audit report generated by the npm audit API may exceed that size. To workaround the stdout is redirected to a file and after read to process the full audit report.

core/src/main/java/org/owasp/dependencycheck/analyzer/PnpmAuditAnalyzer.java
```diff
@@ -0,0 +1,288 @@
...
+            args.add(getPnpm());
+            args.add("audit");
+            if (skipDevDependencies) {
+                args.add("--prod");
+            }
+            // pnpm audit returns a json compliant with NpmAuditParser
+            args.add("--json");
+            // ensure we are using the right registry despite .npmrc
+            args.add("--registry");
+            args.add("https://registry.npmjs.org/");
```
jeremylong, line 211: Should `registry` be a configurable setting (possibly for npm and yarn also)?

jeremylong (approved):
Even if the registry should be configurable - we can approve and merge this now. Then add the configuration in a later release.

jeremylong:
It looks like there are test failures:

https://github.com/jeremylong/DependencyCheck/runs/3983551519?check_suite_focus=true#step:5:2124

jeremylong:
Looks like the error is because `pnpm` is not installed on the build system. We likely need to update the workflow to install pnpm.

jeremylong:
Setting up pnpm should be simple with https://github.com/pnpm/action-setup

jeremylong:
I tried to just add the workflow update via #3755 - however, it appears you did not checking one of the test files (`pnpmaudit/pnpm-audit.json`)?

https://github.com/jeremylong/DependencyCheck/pull/3700/files#diff-bf1b785fb925d2d59a7104b2593a7aedbf3bbfcb81c11e58d9259d0ae6a14ce8R30

core/src/main/java/org/owasp/dependencycheck/analyzer/PnpmAuditAnalyzer.java
iagotomas: Think would be great to have a common registry setting for all, in our company we use them all and plan to use owasp, we only missed pnpm :)

iagotomas:
Sorry missed the file, just added, hope it works now

jeremylong:
Merged via #3755

PR #3702, opened 2021-10-06
Work-around for MSHARED-998 until it's resolved and we can upgrade to a newer transfer-utils

aikebah:
## Description of Change

As a temporary work-around do not apply a filter to artifact-resolution when resolving for a DependencyNode that uses a `<classifier>` which would lead to a NullPointerException deep down in maven's internal logic.

yes, an integration testcase has been added that triggers the error with the old codebase and succeeds with the workaround.

### DIFF
--- a/maven/src/it/3679-classifier-in-dependency/invoker.properties
+++ b/maven/src/it/3679-classifier-in-dependency/invoker.properties
@@ -1,0 +1,19 @@
+#
+# This file is part of dependency-check-maven.
+#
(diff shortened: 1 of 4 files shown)


jeremylong (approved):
I was thinking we would have to use both resolution mechanisms until the upstream bug is resolved. PR looks good - thanks!

jeremylong:
I will try and release this on Friday morning - unfortunately, I can't release before then due to other time commitments (and the next release will also require some updates on the homebrew side)

PR #3711, opened 2021-10-07
Do a null-check for a list that in practice appears to be Nullable

aikebah:
## Description of Change

Add a null-check for a field that in practice has been shown to be `@Nullable` despite not being annotated as such in a library that marks other properties clearly as `@Nullable`

No, as even the original case by now no longer appears to fire a test-case cannot be made that would reliably have a null-valued externalReferences


aikebah:
Better to upgrade the API library instead (just released in response to my ticket in sonatype ossindex-public)

Issue #3710, opened 2021-10-07
Unstable connection to nist database


smiklosovic:
Hi @jeremylong ,

I am from Apache Cassandra project and we would like to use Ant plugin as part of our build pipeline.

It works but every now and then, we are getting this error and it makes it pretty unstable to run the builds.

Are we doing something wrong or nist is unstable?

This is our setup: https://github.com/apache/cassandra/blob/trunk/.build/build-owasp.xml

Thanks

/home/jenkins/jenkins-slave/workspace/Cassandra-devbranch-artifacts/jdk/jdk_1.8_latest/label/cassandra/.build/build-owasp.xml:80: One or more exceptions occurred during analysis:
15:14:52 org.owasp.dependencycheck.exception.ExceptionCollection: One or more exceptions occurred during analysis:
15:14:52 	UpdateException: Unable to download meta file: https://nvd.nist.gov/feeds/json/cve/1.1/nvdcve-1.1-modified.meta
15:14:52 		caused by DownloadFailedException: Download failed, unable to retrieve 'https://nvd.nist.gov/feeds/json/cve/1.1/nvdcve-1.1-modified.meta'; Error downloading file https://nvd.nist.gov/feeds/json/cve/1.1/nvdcve-1.1-modified.meta; unable to connect.
15:14:52 		caused by DownloadFailedException: Error downloading file https://nvd.nist.gov/feeds/json/cve/1.1/nvdcve-1.1-modified.meta; unable to connect.
15:14:52 		caused by DownloadFailedException: Error retrieving https://nvd.nist.gov/feeds/json/cve/1.1/nvdcve-1.1-modified.meta; received response code 503; Service Unavailable
15:14:52 	NoDataException: No documents exist
15:14:52 	at org.owasp.dependencycheck.Engine.throwFatalExceptionCollection(Engine.java:1103)
15:14:52 	at org.owasp.dependencycheck.Engine.analyzeDependencies(Engine.java:612)
15:14:52 	at org.owasp.dependencycheck.taskdefs.Check.callExecuteAnalysis(Check.java:1844)
15:14:52 	at org.owasp.dependencycheck.taskdefs.Check.execute(Check.java:1801)
15:14:52 	at org.apache.tools.ant.UnknownElement.execute(UnknownElement.java:292)
15:14:52 	at sun.reflect.GeneratedMethodAccessor4.invoke(Unknown Source)
[...]

PR #3722, opened 2021-10-11
CVE Download Updates

jeremylong:
1. Make the CVE start year configurable.
2. Add NVD CVE cache so if some downloads work but the update ultimate fails - re-running the update will use the cached data.
3. Add 3 attempts to download JSON files from the NVD with a wait between each attempt.


jeremylong:
This change plus the wait timeout should resolve #3710.

PR #3755, opened 2021-10-23
Pnpm workflow

jeremylong:
## Supersedes #3700

Adds `pnpm` to the workflows so the test cases in #3700 pass during testing.


jeremylong:
Fails due to missing file from original PR. See https://github.com/jeremylong/DependencyCheck/pull/3700#issuecomment-950139305

Issue #3766, opened 2021-10-27
False Positive on protostream-types-4.2.3.Final


ghost:
False positive on library protostream-types-4.3.2.Final.jar 

- reported as cpe:2.3:a:infinispan:infinispan:4.3.2:*:*:*:*:*:*:*

<dependency>
        <groupId>org.infinispan.protostream</groupId>
        <artifactId>protostream-types</artifactId>
        <version>4.3.2.Final</version>
      </dependency>

Showing no known issues:

https://ossindex.sonatype.org/component/pkg:maven/org.infinispan.protostream/protostream-types@4.3.2.Final

The CVEs listed are all for Infinispan itself, not this specific library.

This is another instance of:

https://github.com/jeremylong/DependencyCheck/issues/2977

Guess i should have made the regex looser?

ghost:
PR created

Issue #3738, opened 2021-10-14
False Positive on spring-security-jwt-1.1.1.RELEASE.jar: CVE-2021-22112


st-h:
False positive on library `spring-security-jwt-1.1.1.RELEASE.jar (pkg:maven/org.springframework.security/spring-security-jwt@1.1.1.RELEASE, cpe:2.3:a:pivotal_software:spring_security:1.1.1:release:*:*:*:*:*:*, cpe:2.3:a:vmware:spring_security:1.1.1:release:*:*:*:*:*:*) : CVE-2021-22112`

We are not directly importing spring-security-jwt in our project, but a third party library, which uses spring security as a transitive dependency. This is the output of `man dependency:tree -Dincludes=org.springframework.security:spring-security-jwt`

```
 \- com.third.party:internal-lib:jar:1.0.0:compile
    \- org.springframework.security.oauth.boot:spring-security-oauth2-autoconfigure:jar:2.5.5:compile
       \- org.springframework.security:spring-security-jwt:jar:1.1.1.RELEASE:compile
```

Interestingly, the CVE only lists `Spring Security` itself as being affected and not {spring-security-oauth2-autoconfigure} or {spring-security-jwt}.

Issue #3572, opened 2021-08-10
False Positive on spark-catalyst_2.12-3.0.3


yl-coder:
When reporting a false positive please include:
- The location of the dependency (Maven GAV, URL to download the dependency, etc.)

- The CPE that is believed to be false positive
  - Please report the CPE not the CVE
No CPE detected, it detected pkg:maven/org.spark-project.spark/unused@1.0.0 is vulnerable in sonatype

False positive on library spark-catalyst_2.12-3.0.3.jar/ - reported as vulnerabilities above
```xml
<dependency> <!-- Spark dependency -->
            <groupId>org.apache.spark</groupId>
            <artifactId>spark-sql_2.12</artifactId>
            <version>3.0.3</version>
        </dependency>
```

One or more dependencies were identified with known vulnerabilities in project:

spark-catalyst_2.12-3.0.3.jar/META-INF/maven/org.spark-project.spark/unused/pom.xml (pkg:maven/org.spark-project.spark/unused@1.0.0) : CVE-2017-7678, CVE-2018-1334, CVE-2020-9480

weikangchia:
Have the same issue for another related library as well.

```xml
<dependency>
   <groupId>org.apache.spark</groupId>
   <artifactId>spark-core_2.12</artifactId>
   <version>3.1.2</version>
</dependency>
```

spark-core_2.12-3.1.2.jar/META-INF/maven/org.spark-project.spark/unused/pom.xml: CVE-2017-7678, CVE-2020-9480, CVE-2018-1334

PR #3776, opened 2021-11-02
Improve suppression evaluation

jeremylong:
* Resolves FP on #3738, #3766, and #3572.
* Improves suppression rule handling to separate the vulnerability and CPE suppression rules so that they process slightly faster.

### DIFF
--- a/core/src/main/java/org/owasp/dependencycheck/xml/suppression/SuppressionHandler.java
+++ b/core/src/main/java/org/owasp/dependencycheck/xml/suppression/SuppressionHandler.java
@@ -117,7 +117,6 @@ public class SuppressionHandler extends DefaultHandler {
     public List<SuppressionRule> getSuppressionRules() {
         return suppressionRules;
     }
-    
     public SuppressionHandler(SuppressionRuleFilter filter) {
         this.filter = filter;
     }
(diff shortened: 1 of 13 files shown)


core/src/main/java/org/owasp/dependencycheck/xml/suppression/SuppressionRuleFilter.java
```diff
@@ -0,0 +1,42 @@
...
+     * <code>true</code>
+     */
+    public boolean filter(SuppressionRule rule);
+
+    /**
+     * Returns the name of the filter.
+     *
+     * @return the name of the filters
+     */
+    public String getName();
```
jeremylong:
```suggestion
   String getName();
```

core/src/main/java/org/owasp/dependencycheck/xml/suppression/SuppressionRuleFilter.java
```diff
@@ -0,0 +1,42 @@
...
+public interface SuppressionRuleFilter {
+
+    /**
+     * Determines whether or not to filter a suppression rule.
+     *
+     * @param rule the suppression rule to evaluate
+     * @return <code>true</code> if the rule should be filtered; otherwise
+     * <code>true</code>
+     */
+    public boolean filter(SuppressionRule rule);
```
jeremylong:
```suggestion
   boolean filter(SuppressionRule rule);
```

core/src/main/java/org/owasp/dependencycheck/xml/suppression/SuppressionHandler.java
```diff
@@ -113,11 +117,15 @@
     public List<SuppressionRule> getSuppressionRules() {
         return suppressionRules;
     }
+
```
jeremylong:
```suggestion

```

Issue #3786, opened 2021-11-09
Dependencies with a classifier and SNAPSHOT version cannot be resolved


fmarot:
**Describe the bug**
When depending upon an artifact with a classifier and a SNAPSHOT version, this result in the following error:

> [ERROR] Failed to execute goal org.owasp:dependency-check-maven:6.5.0:check (default) on project ssce-dependency-check: One or more exceptions occurred during dependency-check analysis: One or more exceptions occurred during analysis:
> [ERROR]         DependencyNotFoundException: Expected dependency not found in resolved artifacts for dependency com.fake:fakeartifact:zip:native-linux:0.1-SNAPSHOT of project-artifact ssce-dependency-check

This error does not happen if the artifact is deployed in the LOCAL repository (because the file has the real version) but happens if the artifact comes from a remote repository (because the version is timestamped).

**Version of dependency-check used**
The problem was not there up to version 6.3.1.
Version 6.3.2 led to a NullpointerException
Versions 6.4.0 to 6.5.0 lead to the above error using the Maven plugin.

**To Reproduce**
If you install any file in local repo with SNAPSHOT version and then depend upon it, it works correctly
> mvn install:install-file -Dfile=<anyFile.whatever> -DgroupId=com.fake -DartifactId=fakeartifact -Dversion=0.1-SNAPSHOT -Dpackaging=zip -Dclassifier=native-linux

But if you deploy the same file in a remote repo with SNAPSHOT version it does not work (you have to adapt the command to match your remote repo and launch the build using the mvn -U flag to force resolution)
> mvn deploy:deploy-file -Dfile=<anyFile.whatever> -DgroupId=com.fake -DartifactId=fakeartifact -Dversion=0.1-SNAPSHOT -Dpackaging=zip -Dclassifier=native-linux -DrepositoryId=archiva.snapshots -Durl=https://archiva.olea-medical.local/repository/snapshots

**Code update to solve the problem**

In `BaseDependencyCheckMojo `class, the method `private boolean sameArtifact` should be updated to not use:
> ArtifactResult::getArtifact().getVersion()

but instead:

> ArtifactResult::getArtifact().getBaseVersion()

Because getVersion() returns the timestampped version (something like 0.1-20211109.152740-2 ) whereas getBaseVersion() returns the 'real' version (something like 0.1-SNAPSHOT).

PR #3787, opened 2021-11-09
Fix for Dependencies with classifier and SNAPSHOT

fmarot:
Fixes issue #3786
Dependencies with a classifier and SNAPSHOT version cannot be resolved bug

## Description of Change

Since recent changes to workaround MSHARED-998 the use of the class ArtifactResult appeared in the code. While this class looks like ArtifactCoordinate, we have to take care of the `getVersion()` method because for SNAPSHOT versions it provides the timestamped result, not the 'real' version.

So I only changed the `getVersion()` by `getBaseVersion()` call for class ArtifactResult.

I have tested the change manually with a fake project and it solves the  DependencyNotFoundException as expected.

no, not yet, sorry...

### DIFF
--- a/maven/src/main/java/org/owasp/dependencycheck/maven/BaseDependencyCheckMojo.java
+++ b/maven/src/main/java/org/owasp/dependencycheck/maven/BaseDependencyCheckMojo.java
@@ -1526,7 +1526,7 @@ public abstract class BaseDependencyCheckMojo extends AbstractMojo implements Ma
         }
         boolean result = Objects.equals(res.getArtifact().getGroupId(), theCoord.getGroupId());
         result &= Objects.equals(res.getArtifact().getArtifactId(), theCoord.getArtifactId());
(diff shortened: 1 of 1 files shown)


fmarot:
ping @aikebah as you may want to review my (minor) change on the method `sameArtifact(...)` you recently added to work around MSHARED-998

fmarot:
For an explanation of `getVersion()` VS `getBaseVersion()` here is a comment by Robert Scholte of Maven: https://stackoverflow.com/questions/19196836/use-case-of-artifact-getbaseversion#comment28405796_19196934

aikebah:
LGTM

aikebah:
Thanks for the PR @fmarot overlooked this scenario, which has bitten me in other cases in the past. When I read your ping and the summary it was immediately obvious to me what the change was about

aikebah (approved):

PR #3852, opened 2021-12-11
Fixup configuration of Nexus staging plugin

aikebah:
## Description of Change

My IDE flagged the nexusurl configuration of the staging-plugin as invalid. Research showed that it is camelCased, so it should be nexus**U**rl.
This likely will resolve the NullPointerException on main branch pipeline.

N/A

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -768,7 +768,7 @@ Copyright (c) 2012 - Jeremy Long
                 <extensions>true</extensions>
                 <configuration>
                     <serverId>ossrh</serverId>
-                    <nexusurl>https://oss.sonatype.org/</nexusurl>
+                    <nexusUrl>https://oss.sonatype.org/</nexusUrl>
                     <autoReleaseAfterClose>true</autoReleaseAfterClose>
                 </configuration>
             </plugin>


jeremylong (approved):

jeremylong:
Unfortunately, this did not fix the issue with the publication of staging...

aikebah:
Based on the debug-info my next suspicion would go to the following warning by the plugin-plugin:
[INFO] --- maven-plugin-plugin:3.6.2:helpmojo (generated-helpmojo) @ dependency-check-maven ---
[ERROR] 

Maven dependencies of Maven Plugins should be in provided scope.
Please make sure that all your dependencies declared in POM whose group ID is
org.apache.maven have set '<scope>provided</scope>' as well.
In the future this error will break the build.

The following dependencies are in wrong scope:
 * org.apache.maven:maven-model:jar:3.1.0:compile
 * org.apache.maven:maven-artifact:jar:3.1.0:compile
 * org.apache.maven:maven-model-builder:jar:3.1.0:compile
 * org.apache.maven:maven-compat:jar:3.0:compile

Maybe some of those also leak to the classpath when the nexus staging plugin is run and conflict with the up-to-date maven provided jars

PR #3881, opened 2021-12-15
Let Dependabot check for docker image updates

mprins:
## Description of Change

This should take care of opening PRs against the Dockerfile if an upstream image is bumped

*no*

A a side note, not sure what is special about `azul/zulu-openjdk-alpine:14` is seems that was last updated 2 months ago, while other versions have seen updates.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -8,3 +8,13 @@ updates:
     directory: "/"
     schedule:
       interval: "daily"
+  - package-ecosystem: "docker"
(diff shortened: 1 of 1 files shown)


aikebah:
Regarding the sidenote: Azul's build of openJDK 14 is not an Azul Prime build, that's likely why it gets no update.
https://www.azul.com/products/azul-support-roadmap/

jeremylong (approved):

PR #3902, opened 2021-12-19
Replace rich quote in disclaimer with single quote

smoothreggae:
## Description of Change

The rich quote gets rendered incorrectly in terminals like Git Bash on Windows. Replace the rich quote with a simple single quote.

No.


jeremylong (approved):

jeremylong:
Thanks for the PR - copy/paste error from MS word...

PR #3980, opened 2022-01-11
Consider the nodejs ecosystem when searching for a existing dependency in the rpm ecosystem

tobiasstadler:
## Fixes Issue #

Fixes NodePackageAnalyzer and NodeAuditAnalyzer both adding the same dependency (once with ecosystem nodejs and once with ecosystem npm).

## Description of Change

Search in the npm and nodejs ecosystem to find an existing dependency.

no


jeremylong:
Thank you for the PR. However, I would rather solve this problem by fixing the ecosystem on the Abstract NPM Analyzer - see #3983.

PR #4090, opened 2022-02-18
False Positive workflow automation

jeremylong:
Automates some of the workflow for false positive reports - currently covering maven/gradle, dotnet, and npm.

An example of the workflow in action can be seen here: [odc-falsepositives/issues](https://github.com/jeremylong/odc-falsepositives/issues)

The workflow will:
1. Setup a maven, dotnet, or npm project
2. Run ODC using `dependency-check/Dependency-Check_Action@main`
   - the action contains an up-to-date database so no time is spent updating
3. Creates a comment on the issue with the base suppression rule and a link to the test results that has the HTML report with the FP from step 2.


aikebah:
@jeremylong  have to educate myself some more on the github automation for a final judgement (to see if this automation approach would seem safe to me against abuse attempts by hackers (after all there is all kinds of untrusted-input interpretation followed by embedding it in a pom to then issue building of the generated project and posting data to github issue comments)).

But regarding the general approach I like it and see this as a big improvement to the current state.

In FP handling it always felt cumbersome to try and recreate a project with the dependency, especially if the FP report was not regarding a maven library (with maven I am so acquainted that only missing dependency detail could throw me off - or libraries not hosted by one of the mvnrepository indexed maven repositories).

aikebah:
Looks clean to me from the perspective of exploitability, as the only variables used in the output are parts of a successfully parsed packageURL-input and internal Github workflow variables.

aikebah (approved):

jeremylong:
@aikebah Thanks for questioning the security aspects - after 20+ years of working in security I should know better. I forgot to put permissions on the workflow and it really isn't that hard to prevent some of the possible security issues using environment variables.  See [71ad032](https://github.com/jeremylong/DependencyCheck/pull/4090/commits/71ad0325d5b6c30b5451df90a63c9073ec77a238).

aikebah:
> @aikebah Thanks for questioning the security aspects - after 20+ years of working in security I should know better. I forgot to put permissions on the workflow

Definitely a good improvement

> and it really isn't that hard to prevent some of the possible security issues using environment variables.

Not clear to me where it would add security (as in prevent security issues) in this specific case. Have not been involved deeply enough in security/hacking to see where it might help I guess. My gut feel is that the packageURL parsing and then using the parsed results is the defense-line in the sense of input validation (of course the packageURL parsing routine has then to be trusted upon to only return data in case of fully compliant packageURLs).

.github/workflows/false-positive-ops.yml
```diff
@@ -137,7 +159,7 @@ jobs:
                 purl += '/' + name.replaceAll('.','\\.');
             }
             purl += '@.*$';
-            var cpe = '${{ fromJSON(steps.issue-parser.outputs.jsonString).cpe }}'.split(':');
```
jeremylong: @aikebah as an example of the issue - here was the CPE value from the form being written into server side JavaScript without any encoding, etc. If one put a CPE value of `';await github.rest....//` you could call the GH apis, and do all kinds of stuff.

jeremylong: Also, when considering the "validation" performed by package URL parsing - I'm really not a fan of relying on validation. There is always someone smarter that might know an encoding technique, etc. you didn't think of with the allowed characters.  Best approach for injection attacks is to separate the data and control plan - don't output data into an execution context.

PR #4096, opened 2022-02-20
Fix a performance drain due to multiple executions of the function body

aikebah:
## Description of Change

Immediately return an empty resultset rather than performing  the function body when H2 is only interested in the list of columns that will be return on true execution.

no, tested locally and observed a massive improve in wall-clock processing time as well as total CPU time.


jeremylong (approved):

jeremylong:
@aikebah thank you for figuring this one out!

PR #4102, opened 2022-02-23
#4058 Prevent NPE during race condition

### LINKED ISSUE #4058: NullPointerException checking jackson-databind 2.13.1
**Describe the bug**
A `NullPointerException` is thrown when analyzing one module

The problem occurs using version 6.5.3 of the gradle plugin

**Log file**
```
> Task :xxx:dependencyCheckAnalyze
Verifying dependencies for project xxx

Failed to fetch component-report for: pkg:maven/com.fasterxml.jackson.core/jackson-databind@2.13.1
java.lang.NullPointerException: Cannot invoke "java.util.Map.get(Object)" because "org.owasp.dependencycheck.analyzer.OssIndexAnalyzer.reports" is null
        at org.owasp.dependencycheck.analyzer.OssIndexAnalyzer.enrich(OssIndexAnalyzer.java:235)
        at org.owasp.dependencycheck.analyzer.OssIndexAnalyzer.analyzeDependency(OssIndexAnalyzer.java:158)
```

**To Reproduce**
I haven't created a reproducible example. If needed, I will try to do it.

**Expected behavior**
No exception thrown.

**Additional context**
Java 17

nhumblot:
## Fixes Issue #

Fix #4058 

## Description of Change

This change prevent raising a `NullPointerException` by including the `OssIndexAnalyzer.enrich()` method into the mutex. If not, `reports.get(purl)` can cause the exception.

*yes*

I tried to reproduce the behavior with a unit test and succeeded to do so. On my computer with Eclipse, I was getting a test failure every time with the current test implementation and the `enrich()` call being outside of the mutex. I had to slightly alter the api by making the `enrich()` method package protected instead of private.

I also ran an analysis with the maven installed snapshot to check if we do not encounter a dead lock.

The comment added in the test try to explain the goal of the class and the test. I am not sure if it is clear enough, please do not hesitate to ask questions or indicates if you see an issue with such kind of tests.

Please take extra care while reviewing this PR, due to adding statements in a `synchronized()` block, it might slow things a little. Comments are welcomed. :slightly_smiling_face:

### DIFF
--- a/core/src/main/java/org/owasp/dependencycheck/analyzer/OssIndexAnalyzer.java
+++ b/core/src/main/java/org/owasp/dependencycheck/analyzer/OssIndexAnalyzer.java
@@ -153,12 +153,13 @@ public class OssIndexAnalyzer extends AbstractAnalyzer {
                     throw new AnalysisException("Failed to request component-reports", e);
                 }
             }
+
+            // skip enrichment if we failed to fetch reports
+            if (!failed) {
+                enrich(dependency);
+            }
         }
 
-        // skip enrichment if we failed to fetch reports
-        if (!failed) {
-            enrich(dependency);
-        }
     }
 
     /**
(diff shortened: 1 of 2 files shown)


jeremylong (approved):

jeremylong:
Thanks for the PR.

PR #4108, opened 2022-02-23
Break build on failure to download NVD datastreams (when updates are requested or autoUpdate is enabled)

aikebah:
## Description of Change

Make the build break instead of issuing only a warning that some CVEs may not be reported when there is an error downloading one of the NVD datafeeds.

While working spotted an issue that the year roll-over grace period was not covered for the needsFullUpdate case, so I added it there. And updated the existing logic to only accept the not-found case for failed download (a `too many requests` response would also be acceptedin the grace-period with the original code)

And as I understand it CNAs are using local date to determine the year-part of a CVE, so I think we should start attempting to obtain a new year datafeed as soon as the first timezone enters the new year.

no, tested locally with molesting a local NVD mirror and a temporary extended new year grace period to simulate download errors and validate that a missing new year datafeed still succeeds with just a warning in the grace-period

### DIFF
--- a/core/src/main/java/org/owasp/dependencycheck/data/update/NvdCveUpdater.java
+++ b/core/src/main/java/org/owasp/dependencycheck/data/update/NvdCveUpdater.java
@@ -26,7 +26,7 @@ import java.io.IOException;
 import java.io.InputStreamReader;
 import java.io.OutputStreamWriter;
 import java.net.MalformedURLException;
-import java.util.Calendar;
+import java.time.LocalDate;
 import java.util.HashSet;
 import java.util.Set;
 import java.net.URL;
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/owasp/dependencycheck/data/update/NvdCveUpdater.java
```diff
@@ -441,7 +442,12 @@ private MetaProperties doMetaDownload(String url, boolean throwErrors) throws Up
         if (dbProperties != null && !dbProperties.isEmpty()) {
             try {
                 final int startYear = settings.getInt(Settings.KEYS.CVE_START_YEAR, 2002);
-                final int endYear = Calendar.getInstance().get(Calendar.YEAR);
+                // for establishing the current year use the timezone where the new year starts first
+                // as from that moment on CNAs might start assigning CVEs with the new year depending
+                // on the CNA's timezone
+                ZonedDateTime today = ZonedDateTime.now().withZoneSameInstant(ZoneOffset.ofHours(14));
```
jeremylong: Make checkstyle happier...

```suggestion
                final ZonedDateTime today = ZonedDateTime.now().withZoneSameInstant(ZoneOffset.ofHours(14));
```

jeremylong (approved):

Issue #4151, opened 2022-03-02
Unable to create temp file for suppression file


sergeiwaigant:
Hi all.

I am trying to use the owasp dependency check plugin for Maven and pointing to a suppression file in the pom.xml as below.

```
          <suppressionFiles>
            <suppressionFile>${session.executionRootDirectory}/org.owasp.dependency-check-maven.configuration.suppressions.xml</suppressionFile>
          </suppressionFiles>
```

During the build, I am getting the error `Unable to create temp file for suppression file` ...

```
00:04:01.219  [INFO] Analysis Started
00:04:01.836  [INFO] Finished Archive Analyzer (0 seconds)
00:04:01.836  [INFO] Finished File Name Analyzer (0 seconds)
00:04:02.416  [INFO] Finished Jar Analyzer (0 seconds)
00:04:02.416  [INFO] Finished Dependency Merging Analyzer (0 seconds)
00:04:02.416  [INFO] Finished Version Filter Analyzer (0 seconds)
00:04:02.416  [INFO] Finished Hint Analyzer (0 seconds)
00:04:03.812  [INFO] Created CPE Index (1 seconds)
00:04:03.812  [WARNING] Unable to create temp file for suppression file'/mnt/sdb/cloudbees-agent/001/9f87b2f2/workspace/c_feature_owasp-dependency-check/org.owasp.dependency-check-maven.configuration.suppressions.xml'
00:04:03.812  [ERROR] Exception occurred initializing CPE Analyzer.
```

The build is happening from a Jenkins agent running in a non-privileged user.
The "`java.io.tmpdir`" and "`temp.directory`" is set as `JVM_OPTS` to the WORKSPACE of the Jenkins build, where the user has full access.

Does anyone has an idea where the plugin tries to create a temp file?

Regards
Sergei

jeremylong:
The error message is obviously wrong - it isn't trying to create the temp file. Rather ODC is trying to read the file and it is getting an IOException (I'll update the error message).

Does the suppression file really exist in: /mnt/sdb/cloudbees-agent/001/9f87b2f2/workspace/c_feature_owasp-dependency-check/org.owasp.dependency-check-maven.configuration.suppressions.xml

Or should you be using something other then `${session.executionRootDirectory}`?

PR #4158, opened 2022-03-04
Fix Error Message

jeremylong:
Fixes error message when suppression file cannot be read #4151


core/src/main/java/org/owasp/dependencycheck/analyzer/AbstractSuppressionAnalyzer.java
```diff
@@ -248,7 +248,7 @@ private void loadSuppressionBaseData() throws SuppressionParseException {
         } catch (SuppressionParseException ex) {
             throw ex;
         } catch (IOException ex) {
-            throwSuppressionParseException("Unable to create temp file for suppression file", ex, suppressionFilePath);
+            throwSuppressionParseException("Unable to suppression file", ex, suppressionFilePath);
```
steveruddell, line 251: should we include a verb in this clause? perhaps `"unable to read the supplied suppression file"`

sunsee18: Makes it more understandable

PR #4167, opened 2022-03-04
#4078 Return a dedicated error message when sonatype returns a 403

### LINKED ISSUE #4078: 403 error from ossindex.sonatype.org
**Describe the bug**
When running dependency check via gradle owasp plugin task, OSS Index Analyzer fails on first jar with below debug output. Subsequent jars appear to be analyzed without additional errors until spring-boot-starter-tomcat, which also fails to request component-reports with a 403. The overall analysis step fails.

The problem occurs using version 6.5.3 of the gradle plugin (cli, gradle plugin, maven plugin, etc.)

**Log file**
gist link: https://gist.github.com/ScottDish/f18c4205af6ca290ac719876ba3c48ac

**To Reproduce**
1. Go to 'project root dir'
2. Run ' ./gradlew --debug dependencyCheckAnalyze'
3. See error

**Expected behavior**
Complete analysis with no vulnerabilities found.

**Additional context**
Gradle 6.9.2 
Spring Boot 2.6.3
Problem appears related to network security (running this within a corporate network), but I can't determine the exact issue. I suspect that it might be a problem with SSL certificates, but can't find any certificates from sonatype that make this work. The issue seems to go away (analysis completes without any error) if I run from outside the corporate network.

nhumblot:
## Fixes Issue #

Fix #4078 

## Description of Change

Handle an HTTP 403 Forbidden response from OSS Index and display an explicit error message.

*yes*

No change on the public API.

Comments are welcomed. :slightly_smiling_face: 

edit: I am sorry for the multiple `--force-push-with-lease`, a Maven execution on my workstation ends up in success. I will try to find why the test is failing in the CI.

### DIFF
--- a/core/src/test/java/org/owasp/dependencycheck/analyzer/OssIndexAnalyzerTest.java
+++ b/core/src/test/java/org/owasp/dependencycheck/analyzer/OssIndexAnalyzerTest.java
@@ -19,7 +19,6 @@ import org.owasp.dependencycheck.dependency.naming.Identifier;
 import org.owasp.dependencycheck.dependency.naming.PurlIdentifier;
 import org.owasp.dependencycheck.utils.Settings;
 
-import com.github.packageurl.MalformedPackageURLException;
 import org.sonatype.goodies.packageurl.PackageUrl;
 import org.sonatype.ossindex.service.api.componentreport.ComponentReport;
 import org.sonatype.ossindex.service.client.OssindexClient;
@@ -29,10 +28,11 @@ public class OssIndexAnalyzerTest extends BaseTest {
 
     @Test
     public void should_enrich_be_included_in_mutex_to_prevent_NPE()
-            throws AnalysisException, MalformedPackageURLException {
+            throws Exception {
 
         // Given
         OssIndexAnalyzer analyzer = new SproutOssIndexAnalyzer();
+        analyzer.close();
 
 
         Identifier identifier = new PurlIdentifier("maven", "test", "test", "1.0",
@@ -87,9 +87,10 @@ public class OssIndexAnalyzerTest extends BaseTest {
     }
 
     @Test
-    public void should_analyzeDependency_return_a_dedicated_error_message_when_403_response_from_sonatype() throws MalformedPackageURLException {
+    public void should_analyzeDependency_return_a_dedicated_error_message_when_403_response_from_sonatype() throws Exception {
         // Given
         OssIndexAnalyzer analyzer = new OssIndexAnalyzerThrowing403();
+        analyzer.close();
         analyzer.initialize(getSettings());
 
         Identifier identifier = new PurlIdentifier("maven", "test", "test", "1.0",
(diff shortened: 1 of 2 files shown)


aikebah:
> edit: I am sorry for the multiple `--force-push-with-lease`, a Maven execution on my workstation ends up in success. I will try to find why the test is failing in the CI.

Haven't tried locally yet, but did you run with releasetesting enabled when you ran your local build? 

`mvn clean package verify site -DreleaseTesting --no-transfer-progress --batch-mode`

would be the best effort local reproducer honoring your local settings

`mvn -s settings.xml clean package verify site -DreleaseTesting --no-transfer-progress --batch-mode`

would the fully identical goals run in PR-workflow

nhumblot:
> `mvn -s settings.xml clean package verify site -DreleaseTesting --no-transfer-progress --batch-mode`

Yes this ends in success with this command on my laptop with jdk 8 & jdk11:
```
Apache Maven 3.6.3 (Red Hat 3.6.3-13)
Maven home: /usr/share/maven
Java version: 11.0.14.1, vendor: Red Hat, Inc., runtime: /usr/lib/jvm/java-11-openjdk-11.0.14.1.1-4.fc35.x86_64
Default locale: fr_FR, platform encoding: UTF-8
OS name: "linux", version: "5.16.11-200.fc35.x86_64", arch: "amd64", family: "unix"

Apache Maven 3.6.3 (Red Hat 3.6.3-13)
Maven home: /usr/share/maven
Java version: 1.8.0_322, vendor: Red Hat, Inc., runtime: /usr/lib/jvm/java-1.8.0-openjdk-1.8.0.322.b06-2.fc35.x86_64/jre
Default locale: fr_FR, platform encoding: UTF-8
OS name: "linux", version: "5.16.11-200.fc35.x86_64", arch: "amd64", family: "unix"
```

I plan to try by using a docker container with the same image than the one used in the CI, then enabling debug logs so I can get more data and identify the differences.

nhumblot:
Found the issue, it was due to static fields in the `OssIndexAnalyzer`, keeping state between tests. :slightly_smiling_face:

core/src/test/java/org/owasp/dependencycheck/analyzer/OssIndexAnalyzerTest.java
```diff
@@ -16,16 +19,20 @@
...
 public class OssIndexAnalyzerTest extends BaseTest {
 
     @Test
     public void should_enrich_be_included_in_mutex_to_prevent_NPE()
-            throws AnalysisException, MalformedPackageURLException {
+            throws Exception {
 
         // Given
         OssIndexAnalyzer analyzer = new SproutOssIndexAnalyzer();
+        analyzer.close();
```
aikebah: Closing an analyzer should be performed at the end after it's done with all its work, not immediately after constructing it. And as the SproutOssIndexAnalyzer already runs the close (by way of this.closeAnalyzer() in the enrich) it should not even be needed at all.

nhumblot: Hi, :wave: 

I removed this one as it looked to be unnecessary as you said. Unfortunately, if I remove the one in `OssIndexAnalyzerTest.should_analyzeDependency_return_a_dedicated_error_message_when_403_response_from_sonatype()`, the test fails in the CI (despite working on my laptop with the same Maven command).

If you want me to remove it also, I may be able to look at all the places where this Analyzer is used in tests and try to see where this context is kept.

jeremylong (approved):

PR #4184, opened 2022-03-11
Surpress FPs for cryptomator libraries. Fixes #4177

### LINKED ISSUE #4177: [FP]: Any library of cryptomator.org
### Package URl

pkg:maven/org.cryptomator/dokany-nio-adapter@1.3.3

### CPE

`cpe:2.3:a:cryptomator:cryptomator:*:*:*:*:*:*:*:*`

### CVE

_No response_

### ODC Integration

{"label"=>"Maven Plugin"}

### ODC Version

7.0.0

### Description

This holds also for other libs of the cryptomator organization, e.g.
* [pkg:maven/org.cryptomator/integrations-api@1.1.0-beta1](https://ossindex.sonatype.org/component/pkg:maven/org.cryptomator/integrations-api@1.1.0-beta1?utm_source=dependency-check&utm_medium=integration&utm_content=6.3.1)
* [pkg:maven/org.cryptomator/integrations-win@1.0.0](https://ossindex.sonatype.org/component/pkg:maven/org.cryptomator/integrations-win@1.0.0?utm_source=dependency-check&utm_medium=integration&utm_content=6.3.1)
* [pkg:maven/org.cryptomator/siv-mode@1.4.3](https://ossindex.sonatype.org/component/pkg:maven/org.cryptomator/siv-mode@1.4.3?utm_source=dependency-check&utm_medium=integration&utm_content=6.3.1)
* ... and others

The `cpe:/a:cryptomator:cryptomator` describes the end user application, not any of the underlying used libraries. So a regex supression including any `pkg:maven/org.cryptomator/*` is needed.

infeo:
## Description of Change

Due to [CVE-2022-25366](https://nvd.nist.gov/vuln/detail/CVE-2022-25366), dependency-check reports via its maven plugin every Cryptomator library as vulnerable, even though only the end user application is affected. This PR supresses these FPs by supressing any maven artifact of `org.cryptomator/*` for `cpe:/a:cryptomator:cryptomator`.

no, only tested locally.

### DIFF
--- a/core/src/main/resources/dependencycheck-base-suppression.xml
+++ b/core/src/main/resources/dependencycheck-base-suppression.xml
@@ -4861,7 +4861,7 @@
         <notes><![CDATA[
             Supress FP in cryptomator libraries, because linked CVE-2022-25366 only affects end user application. (#4177)
             ]]></notes>
-        <packageUrl regex="true">^pkg:maven/org\.cryptomator/.*$</packageUrl>
+        <packageUrl regex="true">^pkg:maven/org\.cryptomator/(?!cryptomator).*$</packageUrl>
         <cpe>cpe:/a:cryptomator:cryptomator</cpe>
     </suppress>
 </suppressions>
--- a/core/src/main/resources/dependencycheck-base-suppression.xml
+++ b/core/src/main/resources/dependencycheck-base-suppression.xml
@@ -4857,4 +4857,11 @@
 	<packageUrl regex="true">^pkg:maven/com\.ibm\.etcd/etcd\-java@.*$</packageUrl>
 	<cpe>cpe:/a:etcd:etcd</cpe>
     </suppress>
+    <suppress base="true">
+        <notes><![CDATA[
+            Supress FP in cryptomator libraries, because linked CVE-2022-25366 only affects end user application. (#4177)
+            ]]></notes>
+        <packageUrl regex="true">^pkg:maven/org\.cryptomator/.*$</packageUrl>
+        <cpe>cpe:/a:cryptomator:cryptomator</cpe>
+    </suppress>
 </suppressions>


aikebah:
This suppression is definitely too broad as it also suppresses the desktop app itself

infeo:
I adjusted the regex to exclude [`org.cryptomator.cryptomator`](https://github.com/cryptomator/cryptomator/blob/a32f5bb6d90b163a130d025546657d35eca26d47/pom.xml#L4-L5) from being supressed.

overheadhunter:
The desktop app **is not a maven package** though.

aikebah:
> The desktop app **is not a maven package** though.

But from what I understand the desktop app is packaged from, and versioned identical to the main cryptomator app maven package artifact? And thus any CVE that would be valid for the maven app module (I do see that `CVE-2022-25366` should even be suppressed for the main app library as it manifests in the Mac OS native packaging, not in the app module itself) would be accounted for on the cryptomator CPE by NVD? Or are even those independently versioned?

overheadhunter:
Correct, the vulnerability does not exist in the Maven module, only in the macOS .app based on it.
The CPE is inaccurate, we don't know who added it, as it was not present in the original CVE.

Sadly, the CPE nomenclature only knows vendors and products. Since the product name of the final app is Cryptomator (duh!) no distinction between Maven and binary artifacts is possible.

We're in the process of requesting official CPE names where we can distinguish the apps for several platforms, but for now the problem persists as organization name and product name are identical and basically every artifact published under the orga name is false positive.

Btw the Cryptomator Maven artifact is not published to any Maven repo, it is merely an intermediate build step towards the platform-specific apps.

overheadhunter:
I guess the root cause lies here:

https://github.com/jeremylong/DependencyCheck/blob/e897c58427482b02553bf46c938f5122f77dd2fc/core/src/main/java/org/owasp/dependencycheck/analyzer/JarAnalyzer.java#L642-L645

The maven groupId is used as product name with confidence that can not be beaten by the artifactId.

jeremylong:
> Btw the Cryptomator Maven artifact is not published to any Maven repo, it is merely an intermediate build step towards the platform-specific apps.

If Cryptomator is not published and used by others - shouldn't this just be a suppression rule you use locally as opposed to putting this in the suppression rules of the core engine?

overheadhunter:
> > Btw the Cryptomator Maven artifact is not published to any Maven repo, it is merely an intermediate build step towards the platform-specific apps.
> 
> If Cryptomator is not published and used by others - shouldn't this just be a suppression rule you use locally as opposed to putting this in the suppression rules of the core engine?

1. we publish Maven artifacts with the groupid `org.cryptomator` (one of them being a pretty wide-spread transitive dependency via Nimbus OAuth 2.0 SDK or Graylog 2 Server)
2. we publish Cryptomator (a desktop application, **not** a Maven artifact) via GitHub Releases

The latter had a vulnerability in its macOS package. The former do not even carry "cryptomator" in their artifact names, but the groupId causes FP.

jeremylong:
Another option instead of a suppression rule would be to add a hint rule to remove the offending group id that is causing the FP. See https://github.com/jeremylong/DependencyCheck/blob/963cd91d96887af45e4dfb495c296d793d12e85e/core/src/main/resources/dependencycheck-base-hint.xml#L194-L207

Something like:

```xml
 <hint> 
     <given> 
         <evidence type="product" source="pom" name="groupid" value="org.cryptomator" confidence="HIGHEST"/> 
     </given> 
     <remove> 
         <evidence type="product" source="pom" name="groupid" value="org.cryptomator" confidence="HIGHEST"/> 
     </remove> 
 </hint>

overheadhunter:
Wouldn't that cause false negatives, if there eventually was a vulnerability in a Maven artifact `org.cryptomator:somelib`?

jeremylong:
Whether it would create a false-negative would be up to how the CPE was created for the library... I was just offering a different solution.  I'm okay with the suppression for the maven library the way it is.

jeremylong (approved):

PR #4191, opened 2022-03-12
Makes excludes property configurable in command line

chalmets:
Makes "excludes" configurable in command line like:
mvn clean org.owasp:dependency-check-maven:check -Dodc.excludes=mygroup-1:my-jar-1,mygroup-2:my-jar-2

## Fixes Issue #
Currently to exclude a jar from the dependencies analyses in maven we have to configure the <excludes> section in the pom file. There is no way to be able to configure this exclusions in command line.

## Description of Change
Adding the property **excludes**  configurable in command line
Adding a property name to the existing **excludes** property allow to provided this property when executing the maven plugin in command line like:
`mvn clean org.owasp:dependency-check-maven:check
`With this change we can now use this configuration in command line like:
`mvn clean org.owasp:dependency-check-maven:check -Dodc.excludes=mygroup-1:my-jar-1,mygroup-2:my-jar-2`

*no*

### DIFF
--- a/maven/src/main/java/org/owasp/dependencycheck/maven/BaseDependencyCheckMojo.java
+++ b/maven/src/main/java/org/owasp/dependencycheck/maven/BaseDependencyCheckMojo.java
@@ -911,7 +911,7 @@ public abstract class BaseDependencyCheckMojo extends AbstractMojo implements Ma
      * The list of artifacts (and their transitive dependencies) to exclude from
      * the check.
      */
-    @Parameter(property = "excludes")
+    @Parameter(property = "odc.excludes")
     private List<String> excludes;
 
     /**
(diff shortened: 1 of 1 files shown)


maven/src/main/java/org/owasp/dependencycheck/maven/BaseDependencyCheckMojo.java
```diff
@@ -911,7 +911,7 @@
      * The list of artifacts (and their transitive dependencies) to exclude from
      * the check.
      */
-    @Parameter
+    @Parameter(property = "excludes")
```
aikebah: Can you change this to `odc.excludes` to ensure the command-line property does not clash with properties of [standard maven plugins](https://maven.apache.org/plugins/maven-dependency-plugin/tree-mojo.html#excludes)

chalmets: yes absolutely. thanks for the suggestion

aikebah (approved):

aikebah:
Thanks for the PR

chalmets:
Thank you very much for looking into this PR and merging it into the main branch so quickly.
Do you have visbility on when you will be releasing the next version of your application that will include my change ?

Thank you

PR #4192, opened 2022-03-12
Use a proper parser for PyPA Core Metadata

aikebah:
## Description of Change

Use a proper parser for the PyPA Core Metadata format to process a wheel metadata file. As indicated in the issue the java.util.Properties class is not appropriate to process the file as it is expected to be UTF-8 encoded in a format derived from the internet mail message format.
The format is using header fields for key/value metadata and allows the description to be either a multiline header field or a message body.

**DRAFT** Awaiting feedback from issue-commenters to validate that it also works IRL

Testcases have been added to validate some of the parsing features.

### DIFF
--- a/core/src/main/java/org/owasp/dependencycheck/utils/PyPACoreMetadataParser.java
+++ b/core/src/main/java/org/owasp/dependencycheck/utils/PyPACoreMetadataParser.java
@@ -1,0 +1,136 @@
+/*
+ * This file is part of dependency-check-core.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * Copyright (c) 2013 Jeremy Long. All Rights Reserved.
+ */
+package org.owasp.dependencycheck.utils;
+
+import org.apache.commons.lang3.StringUtils;
+import org.owasp.dependencycheck.analyzer.exception.AnalysisException;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.io.BufferedReader;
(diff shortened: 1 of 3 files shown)


core/src/main/java/org/owasp/dependencycheck/utils/PyPACoreMetadataParser.java
```diff
@@ -0,0 +1,149 @@
...
+import java.util.Properties;
+
+/**
+ * A utility class to handle Python Packaging Authority (PyPA) core metadata files. It was created based on the
+ * <a href="https://packaging.python.org/en/latest/specifications/core-metadata/">specification by PyPA</a> for
+ * version 2.2
+ *
+ * @author Hans Aikema
+ */
+public final class PyPACoreMetadataParser {
```
jeremylong, line 40: I think this works - but in a quick glance through the spec, they reference RFC822 parser. Java has an implementation: [javax.mail.internet.HeaderTokenizer](https://javaee.github.io/javaee-spec/javadocs/javax/mail/internet/HeaderTokenizer.html). 

In reality - I caused this bug. If you look at the blame we see a PR where I was trying to reduce dependencies and didn't realize the real reason behind using the dependency.  I don't think we need to bring the mail API back in. But we may want to look at using the `HeaderTokenizer`.  Thoughts?

https://github.com/jeremylong/DependencyCheck/commit/fbe94b0191c0aabe5ee46364a852856a1925a061

aikebah: Had a brief look at it before coding something myself, but couldn't get my head around how to use its interface to parse the PyPA format - which is specified to handle 'anything that the Python rfc822 parser will handle', which would include non-standard representation of headers that the Python parser accepts (as it's likely coded in a forgiving way given the 'compat' in the referenced policy).

>In the absence of a precise definition, the practical standard is set by what the standard library [email.parser](https://docs.python.org/3/library/email.parser.html#module-email.parser) module can parse using the [compat32](https://docs.python.org/3/library/email.policy.html#email.policy.compat32) policy

Maybe best to start off with this to resolve the currently outstanding issue on URL references in the wheel metadata and open an enhancement issue to try and use the javax.mail parser instead?
Fairly certain that HeaderTokenizer alone won't get it done, as the description can be given as the message body and to my understanding other parts of the javax.mail take care of splitting the headers from the body / MIME parts and HeaderTokenizer gets fed with only the header text.

jeremylong (approved):

jeremylong:
@aikebah can we remove the draft status on this?

PR #4200, opened 2022-03-14
support partial fingerprints for sarif reports

timdodd:
## Fixes Issue #

N/A

## Description of Change

This change is to support [partial fingerprints](https://docs.oasis-open.org/sarif/sarif/v2.0/csprd02/sarif-v2.0-csprd02.html#_Toc10127846) for the sarif report. The partial fingerprint is generated based on the vulnerability name (cve or cwe). 

This change was made to support [github code scanning](https://docs.github.com/en/code-security/code-scanning/integrating-with-code-scanning/sarif-support-for-code-scanning) and prevents duplicate alerts across branches.

> If you upload a SARIF file without fingerprint data using the /code-scanning/sarifs API endpoint, the code scanning alerts will be processed and displayed, but users may see duplicate alerts. To avoid seeing duplicate alerts, you should calculate fingerprint data and populate the partialFingerprints property before you upload the SARIF file.  

Sarif report testing has been added to ReportGeneratorIT; however, the test doesn't inspect the contents.


core/src/main/java/org/owasp/dependencycheck/reporting/ReportGenerator.java
```diff
@@ -258,6 +258,7 @@ private VelocityContext createContext(String applicationName, List<Dependency> d
         ctxt.put("scanDateJunit", scanDateJunit);
         ctxt.put("enc", new EscapeTool());
         ctxt.put("rpt", new ReportTool());
+        ctxt.put("hash", new HashTool());
```
jeremylong: Use existing objects:

```suggestion
        ctxt.put("hash", org.owasp.dependencycheck.utils.Checksum.class);
```

core/src/main/resources/templates/sarifReport.vsl
```diff
@@ -122,6 +122,9 @@ For more information see [How dependency-check works](https://jeremylong.github.
           "message": {
             "text": "$enc.json($vuln.name)#if($vuln.description) - $enc.json($vuln.description)#end"
           },
+          "partialFingerprints": {
+            "vulnerabilityHash": "$hash.md5Hex($vuln.name)"
```
jeremylong: Use existing methods:
```suggestion
            "vulnerabilityHash": "$hash. getMD5Checksum($vuln.name)"
```

core/src/main/java/org/owasp/dependencycheck/reporting/HashTool.java
```diff
@@ -0,0 +1,18 @@
...
+
+import javax.annotation.concurrent.ThreadSafe;
+
+/**
+ * An extremely simple wrapper around hashing functionality
+ *
+ * @author Tim Dodd
+ */
+@ThreadSafe
+public class HashTool {
```
jeremylong: This wrapper is duplicative and should be removed from the PR. See https://github.com/jeremylong/DependencyCheck/blob/3d097c4838d21d861753a0ea77cde92af19fac40/utils/src/main/java/org/owasp/dependencycheck/utils/Checksum.java#L188-L190

jeremylong (changes requested):
Thanks for the PR - however, would you mind updating the PR to use the existing wrapper to generate hashes?

timdodd:
Thanks for the comments and review! The PR has been updated with the requested changes.

jeremylong (approved):

PR #4223, opened 2022-03-17
Remove unnecessary back quotes in SQL statement

alerat:
## Description of Change

Remove unnecessary back quotes in SQL statement

no


jeremylong:
I assure you - the back tics are necessary for H2. If the build would have triggered on this PR you would have seen errors. The correct solution to this problem is to follow the statement in the `TODO` block and move this the SQL statement into the properties files so that we can have a different version of the SQL for the postgres database. See https://github.com/jeremylong/DependencyCheck/tree/main/core/src/main/resources/data

jeremylong:
See https://github.com/jeremylong/DependencyCheck/pull/4236

PR #4231, opened 2022-03-18
Added option for graceful handling of remote errors

sithmein:
## Description of Change

The following new options have been added:

* `ossIndexWarnOnlyOnRemoteErrors` for the Maven plug-in
* `--ossIndexRemoteErrorWarnOnly` for the CLI application
* `ossIndexAnalyzerWarnOnlyOnRemoteErrors` for the Ant task

* yes


jeremylong (approved):

core/src/test/java/org/owasp/dependencycheck/analyzer/OssIndexAnalyzerTest.java
```diff
@@ -112,6 +113,33 @@ public void should_analyzeDependency_return_a_dedicated_error_message_when_403_r
         assertEquals("OSS Index access forbidden", output.getMessage());
     }
 
+    
+    @Test
+    public void should_analyzeDependency_only_warn_when_transport_error_from_sonatype() throws Exception {
+        // Given
+        OssIndexAnalyzer analyzer = new OssIndexAnalyzerThrowing502();
+        analyzer.close();
```
aikebah, line 121: Like the 403-test-case I think this should move to the end of the method after integrating the fix of #4233. #4233 resolves the root cause of the intermittent test-failure this premature close of the analyzer is trying to resolve. This close is compensating for a missing closure of an earlier instance leaving state in the static fields of the superclass.
Should not block this merge, but needs to be cleaned up after integrating #4233

PR #4233, opened 2022-03-19
Ensure closure of OSSIndexAnalyzer at the end of a test method to clear static test state

aikebah:
Ensure closure of analyzer at the end of a test method to prevent leaking static analyzer state breaking other tests in an undeterministic way

## Description of Change

Ensure that the parallel submitted closure of the analyzer for testcases using the SproutOssIndexAnalyzer is processed before ending the testMethod so there is no lock-race for the Mutex between the parallel closure and the analyzeDependency of another testcase.

Also ensure that no failure state leaks from the 403 testcase by closing the analyzer at the end of the test.

no, it's a fix of the testcases


nhumblot (approved):

nhumblot:
Thank you for your help. :slightly_smiling_face:

jeremylong (approved):

Issue #4280, opened 2022-03-28
bad jackson version


Grimoren:
**Describe the bug**

Could not resolve all artifacts for configuration ':classpath'.
   > Could not find com.fasterxml.jackson:jackson-bom:2.13.2.1.
     Searched in the following locations:
       - https://artifacts.apple.com/plugins-release/com/fasterxml/jackson/jackson-bom/2.13.2.1/jackson-bom-2.13.2.1.pom
       - https://artifacts.apple.com/plugins-snapshot/com/fasterxml/jackson/jackson-bom/2.13.2.1/jackson-bom-2.13.2.1.pom
     Required by:
         project : > org.owasp:dependency-check-gradle:7.0.2 > org.owasp:dependency-check-core:7.0.2 > com.fasterxml.jackson.core:jackson-databind:2.13.2.1
   > Could not find com.fasterxml.jackson:jackson-bom:2.13.2.1.
     Searched in the following locations:
       - https://artifacts.apple.com/plugins-release/com/fasterxml/jackson/jackson-bom/2.13.2.1/jackson-bom-2.13.2.1.pom
       - https://artifacts.apple.com/plugins-snapshot/com/fasterxml/jackson/jackson-bom/2.13.2.1/jackson-bom-2.13.2.1.pom
     Required by:
         project : > org.owasp:dependency-check-gradle:7.0.2 > org.owasp:dependency-check-core:7.0.2 > com.fasterxml.jackson.core:jackson-databind:2.13.2.1 > com.fasterxml.jackson.core:jackson-annotations:2.13.2
         project : > org.owasp:dependency-check-gradle:7.0.2 > org.owasp:dependency-check-core:7.0.2 > com.fasterxml.jackson.core:jackson-databind:2.13.2.1 > com.fasterxml.jackson.core:jackson-core:2.13.2

**Version of dependency-check used**
The problem occurs using version 7.0.2 of the gradle (cli, gradle plugin, maven plugin, etc.)

**Log file**

Could not resolve all artifacts for configuration ':classpath'.
   > Could not find com.fasterxml.jackson:jackson-bom:2.13.2.1.
     Searched in the following locations:
       - https://artifacts.apple.com/plugins-release/com/fasterxml/jackson/jackson-bom/2.13.2.1/jackson-bom-2.13.2.1.pom
[...]

PR #4281, opened 2022-03-28
Fix jackson version to use valid Jackson version

Grimoren:
## Fixes Issue #
https://github.com/dependency-check/dependency-check-gradle/issues/260
https://github.com/jeremylong/DependencyCheck/issues/4280
## Description of Change
use valid jackson version

N/A: no new functionality.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -148,7 +148,7 @@ Copyright (c) 2012 - Jeremy Long
         <aho-corasick-double-array-trie.version>1.2.3</aho-corasick-double-array-trie.version>
         <junit.version>4.13.2</junit.version>
         <hamcrest-core.version>2.2</hamcrest-core.version>
-        <jackson.version>2.13.2</jackson.version>
+        <jackson.version>2.13.2.1</jackson.version>
         <jackson-afterburmer.version>2.13.2</jackson-afterburmer.version>
         <jmockit.version>1.49</jmockit.version>
         <mockito-core.version>4.4.0</mockito-core.version>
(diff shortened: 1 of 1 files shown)


Grimoren:
It seems that 2.13.2.1 is valid for most jackson with exception to jackson-bom:
https://mvnrepository.com/artifact/com.fasterxml.jackson/jackson-bom/2.13.2.20220324

Janpopan:
```xml
<dependency>
    <groupId>com.fasterxml.jackson</groupId>
    <artifactId>jackson-bom</artifactId>
    <version>2.13.2.20220324</version>
    <type>pom</type>
</dependency>
```

pom.xml
```diff
@@ -1094,6 +1094,12 @@ Copyright (c) 2012 - Jeremy Long
                 <artifactId>plexus-utils</artifactId>
                 <version>3.4.1</version>
             </dependency>
+            <dependency><!-- Temporary upgrade to fix jackson mess up. Be certain to delete once version is fixed -->
+                <groupId>com.fasterxml.jackson</groupId>
+                <artifactId>jackson-bom</artifactId>
+                <version>2.13.2.20220324</version>
+                <type>pom</type>
+            </dependency>
```
mprins, line 1102: to use a BOM in maven the `<scope>import</scope>` should be used

juergenzimmermann (changes requested):
Meanwhile there is Jackson BOM `2.13.2.20220328`. See https://github.com/FasterXML/jackson-databind/issues/3428

chadlwilson:
According to https://github.com/FasterXML/jackson-databind/issues/3428#issuecomment-1081330160 it seems the right way forward here is to roll forward to `jackson.version` of `2.13.2.2` which corrects the Gradle metadata causing problems for Gradle users (and dependency-check-gradle).

(Or move to a BOM-based approach entirely which would be more work, and involve removing the specific versions for `jackson-databind`)

pom.xml
philsegal: try using Jackson version `2.13.2.2` instead?

jeremylong:
superseded by #4285

PR #4299, opened 2022-04-01
#4293 Handle dependency with unspecified version to prevent NPE when scanning a lock file

### LINKED ISSUE #4293: NullPointerException occurred during package-lock.json analyze
**Describe the bug**
java.lang.NullPointerException occurred during analysis of a package-lock.json

7.0.4

**Log file**
<pre>
[INFO] Analysis Started
[INFO] Finished Archive Analyzer (0 seconds)
[INFO] Finished File Name Analyzer (0 seconds)
[INFO] Finished Dependency Merging Analyzer (0 seconds)
[INFO] Finished Version Filter Analyzer (0 seconds)
[INFO] Finished Hint Analyzer (0 seconds)
[INFO] Created CPE Index (1 seconds)
[INFO] Finished CPE Analyzer (2 seconds)
[INFO] Finished False Positive Analyzer (0 seconds)
[INFO] Finished NVD CVE Analyzer (0 seconds)
[WARN] The Node Package Analyzer has been disabled; the resulting report will only contain the known vulnerable dependency - not a bill of materials for the node project.
[WARN] An unexpected error occurred during analysis of 'C:\Users\TBATTI~1\AppData\Local\Temp\dctemp39951aff-f90a-4c1e-ab61-f13070969377\check12744923193935695818tmp\1\depcheckFrontend\package-lock.json' (Node Audit Analyzer): Cannot invoke "javax.json.JsonString.getString()" because the return value of "org.glassfish.json.JsonObjectBuilderImpl$JsonObjectImpl.getJsonString(String)" is null
[ERROR]
java.lang.NullPointerException: Cannot invoke "javax.json.JsonString.getString()" because the return value of "org.glassfish.json.JsonObjectBuilderImpl$JsonObjectImpl.getJsonString(String)" is null
        at org.glassfish.json.JsonObjectBuilderImpl$JsonObjectImpl.getString(JsonObjectBuilderImpl.java:257)
        at org.owasp.dependencycheck.data.nodeaudit.NpmPayloadBuilder.lambda$build$5(NpmPayloadBuilder.java:152)
        at java.base/java.util.Collections$UnmodifiableMap$UnmodifiableEntrySet.lambda$entryConsumer$0(Collections.java:1625)
        at java.base/java.util.LinkedHashMap$LinkedEntrySet.forEach(LinkedHashMap.java:708)
        at java.base/java.util.Collections$UnmodifiableMap$UnmodifiableEntrySet.forEach(Collections.java:1630)
        at org.owasp.dependencycheck.data.nodeaudit.NpmPayloadBuilder.build(NpmPayloadBuilder.java:147)
        at org.owasp.dependencycheck.analyzer.NodeAuditAnalyzer.legacyAnalysis(NodeAuditAnalyzer.java:249)
        at org.owasp.dependencycheck.analyzer.NodeAuditAnalyzer.analyzeDependency(NodeAuditAnalyzer.java:148)
[INFO] Finished Node Audit Analyzer (0 seconds)
[INFO] Finished Sonatype OSS Index Analyzer (0 seconds)
[INFO] Finished Vulnerability Suppression Analyzer (0 seconds)
[INFO] Finished Dependency Bundling Analyzer (0 seconds)
[INFO] Analysis Complete (2 seconds)
[INFO] Writing report to: C:\Users\tbattisti\Desktop\testDepCheckFE\dependency-check-7.0.4-release\dependency-check\.\dependency-check-report.html
[ERROR] Cannot invoke "javax.json.JsonString.getString()" because the return value of "org.glassfish.json.JsonObjectBuilderImpl$JsonObjectImpl.getJsonString(String)" is null
</pre>

**To Reproduce**
1. Download and unzip dependencyCheck from  https://github.com/jeremylong/DependencyCheck/releases/download/v7.0.4/dependency-check-7.0.4-release.zip
2. Download attacched [package-lock.zip](https://github.com/jeremylong/DependencyCheck/files/8389372/package-lock.zip)
3. Launch command from cmd `bin\dependency-check.bat --disableNodeJS --scan package-lock.zip`

nhumblot:
## Fixes Issue #

Fix #4293 

## Description of Change

Prevent a `NullPointerException` during lock file scan. This change handle the case where a version might not be specified for a dependency.

*yes*, one unit test has been modified to reproduce the NPE.

With the proposed changes, the package-json.zip file provided by the OP is analyzed without any issue and output an html report.

NB : I contributed for over a year to this project. I'll gladly take more responsibility in it if the maintainers wish to by inviting me to join as a collaborator. I think I may be able to assist the maintainers in the issue triage and management. 🙂

### DIFF
--- a/core/src/main/java/org/owasp/dependencycheck/analyzer/NodePackageAnalyzer.java
+++ b/core/src/main/java/org/owasp/dependencycheck/analyzer/NodePackageAnalyzer.java
@@ -37,6 +37,7 @@ import java.security.NoSuchAlgorithmException;
 import java.util.Arrays;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 import javax.annotation.concurrent.ThreadSafe;
 import javax.json.Json;
 import javax.json.JsonException;
@@ -282,7 +283,7 @@ public class NodePackageAnalyzer extends AbstractNpmAnalyzer {
      */
     public static boolean shouldSkipDependency(String name, String version, boolean optional, boolean fileExist) {
         // some package manager can handle alias, yarn for example, but npm doesn't support it
-        if (version.startsWith("npm:")) {
+        if (Objects.nonNull(version) && version.startsWith("npm:")) {
             //TODO make this an error that gets logged
             LOGGER.warn("dependency skipped: package.json contain an alias for {} => {} npm audit doesn't "
                     + "support aliases", name, version.replace("npm:", ""));
@@ -296,7 +297,7 @@ public class NodePackageAnalyzer extends AbstractNpmAnalyzer {
 
         // this seems to produce crash sometimes, I need to tests
         // using a local node_module is not supported by npm audit, it crash
-        if (version.startsWith("file:")) {
+        if (Objects.nonNull(version) && version.startsWith("file:")) {
             LOGGER.warn("dependency skipped: package.json contain an local node_module for {} seems to be "
                     + "located {} npm audit doesn't support locally referenced modules",
                     name, version.replace("file:", ""));
--- a/core/src/main/java/org/owasp/dependencycheck/data/nodeaudit/NpmPayloadBuilder.java
+++ b/core/src/main/java/org/owasp/dependencycheck/data/nodeaudit/NpmPayloadBuilder.java
@@ -20,6 +20,8 @@ package org.owasp.dependencycheck.data.nodeaudit;
 import org.owasp.dependencycheck.analyzer.NodePackageAnalyzer;
 
 import java.util.Map;
+import java.util.Objects;
+import java.util.Optional;
 import java.util.TreeMap;
 import java.util.stream.Collectors;
 import javax.json.Json;
@@ -149,7 +151,10 @@ public final class NpmPayloadBuilder {
                 if (entry.getValue().getValueType() == JsonValue.ValueType.OBJECT) {
                     final JsonObject dep = ((JsonObject) entry.getValue());
                     final String name = entry.getKey();
-                    version = dep.getString("version");
+                    version = Optional.ofNullable(dep.getJsonString("version"))
+                            .map(JsonString::getString)
+                            .orElse(null);
+
                     if (NodePackageAnalyzer.shouldSkipDependency(name, version)) {
                         return;
                     }
@@ -165,7 +170,7 @@ public final class NpmPayloadBuilder {
                         version = tmp;
                     }
                 }
-                requiresBuilder.add(entry.getKey(), "^" + version);
+                requiresBuilder.add(entry.getKey(), Objects.isNull(version) ? "*" : "^" + version);
             });
         }
         payloadBuilder.add("requires", requiresBuilder.build());
@@ -218,7 +223,9 @@ public final class NpmPayloadBuilder {
      */
     private static JsonObject buildDependencies(JsonObject dep, MultiValuedMap<String, String> dependencyMap) {
         final JsonObjectBuilder depBuilder = Json.createObjectBuilder();
-        depBuilder.add("version", dep.getString("version"));
+        Optional.ofNullable(dep.getJsonString("version"))
+                        .map(JsonString::getString)
+                        .ifPresent(version -> depBuilder.add("version", version));
 
         //not installed package (like, dependency of an optional dependency) doesn't contains integrity
         if (dep.containsKey("integrity")) {
--- a/core/src/test/java/org/owasp/dependencycheck/data/nodeaudit/NpmPayloadBuilderTest.java
+++ b/core/src/test/java/org/owasp/dependencycheck/data/nodeaudit/NpmPayloadBuilderTest.java
@@ -18,8 +18,6 @@
 package org.owasp.dependencycheck.data.nodeaudit;
 
 import java.io.InputStream;
-import java.util.HashMap;
-import java.util.Map;
 
 import org.junit.Assert;
 import org.junit.Test;
@@ -52,6 +50,11 @@ public class NpmPayloadBuilderTest {
                                                 .add("integrity", "sha512-nne9/IiQ/hzIhY6pdDnbBtz7DjPTKrY00P/zvPSm5pOFkl6xuGrGnXn/VtTNNfNtAfZ9/1RtehkszU9qcTii0Q==")
                                                 .add("dev", true)
                                 )
+                                .add("node_modules/jest-resolve",
+                                        Json.createObjectBuilder()
+                                                .add("dev", true)
+                                                .add("optional", true)
+                                                .add("peer", true))
                 );
 
         JsonObject packageJson = builder.build();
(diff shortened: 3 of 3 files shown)


aikebah:
Regarding the general approach: wouldn't you agree that supporting a version-less dependency (raising all vulnerabilities for any version of the dependency ) would be a way better approach than skipping the dependency when the package-lock contains no version? (which I would expect to be an improper package-lock by the way, as in my reading of NPMs package-lock documentation there should always be a version field).

nhumblot:
> Regarding the general approach: wouldn't you agree that supporting a version-less dependency (raising all vulnerabilities for any version of the dependency ) would be a way better approach than skipping the dependency when the package-lock contains no version? (which I would expect to be an improper package-lock by the way, as in my reading of NPMs package-lock documentation there should always be a version field).

Hello! :wave: 

> which I would expect to be an improper package-lock by the way, as in my reading of NPMs package-lock documentation there should always be a version field

Yes, absolutely, as stated by https://docs.npmjs.com/cli/v8/configuring-npm/package-lock-json#dependencies.

What made me make this choice was the OP package-lock.json inspection. My comment in #4293 is not detailed enough, I am sorry about that. Here are more details:

So as stated in #4293, the NPE is raised by the following dependency mention:
```
    "node_modules/jest-resolve": {
      "dev": true,
      "optional": true,
      "peer": true
    },
```
This is not the only place in the package-lock.json this dependency is mentioned. We do see it multiple times as it seems to be a transitive dependency of multiple `jest` modules.

For example, here it is mentioned in the `jest-runtime` resolution:

```
"node_modules/@jest/test-sequencer/node_modules/jest-runtime": {
      "version": "27.4.6",
      "resolved": "https://registry.npmjs.org/jest-runtime/-/jest-runtime-27.4.6.tgz",
      "integrity": "sha512-eXYeoR/MbIpVDrjqy5d6cGCFOYBFFDeKaNWqTp0h6E74dK0zLHzASQXJpl5a2/40euBmKnprNLJ0Kh0LCndnWQ==",
      "dev": true,
      "dependencies": {
        "@jest/environment": "^27.4.6",
        "@jest/fake-timers": "^27.4.6",
        "@jest/globals": "^27.4.6",
        "@jest/source-map": "^27.4.0",
        "@jest/test-result": "^27.4.6",
        "@jest/transform": "^27.4.6",
        "@jest/types": "^27.4.2",
        "chalk": "^4.0.0",
        "cjs-module-lexer": "^1.0.0",
        "collect-v8-coverage": "^1.0.0",
        "execa": "^5.0.0",
        "glob": "^7.1.3",
        "graceful-fs": "^4.2.4",
        "jest-haste-map": "^27.4.6",
        "jest-message-util": "^27.4.6",
        "jest-mock": "^27.4.6",
        "jest-regex-util": "^27.4.0",
        "jest-resolve": "^27.4.6",
        "jest-snapshot": "^27.4.6",
        "jest-util": "^27.4.2",
        "slash": "^3.0.0",
        "strip-bom": "^4.0.0"
      },
      "engines": {
        "node": "^10.13.0 || ^12.13.0 || ^14.15.0 || >=15.0.0"
      }
    },
```
`jest-resolve` ends up being resolved to the `27.4.6` version in the sample lock file. We can see it during the resolution of the transitive dependencies, like here for jasmine:

```
"node_modules/jest-jasmine2/node_modules/jest-resolve": {
      "version": "27.4.6",
      "resolved": "https://registry.npmjs.org/jest-resolve/-/jest-resolve-27.4.6.tgz",
      "integrity": "sha512-SFfITVApqtirbITKFAO7jOVN45UgFzcRdQanOFzjnbd+CACDoyeX7206JyU92l4cRr73+Qy/TlW51+4vHGt+zw==",
      "dev": true,
      "dependencies": {
        "@jest/types": "^27.4.2",
        "chalk": "^4.0.0",
        "graceful-fs": "^4.2.4",
        "jest-haste-map": "^27.4.6",
        "jest-pnp-resolver": "^1.2.2",
        "jest-util": "^27.4.2",
        "jest-validate": "^27.4.6",
        "resolve": "^1.20.0",
        "resolve.exports": "^1.1.0",
        "slash": "^3.0.0"
      },
      "engines": {
        "node": "^10.13.0 || ^12.13.0 || ^14.15.0 || >=15.0.0"
      }
    },
```

The cause of the issue is coming from `jest-pnp-resolver`, which reference `jest-resolve` as a peer dependency, without specifying any version:

```
    "node_modules/jest-pnp-resolver": {
      "version": "1.2.2",
      "resolved": "https://registry.npmjs.org/jest-pnp-resolver/-/jest-pnp-resolver-1.2.2.tgz",
      "integrity": "sha512-olV41bKSMm8BdnuMsewT4jqlZ8+3TCARAXjZGT9jcoSnrfUnRCqnMoF9XEeoWjbzObpqF9dRhHQj0Xb9QdF6/w==",
      "dev": true,
      "engines": {
        "node": ">=6"
      },
      "peerDependencies": {
        "jest-resolve": "*"
      },
      "peerDependenciesMeta": {
        "jest-resolve": {
          "optional": true
        }
      }
    },
```

I thought because it was mentioned as a peer dependency, the dependency itself must be declared somewhere else to be imported, so this very specific dependency object declared could be specified as a duplicate.

After you comments and more thought, I'd say there is at least a flaw in my implementation. I should check at least the dependency is flagged as "peer" when ignoring because of no version.

Your point looks also very valid. What if the only mention of a dependency is a wildcard as a peer? This looks kind of blurry as NPM seems to not have the same behavior given it's major version. I think it will end up being resolved by NPM as the latest version but some tests from my side would be needed.

The point I would like to clarify with you is what do you think should we do in the OP case, which is we have a wildcard defined but we end up having a specific version number resolved in the file? Should we ignore the wild card to prevent raising too much false positive?

I am putting this PR as a draft to prevent it being merged yet. Once you have the time to reply, I will work toward a solution that fits your specification (handling no version as wildcard + precision about the OP case).

Thank you for your help!

aikebah:
Checking for peer dependency would definitely be an improvement, however, not sure if it would be sufficient as npm appears to install the peer as needed when not installed (in versions 1.x, 2.x , removed in 3.x and re-added with improvements in 7.x) according to https://github.blog/2020-10-13-presenting-v7-0-0-of-the-npm-cli/

Not deep enough into the NPM ecosystem to be able to say for certain whether or not this would get reflected elsewhere in the package-lock.json or only take effect in the node_modules structure.

nhumblot:
> Checking for peer dependency would definitely be an improvement, however, not sure if it would be sufficient as npm appears to install the peer as needed when not installed (in versions 1.x, 2.x , removed in 3.x and re-added with improvements in 7.x) according to https://github.blog/2020-10-13-presenting-v7-0-0-of-the-npm-cli/
> 
> Not deep enough into the NPM ecosystem to be able to say for certain whether or not this would get reflected elsewhere in the package-lock.json or only take effect in the node_modules structure.

Ok, so I will go with what you suggested in your first comment. It will be enough to fix the bug. I will update this PR and ask for a new review once I am done. Thank you. 👍

jeremylong:
@nhumblot can you email me? jeremy.long @ gmail.com.

nhumblot:
Updated the change so Dependency Check does not ignore a dependency if it does not have a version.

aikebah (approved):
Looks good to me; prefer to have @jeremylong also review before merging in

jeremylong (approved):
Looks good to me too - thanks!

PR #4302, opened 2022-04-03
fix reproducible build issues

hboutemy:
## Description of Change

fix reproducible builds issues found when rebuilding 7.0.4 release: https://github.com/jvm-repo-rebuild/reproducible-central/blob/master/content/org/owasp/dependency-check/dependency-check-parent-7.0.4.diffoscope

there is only 1 unknown issue: why is the `HelpMojo.class` not reproduced? I have no clue yet...

at least with this PR, only 1 jar won't be successful

no

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -488,6 +488,7 @@ Copyright (c) 2012 - Jeremy Long
                         <excludeProperty>git.build.user.*</excludeProperty>
                         <excludeProperty>git.build.host</excludeProperty>
                         <excludeProperty>git.commit.user.*</excludeProperty>
+                        <excludeProperty>git.commit.count</excludeProperty>
                     </excludeProperties>
                 </configuration>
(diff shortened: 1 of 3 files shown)


jeremylong (approved):

pom.xml
```diff
@@ -488,6 +488,7 @@ Copyright (c) 2012 - Jeremy Long
                         <excludeProperty>git.build.user.*</excludeProperty>
                         <excludeProperty>git.build.host</excludeProperty>
                         <excludeProperty>git.commit.user.*</excludeProperty>
+                        <excludeProperty>git.commit.count</excludeProperty>
```
nhumblot, line 491: Should have been `git.total.commit.count` :)

hboutemy: yes :)
and `git.tags` also needs to be excluded

Issue #2221, opened 2019-09-27
Unable to check JAR files with invalid or unverifiable signature


wilx:
**Describe the bug**

We are seeing in our builds some exceptions related to JAR file verifier. Please provide a way to turn off signature verification of opened JAR files.

```
00:22:53.405 [WARNING] An unexpected error occurred during analysis of 'dctemp1109c568-9e6e-4c7d-b275-7909826af292\check7808464953291941319tmp\996\plugins\org.apache.felix.gogo.command_0.10.0.v201209301215.jar' (Jar Analyzer): Invalid signature file digest for Manifest main attributes
00:22:53.405 [ERROR] 
00:22:53.406 java.lang.SecurityException: Invalid signature file digest for Manifest main attributes
00:22:53.406 	at sun.security.util.SignatureFileVerifier.processImpl(SignatureFileVerifier.java:314)
00:22:53.406 	at sun.security.util.SignatureFileVerifier.process(SignatureFileVerifier.java:268)
00:22:53.406 	at java.util.jar.JarVerifier.processEntry(JarVerifier.java:316)
00:22:53.406 	at java.util.jar.JarVerifier.update(JarVerifier.java:228)
00:22:53.406 	at java.util.jar.JarFile.initializeVerifier(JarFile.java:383)
00:22:53.406 	at java.util.jar.JarFile.getInputStream(JarFile.java:450)
00:22:53.406 	at org.owasp.dependencycheck.analyzer.JarAnalyzer.extractPom(JarAnalyzer.java:529)
00:22:53.406 	at org.owasp.dependencycheck.analyzer.JarAnalyzer.analyzePOM(JarAnalyzer.java:408)
00:22:53.406 	at org.owasp.dependencycheck.analyzer.JarAnalyzer.analyzeDependency(JarAnalyzer.java:290)
00:22:53.406 	at org.owasp.dependencycheck.analyzer.AbstractAnalyzer.analyze(AbstractAnalyzer.java:131)
00:22:53.406 	at org.owasp.dependencycheck.AnalysisTask.call(AnalysisTask.java:88)
00:22:53.406 	at org.owasp.dependencycheck.AnalysisTask.call(AnalysisTask.java:37)
00:22:53.406 	at java.util.concurrent.FutureTask.run(FutureTask.java:266)
00:22:53.406 	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1142)
[...]

PR #4424, opened 2022-04-28
Do not verify JAR files when analyzing

jeremylong:
## Fixes Issue #2221

Disable signature verification when analyzing JAR files. See #2221.


aikebah (approved):
Looks good to me. No sense indeed in validating signing as we're only interested in discovery of the dependencies.

Issue #4368, opened 2022-04-14
[FP]: CVE-2021-43138 on com.ibm.async/asyncutil@0.1.0


Felk:
### Package URl

pkg:maven/com.ibm.async/asyncutil@0.1.0

### CPE

`cpe:2.3:a:async_project:async:0.1.0:*:*:*:*:*:*:*`

### CVE

CVE-2021-43138

### ODC Integration

{"label"=>"Maven Plugin"}

### ODC Version

7.0.4

### Description

I believe `asyncutil-0.1.0.jar` has nothing to do with the js library `async`

ghost:
Are these automated, or is a PR required?

marcelstoer:
The bot is so clever in what it already does (from what I see in the comment above) creating a PR for `dependencycheck-base-suppression.xml` wouldn't need much extra I guess 😉 But maybe that's not desirable as vulnerability assessment is no exact science.

ghost:
if the PR was created there is no suggestion that it should be accepted without review...

anyway, if i want these accepted, now i know to create a PR 👍

marcelstoer:
> if the PR was created there is no suggestion that it should be accepted without review..

😄 of course, point taken.

Lars5678:
CPE is cpe:2.3:a:async_project:async:0.1.0:*:*:*:*:*:*:* wich refers to https://github.com/caolan/async

marcelstoer:
Will be included with 7.1.1.

PR #4428, opened 2022-04-29
Suppress asyncutil vs. async_project FP

marcelstoer:
## Fixes Issue #
#4368

## Description of Change

Suppress asyncutil vs. async_project FP

*yes/no*


aikebah (approved):
Thanks for the PR

PR #4511, opened 2022-05-18
fix: Use exit codes between 0 and 255 (inclusive)

candrews:
## Fixes Issue #

Shells express exit codes in the range of 0-255. Anything outside of that range is brought into that range by modding it 256.

For example:
-1 becomes 255
-2 becomes 254
-3 becomes 253

This need to convert exit codes in shell scripts to use dependency check is unexpected and confusing.

For that reason, returning exit codes outside of 0-255 is unusual and generally discouraged. 

See: https://tldp.org/LDP/abs/html/exit-status.html

## Description of Change

Therefore, use exit codes in the 0-255 range by changing all negative exit codes to positive ones (ex, -1 becomes 1).

With this change, exit code 1 was used for both help and CVSS score failure. Therefore, the CVSS score failure exit code was changed to 15.

no


aikebah:
The only handling of return code in a script in my view would be a check for success or failure, which would mean comparison of the exit code to 0 (SUCCESS) and potentially a print of the non-zero exit code received for diagnostic purposes.
That check would be transparent to the exit codes used.

https://unix.stackexchange.com/questions/418784/what-is-the-min-and-max-values-of-exit-codes-in-linux
shows that what exit code ranges are available depend on the shell

Valid exit codes available to POSIX compliant systems is the full range of int as per https://pubs.opengroup.org/onlinepubs/9699919799/functions/_exit.html / https://pubs.opengroup.org/onlinepubs/9699919799/functions/exit.html
 (with typically the least significant byte being made available to shell scripts)

candrews:
> The only handling of return code in a script in my view would be a check for success or failure

My use case is that I want to differentiate between 3 exit reasons when using the `--failOnCVSS` argument:
1. Dependency check ran and found no vulnerabilities
2. Dependency check ran and found vulnerability
3. Dependency checked failed to run for some reason

With this MR, I'd check for exit codes 0, 11, and anything else, respectively.



> https://unix.stackexchange.com/questions/418784/what-is-the-min-and-max-values-of-exit-codes-in-linux
> shows that what exit code ranges are available depend on the shell

Agreed. Most users use `bash`, `dash`, or busybox `sh` - and all of those truncate to 0-255. And that furthers my point that using exit codes outside of the 0-255 range results in unexpected behavior which is why its generally recommended to stick to that range.

> with typically the least significant byte being made available to shell scripts

"Typically" indicates inconsistent behavior, which is generally a good idea to avoid.

cli/src/main/java/org/owasp/dependencycheck/App.java
```diff
@@ -302,7 +302,7 @@ private int determineReturnCode(Engine engine, float cvssFailScore) {
                         || (v.getCvssV3() != null && v.getCvssV3().getBaseScore() >= cvssFailScore)
                         || (v.getUnscoredSeverity() != null && SeverityUtil.estimateCvssV2(v.getUnscoredSeverity()) >= cvssFailScore)
                         || (cvssFailScore <= 0.0f)) { //safety net to fail on any if for some reason the above misses on 0
-                    retCode = 1;
+                    retCode = 11;
```
aikebah:
```suggestion
                    retCode = 15;
```

As you intend to be able to discover 'failure for vulnerabilities' separate from other error cases the returncode should be unique and not overlap with line 196 (Database error)

candrews: Oops! Thank you :)

aikebah:
@jeremylong what do you think. Looks reasonable to me, but think we should postpone integrating for an 8.x release as it will break existing returncode handling in scripted runs.

jeremylong (approved):

jeremylong:
I agree with the PR - but it would be a breaking change. This will get merged, but we might have a release or two happen before we move to 8.0.0.

PR #4568, opened 2022-06-02
Fix compilation error in cli

aikebah:
## Description of Change

Fixup the missing initialization of the score variable in the CLI code to resolve the compilation failure (variable score might not have been initialized)

## Have test cases been added to cover the new functionality?

no

PR #4572, opened 2022-06-06
Fixup XML template for related dependencies Sha256sum

aikebah:
## Description of Change

Fixup an incorrect variable reference in the template that made all relatedDependencies list the sha256 sum of the parent dependency

no, manually verified


aikebah:
Build failure should be resolved when #4568 is integrated

jeremylong (approved):

PR #4624, opened 2022-06-25
4499 infinite loop in cmakeanalyzer

### LINKED ISSUE #4499: Infinite loop in CMakeAnalyzer
**Describe the bug**
CMakeAnalyzer contains an infinite loop when cmake temp variable is used to store old value and then restore it afterwards.
The while loop from https://github.com/jeremylong/DependencyCheck/blob/main/core/src/main/java/org/owasp/dependencycheck/analyzer/CMakeAnalyzer.java#L190 is responsible for a variable substitution from the map, which is calculated with collectDefinedVariables method. Real example can be found here: https://gitlab.com/libtiff/libtiff/-/blob/master/cmake/FindDeflate.cmake.
The cmake file contains the first set to store the old value [set(Deflate_OLD_FIND_LIBRARY_PREFIXES "${CMAKE_FIND_LIBRARY_PREFIXES}")](https://gitlab.com/libtiff/libtiff/-/blob/master/cmake/FindDeflate.cmake#L44) and the second set to restore the old value [set(CMAKE_FIND_LIBRARY_PREFIXES "${Deflate_OLD_FIND_LIBRARY_PREFIXES}")](https://gitlab.com/libtiff/libtiff/-/blob/master/cmake/FindDeflate.cmake#L61). Vars variable (hash map) contains both values and the code continues to replace them in the infinite loop.
Another similar test file: https://gitlab.com/libtiff/libtiff/-/blob/master/cmake/FindJBIG.cmake.

The problem occurs using version 7.1.0 of the cli

**Log file**
[gist](https://gist.github.com/dr0ndv/ab7b8b07c4d5d07c537e0a9079bdb5de)

**To Reproduce**
1. Download the latest DependencyCheck release.
2. Clone libtiff repository: https://gitlab.com/libtiff/libtiff
3. Execute: dependency_check.sh -s <path_to_libtiff> --enableExperimental
4. DependencyCheck hangs

**Expected behavior**
No infinite loop.

nhumblot:
## Fixes Issue #

Fix #4499 

## Description of Change

This change prevent creating an infinite loop during variable resolution when variables references each other. For each variable, it tries to build a resolution chain to see if the initial variable ends up pointing to itself. In such case, this variable resolution is discarded as it cannot be performed.

*yes*, two tests based on the example displayed in the initial issue.

The first one validates the chain resolution method specifically to enforce self referencing variables are not passed to the analyzer to prevent the infinite loop.

The second one validates this new method is called during the analysis process, validating the infinite loop does not occurs.

### DIFF
--- a/core/src/main/java/org/owasp/dependencycheck/analyzer/CMakeAnalyzer.java
+++ b/core/src/main/java/org/owasp/dependencycheck/analyzer/CMakeAnalyzer.java
@@ -174,7 +174,7 @@ public class CMakeAnalyzer extends AbstractFileTypeAnalyzer {
         }
         if (StringUtils.isNotBlank(contents)) {
             final HashMap<String, String> vars = new HashMap<>();
-            collectDefinedVariables(dependency, engine, contents, vars);
+            collectDefinedVariables(contents, vars);
 
             String contentsReplacer = contents;
             Matcher r = INL_VAR_REGEX.matcher(contents);
@@ -241,13 +241,10 @@ public class CMakeAnalyzer extends AbstractFileTypeAnalyzer {
     /**
      * Collect defined CMake variables
      *
-     * @param dependency the dependency being analyzed
-     * @param engine the dependency-check engine
      * @param contents the version information
      * @param vars map of variable replacement tokens
      */
-    private void collectDefinedVariables(Dependency dependency, Engine engine, String contents,
-            HashMap<String, String> vars) {
+    private void collectDefinedVariables(String contents, HashMap<String, String> vars) {
         final Matcher m = SET_VAR_REGEX.matcher(contents);
         int count = 0;
         while (m.find()) {
--- a/core/src/main/java/org/owasp/dependencycheck/analyzer/CMakeAnalyzer.java
+++ b/core/src/main/java/org/owasp/dependencycheck/analyzer/CMakeAnalyzer.java
@@ -355,7 +355,7 @@ public class CMakeAnalyzer extends AbstractFileTypeAnalyzer {
      * @return a new map without infinite chain variables
      */
     Map<String, String> removeSelfReferences(final Map<String, String> vars) {
-        Map<String, String> resolvedVars = new HashMap<>();
+        final Map<String, String> resolvedVars = new HashMap<>();
 
         vars.forEach((key, value) -> {
             if (!isVariableSelfReferencing(vars, key)) {
@@ -367,12 +367,12 @@ public class CMakeAnalyzer extends AbstractFileTypeAnalyzer {
     }
 
     private boolean isVariableSelfReferencing(Map<String, String> vars, String key) {
-        List<String> resolutionChain = new ArrayList<>();
+        final List<String> resolutionChain = new ArrayList<>();
         resolutionChain.add(key);
 
         String nextKey = resolutionChain.get(0);
         do {
-            Matcher matcher = INL_VAR_REGEX.matcher(vars.get(nextKey));
+            final Matcher matcher = INL_VAR_REGEX.matcher(vars.get(nextKey));
             if (!matcher.find()) {
                 break;
             }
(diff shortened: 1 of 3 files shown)


core/src/main/java/org/owasp/dependencycheck/analyzer/CMakeAnalyzer.java
```diff
@@ -340,4 +341,45 @@ private void analyzeSetVersionCommand(Dependency dependency, Engine engine, Stri
...
+        return resolvedVars;
+    }
+
+    private boolean isVariableSelfReferencing(Map<String, String> vars, String key) {
+        List<String> resolutionChain = new ArrayList<>();
+        resolutionChain.add(key);
+
+        String nextKey = resolutionChain.get(0);
+        do {
+            Matcher matcher = INL_VAR_REGEX.matcher(vars.get(nextKey));
```
nhumblot: Fixed by f856df49853437fd9a571f9f75d821c1cd11cbe7

core/src/main/java/org/owasp/dependencycheck/analyzer/CMakeAnalyzer.java
```diff
@@ -340,4 +341,45 @@ private void analyzeSetVersionCommand(Dependency dependency, Engine engine, Stri
...
+            if (!isVariableSelfReferencing(vars, key)) {
+                resolvedVars.put(key, value);
+            }
+        });
+
+        return resolvedVars;
+    }
+
+    private boolean isVariableSelfReferencing(Map<String, String> vars, String key) {
+        List<String> resolutionChain = new ArrayList<>();
```
nhumblot: Fixed by f856df49853437fd9a571f9f75d821c1cd11cbe7

core/src/main/java/org/owasp/dependencycheck/analyzer/CMakeAnalyzer.java
```diff
@@ -340,4 +341,45 @@ private void analyzeSetVersionCommand(Dependency dependency, Engine engine, Stri
...
+     *
+     * This method takes the resolved variables map as an input and will return
+     * a new map, without the keys generate an infinite resolution chain.
+     *
+     * @param vars variables initialization detected in the CMake build file
+     *
+     * @return a new map without infinite chain variables
+     */
+    Map<String, String> removeSelfReferences(final Map<String, String> vars) {
+        Map<String, String> resolvedVars = new HashMap<>();
```
nhumblot: Fixed by f856df49853437fd9a571f9f75d821c1cd11cbe7

core/src/main/java/org/owasp/dependencycheck/analyzer/CMakeAnalyzer.java
```diff
@@ -241,13 +244,10 @@ protected void analyzeDependency(Dependency dependency, Engine engine) throws An
...
      *
-     * @param dependency the dependency being analyzed
-     * @param engine the dependency-check engine
      * @param contents the version information
-     * @param vars map of variable replacement tokens
      */
-    private void collectDefinedVariables(Dependency dependency, Engine engine, String contents,
-            HashMap<String, String> vars) {
+    private Map<String, String> collectDefinedVariables(String contents) {
+        Map<String, String> vars = new HashMap<>();
```
nhumblot: Fixed by 7a0225b5be718004d63961d55e382c858d6a8d3f

core/src/main/java/org/owasp/dependencycheck/analyzer/CMakeAnalyzer.java
```diff
@@ -241,13 +244,10 @@ protected void analyzeDependency(Dependency dependency, Engine engine) throws An
...
      * Collect defined CMake variables
      *
-     * @param dependency the dependency being analyzed
-     * @param engine the dependency-check engine
      * @param contents the version information
-     * @param vars map of variable replacement tokens
      */
-    private void collectDefinedVariables(Dependency dependency, Engine engine, String contents,
-            HashMap<String, String> vars) {
+    private Map<String, String> collectDefinedVariables(String contents) {
```
nhumblot, line 251: Fixed by 7a0225b5be718004d63961d55e382c858d6a8d3f

jeremylong (approved):

Issue #4637, opened 2022-07-01
[FP]: log4j-core@2.17.2 - AWS - CVE-2022-33915


adam-siklosi:
### Package URl

pkg:maven/org.apache.logging.log4j/log4j-core@2.17.2

### CPE

cpe:2.3:a:apache:log4j:2.17.2:*:*:*:*:*:*:*

### CVE

CVE-2022-33915

### ODC Integration

{"label"=>"CLI"}

### ODC Version

7.1.0

### Description

CVE-2022-33915 description says:

> Versions of the Amazon AWS Apache Log4j hotpatch package before log4j-cve-2021-44228-hotpatch-1.3.5 are affected by a race condition that could lead to a local privilege escalation. This Hotpatch package is not a replacement for updating to a log4j version that mitigates CVE-2021-44228 or CVE-2021-45046; it provides a temporary mitigation to CVE-2021-44228 by hotpatching the local Java virtual machines. To do so, it iterates through all running Java processes, performs several checks, and executes the Java virtual machine with the same permissions and capabilities as the running process to load the hotpatch. A local user could cause the hotpatch script to execute a binary with elevated privileges by running a custom java process that performs exec() of an SUID binary after the hotpatch has observed the process path and before it has observed its effective user ID.

I don't think an AWS hotpatch should be related to the general build log4j-core 2.17.2.
Additionally log4j1 and log4j2 shouldn't mixed up.

adam-siklosi:
I'm not sure if the generated suppression is good. Seems like log4j1 and log4j2 has the same cpe. Perhaps that is also the source of the problem. 

So personally I will only suppress the specific CVE. Note that it can also pop up for core and the api package as well.
```
    <suppress>
        <notes><![CDATA[
   FP per issue #4637
   ]]></notes>
        <packageUrl regex="true">^pkg:maven/org\.apache\.logging\.log4j/log4j\-.*@.*$</packageUrl>
        <cve>CVE-2022-33915</cve>
    </suppress>
```

adam-siklosi:
I think i figured it out now... :-) 

Seems like log4j-cve-2021-44228-hotpatch-1.3.5 is about Log4j2 as the mentioned CVE in the CVE-2022-33915 text are all for log4j version 2:
- https://nvd.nist.gov/vuln/detail/CVE-2021-44228
- https://nvd.nist.gov/vuln/detail/CVE-2021-45046
[...]

PR #4638, opened 2022-07-01
Addressing CVE-2022-33915 (log4j AWS)

marcelstoer:
## Fixes Issue

#4637

## Description of Change

From what I understand CVE-2022-33915 is way to broad. The vulnerability appears to be only affecting AWS builds of log4j(2).

*yes/no*


marcelstoer:
Log4j is everywhere... Hence, I would appreciate if this could be addressed asap and then released with 7.1.2.

Issue #4469, opened 2022-05-04
Invalid/inapropriate JUnit report file when using in combination with Jenkins multi-module build and xUnit plugin


haumacher:
I'm trying to use the dependency-check plugin in a multi-module maven build from Jenkins CI. I try to produce a test report in Jenkins from all generated JUnit report files in all directories `*/target/dependency-check-junit.xml` using the xUnit plugin. I'm facing two problems: 

1. The xUnit plugin complains that generated the JUnit report is invalid, because "Attribute 'errors' must appear on element 'testsuite'.". This results in no report generated at all.
2. When fixing problem 1 with an XSLT afterburner, I get a check report, but from the report, I cannot determine the concrete module that has the problematic dependency. This is because the JUnit report does not contain the module that causes the problem. In a build with 100+ modules this is really cumbersome to find the source of the problematic dependency:

![grafik](https://user-images.githubusercontent.com/5607145/166675659-7427e2e2-11d5-4fda-abe8-61cdf17d3e08.png)

**Version of dependency-check used**
```
   <groupId>org.owasp</groupId>
   <artifactId>dependency-check-maven</artifactId>
   <version>7.1.0</version>
```

**Log file**

Not the log file of the dependency check, but the Jenkins log of the xUnit plugin that tries to consume its result:

```
INFO: Processing JUnit
INFO: [JUnit] - 2 test report file(s) were found with the pattern '*/target/dependency-check-junit.xml' relative to '/var/lib/jenkins/workspace/CheckDependencies' for the testing framework 'JUnit'.
WARNING: The file '/var/lib/jenkins/workspace/CheckDependencies/ext.com.meterware.httpunit/target/dependency-check-junit.xml' is an invalid file.
WARNING: At line 1 of file:/var/lib/jenkins/workspace/CheckDependencies/ext.com.meterware.httpunit/target/dependency-check-junit.xml:cvc-complex-type.4: Attribute 'errors' must appear on element 'testsuite'.
WARNING: At line 1 of file:/var/lib/jenkins/workspace/CheckDependencies/ext.com.meterware.httpunit/target/dependency-check-junit.xml:cvc-complex-type.4: Attribute 'errors' must appear on element 'testsuite'.
WARNING: At line 1 of file:/var/lib/jenkins/workspace/CheckDependencies/ext.com.meterware.httpunit/target/dependency-check-junit.xml:cvc-complex-type.4: Attribute 'errors' must appear on element 'testsuite'.
[...]

PR #4664, opened 2022-07-11
Alternative enhancement of the JUnit report

aikebah:
## Description of Change

An alternative implementation to the one proposed in #4476 which in my view better fits the way we use the JUnit report format for reporting on vulnerabilities.
It extends the `stderr` location data with the project references rather than including the project reference in the vulnerability (which, besides being ugly in my view) would not cleanly work for an aggreggate scan that has multiple sources).
Just like the #4476 it also resolves the missing required attributes that in the current reports make the xUnit Jenkins plugin break on XSD violations.

no, screenshots of sample results where added in a [comment](https://github.com/jeremylong/DependencyCheck/issues/4469#issuecomment-1166324079) in the resolved issue

### DIFF
--- a/core/src/main/resources/templates/junitReport.vsl
+++ b/core/src/main/resources/templates/junitReport.vsl
@@ -87,7 +87,17 @@
                 #foreach($ref in $vuln.getReferences())
                     #if($ref.url)$enc.xml($ref.name): $enc.xml($ref.url)#end
                 #end]]>#end</system-out>
-            <system-err>location: $enc.xml($dependency.FilePath)</system-err>
+            #set($referencedProjects="")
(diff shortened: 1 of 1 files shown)


aikebah:
@jeremylong can you take a look at both this and #4476 and see which approach you think would fit best?
Having project reference information somewhere in the JUnit reports I think is beneficial to users of that report. And of course the XSD-violation triggered by tools that use the unofficial, but commonly referenced XSD to validate the report before processing should be resolved.

jeremylong (approved):

jeremylong:
@aikebah I like your approach better.  Thanks!

PR #4687, opened 2022-07-21
Log when non-base suppressions rules are unused

### LINKED ISSUE #4685: Identify suppressions that are no longer needed.
**Description**
We add a suppression when really needed but at some point the suppression may become unnecessary. For example the dependency is updated (often the issue is hidden in a transitive dependency) or it is identified as a false positive and is no longer needed after a plugin update.

When we do an upgrade of our core dependencies every 3 weeks or an upgrade of the dependency check plugin we manually remove all suppressions and put back the ones we need.

This is to keep our suppression list as small as possible since we need to review them periodically.
But this cleaning process is time consuming and error prone.

**Suggested Improvement**
It'd be great to have a flag on the plugin that would report suppressions that are no longer needed. Then we could remove them more easily.

**Other Options Considered**
We considered adding an "until"  date to each suppression but we don't want false positives or vulnerabilities that are not applicable to our software to reappear in future. 
We only use "until" when we really have an issue and want to temporarily suppress to allow builds anyway.

jeremylong:
Fixes #4685

This might be useful to some - a log message has been added when a non-base suppression rule was not used.

### DIFF
--- a/core/src/main/java/org/owasp/dependencycheck/analyzer/AbstractSuppressionAnalyzer.java
+++ b/core/src/main/java/org/owasp/dependencycheck/analyzer/AbstractSuppressionAnalyzer.java
@@ -113,7 +113,12 @@ public abstract class AbstractSuppressionAnalyzer extends AbstractAnalyzer imple
         if (rules.isEmpty()) {
             return;
         }
-        rules.forEach((rule) -> rule.process(dependency));
+        rules.forEach((rule) -> {
+            rule.process(dependency);
+            if (!rule.isMatched() && !rule.isBase()) {
+                LOGGER.debug("Suppression Rule had zero matches: {}", rule.toString());
+            }
+        });
     }
 
     /**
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/owasp/dependencycheck/analyzer/AbstractSuppressionAnalyzer.java
```diff
@@ -113,7 +113,12 @@ protected void analyzeDependency(Dependency dependency, Engine engine) throws An
         if (rules.isEmpty()) {
             return;
         }
-        rules.forEach((rule) -> rule.process(dependency));
+        rules.forEach((rule) -> {
+            rule.process(dependency);
+            if (!rule.isMatched() && !rule.isBase()) {
```
aikebah, line 118: Checking for non-usage should be a post-processing activity, as many rules used in a full analysis will not be hit for a specific dependency (as they apply to a different dependency).

jeremylong: Thank you. I completely forgot that these could run several times. I'll update this shortly.

jeremylong:
superseded by #4709

PR #4701, opened 2022-07-25
Update htmlReport.vsl

mysmlz:
## Fixes Issue #

After the page is rendered, a wrong class on "Published Vulnerabilities" causes inconsistency with the click effect.

## Description of Change

Update htmlReport.vsl

- before fixed:
image

- fixed:
image

yes


jeremylong:
Thanks @mysmlz - however, this change is not complete - as you cannot collapse the `Published Vulnerabilities` section after the update. This is because line 40 needs to be updated from:

```javascript
$(".expandable").click(function (event) {
```

To:

```javascript
$(".expandable, .collapsed").click(function (event) {
```

Any chance you can add the update to this PR?

Thank you!

jeremylong (changes requested):
The JavaScript needs to be updated to account for this change.

PR #4709, opened 2022-07-27
Log when non-base suppressions rules are unused

### LINKED ISSUE #4685: Identify suppressions that are no longer needed.

jeremylong:
Supersedes #4687 

Fixes https://github.com/jeremylong/DependencyCheck/issues/4685

This might be useful to some - a log message has been added when a non-base suppression rule was not used.


aikebah:
Need to get my head around the why, but it appears to break the suppression of https://github.com/jeremylong/DependencyCheck/commit/83e0f647912716a71b58306a7368809cb967d796

causing integration-test 629 to fail as cbor gets bundled with ion as the CVE is no longer suppressed for ion

jeremylong:
I've been off and on looking at this as time permits - I haven't been able to figure out why this PR breaks 629 and does not suppress the CVE for ion...

aikebah:
> I've been off and on looking at this as time permits - I haven't been able to figure out why this PR breaks 629 and does not suppress the CVE for ion...

Added a few pieces of targeted debugging and finally got around the root-cause... the singleton is reused in CPESuppressionAnalyzer and VulnerabilitySuppressionAnalyzer, so rather that 55 CVE suppressions + 3 vulnerabilityName suppressions the VulnerabilitySuppressionAnalyzer runs with the same 508 CPE suppression rules that the CPESuppressionAnalyzer uses
```
[WARNING] CpeSuppressionAnalyzer with 508 suppressionRules
...
[WARNING] VulnerabilitySuppressionAnalyzer with 508 suppressionRules
```

The base suppressions have 560 rules, 49 contain only CVEs, 3 only vulnerabiltyNames, 502 only CPE and 6 a mixture of CPE and CVE. So out of the 58 rules that VulnerabilitySuppressionAnalyzer should take into account only the 6 that also have CPE suppressions are taken into account.

So filtering out the applicable rules needs to move from rules-loading to suppression processing evaluation or we need separate CPE and vulnerabilitySuppression ruleset singletons.

core/src/main/java/org/owasp/dependencycheck/analyzer/AbstractSuppressionAnalyzer.java
```diff
@@ -113,9 +122,24 @@ protected void analyzeDependency(Dependency dependency, Engine engine) throws An
         if (rules.isEmpty()) {
             return;
         }
-        rules.forEach((rule) -> rule.process(dependency));
+        int ctr = 0;
+        for (SuppressionRule rule : rules.list()) {
+            if (filter(rule)) {
+                ctr++;
+                rule.process(dependency);
+            }
```
aikebah: A left-over counter from initial debugging? It is only ever written to, never used.

```suggestion
        for (SuppressionRule rule : rules.list()) {
            if (filter(rule)) {
                rule.process(dependency);
            }
```

core/src/main/java/org/owasp/dependencycheck/analyzer/AbstractSuppressionAnalyzer.java
```diff
@@ -113,9 +122,24 @@ protected void analyzeDependency(Dependency dependency, Engine engine) throws An
...
+        }
     }
 
+    /**
+     * Determines whether or not to filter a suppression rule.
+     *
+     * @param rule the suppression rule to evaluate
+     * @return <code>true</code> if the rule should be filtered; otherwise
+     * <code>true</code>
```
aikebah: @return documented to always retain true. And I think the description can be improved to be more clear on whether it indicates 'filtering out' or 'filtering in' of rules.
```suggestion
     * Determines whether a suppression rule should be retained when filtering a set of suppression rules for a concrete suppression analyzer.
     *
     * @param rule the suppression rule to evaluate
     * @return <code>true</code> if the rule should be retained; otherwise
     * <code>false<code>
```

core/src/main/java/org/owasp/dependencycheck/xml/suppression/SuppressionHandler.java
```diff
@@ -120,13 +116,9 @@ public List<SuppressionRule> getSuppressionRules() {
...
-     *
-     * @param filter The suppression rule filter used when loading the
-     * suppression rules. This is used to differentiate vulnerability
-     * suppression rules from CPE suppression rules.
      */
-    public SuppressionHandler(SuppressionRuleFilter filter) {
-        this.filter = filter;
+    public SuppressionHandler() {
+
     }
```
aikebah: As this is the sole constructor and it's now empty it can be fully left out from the codebase.

aikebah (changes requested):
Looks good, some minor improvements requested

aikebah (approved):

PR #4744, opened 2022-08-11
IntelliJ Code Inspection/Correction

jeremylong:
Finally went through a bunch of warnings from IntelliJ IDEA. The PR fixes several minor issues.


core/src/main/java/org/owasp/dependencycheck/analyzer/DependencyBundlingAnalyzer.java
```diff
@@ -461,16 +461,12 @@ protected boolean isWebJar(Dependency dependency, Dependency nextDependency) {
         final String nextName = nextDependency.getFileName().toLowerCase();
         if (mainName.endsWith(".jar") && nextName.endsWith(".js") && nextName.startsWith(mainName)) {
             return dependency.getSoftwareIdentifiers()
-                    .stream().map(id -> id.getValue()).collect(toSet())
-                    .containsAll(nextDependency.getSoftwareIdentifiers().stream().map(id -> {
-                        return identifierToWebJarForCompairson(id);
-                    }).collect(toSet()));
+                    .stream().map(Identifier::getValue).collect(toSet())
+                    .containsAll(nextDependency.getSoftwareIdentifiers().stream().map(this::identifierToWebJarForCompairson).collect(toSet()));
```
aikebah: May want to fix the typo in `identifierToWebJarForCompairson` by code refactoring while you're at it (should be `identifierToWebJarForComparison`)

core/src/main/java/org/owasp/dependencycheck/analyzer/NvdCveAnalyzer.java
```diff
@@ -193,14 +189,12 @@ private boolean ecosystemMatchesTargetSoftware(String ecosystem, String targetSo
...
                 case "node.js":
+                case "node_js":
                     return true;
                 //not actually in NVD...just future proofing
                 case "npm":
                     return true;
-                case "node_js":
-                    return true;
                 case "node-js":
                     return true;
```
aikebah: Shouldn't these be all combined in a single fall-through case block? (unfortunately github doesn't allow me to suggest the alternate coding due to the deleted lines)

core/src/main/java/org/owasp/dependencycheck/analyzer/PnpmAuditAnalyzer.java
```diff
@@ -148,16 +148,14 @@ protected void prepareFileTypeAnalyzer(Engine engine) throws InitializationExcep
...
-                        LOGGER.warn("The {} has been disabled. Pnpm executable was not found.", getName());
                     default:
                         this.setEnabled(false);
                         LOGGER.warn("The {} has been disabled. Pnpm executable was not found.", getName());
                 }
             }
         } catch (Exception ex) {
             this.setEnabled(false);
-            LOGGER.debug("The {} has been disabled. Pnpm executable was not found.", ex);
+            LOGGER.debug("Pnpm executable was not found.", ex);
```
aikebah: I think like with the YarnAuditAnalyzer this log can be fully removed as we also throw a new exception with ex as the cause, that gets logged as an error in the Engine, so no information is lost here. Keeping it would duplicate the stacktrace in the logs when debug is enabled as the Engine will also log the exception stacktrace at debug level (https://github.com/jeremylong/DependencyCheck/blob/8a0ea14fd8c70c8dc069f1fe85b75be236541740/core/src/main/java/org/owasp/dependencycheck/Engine.java#L802-L806)

core/src/main/resources/data/dbStatements_postgresql.properties
```diff
@@ -14,7 +14,7 @@
...
-UPDATE_ECOSYSTEM=UPDATE CPEENTRY SET ecosystem = (SELECT q.eco FROM CPEENTRY e INNER JOIN (SELECT DISTINCT vendor, product, MIN(ecosystem) eco FROM CPEENTRY WHERE ecosystem is not null GROUP BY vendor, product) q ON q.VENDOR = e.VENDOR AND q.PRODUCT = e.PRODUCT AND e.id = CPEENTRY.id) WHERE exists (SELECT * FROM CPEENTRY e INNER JOIN (SELECT DISTINCT vendor, product, MIN(ecosystem) eco FROM CPEENTRY WHERE ecosystem is not null GROUP BY vendor, product) q ON q.VENDOR = e.VENDOR AND q.PRODUCT = e.PRODUCT AND e.id = CPEENTRY.id) AND CPEENTRY.ecosystem is null;
+#UPDATE_ECOSYSTEM=UPDATE CPEENTRY SET ecosystem = (SELECT q.eco FROM CPEENTRY e INNER JOIN (SELECT DISTINCT vendor, product, MIN(ecosystem) eco FROM CPEENTRY WHERE ecosystem is not null GROUP BY vendor, product) q ON q.VENDOR = e.VENDOR AND q.PRODUCT = e.PRODUCT AND e.id = CPEENTRY.id) WHERE exists (SELECT * FROM CPEENTRY e INNER JOIN (SELECT DISTINCT vendor, product, MIN(ecosystem) eco FROM CPEENTRY WHERE ecosystem is not null GROUP BY vendor, product) q ON q.VENDOR = e.VENDOR AND q.PRODUCT = e.PRODUCT AND e.id = CPEENTRY.id) AND CPEENTRY.ecosystem is null;
```
aikebah: Any reason for not removing the duplicate, but only commenting it?

core/src/test/java/org/owasp/dependencycheck/analyzer/OpenSSLAnalyzerTest.java
```diff
@@ -90,7 +90,7 @@ public void testAccept() {
 
     @Test
     public void testVersionConstantExamples() {
-        final long[] constants = {0x1000203fL, 0x00903000, 0x00903001, 0x00903002l, 0x0090300f, 0x0090301f, 0x0090400f, 0x102031af};
+        final long[] constants = {0x1000203fL, 0x00903000, 0x00903001, 0x00903002L, 0x0090300f, 0x0090301f, 0x0090400f, 0x102031af};
```
aikebah: I would propose to explicitly define all literals as longs in addition to fixing up the inspection warning

```suggestion
        final long[] constants = {0x1000203fL, 0x00903000L, 0x00903001L, 0x00903002L, 0x0090300fL, 0x0090301fL, 0x0090400fL, 0x102031afL};
```

(by default Java integer literals are of type `int` rather than `long`, implicitly widened by Java to the size of a long when stored into the long array - https://docs.oracle.com/javase/specs/jls/se8/html/jls-3.html#jls-3.10.1 / https://docs.oracle.com/javase/specs/jls/se8/html/jls-5.html#jls-5.1.2)

aikebah (commented):
Looks good, with a few improvement suggestions

core/src/main/java/org/owasp/dependencycheck/analyzer/NvdCveAnalyzer.java
jeremylong: yes the fall through should be combined - these were mostly autocorrections by Intellij. I'll combine these.

core/src/main/resources/data/dbStatements_postgresql.properties
jeremylong:
```suggestion
```

aikebah:
@jeremylong did you forget the push of the fixes for the other 2 remarks?

jeremylong:
@aikebah I've been incredibly busy - part of the reason I'm a month late with the release...

aikebah (approved):

PR #4746, opened 2022-08-12
Filter rules at execution instead of at load

aikebah:
## Description of Change

Filter the rules at execution time instead of at load time so that the SuppressionRules singleton holds the rules for both CPESuppressionAnalyzer and VulnerabilitySuppressionAnalyzer.

no

### DIFF
--- a/core/src/main/java/org/owasp/dependencycheck/analyzer/AbstractSuppressionAnalyzer.java
+++ b/core/src/main/java/org/owasp/dependencycheck/analyzer/AbstractSuppressionAnalyzer.java
@@ -123,7 +123,7 @@ public abstract class AbstractSuppressionAnalyzer extends AbstractAnalyzer imple
         if (rules.isEmpty()) {
             return;
         }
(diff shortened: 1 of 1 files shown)


aikebah:
Note: I think following this there may be a good reason to change a bit more (e.g. change the current filter method into a predicate method that indicates whether the rule is to be active rather than whether it needs to be filtered out) , but focused for now on just getting the logic to work well and leave the further judgement on changes to you.

jeremylong:
@aikebah we both started working on the same thing as I have a branch doing something similar - only I'm refactoring things a bit more... I just ran into a test case issue yesterday or I would have created the PR.

aikebah:
Yes, spotted the cleanup branch, didn't check what changes went in besides the code cleanup, so decided to keep this change minimal to reduce risk of difficult to resolve conflicts between the two.

jeremylong:
Updates were included in https://github.com/jeremylong/DependencyCheck/pull/4709

PR #4748, opened 2022-08-13
Vendor evidence Added for PipAnalyzer

### LINKED ISSUE #4747: NIST NVD support for PipAnalyzer
**Description**
- As I could see, when running in an isolated environment, we need to use `--disableOssIndex` to disable OssIndex, which leaves us with NIST NVD.
- With NIST NVD, as I can see the PipAnalyzer does not work well.

- Support for NIST NVD for PipAnalyzer.

- Using OssIndex, vulnerabilities are listed, however this cannot be used as solution for Isolated environments.

**Additional context**
- From debugging sessions, I was able to detect that `VENDOR` entries for PipAnalyzer are not added.

AnupamJuniwal:
## Description
NIST NVD support for PipAnalyzer : https://github.com/jeremylong/DependencyCheck/issues/4747

**Root Cause:** `VENDOR` evidence missing for PipAnalyzer, which eventually led to missed vulnerability detection by NPEAnalyzer
**Fix:** Added the `VENDOR` evidence as the package name.

## Schema changes
- none

### Automated test cases added
- none

### Integration test cases added
- Initialised the `PipAnalyzerIT`: testAnalyzePipAnalyzer

### Manual test cases run
- Verified that the vulnerabilities were listed with `--disableOssIndex` i.e. through NIST NVD index.


aikebah (approved):
Thanks for the PR

PR #4753, opened 2022-08-15
Added local node_module check for NodePackageAnalyzer

yarisvt:
## Description of Change

Ignore packages that start with "./", since those are local packages

*no*

### DIFF
--- a/core/src/main/java/org/owasp/dependencycheck/analyzer/NodePackageAnalyzer.java
+++ b/core/src/main/java/org/owasp/dependencycheck/analyzer/NodePackageAnalyzer.java
@@ -297,7 +297,7 @@ public class NodePackageAnalyzer extends AbstractNpmAnalyzer {
 
         // this seems to produce crash sometimes, I need to tests
         // using a local node_module is not supported by npm audit, it crash
-        if (Objects.nonNull(version) && (version.startsWith("file:") || version.matches("^[.~]*/.*"))) {
+        if (Objects.nonNull(version) && (version.startsWith("file:") || version.matches("^[.~]{0,2}/.*"))) {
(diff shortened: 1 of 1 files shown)


jeremylong:
Do you have an example where `version.matches("^[.~]*/.*")` indicates a local node module is being used?

yarisvt:
@jeremylong 

I am using vaadin to build a web app and it automatically generates js files in the `node_modules` folder. It then puts the following local path in the package.json (without prefixing it with `file:`)
image


According to the [npm docs](https://docs.npmjs.com/cli/v7/configuring-npm/package-json#local-paths), local paths can use four forms:
```
../foo/bar
~/foo/bar
./foo/bar
/foo/bar
```

Therefore, the regex `^[.~]*/.*` matches all four of them:
image

core/src/main/java/org/owasp/dependencycheck/analyzer/NodePackageAnalyzer.java
```diff
@@ -297,10 +297,10 @@ public static boolean shouldSkipDependency(String name, String version, boolean
 
         // this seems to produce crash sometimes, I need to tests
         // using a local node_module is not supported by npm audit, it crash
-        if (Objects.nonNull(version) && version.startsWith("file:")) {
+        if (Objects.nonNull(version) && (version.startsWith("file:") || version.matches("^[.~]*/.*"))) {
```
jeremylong:
```suggestion
        if (Objects.nonNull(version) && (version.startsWith("file:") || version.matches("^[.~]{0,2}/.*"))) {
```

jeremylong (approved):

jeremylong:
Thanks for the detailed response! I made one very minor change to the regex (s/\*/{0,2}/).

yarisvt:
Great!

Issue #4765, opened 2022-08-22
7.1.2 update: the suppression file stopped working


dmitry-weirdo:
**Describe the bug**
After updating from `7.1.1` to `7.1.2`, the false-positives excluded with the file started to fail again.

Failure:
```
[ERROR] Failed to execute goal org.owasp:dependency-check-maven:7.1.2:check (default-cli) on project ins-app: 
[ERROR] 
[ERROR] One or more dependencies were identified with vulnerabilities that have a CVSS score greater than or equal to '8.0': 
[ERROR] 
[ERROR] rhino-1.7.7.2.jar: [sonatype-2018-0590] CWE-611: Improper Restriction of XML External Entity Reference ('XXE')(8.8)
[ERROR] spring-web-5.3.22.jar: CVE-2016-1000027(9.8)
[ERROR] 
[ERROR] See the dependency-check report for more details.
```

Suppression file:
```xml
<?xml version="1.0" encoding="UTF-8"?>
<!--
    An advice from https://github.com/spring-projects/spring-framework/issues/24434#issuecomment-1132621725
-->

<!--
Documentation: https://jeremylong.github.io/DependencyCheck/general/suppression.html
-->
<suppressions xmlns="https://jeremylong.github.io/DependencyCheck/dependency-suppression.1.3.xsd">
    <suppress>
        <notes><![CDATA[Ignored since we are not vulnerable]]></notes>
        <packageUrl regex="true">^pkg:maven/org\.springframework/spring.*$</packageUrl>
        <cve>CVE-2016-1000027</cve>
    </suppress>

    <!-- Known FP, see https://github.com/jeremylong/DependencyCheck/issues/4671 -->
    <suppress>
        <notes><![CDATA[Ignored since it is a known false-positive]]></notes>
        <packageUrl regex="true">^pkg:maven/jakarta.*$</packageUrl>
        <cve>CVE-2022-31569</cve>
    </suppress>

    <!-- todo: remove this when swagger/rhino fixes this CWE -->
    <!--
        Ignore this failure that appeared with Sonartype update at 25.05.2022:
        - rhino-1.7.7.2.jar: [sonatype-2018-0590] CWE-611: Improper Restriction of XML External Entity Reference ('XXE')(8.8)
    -->
    <suppress>
        <notes><![CDATA[TODO: uncomment when swagger/rhino fixes it]]></notes>
[...]

PR #4769, opened 2022-08-23
Fix suppression load

jeremylong:
## Fixes Issue #4765

This was tested with the gradle plugin.


core/src/main/java/org/owasp/dependencycheck/analyzer/AbstractSuppressionAnalyzer.java
```diff
@@ -103,6 +107,11 @@ public Set<String> getSupportedExtensions() {
      */
     @Override
     public synchronized void prepareAnalyzer(Engine engine) throws InitializationException {
+        //check if we have a brand new settings object - if we do the suppression rules could be different
+        boolean loaded = getSettings().getBoolean(SUPPRESSION_LOADED, false);
```
jeremylong:
```suggestion
        final boolean loaded = getSettings().getBoolean(SUPPRESSION_LOADED, false);
```

aikebah (approved):
LGTM

PR #4771, opened 2022-08-24
pom.xml: delete unused config

dhalperi:
## Fixes Issue #

N/A

## Description of Change

Remove unused Maven properties and commented-out dependencies.
Leftover after switching to jackson-bom.

no


pom.xml
```diff
@@ -1102,16 +1100,6 @@ Copyright (c) 2012 - Jeremy Long
                 <type>pom</type>
```
dhalperi, line 1100: FYI, 1 line up is where the jackson deps are really versioned. `jackson-bom`, which does not have a property defined but version inline instead.

jeremylong (approved):

dhalperi:
@jeremylong - Kind of a dumb Q, but am I supposed to do anything after approval? I don't have merge rights.

PR #4773, opened 2022-08-24
Add a PinnedMavenInstallAnalyzer

### LINKED ISSUE #4772: Support for Bazel pinned `maven_install.json` files

This is a new type of Java lockfile (think: `requirements.txt` produced by something like `pip-compile`).

It would be nice if DependencyCheck supported it. Since all dependencies are fully transitively resolved,
I believe that all that needs to be done is creating a virtual dependency corresponding to each dependency.

I thought about:

* producing a bundle jar, and scanning that. Seems heavyweight and relies on the bundling process not to lose any versions.

* auto-generating another manifest (`pom.xml`?). If DependencyCheck supports scanning a pom.xml in isolation, great. But if not, we recurse. Either way, seems simpler to support this format.

**Additional context**

For more info, see: https://github.com/bazelbuild/rules_jvm_external/#pinning-artifacts-and-integration-with-bazels-downloader

Note:  This is not a niche homegrown thing, but a standard used in increasingly many projects. GitHub search of a sentinel used in the pinned file finds 175 repos at the time of writing: https://github.com/search?q=THERE_IS_NO_DATA_ONLY_ZUUL&type=code

dhalperi:
## Fixes Issue #
#4772

## Description of Change

`install.json` is a new type of Maven lockfile commonly used
in Bazel Java projects. Implement virtual dependency scanning
for such files, modeled after the existing PipAnalyzer.

Yes, I think so.

However, I was not able to run `mvn -s settings.xml verify` to success locally, so I did not yet add an IT. I also anticipate some changes to these tests in review.

In addition to the testing added in this PR, it worked on our
install.json file:
https://github.com/batfish/batfish/blob/6688b5b49ea695e7b566a0b70403396f580b2805/maven_install.json


core/src/main/java/org/owasp/dependencycheck/analyzer/PinnedMavenInstallAnalyzer.java
```diff
@@ -0,0 +1,232 @@
...
+                version = pieces[4];
+            }
+
+            if ("sources".equals(classifier)) {
+                LOGGER.debug("Skipping sources jar {}", dep.coord);
+                continue;
+            }
+
+            final Dependency d = new Dependency(dependency.getActualFile(), true);
+            d.setEcosystem(Ecosystem.JAVA);
```
dhalperi, line 176: FYI it was not obvious to me which fields to add. I looked at the Maven plugin, requirements.txt analyzer, and a few other places. At the end of the day, this worked great.

cli/src/test/resources/sample.properties
```diff
@@ -9,6 +9,7 @@ analyzer.python.distribution.enabled=true
 analyzer.python.package.enabled=true
 analyzer.ruby.gemspec.enabled=true
 analyzer.autoconf.enabled=true
+analyzer.maveninstall.enabled=true
```
dhalperi, line 12: FYI, I didn't see a logic to how things were ordered here, so I just put the new code next to `PipAnalyzer`'s corresponding entries everywhere.

cli/src/main/java/org/owasp/dependencycheck/CliParser.java
```diff
@@ -1196,6 +1196,10 @@ public static class ARGUMENT {
          * Disables the Autoconf Analyzer.
          */
         public static final String DISABLE_AUTOCONF = "disableAutoconf";
+        /**
+         * Disables the Maven install Analyzer.
+         */
+        public static final String DISABLE_MAVEN_INSTALL = "disableMavenInstall";
```
dhalperi, line 1202: Would be happy to discuss names as owners see fit.

src/test/resources/maven_install.json
```diff
@@ -0,0 +1,2036 @@
+{
```
dhalperi, line 1: FYI this is a real file copied from rules_jvm_external. (Apache2 license). Would be happy to generate a smaller one if desired, I just didn't want to dummy up a file.

dhalperi (commented):
FYI, I modeled this after `PipAnalyzer`, with a lot of copy and paste.

I did follow my own standards for implementation details and debug logging. E.g., I deleted overridden Javadoc identical to parent.

I anticipate and am happy to iterate on the code to bring it up to local standards.

core/src/main/java/org/owasp/dependencycheck/analyzer/PinnedMavenInstallAnalyzer.java
```diff
@@ -0,0 +1,232 @@
...
+
+    /**
+     * The phase that this analyzer is intended to run in.
+     */
+    private static final AnalysisPhase ANALYSIS_PHASE = AnalysisPhase.INFORMATION_COLLECTION;
+
+    /**
+     * Pattern matching files with "install" in the basename and extension "json".
+     */
+    private static final Pattern MAVEN_INSTALL_JSON_PATTERN = Pattern.compile(".*install.*\\.json");
```
aikebah: I think it would be better to rule out a plain "install.json" from the pattern as that name is also used by cloudflare for defining how to install on their infrastructure, and my impression from a quick read in the docs is that it would not normally appear in bazel project.


see https://www.cloudflare.com/apps/developer/docs/install-json

dhalperi: Hi @aikebah - Did I misunderstand how the plugin "claiming" files work?

FWICT I thought it was okay for the filter to be a little broad, as long as it didn't "claim" the file with `        engine.removeDependency(dependency);`

In this PR, you can see (L123) that the dependency is only removed if a sentinel value is found. Otherwise, the code "returns early" at L120. So this analyzer ought to correctly ignore (and preserve for downstream analyses) such files.

I would be happy to write such a test with a simpler pointer to start from :)

dhalperi: I have pushed a simple unit test that the Cloudflare empty `install.json` is properly ignored, assuming I understand the semantics of removing dependency correctly, that is.

aikebah: > I have pushed a simple unit test that the Cloudflare empty `install.json` is properly ignored, assuming I understand the semantics of removing dependency correctly, that is.

The semantics of removing dependency for the information-collecting analyzers is "this is not a dependency of the project, but a resource that I've used to discover and register the actual dependencies, so it should be completely excluded from further analysis steps", not "I've claimed this dependency". 
The acceptance of certain files is meant to be the primary filter of something that is of interest of a certain analyzer and therefor should be as properly targeted as possible to avoid starting to analyze completely unrelated files, so it should be wide enough to cover the proper files and small enough not to cover unrelated files.
My impression reading up on Bazel is that an actual basel maven_install lockfile would never be simply called install.json. If so it should not be accepted by this analyzer. If it can be, than the regex would be fine.

dhalperi: Thanks for the clarification! That makes a lot of sense.

> If it can be, than the regex would be fine.

I just confirmed that are no filename restrictions on the pinned file: I was able to get things working by naming it `foo_bar_baz.txt`. So `install.json` would be a fine name too.

However, I suspect that most users will follow the `rules_jvm_external` example as convention. Sensible names (other than the example's `maven_install.json`) will probably look like `<string>_install.json` or `maven_install_<string>.json`.

Sampling [GitHub public for the sentinel](https://github.com/search?p=2&q=THERE_IS_NO_DATA_ONLY_ZUUL&type=Code), I see that the vast majority are in fact just named `maven_install.json`. Other examples include: `antlr_install.json`, `jee9_install.json`, `bazel_diff_maven_install.json`, `unsafe_shared_cache_with_pinning_install.json`, `maven_old_install.json`, etc.

My gut is to keep the regex as-is purely for simplicity / code readability, but also happy to change if that is useful.

dhalperi: If we were to change it, something like `(.+install.*|.*install.+)\.json` seem reasonable to you?

core/src/main/java/org/owasp/dependencycheck/analyzer/PinnedMavenInstallAnalyzer.java
```diff
@@ -0,0 +1,229 @@
...
+                continue;
+            }
+
+            final Dependency d = new Dependency(dependency.getActualFile(), true);
+            d.setEcosystem(Ecosystem.JAVA);
+            d.addEvidence(EvidenceType.VENDOR, "project", "groupid", group, Confidence.HIGHEST);
+            d.addEvidence(EvidenceType.PRODUCT, "project", "artifactid", artifact, Confidence.HIGHEST);
+            d.addEvidence(EvidenceType.VERSION, "project", "version", version, Confidence.HIGHEST);
+            d.setName(String.format("%s:%s", group, artifact));
+            d.setVersion(version);
```
aikebah: Think it would be better to include more information that is available from the maven_install json:
- Add the maven PackageURL as a PurlIdentifier from the coord and add that as a SoftwareIdentifier using `Dependency#addSoftwareIdentifier(Identifier)`
- Set the fileName to a string constructed from the maven_install JSON file name and the coord, to get a clearer reporting on the CLI when the `--failOnCVSS` option is used

The packageUrl will be helpful a a hook for false-positive CPE match suppressions and the CLI-output will be more useful when breaking build for CVSSScore

`dependency-check.sh --enableExperimental --failOnCVSS` On commit `05d476205115d7d2f512b8a86ecac84d8e1dc248`  of your project yields:
```
One or more dependencies were identified with vulnerabilities that have a CVSS score greater than or equal to '0.0': 

maven_install.json: CVE-2020-36518(7.5)
maven_install.json: CVE-2022-25647(7.5)
maven_install.json: CVE-2022-33980(9.8)
maven_install.json: CVE-2022-25762(8.6), CVE-2020-8022(7.8), CVE-2022-34305(6.1), CVE-2022-29885(7.5)
```
with the suppress-button for the CPE yielding an unusable
```xml
<suppress>
   <notes><![CDATA[
   file name: com.fasterxml.jackson.core:jackson-databind:2.11.4
   ]]></notes>
   <sha1>$enc.html($dependency.Sha1sum)</sha1>
   <cpe>cpe:/a:fasterxml:jackson-databind</cpe>
</suppress>
```
after adding the packageURL and setting a constructed filename on the dependency (I opted for `<maven_install path>'>>'<coord>`, feel free to choose a pattern that shorter but still sufficiently informative on the CLI-output):
```
One or more dependencies were identified with vulnerabilities that have a CVSS score greater than or equal to '0.0': 

/Users/aikebah/Projects/dcTesting/bazel/batfish/maven_install.json>>com.fasterxml.jackson.core:jackson-databind:2.11.4: CVE-2020-36518(7.5)
/Users/aikebah/Projects/dcTesting/bazel/batfish/maven_install.json>>com.google.code.gson:gson:2.8.5: CVE-2022-25647(7.5)
/Users/aikebah/Projects/dcTesting/bazel/batfish/maven_install.json>>junit:junit:4.12: CVE-2020-15250(5.5)
/Users/aikebah/Projects/dcTesting/bazel/batfish/maven_install.json>>org.apache.commons:commons-configuration2:2.7: CVE-2022-33980(9.8)
/Users/aikebah/Projects/dcTesting/bazel/batfish/maven_install.json>>org.apache.tomcat.embed:tomcat-embed-core:8.5.75: CVE-2022-25762(8.6), CVE-2022-34305(6.1), CVE-2022-29885(7.5)
/Users/aikebah/Projects/dcTesting/bazel/batfish/maven_install.json>>org.glassfish.jersey.core:jersey-common:2.27: CVE-2021-28168(5.5)
```

with the button for the CPE yielding the usable
```xml
<suppress>
   <notes><![CDATA[
   file name: com.fasterxml.jackson.core:jackson-databind:2.11.4
   ]]></notes>
   <packageUrl regex="true">^pkg:maven/com\.fasterxml\.jackson\.core/jackson\-databind@.*$</packageUrl>
   <cve>CVE-2020-36518</cve>
</suppress>
```
```suggestion
            final Dependency d = new Dependency(dependency.getActualFile(), true);
            d.setEcosystem(Ecosystem.JAVA);
            d.addEvidence(EvidenceType.VENDOR, "project", "groupid", group, Confidence.HIGHEST);
            d.addEvidence(EvidenceType.PRODUCT, "project", "artifactid", artifact, Confidence.HIGHEST);
            d.addEvidence(EvidenceType.VERSION, "project", "version", version, Confidence.HIGHEST);
            d.setName(String.format("%s:%s", group, artifact));
            d.setVersion(version);
            d.setFileName(String.format("%s>>%s", dependency.getActualFile(), dep.coord));
            try {
                final PackageURLBuilder purl = PackageURLBuilder.aPackageURL()
                        .withType(PackageURL.StandardTypes.MAVEN)
                        .withNamespace(group)
                        .withName(artifact)
                        .withVersion(version);
                if (classifier != null) {
                    purl.withQualifier("classifier", classifier);
                }
                d.addSoftwareIdentifier(new PurlIdentifier(purl.build(), Confidence.HIGHEST));
            } catch (MalformedPackageURLException e) {
                d.addSoftwareIdentifier(new GenericIdentifier("maven_install JSON coord " + dep.coord, Confidence.HIGH));
            }
```

core/src/main/java/org/owasp/dependencycheck/analyzer/PinnedMavenInstallAnalyzer.java
aikebah: Seems like an appropriate pattern, given the fact that Bazel allows customized names for the file I can imagine people naming it within that pattern, but would be suprised when they fully shorten it to install.json as the json file is meant to pin maven versions. 
If I was in charge of the buiild engineering of Bazel projects I think I would even opt for convention-enforcement of using at least the full string of maven_install within any maven_install pin-file filename.

core/src/main/java/org/owasp/dependencycheck/analyzer/PinnedMavenInstallAnalyzer.java
dhalperi: Thanks! I did this, except that I also set `FilePath` so that the file path is visible in the HTML.

src/test/resources/maven_install.json
dhalperi: Replaced with a homegrown but real file (that is also much shorter). Just to make sure there's no licensing issues.

aikebah:
> FYI, I modeled this after `PipAnalyzer`, with a lot of copy and paste.
> 
> I did follow my own standards for implementation details and debug logging. E.g., I deleted overridden Javadoc identical to parent.
> 
> I anticipate and am happy to iterate on the code to bring it up to local standards.

I'll leave the codestyle review to the various automated code reviews... results should appear soon now that I triggered the PR workflow. Hope to take a second look at the coding over the weekend.

@jeremylong any not automatically checked coding standards you'd like to enforce on the project?

core/src/main/java/org/owasp/dependencycheck/analyzer/PinnedMavenInstallAnalyzer.java
dhalperi: (This has been much improved with excellent review feedback.)

aikebah:
@dhalperi Up to now no issue, I've only glanced over and picked out some items, but going forward: please do not force-push, as that breaks the review cycle. With regular pushes only the changed bits need (re)review, with a force-push the entire change needs to be (re)reviewed as the delta to the already reviewed code is unclear.

core/src/main/java/org/owasp/dependencycheck/analyzer/PinnedMavenInstallAnalyzer.java
```diff
@@ -0,0 +1,262 @@
...
+            } else if (pieces.length == 4) {
+                classifier = pieces[2];
+                version = pieces[3];
+            } else {
+                // length == 5 as guaranteed above.
+                classifier = pieces[3];
+                version = pieces[4];
+            }
+
+            if ("sources".equals(classifier)) {
```
jeremylong: likely rare - but if we are skipping `sources` we should also skip `javadoc`:

```suggestion
            if ("sources".equals(classifier) || "javadoc".equals(classifier)) {
```

dhalperi:
Thanks for all the great feedback and useful improvements!

What else can I do?

jeremylong (approved):

jeremylong:
@dhalperi thanks for the PR!

Issue #4723, opened 2022-08-02
Hosted Suppression File


jeremylong:
Due to situations like https://github.com/jeremylong/DependencyCheck/issues/4670, https://github.com/jeremylong/DependencyCheck/issues/4671, https://github.com/jeremylong/DependencyCheck/issues/4677, https://github.com/jeremylong/DependencyCheck/issues/4690 - ODC needs to be able to respond faster and provide an updated suppression file when situations like this occur.

1. Implement a hosted suppression file - ODC will download the file to its local cache daily. This will be controllable by a switch so those working in an offline mode can disable the feature.
2. Update the [issue-ops for false positives](https://github.com/jeremylong/DependencyCheck/blob/main/.github/workflows/false-positive-ops.yml) so that if/when a core maintainer (@aikebah, @nhumblot, or myself (@ctxt)) gives the github action bot's comment a thumbs up on a the generated suppression rule (e.g., [comment on #4722](https://github.com/jeremylong/DependencyCheck/issues/4722#issuecomment-1201839476)) the generated suppression rule is then copied into the hosted suppression file, the updated suppression file is tested (to ensure it does not break ODC by not being parsable for some reason), and then the updated file will be automatically published.
    - Note, this will allow the team at release time to copy the suppressions from the hosted file into the release. 
    - A schedule for removing the suppressions from the hosted file into the base suppression rules will need to be determined. We don't want to force someone to upgrade, but we also don't want to maintain the rules in two places.

albuch:
Please make this resilient against network failures and unavailability of the resource (use suppression file that is bundled instead).
We already have a long enough history of data source availability impacting DependencyCheck.

OrangeDog:
Just to note that the bot-generated suppression is often not good, as it doesn't understand which aspect is wrong, or which additional versions and/or packages would be affected.

aikebah:
I would say it's often fine, though sometimes multiple FP reports can be combined for a single rule.  Many of the reports I could, after review take over the generated suppression, some I had to modify and some were plain wrong.
Having a quick path for the approval and processing of the obviously correct ones eases the process. Those that require special handling would follow a regular code-and-PR process, but the obvious ones would be as plain and simple (or even simpler I guess) as various dependabot PR's that I was able to quickly merge in after changelog review from a mobile device by pushing a few approve/merge buttons

jeremylong:
[...]

PR #4784, opened 2022-08-27
IssueOps: Generate Hosted Suppression File

jeremylong:
Implements the first half of #4723 - the workflow will commit the previously generated suppression rule to the `generatedSuppressions` suppression branch and publish the suppression file to `gh-pages` **if approved** by one of the core maintainers. Core maintainers can simply reply "approved" to the generated suppression rule in an issue labeled as a `FP Report` to kick off this workflow.

### DIFF
--- a/.github/workflows/false-positive-approvals.yml
+++ b/.github/workflows/false-positive-approvals.yml
@@ -10,7 +10,9 @@ jobs:
     if: ${{ !github.event.issue.pull_request && 
       contains(github.event.issue.labels.*.name, 'FP Report') && 
       contains(github.event.comment.body,'approved') &&
-      contains(github.event.comment.user.login, 'jeremylong') }}
+      (contains(github.event.comment.user.login, 'jeremylong') ||
+      contains(github.event.comment.user.login, 'aikebah') || 
+      contains(github.event.comment.user.login, 'nhumblot'))  }}
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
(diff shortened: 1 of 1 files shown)


jeremylong:
Development and testing was done here: https://github.com/jeremylong/workflow-test/

Duplicate reports see: https://github.com/jeremylong/workflow-test/issues/4
Approving a suppression rule see: https://github.com/jeremylong/workflow-test/issues/6

.github/workflows/false-positive-approvals.yml
```diff
@@ -0,0 +1,158 @@
...
+
+jobs:
+  update_suppression:
+    name: Update Suppression Rules
+    if: ${{ !github.event.issue.pull_request && 
+      contains(github.event.issue.labels.*.name, 'FP Report') && 
+      contains(github.event.comment.body,'approved') &&
+      (contains(github.event.comment.user.login, 'jeremylong') ||
+      contains(github.event.comment.user.login, 'aikebah') || 
+      contains(github.event.comment.user.login, 'nhumblot'))  }}
```
aikebah:
```suggestion
      (github.event.comment.user.login === 'jeremylong') ||
      github.event.comment.user.login === 'aikebah') || 
      github.event.comment.user.login === 'nhumblot'))  }}
```

Stronger check required to avoid people with github logins like `aikebah1337` impersonating as me. Derived the alternative check from the filter applied at line 41 to check whether a comment was made by the github actions bot. Not sure whether exact same check can be used for mere mortals with a github account

aikebah (changes requested):
Looks good to me, except for the approving-user-guard on the approve comment

jeremylong:
@aikebah - thanks for catching the issue with the approvers!

aikebah:
@jeremylong The `===` syntax is invalid.... see https://github.com/jeremylong/DependencyCheck/actions/runs/2984591036

PR #4795, opened 2022-08-31
Fixing file separator bug for unit test on windows machines

echalone:
## Fixes Issue #
Fixed test for windows machines (backslash would escape a character, therefore it needs to be replaced with a double backslash if file separator is a backslash).

## Description of Change
Replacing potential backslash in File.separator property with double-backslash so it won't throw an exception on windows machines (or other OS using backslash instead of slash as path separator).

Technically no, because it fixes a test case itself.

### DIFF
--- a/core/src/test/java/org/owasp/dependencycheck/analyzer/DependencyCheckPropertiesTest.java
+++ b/core/src/test/java/org/owasp/dependencycheck/analyzer/DependencyCheckPropertiesTest.java
@@ -89,7 +89,7 @@ public class DependencyCheckPropertiesTest {
     }
 
     public Set<Class<?>> findAllClasses(String packageName) throws IOException {
-        String parsedPackageName = packageName.replaceAll("[.]", File.separator.replaceAll("\\\\","\\\\\\\\"));
+        String parsedPackageName = packageName.replaceAll("[.]", "/");
 
         Set<Class<?>> classes = new HashSet<>();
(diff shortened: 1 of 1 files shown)


core/src/test/java/org/owasp/dependencycheck/analyzer/DependencyCheckPropertiesTest.java
```diff
@@ -89,7 +89,7 @@ private boolean isATestAnalyzer(Class<?> clazz) {
     }
 
     public Set<Class<?>> findAllClasses(String packageName) throws IOException {
-        String parsedPackageName = packageName.replaceAll("[.]", File.separator);
+        String parsedPackageName = packageName.replaceAll("[.]", File.separator.replaceAll("\\\\","\\\\\\\\"));
```
aikebah:
```suggestion
        String parsedPackageName = packageName.replaceAll("[.]", "/");
```

As the resourceloader is used the proper fix should be replacing `.` by forward slash see [getResources javadoc](https://docs.oracle.com/javase/8/docs/api/java/lang/ClassLoader.html#getResources-java.lang.String-)

> The name of a resource is a /-separated path name that identifies the resource.

aikebah (changes requested):

jeremylong (approved):

aikebah (approved):

Issue #4770, opened 2022-08-23
AnalysisException: java.lang.NullPointerException after 7.1.2 Update


helloingob:
**Describe the bug**
We use Dependency Check as an Azure Task and after the update from 7.1.1 to 7.1.2 it fails with a java.lang.NullPointerException in MSBuildProjectAnalyzer. We **DID NOT** update/change our code, especially not the csproj file.

**Version of dependency-check used**
owasp-dependency-check 7.1.2 x64

```
- task: OWASPDependencyCheck@0
  inputs:
    outputDirectory: '$(Agent.TempDirectory)/dependency-scan-results'
    scanDirectory: '$(Build.SourcesDirectory)'
    outputFormat: 'ALL'
    useSonarQubeIntegration: true
    additionalArguments: |
      --log $(Build.ArtifactStagingDirectory)/owasp.log
```

**Log**
```
WARN  - An error occurred while analyzing 'D:\a\1\s\XXX\XXX.csproj' (MSBuild Project Analyzer).
2022-08-23 14:01:45,990 org.owasp.dependencycheck.AnalysisTask:91
DEBUG - 
org.owasp.dependencycheck.analyzer.exception.AnalysisException: java.lang.NullPointerException
	at org.owasp.dependencycheck.analyzer.MSBuildProjectAnalyzer.analyzeDependency(MSBuildProjectAnalyzer.java:174)
	at org.owasp.dependencycheck.analyzer.AbstractAnalyzer.analyze(AbstractAnalyzer.java:131)
	at org.owasp.dependencycheck.AnalysisTask.call(AnalysisTask.java:88)
	at org.owasp.dependencycheck.AnalysisTask.call(AnalysisTask.java:37)
	at java.util.concurrent.FutureTask.run(FutureTask.java:266)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
	at java.lang.Thread.run(Thread.java:750)
Caused by: java.lang.NullPointerException: null
	at org.owasp.dependencycheck.data.nuget.XPathMSBuildProjectParser.parse(XPathMSBuildProjectParser.java:74)
	at org.owasp.dependencycheck.analyzer.MSBuildProjectAnalyzer.analyzeDependency(MSBuildProjectAnalyzer.java:120)
	... 7 common frames omitted

[...]

PR #4796, opened 2022-08-31
Checking for null value from MSBuild xpath method before accessing it

echalone:
Checking for null value from xpath method in MSBuild project analysis before accessing its result so we don't get a NullPointerException in case of missing version element or attribute. Also added package references without version to test-project to test for this case in the future.

## Fixes Issue 
Fixed issue #4770

## Description of Change
Concerning MSBuild project analysis:
Added check if version attribute or element was found before trying to access its value (because otherwise this may result in a NullPointerException). This bug was introduced with PR https://github.com/jeremylong/DependencyCheck/pull/4678
Will no longer throw a NullPointerException on missing version element/attribute or if version was written in lower case.
Will however not interpret in any version element or attribute that was written in lower-case.
Note: This is the same behaviour as was observed before version 7.1.2, in the future it would probably be wise to check the MSBuild project file for all the elements and attributes case insensitive. But that would go beyond this bugfix.

Yes, PackageReferences without any version have been added to the test-project so test will fail in future if this happenes again.

### DIFF
--- a/core/src/main/java/org/owasp/dependencycheck/data/nuget/XPathMSBuildProjectParser.java
+++ b/core/src/main/java/org/owasp/dependencycheck/data/nuget/XPathMSBuildProjectParser.java
@@ -69,9 +69,14 @@ public class XPathMSBuildProjectParser implements MSBuildProjectParser {
                 final NamedNodeMap attrs = node.getAttributes();
 
                 final String include = attrs.getNamedItem("Include").getNodeValue();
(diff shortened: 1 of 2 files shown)


mprins (approved):

echalone:
whoops, sorry, clicked on the wrong button so now I've re-requested an approval? Oo

jeremylong (approved):

jeremylong:
Thanks for the PR!

PR #4800, opened 2022-09-01
Allow custom report templates via CLI

gesellix:
## Description of Change

This makes the CliParser more tolerant for the `--format` argument. It still allows all values from `Format` and additionally accepts existing file paths. It won't validate if the file is a valid Velocity template.

no (not yet)


jeremylong (approved):
Thanks for the PR!

PR #4804, opened 2022-09-02
CLI,completion: add --disableMavenInstall

dhalperi:
Omission in #4773


aikebah (approved):
Thanks for the PR. Would you mind also updating the corresponding doc on the site? Those would be markdown documents in the various src/main/site folders like, for the CLI

https://github.com/jeremylong/DependencyCheck/blob/main/cli/src/site/markdown/arguments.md

That way the options to control the new analyzer will also be properly listed at the [github pages](https://jeremylong.github.io/DependencyCheck/)

jeremylong (approved):

PR #4901, opened 2022-10-02
feat: Use hosted suppressions

aikebah:
## Description of Change

Adds the hosted base-suppressions file to the suppression rules used by the CPE and CVE suppression analyzers

No, would be happy to add some if I had a clue how to do some proper tests that make sense.

For now the only testing is performed by the fact that it doesn't break the existing tests.

Locally performed some manual tests that inability to download a hosted base-suppressions file from an improper location or a corrupted download (simulated by taking a download that's not even XML) will only log a warning (including stacktrace with root cause), but continue the analysis with only a higher chance on encountering FPs.


cli/src/main/java/org/owasp/dependencycheck/CliParser.java
```diff
@@ -1481,5 +1487,18 @@ public static class ARGUMENT {
          * when generating the JUNIT report format.
          */
         public static final String FAIL_JUNIT_ON_CVSS = "junitFailOnCVSS";
+        /**
+         * The CLI argument to set the number of hours to wait before re-checking hosted suppressions file for updates
```
jeremylong: Make checkstyle happy...
```suggestion
         * The CLI argument to set the number of hours to wait before re-checking hosted suppressions file for updates.
```

jeremylong:
@aikebah any idea why the two test cases failed? https://github.com/jeremylong/DependencyCheck/actions/runs/3169240938/jobs/5161011285#step:7:3931

aikebah:
> @aikebah any idea why the two test cases failed? https://github.com/jeremylong/DependencyCheck/actions/runs/3169240938/jobs/5161011285#step:7:3931

No, not yet... had some busy days, so no time yet to look into them... local build (branch only) went fine, so I was surprised by the failing test on PR... trying a merged local branch now to dig deeper

aikebah:
Logical that local tests do not break when you build up to the package phase rather than verify.... the error was all in my alias to ease starting the build including releaseTesting.....

aikebah:
> @aikebah any idea why the two test cases failed? https://github.com/jeremylong/DependencyCheck/actions/runs/3169240938/jobs/5161011285#step:7:3931

Caused by two bugs in the new code (the defensive copy part... which I failed to properly test afterwards)
* code was unintendedly disabling the suppression-analyzer on failures loading the hosted rules - the direct cause of the test-failure
* the code failed to signal to Files.copy that it was OK to replace an existing file (created just before as an empty file by Files.createTempFile) - which was the trigger to surface the disable-bug: a failure to copy to an already existing file (a swallowed IOException).

aikebah:
@jeremylong Working on a PR for dependency-check-gradle to add configurability there as well which is almost finished, so expecting to add another commit here to update the gradle plugin docs for the site accordingly.

aikebah:
@jeremylong done with my changes, so ready for merge if there are no more review findings

aikebah:
@jeremylong failing test is due to a mysteriously missing file in `/home/runner/.m2` halfway through the pipeline. No clue what happened there - two pipelines running from the same `/home/runner/.m2` and the other one clearing out that file?

jeremylong:
@aikebah the m2 directory is cached... wonder if something weird is going on there?

aikebah:
> @aikebah the m2 directory is cached... wonder if something weird is going on there?

The file concerned holds the masterpassword for secret encryption in settings.xml. I can imagine that that is automatically cleared at the end of the pipeline to expose the secrets only during build.

jeremylong (approved):

jeremylong:
Sorry - apparently what I thought was a simple fix of the merge conflicts caused a bit of an issue.  I have to step away for a few hours - but I'll fix this later today.

aikebah:
@jeremylong Propose to merge this one as the last of the three... as it's modifications to existing codebase are smallest and therefor the easiest one to resolve the conflicts should any arise (which I think will... volunteering to resolve those after the merge of the other PRs)

Issue #2793, opened 2020-09-07
An illegal reflective access operation has occurred


OrangeDog:
**Describe the bug**
WARNING: An illegal reflective access operation has occurred
WARNING: Illegal reflective access by com.fasterxml.jackson.module.afterburner.util.MyClassLoader (file:/C:/Users/me/.m2/repository/com/fasterxml/jackson/module/jackson-module-afterburner/2.11.2/jackson-module-afterburner-2.11.2.jar) to method java.lang.ClassLoader.findLoadedClass(java.lang.String)
WARNING: Please consider reporting this to the maintainers of com.fasterxml.jackson.module.afterburner.util.MyClassLoader
WARNING: Use --illegal-access=warn to enable warnings of further illegal reflective access operations
WARNING: All illegal access operations will be denied in a future release

**Version of dependency-check used**
The problem occurs using version 6.0.0 of the maven plugin

**To Reproduce**
1. Run on JDK 9+

**Expected behavior**
No warnings about reflective access violations

**Additional context**
Upstream issue: https://github.com/FasterXML/jackson-modules-base/issues/37

Options:
 - Don't use afterburner
 - Use [blackbird](https://github.com/stevenschlansker/jackson-blackbird) instead (e.g. with a profile activation)
 - Tell everyone to ignore it until afterburner fixes it

migueldolomandin:
This happens intermittently too using the latest docker image
https://hub.docker.com/r/owasp/dependency-check

jeremylong:
We will upgrade the library when it is available. However, this issue is from a dependency: https://github.com/FasterXML/jackson-modules-base/issues/37

calebwhitt:
Is there a way to get round this issue in the meantime? It seems the issue with the dependency is ongoing... I too am getting this error when using the latest docker image

jeremylong:
I've run scans using the latest docker image and do not see the warning.  @calebwhitt to be clear - this is a warning not an error.

Disabling afterburner had an impact in some very quick tests I ran of ~10-15 seconds in the update process... We can't use blackbird as we still support Java 8. For now - I'm okay with ignoring the warning... However, we will continue to monitor...

OrangeDog:
It'll only come up if it actually does some JSON processing. Purge the database to guarantee reproduction.

Is it possible to do a layered jar or something in the plugin definition so Java 8 uses Afterburner and 9+ uses Blackbird?

calebwhitt:
[...]

PR #4905, opened 2022-10-05
Switch to Jackson Blackbird

jeremylong:
Per #2793 switch from afterburner to blackbird.

### DIFF
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -268,17 +268,14 @@ Copyright (c) 2012 Jeremy Long. All Rights Reserved.
         <dependency>
             <groupId>com.fasterxml.jackson.core</groupId>
             <artifactId>jackson-databind</artifactId>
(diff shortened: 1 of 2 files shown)


aikebah (approved):

jpcmonster:
just to confirm - with this change, 7.3.0 is no longer JDK8 compatible - or?

aikebah:
It still is, but will give you a warning, see also https://github.com/jeremylong/DependencyCheck/issues/4966

jeremylong:
The warning will be removed with 7.3.1.

jpcmonster:
@aikebah @jeremylong - thanks for the answer.  And thanks again for a great product; even though we have GitHub Enterprise with Dependabot, it doesn't execute Gradle and will not yet report any vulnerabilities for the Gradle package ecosystem.

Issue #4888, opened 2022-09-28
Invalid column name 'cveId'


dmowfo:
We have setup a MSSQL database and connection to this database is working fine. The database is updated as expected though we are seeing the following error later on in the process when trying to check vulnerabilities found against entries found in the DB 

Have you see this before? there are no references to cveId within the database

Caused by: org.owasp.dependencycheck.data.nvdcve.DatabaseException: Exception retrieving vulnerability for cpe:2.3:a:microsoft:.net_core:4.700.19.56404:*:*:*:*:*:*:*
	at org.owasp.dependencycheck.data.nvdcve.CveDB.getVulnerabilities(CveDB.java:635)
	at org.owasp.dependencycheck.analyzer.NvdCveAnalyzer.lambda$analyzeDependency$2(NvdCveAnalyzer.java:65)
	... 18 common frames omitted
Caused by: com.microsoft.sqlserver.jdbc.SQLServerException: **Invalid column name 'cveId'.**
	at com.microsoft.sqlserver.jdbc.SQLServerException.makeFromDatabaseError(SQLServerException.java:262)
	at com.microsoft.sqlserver.jdbc.SQLServerStatement.getNextResult(SQLServerStatement.java:1621)
	at com.microsoft.sqlserver.jdbc.SQLServerPreparedStatement.doExecutePreparedStatement(SQLServerPreparedStatement.java:592)
	at com.microsoft.sqlserver.jdbc.SQLServerPreparedStatement$PrepStmtExecCmd.doExecute(SQLServerPreparedStatement.java:522)
	at com.microsoft.sqlserver.jdbc.TDSCommand.execute(IOBuffer.java:7194)
	at com.microsoft.sqlserver.jdbc.SQLServerConnection.executeCommand(SQLServerConnection.java:2935)
	at com.microsoft.sqlserver.jdbc.SQLServerStatement.executeCommand(SQLServerStatement.java:248)
	at com.microsoft.sqlserver.jdbc.SQLServerStatement.executeStatement(SQLServerStatement.java:223)
	at com.microsoft.sqlserver.jdbc.SQLServerPreparedStatement.executeQuery(SQLServerPreparedStatement.java:444)
	at org.apache.commons.dbcp2.DelegatingPreparedStatement.executeQuery(DelegatingPreparedStatement.java:122)
	at org.apache.commons.dbcp2.DelegatingPreparedStatement.executeQuery(DelegatingPreparedStatement.java:122)
[...]

PR #4908, opened 2022-10-05
Fix Case Issue with SQL

jeremylong:
## Fixes Issue #4888

Corrects the casing of the cveid column. Long term we should move to standard Screaming Query Language - and use liquibase or similar to manage the database creation/upgrades.


aikebah (approved):

aikebah:
somewhere deep down I always knew that the `S` didn't stand for structured.... now I finally know for sure :)

Issue #4652, opened 2022-07-06
[FP]: jruby-rack confused for jruby itself


chadlwilson:
### Package URl

pkg:maven/org.jruby.rack/jruby-rack@1.1.21

### CPE

cpe:/a:jruby:jruby

### ODC Integration

{"label"=>"Gradle Plugin"}

### ODC Version

7.1.1

### Description

jruby-rack is an independent add-on to JRuby, and not the same as JRuby itself. See https://github.com/jruby/jruby-rack and https://mvnrepository.com/artifact/org.jruby.rack/jruby-rack

aikebah:
approved

Issue #4647, opened 2022-07-06
[FP]: Shaded JRuby dirgra being confused for JRuby itself


chadlwilson:
### Package URl

pkg:maven/org.jruby/dirgra@0.3

### CPE

cpe:/a:jruby:jruby

### ODC Integration

{"label"=>"Gradle Plugin"}

### ODC Version

7.1.1

### Description

```
  <suppress>
    <notes><![CDATA[
    Suppressing false positive caused by OWASP Dependency Check thinking the shaded/packaged dirgra library is the same
    as the JRuby version. These are versioned independently and not the same thing.
    ]]></notes>
    <packageUrl regex="true">^pkg:maven/org\.jruby/dirgra@.*$</packageUrl>
    <cpe>cpe:/a:jruby:jruby</cpe>
  </suppress>
```

See https://github.com/jruby/dirgra

aikebah:
was merged and released with 7.1.2

chadlwilson:
Hi @aikebah - are you able to point me to how this was resolved or the relevant suppression?

I am still getting this and needing to suppress with `7.2.1` via the Gradle plugin so will check if there is something else going on here.

https://github.com/gocd/gocd/blob/75e13170e1cd5eb9f07015bf69719bfb17147043/buildSrc/dependency-check-suppress.xml#L19-L26

```
dependency-check version: 7.2.1
Report Generated On: Sun, 9 Oct 2022 03:35:25 +0530
Dependencies Scanned: 320 (317 unique)
Vulnerable Dependencies: 0
Vulnerabilities Found: 0
Vulnerabilities Suppressed: 115

jruby-complete-9.3.8.0.jar (shaded: org.jruby:dirgra:0.3)
Description: Simple Directed Graph
License: EPL: http://www.eclipse.org/legal/epl-v10.html
File Path: /go/.gradle/caches/modules-2/files-2.1/org.jruby/jruby-complete/9.3.8.0/8e11191265ab501930125081d8c21a3f55f1b8cd/jruby-complete-9.3.8.0.jar/META-INF/maven/org.jruby/dirgra/pom.xml
MD5: 4d7f76247a22e56064ab9db464794cd4
SHA1: 91c78b3f134c5b1f04d3a6447d246cf0a0d9a8e2
SHA256: d0f49f7eaf14307bc8ce44b14fe999c1330e029043f6e8a125b5a9f7ed1c417a

Suppressed Identifiers

[...]

PR #4913, opened 2022-10-08
Suppress false positives for common jruby rubygems

### LINKED ISSUE #4648: [FP]: jruby-readline confused for jruby itself
### Package URl

pkg:maven/rubygems/jruby-readline@1.3.7

### CPE

`cpe:/a:jruby:jruby`

### CVE

_No response_

### ODC Integration

{"label"=>"Gradle Plugin"}

### ODC Version

7.2.1

### Description

https://github.com/jruby/jruby-readline is an independently versioned project (also see https://mvnrepository.com/artifact/rubygems/jruby-readline/1.3.7)

### LINKED ISSUE #4649: [FP]: jruby-openssl confused for jruby and openssl individually
### Package URl

pkg:maven/rubygems/jruby-openssl@0.13.0

### CPE

`cpe:/a:jruby:jruby`
`cpe:/a:openssl:openssl`

### CVE

_No response_

### ODC Integration

{"label"=>"Gradle Plugin"}

### ODC Version

7.2.1

### Description

jruby-openssl is an emulation library that uses Java native crypto. It has nothing to do with openssl and is different to versioned independently of JRuby itself: https://github.com/jruby/jruby-openssl

chadlwilson:
## Fixes Issue 

- fixes #4648
- fixes #4649

## Description of Change

Adds suppressions for two libraries which the suppressions bot cannot handle due to use of non-standard (but common) RubyGems maven repo for use with JRuby.

These are located via JRuby uber-jars, e.g `jruby-complete-9.3.8.0.jar: readline.jar (shaded: rubygems:jruby-readline:1.3.7)` (and similar for jruby-openssl)

Dependency Check can locate the nested jar at `File Path: /go/.gradle/caches/modules-2/files-2.1/org.jruby/jruby-complete/9.3.8.0/8e11191265ab501930125081d8c21a3f55f1b8cd/jruby-complete-9.3.8.0.jar/META-INF/jruby.home/lib/ruby/stdlib/readline.jar/META-INF/maven/rubygems/jruby-readline/pom.xml`

N/A

### DIFF
--- a/core/src/main/resources/dependencycheck-base-suppression.xml
+++ b/core/src/main/resources/dependencycheck-base-suppression.xml
@@ -5403,4 +5403,18 @@
         <packageUrl regex="true">^pkg:npm/parseurl@.*$</packageUrl>
         <cpe>cpe:/a:parse-url_project:parse-url</cpe>
     </suppress>
+    <suppress base="true">
+        <notes><![CDATA[
+        FP per issue #4648 and #4649
+        ]]></notes>
+        <packageUrl regex="true">^pkg:maven/rubygems/jruby\-.*@.*$</packageUrl>
+        <cpe>cpe:/a:jruby:jruby</cpe>
+    </suppress>
+    <suppress base="true">
+        <notes><![CDATA[
+        FP per issue #4649
+        ]]></notes>
+        <packageUrl regex="true">^pkg:maven/rubygems/jruby\-openssl@.*$</packageUrl>
+        <cpe>cpe:/a:openssl:openssl</cpe>
+    </suppress>
 </suppressions>


core/src/main/resources/dependencycheck-base-suppression.xml
```diff
@@ -5403,4 +5403,19 @@
         <packageUrl regex="true">^pkg:npm/parseurl@.*$</packageUrl>
         <cpe>cpe:/a:parse-url_project:parse-url</cpe>
     </suppress>
+    <suppress base="true">
+        <notes><![CDATA[
+        FP per issue #4648
+        ]]></notes>
+        <packageUrl regex="true">^pkg:maven/rubygems/jruby\-readline@.*$</packageUrl>
```
aikebah:
```suggestion
        <packageUrl regex="true">^pkg:maven/rubygems/jruby\-.*@.*$</packageUrl>
```

Wouldn't it be better to immediately prevent a FP CPE match on any of the jruby gems? 

From what I can quickly gather at the jruby github org it would be safe to wildcard suppress it as jruby itself gets published under significantly different coordinates (org.jruby:jruby-*)

chadlwilson: I suppose this is possible, however if you look at https://rubygems.org/search?query=jruby- there is also `jruby-jars` which is indeed versioned alongside JRuby itself, and which (if it contained a CVE) would probably be marked against the same CPE as jruby.

I also didn't really want to guess about the content/intention of other (java) RubyGems that might be published with the `jruby-` prefix in future and may not be false positives?

So I felt it was better to be explicit about which particular java RubyGems are being referred to here.

Additionally, as can be seen by the `openssl` false positive, it's also conceptually possible that this rule could end up needing a CPE suppression for something like `gnu:readline` (CPEs [here](https://nvd.nist.gov/products/cpe/search/results?namingFormat=2.3&keyword=readline)).

WDYT?

aikebah: From what I can see jruby-jars is only a gem (easing dependency management for gems on the core JRuby libraries of the org.jruby group), not an artifact in maven, so that scenario is currently safe, and would be for the future because the org.jruby group is clearly used to publish any maven artifact from the Jruby code repository (any maven artifact that is considered an intrinsic part of Jruby itself), with rubygems group being there for publishing libraries related to rubygems that are not an intrinsic part of JRuby.

With respect to the potential of false positives on readline: that type of case might occur, which would lead to the addition of a suppression rule for readline at that point in time, but would not necessitate an individual rule per jrubygem for the suppression of the jruby cpe.
Did not include it in the suggestion, but with the suggestion applied the org.jruby:jruby suppression could also be removed from your second rule, keeping it in only for the suppression of the additional FP CPE match on OpenSSL.

In general I like to see base suppressions combining being:
* specific enough to not avoid triggering false negatives or hitting so broadly that many absolutely unrelated non-FP libraries' CPE-matches need to be checked against the suppressed CPE.
* broad enough to avoid the occurrence of very related (such as the artifactId starts with jruby- so that the text-matching easily confuses it with the Jruby language project) suppressions suppressing the same FP CPE.

So in a case like this I would prefer to preventively suppress the JRuby CPE for all jruby rubygem-supporting libraries over repeating the jruby suppression for each jruby rubygem-supporting library on its own rule.

chadlwilson: OK, thanks @aikebah - you've convinced me :-) Have made both those changes now.

chadlwilson:
This PR is broadly related to #4647 and #4652 which are also (bot-compatible) FPs on JRuby (this time with non-gem jruby add-ons) required to get a clean-ish scan on a pretty barebones JRuby web project.

aikebah (approved):
Thanks for the PR

aikebah:
> This PR is broadly related to #4647 and #4652 which are also (bot-compatible) FPs on JRuby (this time with non-gem jruby add-ons) required to get a clean-ish scan on a pretty barebones JRuby web project.

4647 was already resolved, but not marked as such, in v7.1.2. 

4652 I've just triggered the bot-processing so it should be available in the hosted suppressions soon (which will be default be used in addition to the packaged suppressions as of v8 of dependencycheck, but will require explicit suppressionfile configuration for now)

Issue #4616, opened 2022-06-20
[FP]: r2dbc-postgresql


mikehalmamoj:
### Package URl

pkg:maven/org.postgresql/r2dbc-postgresql@0.9.1.RELEASE

### CPE

`cpe:/a:postgresql:postgresql`

### ODC Integration

{"label"=>"Gradle Plugin"}

### ODC Version

7.1.0.1

### Description

r2dbc-postgresql has changed Maven coordinates - the group has changed from `io.r2dbc` to `org.postgresql`.

So [this false positive](https://github.com/jeremylong/DependencyCheck/issues/3004) has come up again - and needs a similar fix.

mikehalmamoj:
Like the GitHub Action 😄. Just be careful, it seems to have slipped in an extra backtick on the cpe.

chadlwilson:
Hmm, this one has been lingering around for a while :-( (the github actions issue with backticks is related to not having the full `:*:*:*` CPE string. If you remove the backticks in your case it will work correctly.

aikebah:
Resolved by #4755

PR #4914, opened 2022-10-09
Correct false positive ops CPE parsing when CPE is already a well-formed CPE name

chadlwilson:
## Description of Change

Users are asked to include backticks so the CPE displays correctly, but the code doesn't remove them before trying to parse.

Previous code works fine if the CPE is in formatted string binding format (e.g `cpe:2.3:a:microsoft:internet_explorer:8.0.6001:beta:*:*:*:*:*:*`) because it ignores content of the first and last tokens, but not if it's already in WFN format without version since the product token will have a backtick in it (e.g `cpe:2.3:/a:microsoft:internet_explorer`)

Suspect that users may often enter the CPE in the WFN format (without versions) because they are translating from a suppression they already have in their own config, rather than from the formatted strings that are typically included in the reports.

Example of issue with this problem: https://github.com/jeremylong/DependencyCheck/issues/4616#issuecomment-1160612100

N/A

### DIFF
--- a/.github/workflows/false-positive-ops.yml
+++ b/.github/workflows/false-positive-ops.yml
@@ -160,7 +160,7 @@ jobs:
                 purl += '/' + name.replaceAll('.','\\.');
             }
             purl += '@.*$';
-            var cpe = process.env.CPE.split(':');
+            var cpe = process.env.CPE.trim().replaceAll(/^`|`$/g,'').split(':');
             var matchCpe;
             if (cpe[1] == '2.3') {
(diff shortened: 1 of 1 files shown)


jeremylong:
Thanks! This probably needs to be done on line 52 as well - where the package URL is parsed.

jeremylong (approved):

chadlwilson:
For the packages we don't specifically ask folks to surround with backticks in the template, and if I recall correctly from personal experience it might not work properly if you *do* supply them, so it might be a good idea to make it work if you do include them.

PR #4935, opened 2022-10-15
fix concurrent modification exception

jeremylong:
Supersedes #4923.

@aikebah this is what I was thinking. currently the persistent object store is on the `engine` - if it was moved to the settings object there would likely be fewer code changes needed. Thoughts on this approach?


aikebah:
> Supersedes #4923.
> 
> @aikebah this is what I was thinking. currently the persistent object store is on the `engine` - if it was moved to the settings object there would likely be fewer code changes needed. Thoughts on this approach?

Your approach of having it in the engine feels like the clean approach to me. Putting it in settings feels like making the settings object cross the boundary implicitly set by its classname.

core/src/main/java/org/owasp/dependencycheck/analyzer/AbstractSuppressionAnalyzer.java
```diff
@@ -107,41 +84,46 @@ public Set<String> getSupportedExtensions() {
...
-        final boolean loaded = getSettings().getBoolean(SUPPRESSION_LOADED, false);
-        if (!loaded) {
-            SuppressionRules.getInstance().list().clear();
+        if (engine.hasObject(SUPPRESSION_OBJECT_KEY)) {
+            return;
         }
-        if (rules.isEmpty()) {
+        @SuppressWarnings("unchecked")
+        List<SuppressionRule> rules = (List<SuppressionRule>) engine.getObject(SUPPRESSION_OBJECT_KEY);
+        if (rules == null || rules.isEmpty()) {
```
aikebah: superfluous check as per lines 87-88 rules will be `null` (if not the method would've returned)

aikebah:
The significant amount of additional required `engine` method parameters could be avoided by allowing the analyzers to store a reference to the engine in its `AbstractAnalyzer#prepareAnalyzer(Engine engine)` implementation.
If done like that the `AbstractAnalyzer#analyzeDependency(Settings settings, Engine engine)` and `AbstractAnalyzer#analyze(Settings settings, Engine engine)`  could also have the second argument removed.

That refactoring could be done in a separate and later change. To me it feels a bit weird to first request an analyzer to prepare itself for running on behalf of an engine instance and then requiring the same engine instance to be supplied again on each and every analysis.

core/src/main/java/org/owasp/dependencycheck/analyzer/AbstractSuppressionAnalyzer.java
```diff
@@ -107,41 +84,46 @@ public Set<String> getSupportedExtensions() {
...
             } catch (SuppressionParseException ex) {
                 throw new InitializationException("Warn initializing the suppression analyzer: " + ex.getLocalizedMessage(), ex, false);
             }
-            getSettings().setBoolean(SUPPRESSION_LOADED, true);
         }
     }
 
     @Override
-    protected void analyzeDependency(Dependency dependency, Engine engine) throws AnalysisException {
+    protected synchronized void analyzeDependency(Dependency dependency, Engine engine) throws AnalysisException {
```
aikebah: I think it no longer needs to be synchronized as the rules of the engine will be initialized once by the analyzer instance that wins the race for the handle to run the synchronized prepareAnalyzer. 
Subsequent execution by parallel running analyzers of the prepareAnalyzer method will detect that the rules have already been loaded and return early without modifying the ruleset, so the ruleset will not change once initialized and the later iteration over it can no longer run into a ConcurrentModificationException.

core/src/main/java/org/owasp/dependencycheck/analyzer/AbstractSuppressionAnalyzer.java
```diff
@@ -107,41 +84,46 @@ public Set<String> getSupportedExtensions() {
...
-        //check if we have a brand new settings object - if we do the suppression rules could be different
-        final boolean loaded = getSettings().getBoolean(SUPPRESSION_LOADED, false);
-        if (!loaded) {
-            SuppressionRules.getInstance().list().clear();
+        if (engine.hasObject(SUPPRESSION_OBJECT_KEY)) {
+            return;
         }
+        @SuppressWarnings("unchecked")
+        final List<SuppressionRule> rules = (List<SuppressionRule>) engine.getObject(SUPPRESSION_OBJECT_KEY);
         if (rules.isEmpty()) {
```
aikebah: Here you changed the code to trigger a guaranteed NPE. lines 87 and 88 guarantee that rules will be null (if rules was non-null here then the implementation of Engine#hasObject would be flawed). My review comment concerned the full removal of the if-statement, as the first clause was a guaranteed true and therefor the entire if clause was obsolete.

jeremylong:
@aikebah thank you for the feedback on the PR. The abstract suppression analyzer now stores a reference to the engine and I also moved the call in the engine to report the unused suppression rules to a new analyzer - so that it is cleaner and not a weird one off solution.

aikebah:
@jeremylong LGTM apart from the nitpicking review comments of Checkstyle

aikebah (approved):

Issue #4966, opened 2022-10-19
Getting a java error in a CVE check on launch


mushu999:
Is this something wrong on *my* system? Here is the error:
```
[INFO] Processing Started for NVD CVE - 2021
Oct 19, 2022 1:48:15 PM com.fasterxml.jackson.module.blackbird.util.ReflectionHack$Java9Up init
WARNING: Unable to find Java 9+ MethodHandles.privateLookupIn.  Blackbird is not performing optimally!
java.lang.NoSuchMethodException: no such method: java.lang.invoke.MethodHandles.privateLookupIn(Class,Lookup)Lookup/invokeStatic
        at java.lang.invoke.MemberName.makeAccessException(Unknown Source)
. . .
```
So do I need to update my Java installation? This is version 7.3.0 of dependency-check code on Windows 10. I run it like I always do. Just updated from ver 7.1.1 and left the **data** directory from the earlier version.

mprins:
Where is the ERROR? I only see a WARNING...

> So do I need to update my Java installation?

you have not told us which version you are using, so it is difficult to advise

jeremylong:
This is an expected warning when running on Java 8. See other projects responses: https://issuehint.com/issue/dropwizard/dropwizard/5318#84997997

Everything will still work as expected.

PR #4971, opened 2022-10-20
Use Afterburner if JRE 8

jeremylong:
## Fixes Issue #4966

If using JRE 8 use afterburner instead of blackbird.

### DIFF
--- a/core/src/main/java/org/owasp/dependencycheck/data/update/nvd/NvdCveParser.java
+++ b/core/src/main/java/org/owasp/dependencycheck/data/update/nvd/NvdCveParser.java
@@ -131,7 +131,7 @@ public final class NvdCveParser {
      * @return the Java major version as a whole number
      */
     private static int getJavaVersion() {
-        String version = System.getProperty("java.version");
+        String version = System.getProperty("java.specification.version");
         if (version.startsWith("1.")) {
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/owasp/dependencycheck/data/update/nvd/NvdCveParser.java
```diff
@@ -117,6 +125,24 @@ public void parse(File file) throws UpdateException, CorruptedDatastreamExceptio
         }
     }
 
+    /**
+     * Returns the Java major version as a whole number.
+     *
+     * @return the Java major version as a whole number
+     */
+    private static int getJavaVersion() {
+        String version = System.getProperty("java.version");
```
aikebah:
```suggestion
        String version = System.getProperty("java.specification.version");
```

I think it will be better to use the java.specification.version


So that we don't have to change the logic to support the early-access releasese `<major>-ea` (see [JEP-223](https://openjdk.org/jeps/223))

aikebah (changes requested):
Needs either my suggestion or a fix to work with the early-access version pattern to prevent a NumberFormatException on early-access JREs

aikebah (approved):

Issue #4972, opened 2022-10-21
Exception occurred initializing Yarn Audit Analyzer


antonysmith-mandogroup:
@jeremylong according to #4049, this should have been fixed in 7.0.0, but I still appear to be seeing it in 7.1.0.

I'm getting the following in a project that only has references to yarn.lock files in node_modules: 

```
[WARN] The Yarn Audit Analyzer has been disabled. Yarn executable was not found.
[ERROR] Exception occurred initializing Yarn Audit Analyzer.
```

As suggested above, I've run ODC with --log and this is the output:

```
	Line     48: DEBUG - Setting: analyzer.yarn.audit.enabled='true'
	Line  43436: DEBUG - Found file ***\node_modules\@yarnpkg\lockfile\README.md
	Line  43438: DEBUG - Found file ***\node_modules\@yarnpkg\lockfile\index.js
	Line  43440: DEBUG - Found file ***\node_modules\@yarnpkg\lockfile\package.json
	Line  51178: DEBUG - Found file ***\node_modules\browserify-zlib\yarn.lock
	Line  51480: DEBUG - Found file ***\node_modules\bs-recipes\recipes\webpack.preact-hot-loader\yarn.lock
	Line  55474: DEBUG - Found file ***\node_modules\chrome-trace-event\yarn.lock
	Line  66882: DEBUG - Found file ***\node_modules\debug-fabulous\yarn-error.log
	Line  66884: DEBUG - Found file ***\node_modules\debug-fabulous\yarn.lock
	Line  67694: DEBUG - Found file ***\node_modules\dom-event-types\yarn.lock
	Line  82150: DEBUG - Found file ***\node_modules\fileset\yarn.lock
	Line  82842: DEBUG - Found file ***\node_modules\form-data\yarn.lock
	Line  84908: DEBUG - Found file ***\node_modules\gulp-accessibility\yarn.lock
	Line  85274: DEBUG - Found file ***\node_modules\gulp-htmlhint\yarn.lock
	Line  88970: DEBUG - Found file ***\node_modules\has-yarn\index.d.ts
	Line  88972: DEBUG - Found file ***\node_modules\has-yarn\index.js
	Line  88974: DEBUG - Found file ***\node_modules\has-yarn\license
	Line  88976: DEBUG - Found file ***\node_modules\has-yarn\package.json
	Line  88978: DEBUG - Found file ***\node_modules\has-yarn\readme.md
[...]

PR #4974, opened 2022-10-21
Exclude node_modules from Maven's scan path

jeremylong:
## Fixes Issue #4972

99% positive the error reported in #4972 is due to how file sets work with Maven. Other places in the code we filter out scanning the node_modules directory - filtering here should prevent the reported error regarding yarn.


aikebah (approved):
I share your analysis that it would solve the issue and is an appropriate exclude for maven projects

Issue #4973, opened 2022-10-21
CPE CVE Mapping


parimala-kathir:
[hadoop-0.13.0-core.jar] This is my jar. We need to analyze whether this has known vulnerabilities or not. 

As per Dependency checker(5.3.2) the corresponding CPE mapped with 5 CVEs
-------------------cpe:2.3:a:apache:hadoop:0.13.0:*:*:*:*:*:*:* |   | CRITICAL | 5

But as per Dependency checker(7.3.0) the corresponding CPE mapped with 0 CVEs
-------------------hadoop-0.13.0-core.jar |   |   |   | 0

jeremylong:
So - have you actually looked at the CVEs to know if they are correct? 5.3.2 is old and unsupported - many bugs and false positives have been fixed since then.

parimala-kathir:
NVD also shows 5 CVEs for CPE (cpe:2.3:a:apache:hadoop:0.13.0:*:*:*:*:*:*:*)

parimala-kathir:
If NVD shows crct result then is there any bug/issue in 7.3.0

PR #4997, opened 2022-10-28
Fix fn hadoop update

jeremylong:
### Supersedes #4991

The original PR has been updated so a database schema version is no longer needed - instead if we have additional entries to update we can simply add them to the [dbEcosystemCacheUpdates.sql](https://github.com/jeremylong/DependencyCheck/compare/fixFN-hadoop-update?expand=1#diff-d8e20743037bc8c22dc50534ac9409f12bba0574fa0e4f5f32ac9437c286f044) file. This allows easier updates for false negatives caused by the ecosystem.

### Fixes Issue #4973

Hadoop was incorrectly marked as `native` in the database. Update the ecosystem to MULTIPLE and the false negative is resolved.


aikebah (approved):
Better solution for such updates indeed

PR #5004, opened 2022-10-31
Update Suppression File

jeremylong:
Update the base suppression file from the `generatedSuppressions` branch to prepare for the 7.3.1 release.
- Delineator added to `generatedSuppressions.xml`:  https://github.com/jeremylong/DependencyCheck/pull/new/updateSuppressions
- Intent is to delete these entries from the generated suppressions once 7.3.1 is no longer supported.

The purpose of this is to keep the base suppression file current - even when users do not utilize the hosted suppression file (some run ODC in an offline mode).


jeremylong:
@aikebah the intent is to ship 7.3.1 this week.

aikebah (approved):
I see we are like-minded on how to manage the hosted suppressions.

PR #5007, opened 2022-11-01
Add CLI option to disable Dart analyzer

Hallot:
## Description of Change

Add a CLI option to disable the new Dart analyzer introduced in 7.3.0. (Because it crashes in my case.)

No, feel free to point to where I can do it.


utils/src/main/java/org/owasp/dependencycheck/utils/Settings.java
```diff
@@ -421,6 +421,10 @@ public static final class KEYS {
          * The properties key for whether the Golang Dep analyzer is enabled.
          */
         public static final String ANALYZER_GOLANG_DEP_ENABLED = "analyzer.golang.dep.enabled";
+        /**
+         * The properties key for whether the Dart analyzer is enabled.
+         */
+        public static final String ANALYZER_DART_ENABLED = "analyzer.dart.enabled";
```
jeremylong: This was already defined on line 740.
```suggestion
```

jeremylong (approved):
Thanks - I can't believe I forgot to add these when the analyzer was added...

Issue #5008, opened 2022-11-01
Dart analyzer NPE


Hallot:
**Describe the bug**
Dart analyzer crashes with a NPE when analyzing a pubspec.yaml file.

**Version of dependency-check used**
The problem occurs using version 7.3.0 of the cli

**Log file**

> [WARN] An unexpected error occurred during analysis of '/tmp/dev/.cache/bazel/_bazel/45687a9fc51e3180e3ee6f957bfa2b6a/external/com_google_protobuf/examples/pubspec.yaml' (Dart Package Analyzer): null
> [ERROR]
> java.lang.NullPointerException: null
> 	at org.owasp.dependencycheck.analyzer.DartAnalyzer.analyzeYamlFileDependencies(DartAnalyzer.java:137)
> 	at org.owasp.dependencycheck.analyzer.DartAnalyzer.analyzeDependency(DartAnalyzer.java:117)
> 	at org.owasp.dependencycheck.analyzer.AbstractAnalyzer.analyze(AbstractAnalyzer.java:131)
> 	at org.owasp.dependencycheck.AnalysisTask.call(AnalysisTask.java:88)
> 	at org.owasp.dependencycheck.AnalysisTask.call(AnalysisTask.java:37)
> 	at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:264)
> 	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1128)
> 	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:628)
> 	at java.base/java.lang.Thread.run(Thread.java:829)
> [INFO] Finished Dart Package Analyzer (0 seconds)

**To Reproduce**
pubspec.yaml content:
```
name: addressbook
description: dartlang.org example code.

dependencies:
  protobuf:
```

**Expected behavior**
No crashes happen.

**Additional context**
The analyzer is run in a Linux Docker container running on aarch64 ARM Mac.

I have outOfMemory errors with the CMake analyzer in the same run, but it is probably not related.

> [WARN] An unexpected error occurred during analysis of '/tmp/dev/.cache/bazel/_bazel/45687a9fc51e3180e3ee6f957bfa2b6a/external/cmake-3.23.2-linux-aarch64/share/cmake-3.23/Modules/FindGTK2.cmake' (CMake Analyzer): Java heap space
> [ERROR]
[...]

PR #5011, opened 2022-11-02
Update Dart Analyzer

jeremylong:
## Fixes Issue #5008

* Adds checks for empty/non-existent elements.
* Don't process yaml if lock is present (yaml may have blank versions)


core/src/main/java/org/owasp/dependencycheck/analyzer/DartAnalyzer.java
```diff
@@ -212,11 +223,13 @@ private void addYamlDependenciesToEngine(Iterator<Map.Entry<String, JsonNode>> d
     }
 
     private void addYamlDartDependencyToEngine(JsonNode rootNode, File file, Engine engine) throws AnalysisException {
-        final String dartVersion = rootNode.get("environment").get("sdk").textValue();
-        final String minimumVersion = extractMinimumVersion(dartVersion);
+        if (rootNode.hasNonNull("environment")) {
+            final String dartVersion = rootNode.get("environment").get("sdk").textValue();
```
aikebah, line 227: Think this should have null checks as well? 
Not familiar with the Dart ecosystem, but https://dart.dev/tools/pub/pubspec
and the readme for environment property hint to me that the elements such as sdk underneath environment may also be unavailable (for packages for Dart version <2.0)

aikebah (changes requested):

core/src/main/java/org/owasp/dependencycheck/analyzer/DartAnalyzer.java
jeremylong: Good catch - thanks!

aikebah (approved):

Issue #3666, opened 2021-09-22
java.lang.ClassNotFoundException: org.apache.commons.jcs.auxiliary.disk.indexed.IndexedDiskElementDescriptor


smiklosovic:
@jeremylong this is still happening, the older issue was autoclosed.

**Describe the bug**
I am just doing ant dependency-check

**Version of dependency-check used**
6..3.1

**Log file**

    [dependency-check] Sep 22, 2021 8:04:15 PM org.apache.commons.jcs.auxiliary.disk.indexed.IndexedDiskCache loadKeys
    [dependency-check] SEVERE: Region [CENTRAL] Problem loading keys for file CENTRAL
    [dependency-check] java.lang.ClassNotFoundException: org.apache.commons.jcs.auxiliary.disk.indexed.IndexedDiskElementDescriptor
    [dependency-check]      at java.net.URLClassLoader.findClass(URLClassLoader.java:382)
    [dependency-check]      at java.lang.ClassLoader.loadClass(ClassLoader.java:418)
    [dependency-check]      at org.codehaus.classworlds.RealmClassLoader.loadClassDirect(RealmClassLoader.java:195)
    [dependency-check]      at org.codehaus.classworlds.DefaultClassRealm.loadClass(DefaultClassRealm.java:255)
    [dependency-check]      at org.codehaus.classworlds.RealmClassLoader.loadClass(RealmClassLoader.java:214)
    [dependency-check]      at java.lang.ClassLoader.loadClass(ClassLoader.java:351)
    [dependency-check]      at java.lang.Class.forName0(Native Method)
    [dependency-check]      at java.lang.Class.forName(Class.java:348)
    [dependency-check]      at org.apache.commons.jcs.io.ObjectInputStreamClassLoaderAware.resolveClass(ObjectInputStreamClassLoaderAware.java:37)
    [dependency-check]      at java.io.ObjectInputStream.readNonProxyDesc(ObjectInputStream.java:1867)
    [dependency-check]      at java.io.ObjectInputStream.readClassDesc(ObjectInputStream.java:1750)
    [dependency-check]      at java.io.ObjectInputStream.readOrdinaryObject(ObjectInputStream.java:2041)
    [dependency-check]      at java.io.ObjectInputStream.readObject0(ObjectInputStream.java:1572)
    [dependency-check]      at java.io.ObjectInputStream.readObject(ObjectInputStream.java:430)
[...]

PR #5046, opened 2022-11-14
Set the context class loader when executing ant tasks

danielhodder:
Issue: #3666

## Description of Change

Adds a context class loader to all the ant tasks. Other build tools set the ContextClassLoader property on the current Thread instance before calling into the plugin, ant does not.

The JCS library has some logic down in it's deserializer which finds the classloader to use. If it used the 'default' one this would work; but since it uses the Context Class Loader it fails since that classloader is something else.

*no*


jeremylong (approved):

danielhodder:
I ran the step the failed on my machine and it worked. It looks like the build system had a hiccup.

jeremylong:
I kicked off the failed job again.

jeremylong:
Thanks for the PR!

Issue #5039, opened 2022-11-09
Create HTML report that works without JavaScript


chovyy:
**Is your feature request related to a problem? Please describe.**
Currently it is not possible to embed a proper working HTML report of Dependency Check into Jenkins. This is because the HTML report relys on some JavaScript functionalities (e.g. for expanding/collapsing some content etc.), but Jenkins has a [Content Security Policy (CSP)](https://www.jenkins.io/doc/book/security/configuring-content-security-policy/) that prohibits the execution of embedded JavaScript, see also #2213. This might applies to other build or SLM tools.

**Describe the solution you'd like**
It would be nice to have a least the option to create a JavaScript-free HTML report that shows all content without the need of JavaScript-based expanding/collapsing and so on. Some minor functionalities might be left out.

**Describe alternatives you've considered**
An alternative is to loosen the Jenkins CSP, but that possibly leads to security issues in other contexts.

**Additional context**
The problem behind this issue is already discussed in #2213.

jeremylong:
As mentioned previously - the CSP can be modified: https://www.jenkins.io/doc/book/security/configuring-content-security-policy/

Many Jenkins users also generate the junit report so the CVEs end up showing as failures - but I do see the value in an HTML report that does not include JavaScript.

OrangeDog:
Note that there is also [a plugin](https://plugins.jenkins.io/dependency-check-jenkins-plugin/) that will generate integrated Jenkins reports from the xml output.

chovyy:
That's right, but afaik this plugin has no pipeline support so far.

UPDATE: Sorry, that was wrong. There is actually a [pipeline reference page](https://www.jenkins.io/doc/pipeline/steps/dependency-check-jenkins-plugin/).

chovyy:
@jeremylong: Great work, thanks a lot! 👍  But afais, the new format option `JENKINS` is still missing in the docs.

PR #5053, opened 2022-11-15
feat: add HTML report without JS

jeremylong:
## Fixes Issue #5039

Add a JENKINS report type - which is a copy of the HTML report without the JavaScript. As such the HTML Report for Jenkins would not have the capability to create suppression rules.

### DIFF
--- a/ant/src/main/java/org/owasp/dependencycheck/taskdefs/Check.java
+++ b/ant/src/main/java/org/owasp/dependencycheck/taskdefs/Check.java
@@ -229,12 +229,12 @@ public class Check extends Update {
      */
     private Boolean autoUpdate;
     /**
-     * The report format to be generated (HTML, XML, JUNIT, CSV, JSON, SARIF,
+     * The report format to be generated (HTML, XML, JUNIT, CSV, JSON, SARIF, JENKINS,
      * ALL). Default is HTML.
      */
     private String reportFormat = "HTML";
     /**
-     * The report format to be generated (HTML, XML, JUNIT, CSV, JSON, SARIF,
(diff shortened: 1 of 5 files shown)


ant/src/main/java/org/owasp/dependencycheck/taskdefs/Check.java
```diff
@@ -229,12 +229,12 @@ public class Check extends Update {
...
     private Boolean autoUpdate;
     /**
-     * The report format to be generated (HTML, XML, JUNIT, CSV, JSON, SARIF,
+     * The report format to be generated (HTML, XML, JUNIT, CSV, JSON, SARIF, JENKINS,
      * ALL). Default is HTML.
      */
     private String reportFormat = "HTML";
     /**
-     * The report format to be generated (HTML, XML, JUNIT, CSV, JSON, SARIF,
+     * The report format to be generated (HTML, XML, JUNIT, CSV, JSON, SARIF, JENKINs,
```
nhumblot: Hi 👋

Shouldn't it be a caps locked `JENKINS` value instead? 🤔

nhumblot (approved):
Hi 👋

Looks good to me except a typo. I also went to look into the website to see if there would be some place where the documentation should be changed but I don't think so.

Thank you for your work! 🤗

ant/src/main/java/org/owasp/dependencycheck/taskdefs/Check.java
jeremylong:
```suggestion
     * The report format to be generated (HTML, XML, JUNIT, CSV, JSON, SARIF, JENKINS,
```

jeremylong:
Thanks for catching the typo!

PR #5068, opened 2022-11-19
fix: automation should use conventional commits

jeremylong:
While I'm not perfect with this - I've been trying to follow conventional commits: https://www.conventionalcommits.org/en/v1.0.0/

This PR updates the release plugin to use conventional commit prefix when making commits during the release process.

### DIFF
--- a/.github/workflows/false-positive-approvals.yml
+++ b/.github/workflows/false-positive-approvals.yml
@@ -129,7 +129,7 @@ jobs:
                 if (err) throw err;
               });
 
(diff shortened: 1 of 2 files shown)


aikebah (approved):

aikebah:
Maybe reword the PR to "automation should use conventional commits" as the PR concerns both mvn release and the fp gitops commits

aikebah (approved):

jeremylong:
Thanks for pointing that out - the gitops was (obviously) added after I created the PR...

PR #5072, opened 2022-11-20
fix: prevent duplicate, non-built (reactor) virtual dependencies

### LINKED ISSUE #290: Performance Degradation On Large Multi-Projects With Nested Transitive Project Dependencies
Observed after moving from 6.5.3 to 7.0.0.

### Fault Description

Running ODC against a large Gradle project with several hundred sub-projects having nested cross-references (i.e. sub A depends on sub B which again depends on sub C and so on) results in very long execution times. That makes it difficult to use the plugin for a quick check (e.g. after adjusting some vulnerability exposing dependencies).

As a reference: I'm used to see ODC complete after 3 minutes. Now I have to wait for more than 30 minutes (not including DB update time).

I attach [mwe.zip](https://github.com/dependency-check/dependency-check-gradle/files/10048866/mwe.zip) for reproduction. Tested with Gradle 7.1.

Debugging into Depedency Check Core Engine you can see the generated workload which leads to increased execution time for the following analyzers.
<table>
<tr>
<th>Analyzer</th>
<th>Analysis Task Count v6.5.3</th>
<th>Analysis Task Count v7.0.0</th>
<th>Mean Execution time [s] v6.5.3</th>
<th>Mean Execution time [s] v7.0.0</th>
</tr>
<tr>
<td>Dependency Bundling Analyzer</td>
<td rowspan="4" align="center">496</td>
<td rowspan="4" align="center">105,607</td>
<td>&lt; 1</td>
<td>1,900</td>
</tr>
<tr>
<td>Dependency Merging Analyzer</td>
<td>&lt; 1</td>
<td>90</td>
</tr>
<tr>
<td>CPE Analyzer</td>
<td>2</td>
<td>52</td>
</tr>
<tr>
<td>Sonatype OSS Index Analyzer</td>
<td>&lt; 1</td>
<td>30</td>
</tr>
</table>

While different platforms may result in different execution times the point here is that we have an increase by at least factor 10 on a fixed platform. This behavior hasn't changed since version 7.0.0.

### Analysis

The cause seems to be a change in `AbstractAnalyze#addDependency()`. See #239, #245 and commit [5f512b4](https://github.com/dependency-check/dependency-check-gradle/commit/5f512b45829ad0afd55c9067e6f0831713ecb406) for reference.

https://github.com/dependency-check/dependency-check-gradle/blob/fed02e4456a3dc6c6cbd7ae9ace51574c7ac69e2/src/main/groovy/org/owasp/dependencycheck/gradle/tasks/AbstractAnalyze.groovy#L526-L530

Judging from the issue description

> The false positives seem to arise because the plugin has lost the link from a given dependency jar to the actual Maven package it came from.

I would say the `MavenArtifact` creation is correct since it implicitly adds the package URL to the dependency. Always adding the dependency to the list of dependencies from the `Engine`, however, may create duplicates because it might already be in the list.

### Suggested Fix

Use the former approach to check whether a dependency already exists. If so, update solely the project reference since the only reason for a recurring dependency is a different project or configuration.

jeremylong:
Resolves https://github.com/dependency-check/dependency-check-gradle/issues/290

The gradle plugin and maven plugin could create duplicate dependencies that would be present during the analysis phase and only be cleaned up in post-processing. This can cause large projects that do not run a build before analysis to take a very long time to complete.

### DIFF
--- a/core/src/main/java/org/owasp/dependencycheck/Engine.java
+++ b/core/src/main/java/org/owasp/dependencycheck/Engine.java
@@ -83,6 +83,7 @@ import static org.owasp.dependencycheck.analyzer.AnalysisPhase.PRE_FINDING_ANALY
 import static org.owasp.dependencycheck.analyzer.AnalysisPhase.PRE_IDENTIFIER_ANALYSIS;
 import static org.owasp.dependencycheck.analyzer.AnalysisPhase.PRE_INFORMATION_COLLECTION;
 import org.owasp.dependencycheck.analyzer.DependencyBundlingAnalyzer;
+import org.owasp.dependencycheck.dependency.naming.Identifier;
 
 /**
  * Scans files, directories, etc. for Dependencies. Analyzers are loaded and
@@ -250,13 +251,31 @@ public class Engine implements FileFilter, AutoCloseable {
     }
 
     /**
-     * Adds a dependency.
+     * Adds a dependency. In some cases, when adding a virtual dependency, the
+     * method will identify if the virtual dependency was previously added and
+     * if so the existing dependency will be returned.
      *
      * @param dependency the dependency to add
+     * @return the dependency added; in the case of virtual dependencies this
+     * may not be the same as the dependency passed in as an argument
      */
-    public synchronized void addDependency(Dependency dependency) {
+    public synchronized Dependency addDependency(Dependency dependency) {
+        if (dependency.isVirtual()) {
+            for (Dependency existing : dependencies) {
+                if (existing.isVirtual()
+                        && existing.getSha256sum() != null
+                        && existing.getSha256sum().equals(dependency.getSha256sum())
+                        && existing.getDisplayFileName() != null
+                        && existing.getDisplayFileName().equals(dependency.getDisplayFileName())
+                        && identifiersMatch(existing.getSoftwareIdentifiers(), dependency.getSoftwareIdentifiers())) {
+                    DependencyBundlingAnalyzer.mergeDependencies(existing, dependency, null);
+                    return existing;
+                }
+            }
+        }
         dependencies.add(dependency);
         dependenciesExternalView = null;
+        return dependency;
     }
 
     /**
@@ -1247,6 +1266,22 @@ public class Engine implements FileFilter, AutoCloseable {
         }
     }
     //CSON: LineLength
+
+    private boolean identifiersMatch(Set<Identifier> left, Set<Identifier> right) {
+        if (left != null && right != null && left.size() > 0 && left.size() == right.size()) {
+            int count = 0;
+            for (Identifier l : left) {
+                for (Identifier r : right) {
+                    if (l.getValue().equals(r.getValue())) {
+                        count += 1;
+                        break;
+                    }
+                }
+            }
+            return count == left.size();
+        }
+        return false;
+    }
 
     /**
      * {@link Engine} execution modes.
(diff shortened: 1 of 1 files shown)


core/src/main/java/org/owasp/dependencycheck/Engine.java
```diff
@@ -250,13 +251,31 @@ public List<Analyzer> getAnalyzers(AnalysisPhase phase) {
...
+                        && existing.getDisplayFileName().equals(dependency.getDisplayFileName())
+                        && identifiersMatch(existing.getSoftwareIdentifiers(), dependency.getSoftwareIdentifiers())) {
+                    DependencyBundlingAnalyzer.mergeDependencies(existing, dependency, null);
+                    return existing;
+                }
+            }
+        }
         dependencies.add(dependency);
         dependenciesExternalView = null;
+        return dependency;
```
aikebah:
```suggestion
    public synchronized void addDependency(Dependency dependency) {
        if (dependency.isVirtual()) {
            for (Dependency existing : dependencies) {
                if (existing.isVirtual()
                        && existing.getSha256sum() != null
                        && existing.getSha256sum().equals(dependency.getSha256sum())
                        && existing.getDisplayFileName() != null
                        && existing.getDisplayFileName().equals(dependency.getDisplayFileName())
                        && identifiersMatch(existing.getSoftwareIdentifiers(), dependency.getSoftwareIdentifiers())) {
                    DependencyBundlingAnalyzer.mergeDependencies(existing, dependency, null);
                    return;
                }
            }
        }
        dependencies.add(dependency);
        dependenciesExternalView = null;
```

As we never use the result there is no use in returning the dependency. We can just early-return from the method.

aikebah (commented):
Existing code would also be fine, but the method signature doesn't have to change in my view

jeremylong:
I was debating whether to update the signature of the method or not. I agree there was no reason to do that.

core/src/main/java/org/owasp/dependencycheck/Engine.java
```diff
@@ -250,11 +251,28 @@ public List<Analyzer> getAnalyzers(AnalysisPhase phase) {
...
 
     /**
-     * Adds a dependency.
+     * Adds a dependency. In some cases, when adding a virtual dependency, the
+     * method will identify if the virtual dependency was previously added and
+     * if so the existing dependency will be returned.
      *
      * @param dependency the dependency to add
+     * @return the dependency added; in the case of virtual dependencies this
+     * may not be the same as the dependency passed in as an argument
```
jeremylong: Updating javadoc per earlier suggestion.
```suggestion
```

core/src/main/java/org/owasp/dependencycheck/Engine.java
```diff
@@ -250,11 +251,28 @@ public List<Analyzer> getAnalyzers(AnalysisPhase phase) {
     }
 
     /**
-     * Adds a dependency.
+     * Adds a dependency. In some cases, when adding a virtual dependency, the
+     * method will identify if the virtual dependency was previously added and
+     * if so the existing dependency will be returned.
```
jeremylong:
```suggestion
     * Adds a dependency. In some cases, when adding a virtual dependency, the
     * method will identify if the virtual dependency was previously added and update
     * the existing dependency rather then adding a duplicate.
```

aikebah (approved):
LGTM

Issue #4031, opened 2022-01-28
Angular app - Package-Lock files throw errors


crypticraven:
**Describe the bug**
When scanning an Angular application, the dependency checker application throws an error when it comes across package-lock files.

**Version of dependency-check used**
6.5.3

**Log file**

[INFO] Finished NVD CVE Analyzer (0 seconds)
[ERROR] NodeAuditAnalyzer failed on C:\a\1\s\NextGen\UserDriven\Website\SPA\package-lock.json
[WARN] An error occurred while analyzing 'C:\a\1\s\NextGen\UserDriven\Website\SPA\package-lock.json' (Node Audit Analyzer).
[INFO] Finished Node Audit Analyzer (0 seconds)
00:00  INFO: Vulnerability found: jquery below 1.12.0
00:00  INFO: Vulnerability found: jquery below 1.12.0
00:00  INFO: Vulnerability found: jquery below 3.4.0
00:00  INFO: Vulnerability found: jquery below 3.5.0
00:00  INFO: Vulnerability found: jquery below 3.5.0
00:00  INFO: Vulnerability found: jquery below 1.12.0
00:00  INFO: Vulnerability found: jquery below 1.12.0
00:00  INFO: Vulnerability found: jquery below 3.4.0
00:00  INFO: Vulnerability found: jquery below 3.5.0
00:00  INFO: Vulnerability found: jquery below 3.5.0
00:00  INFO: Vulnerability found: jquery below 1.12.0
00:00  INFO: Vulnerability found: jquery below 1.12.0
00:00  INFO: Vulnerability found: jquery below 3.4.0
00:00  INFO: Vulnerability found: jquery below 3.5.0
00:00  INFO: Vulnerability found: jquery below 3.5.0
[INFO] Finished RetireJS Analyzer (1 seconds)
[INFO] Finished Sonatype OSS Index Analyzer (4 seconds)
[INFO] Finished Vulnerability Suppression Analyzer (0 seconds)
[INFO] Finished Dependency Bundling Analyzer (0 seconds)
[INFO] Analysis Complete (24 seconds)
[INFO] Writing report to: C:\a\1\TestResults\dependency-check\dependency-check-report.xml
[INFO] Writing report to: C:\a\1\TestResults\dependency-check\dependency-check-report.html
[INFO] Writing report to: C:\a\1\TestResults\dependency-check\dependency-check-junit.xml
[...]

Issue #3717, opened 2021-10-08
Could not perform Node Audit analysis


sranney:
My team has been working over the last couple of days to understand why this error is being thrown in our Jenkins build pipeline:
```
[2021-10-08T18:09:32.862Z] [DependencyCheck] [INFO] Checking for updates
[2021-10-08T18:09:32.869Z] [DependencyCheck] [INFO] Skipping NVD check since last check was within 4 hours.
[2021-10-08T18:09:32.958Z] [DependencyCheck] [INFO] Check for updates complete (96 ms)
[2021-10-08T18:09:33.112Z] [DependencyCheck] [INFO] 
[2021-10-08T18:09:33.112Z] [DependencyCheck] 
[2021-10-08T18:09:33.112Z] [DependencyCheck] Dependency-Check is an open source tool performing a best effort analysis of 3rd party dependencies; false positives and false negatives may exist in the analysis performed by the tool. Use of the tool and the reporting provided constitutes acceptance for use in an AS IS condition, and there are NO warranties, implied or otherwise, with regard to the analysis or its use. Any use of the tool and the reporting provided is at the user’s risk. In no event shall the copyright holder or OWASP be held liable for any damages whatsoever arising out of or in connection with the use of this tool, the analysis performed, or the resulting report.
[2021-10-08T18:09:33.112Z] [DependencyCheck] 
[2021-10-08T18:09:33.112Z] [DependencyCheck] 
[2021-10-08T18:09:33.112Z] [DependencyCheck]    About ODC: https://jeremylong.github.io/DependencyCheck/general/internals.html
[2021-10-08T18:09:33.112Z] [DependencyCheck]    False Positives: https://jeremylong.github.io/DependencyCheck/general/suppression.html
[2021-10-08T18:09:33.112Z] [DependencyCheck] 
[2021-10-08T18:09:33.112Z] [DependencyCheck] 💖 Sponsor: https://github.com/sponsors/jeremylong
[2021-10-08T18:09:33.112Z] [DependencyCheck] 
[2021-10-08T18:09:33.112Z] [DependencyCheck] 
[2021-10-08T18:09:33.112Z] [DependencyCheck] [INFO] Analysis Started
[2021-10-08T18:09:33.117Z] [DependencyCheck] [INFO] Finished Archive Analyzer (0 seconds)
[2021-10-08T18:09:33.125Z] [DependencyCheck] [INFO] Finished File Name Analyzer (0 seconds)
[...]

Issue #3716, opened 2021-10-08
Invalid payload submitted to Node Audit API. Received response code: 400 Bad Request


rowe42:
In our project the dependency-check fails on the `Node Audit` analyzer with the error:
```
Could not perform Node Audit analysis. Invalid payload submitted to Node Audit API.
```
However, this is NOT the same problem as in #2641

Investigating I found out that in fact the payload, which is generated in https://github.com/jeremylong/DependencyCheck/blob/main/core/src/main/java/org/owasp/dependencycheck/analyzer/NodeAuditAnalyzer.java#L184, seems not to be accepted by the NPM REST API.

What I did was 
- run the dependency check (in my case throught the CLI like this: `dependency-check -s . --disableYarnAudit --log dependencycheck.log`
- open the file `dependencycheck.log` and find the payload in there (close to the line `Node Audit Payload`) and copy it
- run the call to `https://registry.npmjs.org/-/npm/v1/security/audits` manually with curl or Postman like this
```bash
curl --location --request POST 'https://registry.npmjs.org/-/npm/v1/security/audits' \
--header 'Content-Type: application/json' \
--data-raw 'PLACE-PAYLOAD-HERE'
```

and I get the same error

```json
{
    "statusCode": 400,
    "error": "Bad Request",
    "message": "Invalid package tree, run  npm install  to rebuild your package-lock.json"
}
```

When I run `npm audit` it works though, so I think it's not a problem with my `package-lock.json` or with NPM.

Here is the payload in our example
[nodeaudit_payload.txt](https://github.com/jeremylong/DependencyCheck/files/7311335/nodeaudit_payload.txt)

So, the problem must be somewhere in `NpmPayloadBuilder.java`.

Stripping down the payload on a trial-and-error basis, the error persists and only goes away when the entries in section `requires` resemble the entries in section `dependencies` - but not sure if that is really the rule.

magob-sagefr:
Hello,
I have the very same behavior, meaning that npm audit works fine on the contrary to dependency checker in the node audit analyzer which fails both on my Jenkins server and locally.
No more specific information thant "Invalid payload submitted to node audit analyzer".
The used API URL is the default one : https://registry.npmjs.org/-/npm/v1/security/audits
Any idea ?

andrzejcichonroche:
[...]

PR #5078, opened 2022-11-21
fix: add package-lock v2 and v3

phoenix741:
Signed-off-by: Ulrich Van Den Hekke <ulrich.vdh@shadoware.org>

## Fixes Issue #

Fix issue related to version of package-lock.json :

#4031
#3717 (maybe)
#3716 (maybe)

## Description of Change

There is two problem actually with recent version of npm:
- the structure of package-lock.json change and in the last version can't be used to make the audit
- The error occure because of a bug of npm that remove integrity of package-lock.json (https://github.com/npm/cli/issues/4460). "npm audit" dont have the problem because the endpoint called now different and doesn't need the integrity in package-lock.json.

This PR change the first point (i will maybe create a PR later for the second point)

yes

### DIFF
--- a/core/src/main/java/org/owasp/dependencycheck/analyzer/NodePackageAnalyzer.java
+++ b/core/src/main/java/org/owasp/dependencycheck/analyzer/NodePackageAnalyzer.java
@@ -93,6 +93,10 @@ public class NodePackageAnalyzer extends AbstractNpmAnalyzer {
      */
     public static final String SHRINKWRAP_JSON = "npm-shrinkwrap.json";
     /**
+     * The name of the directory that contains node modules
+     */
+    public static final String NODE_MODULES_DIRNAME = "node_modules";
+    /**
      * Filter that detects files named "package.json", "package-lock.json", or
      * "npm-shrinkwrap.json".
      */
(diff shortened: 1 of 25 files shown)


aikebah:
Would it be possible to construct the testcases with fewer packages/dependencies? Feels to me like overkill, but maybe that's just a sign of how finegrained and interconnected the node ecosystem is.

I've only seen its usages in frontend framework development where the packages quickly accumulate to the many hundreds, a count that I myself consider a completely unmanageable spaghetti-meatball dish that I just have to trust to "do the right thing and don't do evil". 

Compared to that this set appears very modest, I just wonder whether it is not feasible to reduce the set even further (but keep the diversity needed to test the functionality properly).

phoenix741:
Hi,

I reduce the number of dependencies by only putting two dependencies on the package.json. The goal is mainly to test the v2 and v3 format of the package-lock.json. (Other cases being tested by other tests)

In nodejs (used in front, but also on the server), some dependencies import smaller modules (sometimes 3 lines), which can quickly have an effect on the number of libraries in a project.

I see indeed simple node.js project (angular, view, nest.js) easily composed of a hundred modules

jeremylong (approved):

PR #5095, opened 2022-11-27
fix: document scanning dependency-check with dependency-check

jeremylong:
- upgrade and exclude dependencies.
- move optional, test dependencies to a profile called `test-dependencies`.
- document how to run dependency-check on dependency-check by excluding the `test-dependencies` profile.

### DIFF
--- a/maven/pom.xml
+++ b/maven/pom.xml
@@ -166,16 +166,6 @@ Copyright (c) 2013 Jeremy Long. All Rights Reserved.
             <artifactId>maven-artifact</artifactId>
             <scope>provided</scope>
         </dependency>
-        <dependency>
-            <groupId>org.apache.maven</groupId>
-            <artifactId>maven-model-builder</artifactId>
-            <scope>provided</scope>
-        </dependency>
-        <dependency>
-            <groupId>org.apache.maven</groupId>
-            <artifactId>maven-compat</artifactId>
-            <scope>provided</scope>
-        </dependency>
     </dependencies>
     <profiles>
         <profile>
--- a/maven/pom.xml
+++ b/maven/pom.xml
@@ -185,6 +185,9 @@ Copyright (c) 2013 Jeremy Long. All Rights Reserved.
                     <name>releaseTesting</name>
                 </property>
             </activation>
+            <properties>
+                <maven.api.version>3.1.0</maven.api.version>
+            </properties>
             <build>
                 <plugins>
                     <plugin>
--- a/maven/pom.xml
+++ b/maven/pom.xml
@@ -65,7 +65,6 @@ Copyright (c) 2013 Jeremy Long. All Rights Reserved.
                 <artifactId>maven-plugin-plugin</artifactId>
                 <version>${version.maven-plugin-plugin}</version>
                 <configuration>
-                    <skipErrorNoDescriptorsFound>true</skipErrorNoDescriptorsFound>
                     <goalPrefix>dependency-check</goalPrefix>
                 </configuration>
                 <executions>
--- a/maven/pom.xml
+++ b/maven/pom.xml
@@ -38,7 +38,7 @@ Copyright (c) 2013 Jeremy Long. All Rights Reserved.
         <tag>v6.4.1</tag>
     </scm>
     <prerequisites>
-        <maven>${maven.api.version}</maven>
+        <maven>3.1.0</maven>
     </prerequisites>
     <build>
         <resources>
--- a/README.md
+++ b/README.md
@@ -241,11 +241,24 @@ docker run --rm ^
 ```
 
 Building From Source
--------------
+--------------------
+
 To build dependency-check (using Java 8) run the command:
 
 ```
 mvn -s settings.xml install
+```
+
+Running dependency-check on dependency-check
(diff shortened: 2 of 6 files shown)


maven/pom.xml
```diff
@@ -38,7 +38,7 @@ Copyright (c) 2013 Jeremy Long. All Rights Reserved.
         <tag>v6.4.1</tag>
     </scm>
     <prerequisites>
-        <maven>${maven.api.version}</maven>
+        <maven>3.1.0</maven>
```
aikebah, line 41: How would you ensure that we don't use items of the new API if you decouple maven prerequisite of the plugin from the version of the api we build on (which you updated to 3.8.1, which means we risk using new API features introduced after 3.1.0)

jeremylong: gah... good point. I upgraded so we no longer had findings against ODC. Maybe we can downgrade to 3.1.0 in the `releaseTesting` profile? thoughts?

aikebah: I'm tempted to say that 'people using DependencyCheck to scan their maven projects for vulnerable dependencies' should already be using maven 3.8.1 or later, but not sure whether we would want to force that upon them.

They could also be using the 'use a secure proxy for all' strategy to mitigate the vulnerability (IIRC there's only the 'plain http repository-url defined in a (transitive) dependency pom' vulnerability in-between 3.1.0 and 3.8.1?) and have some projects that have other plugins that prevent them from running on up-to-date maven versions. (which could be considered a vulnerability issue of the project on its build toolchain given OWASP TOP10 2021 A06 considers outdated components in addition to vulnerable components)

Downgrading to 3.1.0 in the releaseTesting profile would only be feasible if we set the default (3.8.1) API version in a profile that gets deactivated when releaseTesting profile is active I think.

Doing so would be an acceptable work-around to me.

jeremylong: See [1d2988c](https://github.com/jeremylong/DependencyCheck/pull/5095/commits/1d2988c00a1f6f372875b44424a4904f46f550c1) - the `maven.api.version` has been moved to the profiles so that 3.1.0 is used during regression testing.

aikebah (approved):
LGTM

Checking the output of `mvn dependency:tree -Dincludes=org.apache.maven -DreleaseTesting` I did spot a dependency that we currently do not govern by the maven API version that I think we should (it's even more modern at 3.8.6): `org.apache.maven:maven-compat:jar:3.8.6:provided`

aikebah:
> LGTM
> 
> Checking the output of `mvn dependency:tree -Dincludes=org.apache.maven -DreleaseTesting` I did spot a dependency that we currently do not govern by the maven API version that I think we should (it's even more modern at 3.8.6): `org.apache.maven:maven-compat:jar:3.8.6:provided`

Given its purpose it would be even better if we could do without it (its a maven 2.x compatibility layer to ease the transition from 2 to 3, nowadays we're close to reaching the 3 to 4 transition)

aikebah:
> > LGTM
> > Checking the output of `mvn dependency:tree -Dincludes=org.apache.maven -DreleaseTesting` I did spot a dependency that we currently do not govern by the maven API version that I think we should (it's even more modern at 3.8.6): `org.apache.maven:maven-compat:jar:3.8.6:provided`
> 
> Given its purpose it would be even better if we could do without it (its a maven 2.x compatibility layer to ease the transition from 2 to 3, nowadays we're close to reaching the 3 to 4 transition)

According to the maven dependency plugin it appears we can....

```
[INFO] --- maven-dependency-plugin:3.3.0:analyze (default-cli) @ dependency-check-maven ---
[WARNING] Used undeclared dependencies found:
[WARNING]    org.slf4j:slf4j-api:jar:1.7.36:compile
[WARNING]    org.apache.maven.doxia:doxia-sink-api:jar:1.11.1:compile
[WARNING]    org.hamcrest:hamcrest:jar:2.2:test
[WARNING]    org.apache.maven.resolver:maven-resolver-api:jar:1.6.2:provided
[WARNING]    org.apache.commons:commons-lang3:jar:3.12.0:compile
[WARNING]    com.github.package-url:packageurl-java:jar:1.4.1:compile
[WARNING]    org.apache.maven.shared:maven-common-artifact-filters:jar:3.1.0:compile
[WARNING] Unused declared dependencies found:
[WARNING]    org.apache.maven:maven-model-builder:jar:3.8.1:provided
[WARNING]    org.apache.maven:maven-compat:jar:3.8.6:provided
[WARNING]    org.hamcrest:hamcrest-core:jar:2.2:test
[WARNING]    org.jetbrains:annotations:jar:23.0.0:compile
```

jeremylong:
Can we just remove `maven-compat`?

BTW - I'm planning on releasing 7.4.1 tomorrow (Friday) - and i'll hopefully get the gradle release correct this time.

aikebah:
@jeremylong I'll run a check today on a build without maven-compat to be fully sure, but fairly certain that it's really unused.

aikebah:
Awaiting the outcome of a final local test and then I'll file a PR onto this branch for the removal of the obsolete dependencies

jeremylong:
@aikebah Thanks!

PR #5112, opened 2022-12-02
Optimize file checksums calculation

koscejev:
## Fixes Issue #
I noticed checksums can be very slow to calculate for some larger files. One of the reasons is that the file is read 3 times - once per each checksum. (See `Dependency.calculateChecksums(File)`.) This is easy to optimize by reading the file only once and calculating all 3 checksums together.

## Description of Change
There are two commits:
1) calculates (and caches) all supported checksums together for files
2) (optional) removes dependency on commons-codec since I realized it's barely used and the methods that actually are used are already implemented in this `Checksum` class

No new functionality added, and existing tests already cover the changed functionality.

### DIFF
--- a/utils/pom.xml
+++ b/utils/pom.xml
@@ -55,10 +55,6 @@ Copyright (c) 2014 - Jeremy Long. All Rights Reserved.
             <version>${jackson.version}</version>
         </dependency>
         <dependency>
-            <groupId>commons-codec</groupId>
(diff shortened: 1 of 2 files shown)


jeremylong (approved):

jeremylong:
CodeQL flagged the use of MD5 - this is not used as a security mechanism.

PR #5120, opened 2022-12-06
fix: ensure suppression file is not removed during publish

jeremylong:
The generated suppression file in `gh-pages` was `cleaned` by the `JamesIves/github-pages-deploy-action@v4.4.1`. This PR resolves the issue.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -276,3 +276,4 @@ jobs:
         with:
           branch: gh-pages
           folder: target/staging
+          clean: false


aikebah (approved):
Wondering whether it would be possible to retrieve the current published suppressions and integrate them into the site archive files instead of preventing the cleanup of the gh pages site (which would ensure removal of pages should we remove them). However I don't expect many page-removals, so at least for now it's safer to skip the cleaning to ensure we keep the hosted suppressions in place.

PR #5129, opened 2022-12-08
Remove unused dependencies on maven libraries

aikebah:
## Description of Change

Processed the 'unused declared dependencies' mentioned for the maven plugin in response to my remarks of an obsolete maven-compat dependency.

org.hamcrest:hamcrest-core was listed as unused as it contains only deprecated classes and a note on the artifact itself being deprecated, with a transitive dependency on org.hamcrest:hamcrest, which is the up-to-date coordinate of hamcrest, so I've updated that coordinate to its future-proof coordinate and excluded the old artifactId from the transitive dependencies of JUnit.

No, existing testcases have been used to validate the changes.

### DIFF
--- a/maven/pom.xml
+++ b/maven/pom.xml
@@ -166,16 +166,6 @@ Copyright (c) 2013 Jeremy Long. All Rights Reserved.
             <artifactId>maven-artifact</artifactId>
             <scope>provided</scope>
         </dependency>
-        <dependency>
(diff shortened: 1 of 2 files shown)


aikebah:
@jeremylong Codacy error is just because I decided to file a PR towards your branch rather than directly committing to it (target branch not configured for Codacy scan)

jeremylong (approved):

PR #5152, opened 2022-12-15
fix: maven compatability

### LINKED ISSUE #5151: Incompatibility with maven 3.3.9
I have just upgraded to dependency-check 7.4.1 to eliminate some false positive.  Running the plugin with maven 3.3.9 I get the same error described in https://github.com/jeremylong/DependencyCheck/issues/1054.  If I use maven 3.5.3 it works fine.
Documentation at https://jeremylong.github.io/DependencyCheck/dependency-check-maven/index.html still says "The plug-in requires Maven 3.1 or higher".  Is this a regression that will be fixed?  If not, please, update the doc with the new minimum maven version.

jeremylong:
Resolves #5151

Reverts upgrade to maven dependencies.

### DIFF
--- a/README.md
+++ b/README.md
@@ -252,13 +252,13 @@ mvn -s settings.xml install
 Running dependency-check on dependency-check
 --------------------------------------------
 
-Dependency-check references several vulnerables dependencies that are never used
+Dependency-check references several vulnerable dependencies that are never used
 except as test resources. All of these optional test dependencies are included in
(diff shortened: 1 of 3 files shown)


aikebah (approved):

aikebah:
I wonder whether it would also be sufficient to mark the other maven library dependencies `<scope>provided</scope>`, but I think taking this conservative approach is a better safeguard anyhow to avoid incompatibilities.

PR #5156, opened 2022-12-15
Fp fixes

### LINKED ISSUE #5153: [FP]:  CVE-2011-0509 creating false report for vaadin artifacts.
### Package URl

pkg:maven/com.vaadin/collaboration-engine@5.3.0

### CPE

cpe:2.3:a:vaadin:vaadin:5.3.0:*:*:*:*:*:*

### CVE

CVE-2011-0509

### ODC Integration

{"label"=>"Maven Plugin"}

### ODC Version

7.4.0

### Description

the CVE-2011-0509, has been created for Vaadin 6, the listed artifacts are for Vaadin 14+, which doesn't have any connection with Vaadin 6.

the same FP has also been reported for other vaadin artifacts

- pkg:maven/com.vaadin/license-checker@1.11.2
- pkg:maven/com.vaadin.servletdetector/throw-if-servlet5@1.0.0
- pkg:maven/com.vaadin.servletdetector/throw-if-servlet3@1.0.0

we have other reporter using ODC in their gradle project, so probably add the gradle plugin too.

### LINKED ISSUE #5155: [FP]: CVE-2019-25028 creates false report for vaadin artifacts
### Package URl

pkg:maven/com.vaadin/open@8.4.0.3

### CPE

cpe:2.3:a:vaadin:vaadin:8.4.0.3:*:*:*:*:*:*:*

### CVE

CVE-2019-25028

### ODC Integration

{"label"=>"Maven Plugin"}

### ODC Version

7.4.0

### Description

similar as #5153 ， the CVE-2019-25028 was created for Vaadin 7 and 8.  the list artifacts are used for Vaadin 14, which doesn't have connections with the ill version

the same FP has been reported for other vaadin artifacts
- pkg:maven/com.vaadin/vaadin-testbench-core@8.2.0

ZheSun88:
## Fixes Issue #
Fixes #5153 
Fixes #5155

## Description of Change
Add FP suppressions for raised false positives that could not be handled by the automation and add all automatically generated FP suppressions up to now

no

### DIFF
--- a/core/src/main/resources/dependencycheck-base-suppression.xml
+++ b/core/src/main/resources/dependencycheck-base-suppression.xml
@@ -2900,6 +2900,20 @@
     </suppress>
     <suppress base="true">
         <notes><![CDATA[
+        FP per #5155 #5153
+        ]]></notes>
+        <packageUrl regex="true">^pkg:maven/com\.vaadin/.*(collaboration|license|open|testbench).*$</packageUrl>
+        <cpe>cpe:/a:vaadin:vaadin</cpe>
+    </suppress>
+    <suppress base="true">
+        <notes><![CDATA[
+        FP per #5153
+        ]]></notes>
+        <packageUrl regex="true">^pkg:maven/com\.vaadin\.servletdetector/throw-if.*$</packageUrl>
+        <cpe>cpe:/a:vaadin:vaadin</cpe>
+    </suppress>
+    <suppress base="true">
+        <notes><![CDATA[
         FP per #2191
         ]]></notes>
         <packageUrl regex="true">^pkg:maven/com\.vaadin/.*(sass|flow|theme|mixin|style|widget).*$</packageUrl>
--- a/core/src/main/resources/dependencycheck-base-suppression.xml
+++ b/core/src/main/resources/dependencycheck-base-suppression.xml
@@ -37,13 +37,6 @@
     </suppress>
     <suppress base="true">
         <notes><![CDATA[
-            FP per issue #4180, #4188, #4189, #4190
-            ]]></notes>
-        <packageUrl regex="true">^pkg:maven/com\.vaadin/vaadin-spring.*@.*$</packageUrl>
-        <cpe>cpe:/a:vaadin:vaadin</cpe>
-    </suppress>
-    <suppress base="true">
-        <notes><![CDATA[
             FP per issue #4199
             ]]></notes>
         <packageUrl regex="true">^pkg:maven/org\.sonarsource\.scanner\.gradle/sonarqube-gradle-plugin@.*$</packageUrl>
@@ -2886,13 +2879,6 @@
     </suppress>
     <suppress base="true">
         <notes><![CDATA[
-        FP per #2651
-        ]]></notes>
-        <packageUrl regex="true">^pkg:maven/com\.vaadin/vaadin[^@]+testbench@.*$</packageUrl>
-        <cpe>cpe:/a:vaadin:vaadin</cpe>
-    </suppress>
-    <suppress base="true">
-        <notes><![CDATA[
         FP per #2191
         ]]></notes>
         <packageUrl regex="true">^pkg:maven/com\.vaadin\.(addon|external).*$</packageUrl>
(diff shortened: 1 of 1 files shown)


ZheSun88:
I hope the regex i used is correct. if not, thanks for updating this.

core/src/main/resources/dependencycheck-base-suppression.xml
```diff
@@ -2898,6 +2898,20 @@
...
+        <packageUrl regex="true">^pkg:maven/com\.vaadin/.*(collaboration|license|open|testbench).*$</packageUrl>
+        <cpe>cpe:/a:vaadin:vaadin</cpe>
+    </suppress>
+    <suppress base="true">
+        <notes><![CDATA[
+        FP per #5153
+        ]]></notes>
+        <packageUrl regex="true">^pkg:maven/com\.vaadin\.servletdetector/throw-if.*$</packageUrl>
+        <cpe>cpe:/a:vaadin:vaadin</cpe>
+    </suppress>
```
aikebah:
```suggestion
    <suppress base="true">
        <notes><![CDATA[
        FP per #5155 #5153
        ]]></notes>
        <packageUrl regex="true">^(?!pkg:maven/com\.vaadin/vaadin(-core)?@).*$</packageUrl>
        <cpe>cpe:/a:vaadin:vaadin</cpe>
    </suppress>
```

Thanks for the PR... 
Have been looking into Vaadin a bit to ensure we're not improperly suppressing with your proposed regexes, and I think the suppression could be further optimised.
CPE `cpe:/a:vaadin:vaadin` appears to be in use for vaadin / vaadin-core components only (e.g. vaadin-flow receives `cpe:/a:vaadin:flow` (https://nvd.nist.gov/vuln/detail/CVE-2021-31412)

aikebah: Would like to hear your thoughts on that proposal before modifying it

ZheSun88: Hi @aikebah , Thanks for taking care of this PR. I have just tested with your proposal and It worked as expected. 

Nice Regex (I am not an expert on that :-p )

manolo: so @aikebah why do you exclude `core`?  shouldn't be enough the regex `^(?!pkg:maven/com\.vaadin/vaadin@).*$`   
there is no `vaadin-core` version under `10.0` so they never will be marked as vulnerables, but probably I'm missing something.

ZheSun88: by the way, we saw there are some rules using the `<cve>` tag, what is this tag for and do we need that too? 

if so, the following cve should be listed. 

```
<cve>CVE-2011-0509</cve>
<cve>CVE-2021-31409</cve> 
<cve>CVE-2019-25028</cve>
<cve>CVE-2021-33609</cve> 
<cve>CVE-2021-31403</cve>
```

manolo: @ZheSun88 at the same you could remove other occurrences of vaadin in other parts of the file, I think those were added because of the same problem, and the generic regex suggested by @aikebah should fix those as well

aikebah: @ZheSun88 cve suppresses a single vulnerability, cpe suppresses all vulnerabilitie of a product

Also see https://jeremylong.github.io/DependencyCheck/general/suppression.html

In general the suppression rules in the base-suppressions will be about an invalid CPE detection and suppress that CPE (and thereby all vulnerabilities of it). 

In some cases we do want to honor suppression of a specific vulnerability and suppress a specific vulnerability by CVE (as they are known to never apply), but most of the time the FPs that can be suppressed without knowing the usage in sources are only improper product detection.

@manolo I include everything that does not start with either `pkg:maven/com\.vaadin/vaadin@` or `pkg:maven/com\.vaadin/vaadin-core@`. That way `cpe:/a:vaadin:vaadin` as a detected CPE will only be retained for those two libraries. 
I excluded vaadin-core in there as the suppression is suppressing product-linkage for all versions. The absence of *known vulnerabilities* in current vaadin versions does not mean that in future no vulnerability will be uncovered (remember how there was a CVE in OpenSSL a while ago that went unnoticed by everyone for years?) or unintendedly introduced, so vaadin-core is to be linked to the CPE, so that if/when new vulnerabilities emerge they will be associated with both vaadin and vaadin-core.

ZheSun88: Thanks for the explanation.  👍

aikebah (approved):

aikebah:
@jeremylong As I widened the regex I think it would be good that you review the updated suppression as well

ZheSun88:
@aikebah , there are also other rules related to Vaadin artifacts in the suppression, maybe we should remove those after the widened regex?

aikebah:
@ZheSun88 Would be fine yo remove those, feel free to adapt your PR further to remove them right away. If not done directly we can always remove them later on in a follow-up housekeeping on these rules.

ZheSun88:
sure, I will clean up the vaadin rules with a new commit. 👍 
thanks for all the helps.

aikebah (approved):

jeremylong (approved):

jeremylong:
LGTM...

PR #5165, opened 2022-12-19
fix(FP): Suppress pluto CPE for the portlet-API library

jan-tosovsky-cz:
## Description of Change

*Apache Pluto is defined as the parent in the [pom.xml](https://repo1.maven.org/maven2/javax/portlet/portlet-api/3.0.0/portlet-api-3.0.0.pom) so any vulnerabilities detected in it has no effect on the portlet-api library.*

*no*


aikebah (approved):

aikebah:
Thanks for the PR

Issue #5036, opened 2022-11-08
Configuration option to set baseUrl for OSS Index


chovyy:
**Is your feature request related to a problem? Please describe.**
Since it is currenty not possible to use the OssIndexAnalyzer without internet connection (see also #4881), it would be nice to be able to use a proxy of the [OSS Index Service](https://ossindex.sonatype.org/rest).

**Describe the solution you'd like**
Introduce a configuration option to set the baseUrl of the OSS Index Service, which is `https://ossindex.sonatype.org/` by default.

**Describe alternatives you've considered**
Currenty, the only option is to disable the OssIndexAnalyzer when you don't have internet connection.

**Additional context**
[The Oss Index Maven Plugin do have this option](https://sonatype.github.io/ossindex-maven/maven-plugin/ossindex-audit/). Since DepencyCheck seems to make use of the OssClient library by Sonatype, the configuration of the baseUrl shouldn't be an issue.
The default can be seen here: [OssindexClientConfiguration.java](https://github.com/sonatype/ossindex-public/blob/main/client/src/main/java/org/sonatype/ossindex/service/client/OssindexClientConfiguration.java)

chovyy:
Works great. Thanks a lot for the quick response! 👍

PR #5176, opened 2022-12-22
Make OSS Index URL configurable via CLI

matzi:
## Fixes Issue #
Makes the OSS Index URL configurable in the CLI via --ossIndexUrl (#5036)

## Description of Change
URL can be defined via --ossIndexUrl otherwise the default is used.

No


aikebah:
Thanks for the PR. CLI options also need to be maunually documented within https://github.com/jeremylong/DependencyCheck/blob/main/cli/src/site/markdown/arguments.md. Would you be willing to include that update in your PR?

jeremylong:
documentation added via #5180. Messed up trying to update this PR - so I just created a new one.

PR #5179, opened 2022-12-24
fix: synchronization issue

### LINKED ISSUE #4374: Node.js Package Analyzer may run into ConcurrentModificationException
As can be seen in the conversation of #4089 there is a potential for the Node.js Package Analyzer to run into a ConcurrentModificationException under for now unknown circumstances.

Raising a new ticket as #4089 was about the NullPointer that emerged from it while creating the nestedCauseList for an ExceptionCollection when the ConcurrentModificationException (which has a null-valued message) is the first exception in the ExceptionCollection.

As the ConcurrentModificationException occurs in unknown circumstances we had to split fixing the NPE and further research into the ConcurrentModificationException.

jeremylong:
Resolves #4374

### DIFF
--- a/core/src/main/java/org/owasp/dependencycheck/analyzer/DependencyMergingAnalyzer.java
+++ b/core/src/main/java/org/owasp/dependencycheck/analyzer/DependencyMergingAnalyzer.java
@@ -152,9 +152,11 @@ public class DependencyMergingAnalyzer extends AbstractDependencyComparingAnalyz
             relatedDependency.getEvidence(EvidenceType.PRODUCT).forEach((e) -> dependency.addEvidence(EvidenceType.PRODUCT, e));
             relatedDependency.getEvidence(EvidenceType.VERSION).forEach((e) -> dependency.addEvidence(EvidenceType.VERSION, e));
 
-            relatedDependency.getRelatedDependencies()
-                    .forEach(dependency::addRelatedDependency);
-            relatedDependency.clearRelatedDependencies();
+            synchronized (relatedDependency) {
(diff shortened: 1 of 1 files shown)


core/src/main/java/org/owasp/dependencycheck/analyzer/DependencyMergingAnalyzer.java
```diff
@@ -152,9 +152,11 @@ public static void mergeDependencies(final Dependency dependency, final Dependen
             relatedDependency.getEvidence(EvidenceType.PRODUCT).forEach((e) -> dependency.addEvidence(EvidenceType.PRODUCT, e));
             relatedDependency.getEvidence(EvidenceType.VERSION).forEach((e) -> dependency.addEvidence(EvidenceType.VERSION, e));
 
-            relatedDependency.getRelatedDependencies()
-                    .forEach(dependency::addRelatedDependency);
-            relatedDependency.clearRelatedDependencies();
+            synchronized (relatedDependency) {
+                relatedDependency.getRelatedDependencies()
```
aikebah: The nested synchronisation (Lines 148 and 155) feels dangerous to me as both are on a dependency (risk of deadlock). I think a global lock 'I want to mess around with related dependencies' obtained for the outer level at line 148 would be a safer bet.

jeremylong: Thanks! I just updated the PR

aikebah:
LGTM

aikebah (approved):

PR #5190, opened 2022-12-29
fix: Upgrade semver4j to 4.1.1 to prevent NullPointerException

### LINKED ISSUE #5158: NullPointerException in Semver for alternative NPM versions
**Describe the bug**
Follow-up to #5128, seems not to be fully fixed yet: I receive a `NullPointerException` from the NodePackageAnalyzer when analyzing a package-lock.json. Exception message does slightly differ from the trace in the linked earlier issue though:

```
[WARN] An unexpected error occurred during analysis of '/src/package-lock.json' (Node.js Package Analyzer): Cannot invoke "String.isEmpty()" because "string" is null
[ERROR] 
java.lang.NullPointerException: Cannot invoke "String.isEmpty()" because "string" is null
        at org.semver4j.Range$RangeOperator.value(Range.java:138)
        at org.semver4j.RangesListFactory.addRanges(RangesListFactory.java:53)
        at org.semver4j.RangesListFactory.create(RangesListFactory.java:26)
        at org.semver4j.Semver.satisfies(Semver.java:445)
        at org.owasp.dependencycheck.analyzer.DependencyBundlingAnalyzer.npmVersionsMatch(DependencyBundlingAnalyzer.java:628)
        at org.owasp.dependencycheck.analyzer.AbstractNpmAnalyzer.findDependency(AbstractNpmAnalyzer.java:296)
        at org.owasp.dependencycheck.analyzer.NodePackageAnalyzer.processDependencies(NodePackageAnalyzer.java:454)
        at org.owasp.dependencycheck.analyzer.NodePackageAnalyzer.processDependencies(NodePackageAnalyzer.java:402)
        at org.owasp.dependencycheck.analyzer.NodePackageAnalyzer.analyzeDependency(NodePackageAnalyzer.java:270)
```

latest main as of writing -> e84cb804aa1510bd50515ad600fec2034decabd6 (run via the CLI within a container, using the included Dockerfile)

**To Reproduce**

Adding a few quick debug logs in _DependencyBundlingAnalyzer_ shows that in my current/specific case `npmVersionsMatch` is invoked with arguments `current = "^14.14.20 || ^16.0.0"` and `next = "18.11.5"`, which seems to originate from the following snippet in the `package-lock.json`:

```json
    "@storybook/angular": {
      "version": "6.5.13",
      // ...
      "requires": {
        // ...
        "@types/node": "^14.14.20 || ^16.0.0",
```

Similar to https://github.com/jeremylong/DependencyCheck/issues/5128#issuecomment-1343080426, this seems to be sufficient to trigger the NPE:

```java
Semver semver = new Semver("18.11.5");
System.out.println(semver.satisfies("^14.14.20 || ^16.0.0"));
```

**Note**

Based on the comment in the code I'm not sure if the check in https://github.com/jeremylong/DependencyCheck/blob/e84cb804aa1510bd50515ad600fec2034decabd6/core/src/main/java/org/owasp/dependencycheck/analyzer/DependencyBundlingAnalyzer.java#L616-L617 should be also preventing this, if so it doesn't catch this specific parameter combination as the next version doesn't include a space.

nhumblot:
## Fixes Issue #

Fix #5158 

## Description of Change

This PR fix an issue in semver4j causing a NullPointerException by upgrading it to the latest version

*no*

### DIFF
--- a/core/src/test/java/org/owasp/dependencycheck/utils/SemverTest.java
+++ b/core/src/test/java/org/owasp/dependencycheck/utils/SemverTest.java
@@ -13,6 +13,7 @@
  */
 package org.owasp.dependencycheck.utils;
 
+import static org.junit.Assert.assertFalse;
 import static org.junit.Assert.assertTrue;
 
 import org.junit.Test;
@@ -31,5 +32,16 @@ public class SemverTest {
     public void testSemver() {
         Semver semver = new Semver("3.1.4");
         assertTrue(semver.satisfies("^3.0.0-0"));
+    }
+    /**
+     * Test of semver4j. See https://github.com/jeremylong/DependencyCheck/issues/5158
+     */
+    @Test
+    public void testSemverComplex() {
+        Semver semver = new Semver("18.11.5");
+        assertFalse(semver.satisfies("^14.14.20 || ^16.0.0"));
+        
+        semver = new Semver("14.15.0");
+        assertTrue(semver.satisfies("^14.14.20 || ^16.0.0"));
     }
 }
--- a/pom.xml
+++ b/pom.xml
@@ -1069,7 +1069,7 @@ Copyright (c) 2012 - Jeremy Long
             <dependency>
                 <groupId>org.semver4j</groupId>
                 <artifactId>semver4j</artifactId>
-                <version>4.1.0</version>
+                <version>4.1.1</version>
             </dependency>
             <dependency>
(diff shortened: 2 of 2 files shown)


jeremylong (approved):

jeremylong:
Should have added the test case on this earlier...  Thanks for the PR - I'll release this fix shortly.

PR #5198, opened 2022-12-31
fix: update documentation

### LINKED ISSUE #5196: Cli execution with params file ignores certain paramaters
**Describe the bug**
When using the properties file as a parameter, certain settings get ignored

CLI 7.4.2

**Log file**
[<script src="https://gist.github.com/pteichner/92c61d3782d44157af38f2a587c92cbe.js"></script>](https://gist.github.com/pteichner/92c61d3782d44157af38f2a587c92cbe)

**To Reproduce**
1. Use the following properties file
```
scan=${WORKSPACE}/package.json
scan=${WORKSPACE}/package-lock.json
exclude=${WORKSPACE}/node_modules
cve.url.base=${owaspRepoURI}/nvdcve-1.1-%d.json.gz
cve.url.modified=${owaspRepoURI}/nvdcve-1.1-modified.json.gz
retireJsUrl=${owaspRepoURI}/jsrepository.json
cve.user=${USERNAME}
cve.password=${PASSWORD}
disableOssIndex=true
ossindexAnalyzerEnabled=false
junitFailOnCVSS=10
format=JSON,XML,HTML
disableAssembly=true
disableMSBuild=true
disableRetireJS=true
disableYarnAudit=true
log=owasp_log.log
out=owasp_reports
```

**Expected behavior**
All those settings **disableMSBuild**, **disableRetireJS** should be taken into account

**Additional context**

jeremylong:
resolves #5196


cli/src/site/markdown/arguments.md
```diff
@@ -35,7 +35,7 @@ Advanced Options
...
 |       | \-\-hints                             | \<file\>        | The file path to the XML hints file \- used to resolve [false negatives](../general/hints.html)                                                                                                        | &nbsp; |
-| \-P   | \-\-propertyfile                      | \<file\>        | Specifies a file that contains properties to use instead of application defaults.                                                                                                                      | &nbsp; |
+| \-P   | \-\-propertyfile                      | \<file\>        | Specifies a file that contains properties to use instead of application defaults. The key values used in the properties file are not the same as the arguments listed on this page; use the keys here: https://github.com/jeremylong/DependencyCheck/blob/main/core/src/main/resources/dependencycheck-cache.properties | &nbsp; |
```
nhumblot: Hi :wave: 

Shouldn't the url be `https://github.com/jeremylong/DependencyCheck/blob/main/core/src/main/resources/dependencycheck.properties` instead? `dependency-cache.properties` is only about the cache properties and the two properties given as examples in the issue are not listed in it. For example, the `disableMSBuild` property used by the user should be instead `analyzer.msbuildproject.enabled` based on the `dependencycheck.properties` file. :thinking:

jeremylong:
```suggestion
| \-P   | \-\-propertyfile                      | \<file\>        | Specifies a file that contains properties to use instead of application defaults. The key values used in the properties file are not the same as the arguments listed on this page; use the keys here: https://github.com/jeremylong/DependencyCheck/blob/main/core/src/main/resources/dependencycheck.properties | &nbsp; |
```

jeremylong:
@nhumblot thanks for catching the wrong url...

nhumblot (approved):

PR #5235, opened 2023-01-06
fix: crash when analyzing node package with a local dependency

rturner-edjuster:
## Description of Change

When a package has a local dependency, the package-lock.json file contains an entry without a version number, with "link": "true". This lack of version attribute causes the NodePackageAnalyzer class to crash.

These entries should be skipped instead as they cannot be processed.

yes

### DIFF
--- a/core/src/main/java/org/owasp/dependencycheck/analyzer/NodePackageAnalyzer.java
+++ b/core/src/main/java/org/owasp/dependencycheck/analyzer/NodePackageAnalyzer.java
@@ -390,6 +390,7 @@ public class NodePackageAnalyzer extends AbstractNpmAnalyzer {
                     // Ignore/skip linked entries (as they don't have "version" and
                     // later logic will crash)
                     if (jo.getBoolean("link", false)) {
+                        LOGGER.warn("Skipping `" + name + "` because it is a link dependency");
                         continue;
                     }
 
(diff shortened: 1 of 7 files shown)


core/src/main/java/org/owasp/dependencycheck/analyzer/NodePackageAnalyzer.java
```diff
@@ -386,6 +386,13 @@ private void processDependencies(JsonObject json, File baseDir, File rootFile,
 
                 if (entry.getValue() instanceof JsonObject) {
                     jo = (JsonObject) entry.getValue();
+
+                    // Ignore/skip linked entries (as they don't have "version" and
+                    // later logic will crash)
+                    if (jo.getBoolean("link", false)) {
+                        continue;
```
jeremylong, line 394:
```suggestion
                        LOGGER.warn("Skipping `" + name + "` because it is a link dependency");
                        continue;
```

jeremylong (approved):

core/src/main/java/org/owasp/dependencycheck/analyzer/NodePackageAnalyzer.java
rturner-edjuster: thanks.

rturner-edjuster:
@jeremylong Let me know if there is anything else I can do on this? (like update to merge in latest master, etc).

jeremylong:
Thanks! The change is good as is - I'm just debating on the release schedule. I'll likely put in one more release before 8.0.0.

rturner-edjuster:
@jeremylong  I would consider this for a patch -- it's pretty debilitating. We've had to completely disable use of DependencyCheck until it's resolved (largely in combination with the need for the fix for the field length + data issue that you released a fix for recently).

MenschNestor:
Seconded regarding the patch release – we were caught between a rock and a hard place because we needed the 7.4.4 fix for using DependencyCheck at all and then had to disable the Node package analyzer or we would have had no dependency check at all.

Issue #4671, opened 2022-07-15
[FP]: CVE-2022-31569 jakarta.annotation-api@1.3.5 & javax.jms-api-2.0


jpcmonster:
### Package URl

pkg:maven/jakarta.annotation/jakarta.annotation-api@1.3.5, pkg:maven/javax.jms/javax.jms-api@2.0

### CPE

`cpe:2.3:a:oracle:java_se:1.3.5:*:*:*:*:*:*:*, cpe:2.3:a:oracle:projects:1.3.5:*:*:*:*:*:*:*, cpe:2.3:a:projects_project:projects:1.3.5:*:*:*:*:*:*:*`

### CVE

CVE-2022-31569

### ODC Integration

{"label"=>"Gradle Plugin"}

### ODC Version

7.1.1

### Description

overly inclusive CPE?
https://nvd.nist.gov/vuln/detail/CVE-2022-31569
https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2022-31569

mjeffrey:
We hit the same thing (including CVE-2022-31548 ) 
```
[ERROR] jakarta.annotation-api-1.3.5.jar: CVE-2022-31569(9.3)
[ERROR] jakarta.transaction-api-1.3.3.jar: CVE-2022-31569(9.3)
[ERROR] querydsl-core-5.0.0.jar: CVE-2022-31548(9.3)
[ERROR] spring-cloud-commons-3.1.3.jar: CVE-2022-31569(9.3)
[ERROR] spring-security-rsa-1.0.10.RELEASE.jar: CVE-2022-31569(9.3)
```
These CVE descriptions both seem to reference github repositories
> The RipudamanKaushikDal/projects repository through 2022-04-03 on GitHub allows absolute path traversal because the Flask send_file function is used unsafely. 

and
> The nrlakin/homepage repository through 2017-03-06 on GitHub allows absolute path traversal because the Flask send_file function is used unsafely.

Has anyone else noticed the quality of the CVEs has gone down recently? There was a CVE for a documentation bug recently (in some example code!). 

This seems a bit extreme and if, like us, it breaks the build it generates a lot of effort.

sammyhk:
We got the same issue on
javax.jms-api-2.0.1.jar: CVE-2022-31569(9.3)
stax-ex-1.8.3.jar: CVE-2022-31569(9.3)

AB-xdev:
Same problem here.

[...]

PR #5237, opened 2023-01-07
fix(FP): Fix #5213 on generic utils project

### LINKED ISSUE #5213: [FP]: Dependencies with "utils" in name are matched (CVE-2021-4277)
### Package URl

pkg:maven/software.amazon.awssdk/utils@2.19.8

### CPE

cpe:2.3:a:utils_project:utils:*:*:*:*:*:*:*:*

### CVE

CVE-2021-4277

### ODC Integration

{"label"=>"Maven Plugin"}

### ODC Version

7.4.3

### Description

Although I put aws utils library on the package URI, that doesn't limit to that library. I have a private library with artifactId "java-utils". It is matched if I import it to another project. I reproduce that by creating a dummy library with artifactId "helloworld-utils" and it is also matched. Currently I suppress it on my side first.

Does DependencyCheck do partial match? I know it uses Lucene but I don't know how it works in details.

Thanks.

chadlwilson:
## Description of Change

Suppresses all Maven artifacts for this CPE, which is not actually a real CPE, see https://github.com/jeremylong/DependencyCheck/issues/5213#issuecomment-1372442259

Willing to take input on whether this is the right approach.

no

### DIFF
--- a/core/src/main/resources/dependencycheck-base-suppression.xml
+++ b/core/src/main/resources/dependencycheck-base-suppression.xml
@@ -6106,4 +6106,11 @@
         <packageUrl regex="true">^pkg:maven/org\.apache\.james/apache-jsieve-core@.*$</packageUrl>
         <cpe>cpe:/a:apache:james</cpe>
     </suppress>
+    <suppress base="true">
+        <notes><![CDATA[
+        FP per issue #5213, CVEs for scripting pet projects
+        ]]></notes>
+        <packageUrl regex="true">^.*$</packageUrl>
+        <cve>CVE-2021-4277</cve>
+    </suppress>
 </suppressions>


aikebah:
I think a CVE-suppression, similar to what we did for #4671 https://github.com/jeremylong/DependencyCheck/blob/8135326151248c0c0976db99daf91bf7a3ffd0fc/core/src/main/resources/dependencycheck-base-suppression.xml#L5440-L5445 would be the safer choice. Base suppressions are more-or-less permanent and when NVD decides to cleanup this apparently inappropriate`utils_project:utils` allocation to a private project and re-allocate it to some new emerging maven package for which the CPE-coordinates would be approrpriate we would suppress any CVE found for it.

chadlwilson:
> I think a CVE-suppression [...] would be the safer choice.

OK, fair enough, switched to that. :-)

aikebah (approved):
Thanks for the PR

albertwangnz:
Hi @aikebah  Where are we going to have the fix in? 7.4.5, 7.5.0 or 8.0.0?

aikebah:
@albertwangnz, will be in next release. not sure yet which version, depends on how far the other changes of 8.0.0 are (how much work is left to finish them).
In any case the suppression has also been added to our hosted suppressions, so if you use that you should also have them resolved.

jensrutschmann:
@aikebah Is this 'hosted suppressions' thing documented anywhere?

Are you simply referring to this URL? https://github.com/jeremylong/DependencyCheck/blob/main/core/src/main/resources/dependencycheck-base-suppression.xml

aikebah:
@jensrutschmann It's referenced in the comments of #4723 and located at https://jeremylong.github.io/DependencyCheck/suppressions/publishedSuppressions.xml

albertwangnz:
> 

Hi @aikebah  thank you for your reply. May I ask what are your hosted suppressions? Cheers.

albertwangnz:
> @jensrutschmann It's referenced in the comments of #4723 and located at https://jeremylong.github.io/DependencyCheck/suppressions/publishedSuppressions.xml

Hi @aikebah  thank you for your reply. Just to confirm my understanding. In 7.4.x, the hosted suppression file will not be used by ODC automatically, and I need to manually copy the file and apply those suppressions to our scans.

In 8.0.0, it is likely that ODC will automatically update and use the hosted suppression file.

We are using the ODC maven plugin in a `mvn xxxx` command. I am not sure if my understanding is correct.

Cheers.

Albert

aikebah:
@albertwangnz Correct, in 8.0.0 we intend to automatically use the hosted suppression file. 

For now we expect to be able finally integrate the changes intended for 8.0.0 later this week and cut a release from it.

Up to its availability you should be able to use something along the lines of:

`mvn org.owasp:dependency-check-maven:7.4.4:check -DsuppressionFiles="your-filter-file.xml,https://jeremylong.github.io/DependencyCheck/suppressions/publishedSuppressions.xml"`

To run with both your own filters and the hosted suppressions file

albertwangnz:
> @albertwangnz Correct, in 8.0.0 we intend to automatically use the hosted suppression file.
> 
> For now we expect to be able finally integrate the changes intended for 8.0.0 later this week and cut a release from it.
> 
> Up to its availability you should be able to use something along the lines of:
> 
> `mvn org.owasp:dependency-check-maven:7.4.4:check -DsuppressionFiles="your-filter-file.xml,https://jeremylong.github.io/DependencyCheck/suppressions/publishedSuppressions.xml"`
> 
> To run with both your own filters and the hosted suppressions file

This is great! Thanks a lot :-)

Albert

PR #5241, opened 2023-01-07
chore: Bump minor version

aikebah:
To signal the feature addition of severity in the SARIF report


aikebah:
Spotted that the plugin PR was merged, so the minor version bump is outdated

aikebah:
should read more closely... main was merged into the plugin branch rather than the other way around.

@jeremylong close to finish up on the plugin? or should we first go for 7.5.0?

Issue #5220, opened 2023-01-06
DatabaseException: Error updating 'CVE-2020-36569'


fdspbsteam:
**Describe the bug**
     classpath 'com.android.tools.build:gradle:4.1.1'
gradle version is: 7.0.2

after run ./gradlew dependencyCheckAnalyze
in logs falls down, this error:

```> Task :dependencyCheckUpdate
org.owasp.dependencycheck.data.nvdcve.DatabaseException: Error updating 'CVE-2020-36569'
org.owasp.dependencycheck.data.update.exception.UpdateException: org.owasp.dependencycheck.data.nvdcve.DatabaseException: Error updating 'CVE-2020-36569'
	at org.owasp.dependencycheck.data.update.nvd.ProcessTask.processFiles(ProcessTask.java:157)
	at org.owasp.dependencycheck.data.update.nvd.ProcessTask.call(ProcessTask.java:114)
	at org.owasp.dependencycheck.data.update.nvd.ProcessTask.call(ProcessTask.java:41)
	at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:264)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1128)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:628)
	at java.base/java.lang.Thread.run(Thread.java:829)
Caused by: org.owasp.dependencycheck.data.nvdcve.DatabaseException: Error updating 'CVE-2020-36569'
	at org.owasp.dependencycheck.data.nvdcve.CveDB.updateVulnerability(CveDB.java:823)
	at org.owasp.dependencycheck.data.update.nvd.NvdCveParser.parse(NvdCveParser.java:114)
Caused by: org.owasp.dependencycheck.data.nvdcve.DatabaseException: Error updating 'CVE-2020-36569'

	at org.owasp.dependencycheck.data.update.nvd.ProcessTask.importJSON(ProcessTask.java:141)
	at org.owasp.dependencycheck.data.update.nvd.ProcessTask.processFiles(ProcessTask.java:154)
	... 6 more
[...]

PR #5266, opened 2023-01-10
fix: name calculation

khoelldobler:
## Description of Change

The current name calculation just looks for the last occurence of the string "node_modules" and calculates a substring after the occurence plus 1 arbitrary additional character. However, if this string is used as part of the package name the calculation ist wrong and the dependency check crashes with:
 "Could not perform Node Audit analysis. Invalid payload submitted to Node Audit API". 

We ran into this problem as we updated to the latest gradle plugin version due to #5220 and having "[node_modules-path](https://www.npmjs.com/package/node_modules-path)" as a transitive dependency. 

I added the slash to the indexOf check to ensure the right pattern is matched.

*no*


jeremylong (approved):

jeremylong:
thanks for the PR!

PR #5268, opened 2023-01-10
fix: update rejected CVEs check

### LINKED ISSUE #5247: Is there a better mechanism for detecting rejected CVEs?
I've started getting a large number of false positives on one of my projects that uses Camel because of [CVE-2022-45046](https://nvd.nist.gov/vuln/detail/CVE-2022-45046). Apache have withdrawn/rejected the CVE, but I still get it in my scans seemingly because DependencyCheck is only reading the description in a very specific way to identify rejected CVEs: 

https://github.com/jeremylong/DependencyCheck/blob/175a00d499f243e8250ab5b86685b44c9c368f86/core/src/main/java/org/owasp/dependencycheck/data/nvdcve/CveItemOperator.java#L199-L208

https://github.com/jeremylong/DependencyCheck/blob/175a00d499f243e8250ab5b86685b44c9c368f86/core/src/main/java/org/owasp/dependencycheck/data/nvdcve/CveDB.java#L807-L809

To workaround the issue I've added some suppressions for the CVE, but I'm wondering if there's a better way? I'm not knowledgeable enough about this space to know whether the problem stems from Apache not following the "standard" description when rejecting the CVE or if DependencyCheck is too naïve in its approach.

jeremylong:
resolves #5247

### DIFF
--- a/core/src/main/java/org/owasp/dependencycheck/data/nvdcve/CveItemOperator.java
+++ b/core/src/main/java/org/owasp/dependencycheck/data/nvdcve/CveItemOperator.java
@@ -204,7 +204,7 @@ public class CveItemOperator {
      * <code>false</code>
      */
     public boolean isRejected(String description) {
-        return description.startsWith("** REJECT **");
+        return description.startsWith("** REJECT **") || description.startsWith("DO NOT USE THIS CANDIDATE NUMBER");
     }
 
     /**


nhumblot (approved):

aikebah:
LGTM

PR #5274, opened 2023-01-11
docs: update documentation for hostd suppressions

jeremylong:


aikebah:
@jeremylong wonder how I've managed to miss that in my update... thanks for catching the omission

aikebah (approved):

jeremylong:
@aikebah there are two locations in the gradle plugin documentation where the info is located - you had updated one of them.

PR #5299, opened 2023-01-14
docs: updates for 8.0.0

jeremylong:


SECURITY.md
```diff
@@ -4,8 +4,8 @@
 
 | Version  | Supported          |
 | ---------|--------------------|
-| 6.1.6+   | :white_check_mark: |
-| <= 6.1.5 | :x:                |
+| 7.4.4+   | :white_check_mark: |
```
aikebah: Do we want 7.4.4+ or 8.0.0+ after release? 7.4.4+ indicates we should be prepared for a security-fix on 7.4.x

jeremylong: Completely agree.

SECURITY.md
```diff
@@ -4,8 +4,8 @@
...
+| <= 7.4.3 | :x:                |
```
jeremylong:
```suggestion
| 8.0.0+   | :white_check_mark: |
| <= 7.4.4 | :x:                |
```

PR #5300, opened 2023-01-14
fix: disable OSS Index Analyzer when transport errors occur

jeremylong:
- disable OSS Index Analyzer to prevent hundreds of exceptions in report
- update documentation for the OSS Index Analyzer


aikebah (approved):
LGTM

EugRomanchenko:
@jeremylong can the OSS Index analyzer be disabled automatically in case of errors with code 503?

PR #5302, opened 2023-01-14
fix: add missing args to completion

jeremylong:


aikebah (approved):
Good one... hold off on the release... I'll have a few more to add beyond these as I overlooked this in the hosted suppressions file updates (and spotted a typo in the CLI docs of it)

PR #5314, opened 2023-01-16
fix: correct typos and sql syntax

### LINKED ISSUE #5311: merge_knownexploited is expected to be a procedure using PostgreSQL
**Describe the bug**
While running version 8.0.0 using PostgreSQL the maven plugin fails due to `merge_knownexploited` being a function instead of a procedure. This seems to be a bug for PostgreSQL only because `merge_knownexploited` is declared as a procedure for Oracle, MySQL and others.

The problem occurs using version 8.0.0 of the maven plugin.

**Log file**
```
[ERROR] Failed to execute goal org.owasp:dependency-check-maven:8.0.0:check (default-cli) on project product-parent: Fatal exception(s) analyzing product-parent: One or more exceptions occurred during analysis:
[ERROR] 	UpdateException: java.sql.BatchUpdateException: Batch entry 0 CALL merge_knownexploited('CVE-2021-27104','Accellion','FTA','Accellion FTA OS Command Injection Vulnerability','2021-11-03','Accellion FTA 9_12_370 and earlier is affected by OS command execution via a crafted POST request to various admin endpoints.','Apply updates per vendor instructions.','2021-11-17',NULL) was aborted: ERROR: merge_knownexploited(character varying, character varying, character varying, character varying, character varying, character varying, character varying, character varying, character varying) is not a procedure
```

**To Reproduce**
1. Create an empty PostgreSQL database.
2. Add missing to `initialize_postgres.sql` inside of `merge_knownexploited`, see additional context, and execute on newly created database.
3. Run dependency check maven goal.

**Expected behavior**
The plugin successfully executes updates and checks using PostgreSQL.

**Additional context**
`initialize_postgres.sql` also seems to be missing a semicolon inside `merge_knownexploited` which prevents the migration file from succeeding initially.

jeremylong:
resolves #5311

### DIFF
--- a/core/src/main/resources/data/initialize_mssql.sql
+++ b/core/src/main/resources/data/initialize_mssql.sql
@@ -26,8 +26,8 @@ if exists (SELECT 1 FROM sysobjects WHERE name='insert_software' AND xtype='P')
     drop procedure insert_software;    
 if exists (SELECT 1 FROM sysobjects WHERE name='knownExploited' AND xtype='U')
     drop table knownExploited;
-if exists (SELECT 1 FROM sysobjects WHERE name='merge_knownexpoited' AND xtype='P')
-    drop procedure merge_knownexpoited;    
+if exists (SELECT 1 FROM sysobjects WHERE name='merge_knownexploited' AND xtype='P')
+    drop procedure merge_knownexploited;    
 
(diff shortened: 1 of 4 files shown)


aikebah:
Hmmmm.. thought we had these tackled somewhere in the review... did I overlook them, or where they somehow reintroduced in rebasing/merging?

jeremylong:
I think a couple were missed... but I could be wrong. no worries either way.

PR #5330, opened 2023-01-17
fix: use proxy if configured

### LINKED ISSUE #5313: Version 8.0.0. KnownExploitedDataSource proxy HTTP 403
KnownExploitedDataSource Line: 78 

https://github.com/jeremylong/DependencyCheck/blob/f3d806de22002d932fe12a706f4b9fa91ed3322d/core/src/main/java/org/owasp/dependencycheck/data/update/KnownExploitedDataSource.java

```
                //TODO - add all the proxy config, likely use the same as configured for NVD
                final HttpResourceConnection conn = new HttpResourceConnection(settings);
```

In class `HostedSuppressionsDataSource` i do not see such a commonet:

https://github.com/jeremylong/DependencyCheck/blob/f3d806de22002d932fe12a706f4b9fa91ed3322d/core/src/main/java/org/owasp/dependencycheck/data/update/HostedSuppressionsDataSource.java

Logs:

```
[ERROR] Failed to execute goal org.owasp:dependency-check-maven:8.0.0:check (default-cli) on project iaml: Fatal exception(s) analyzing aaaaaaaaaa: One or more exceptions occurred during analysis:
[ERROR] 	UpdateException: org.owasp.dependencycheck.utils.DownloadFailedException: Error downloading file https://www.cisa.gov/sites/default/files/feeds/known_exploited_vulnerabilities.json; unable to connect.
[ERROR] 		caused by DownloadFailedException: Error downloading file https://www.cisa.gov/sites/default/files/feeds/known_exploited_vulnerabilities.json; unable to connect.
[ERROR] 		caused by IOException: Unable to tunnel through proxy. Proxy returns "HTTP/1.1 403 Forbidden"
[ERROR] 	NoDataException: No documents exist
[ERROR] -> [Help 1]
[ERROR] 
```

jeremylong:
Resolves #5313

### DIFF
--- a/core/src/main/java/org/owasp/dependencycheck/data/update/KnownExploitedDataSource.java
+++ b/core/src/main/java/org/owasp/dependencycheck/data/update/KnownExploitedDataSource.java
@@ -75,8 +75,7 @@ public class KnownExploitedDataSource implements CachedWebDataSource {
             try {
                 final URL url = new URL(settings.getString(Settings.KEYS.KEV_URL, DEFAULT_URL));
                 LOGGER.info("Updating CISA Known Exploited Vulnerability list: " + url.toString());
-                //TODO - add all the proxy config, likely use the same as configured for NVD
-                final HttpResourceConnection conn = new HttpResourceConnection(settings);
+                final HttpResourceConnection conn = new HttpResourceConnection(settings, true);
                 try (InputStream in = conn.fetch(url)) {
                     final KnownExploitedVulnerabilityParser parser = new KnownExploitedVulnerabilityParser();
                     final KnownExploitedVulnerabilitiesSchema data = parser.parse(in);


core/src/main/java/org/owasp/dependencycheck/data/update/KnownExploitedDataSource.java
```diff
@@ -75,8 +75,7 @@ public boolean update(Engine engine) throws UpdateException {
             try {
                 final URL url = new URL(settings.getString(Settings.KEYS.KEV_URL, DEFAULT_URL));
                 LOGGER.info("Updating CISA Known Exploited Vulnerability list: " + url.toString());
-                //TODO - add all the proxy config, likely use the same as configured for NVD
-                final HttpResourceConnection conn = new HttpResourceConnection(settings);
+                final HttpResourceConnection conn = new HttpResourceConnection(settings, true);
```
aikebah, line 78: That won't help given that the settings-only constructor already defaults to proxy usage.

Have seen it work in my (authentication-less) proxy environment, so my gut feel is that reporter uses an authenticated proxy or there are proxy-rules in place that do not grant access to the CISA website

https://github.com/jeremylong/DependencyCheck/blob/e63738ae3d87fd9407acdf3b6a971379e43564c5/utils/src/main/java/org/owasp/dependencycheck/utils/HttpResourceConnection.java#L94-L96

PR #5345, opened 2023-01-18
fix: Add double quote for "properties.security-severity" to fix processing of SARIF file in GitHub

mrclrchtr:
## Fixes Issue #

I'm so sorry. I somehow forgot the quotation marks in my PR here https://github.com/jeremylong/DependencyCheck/pull/5227 

I don't know how that could have happened. This causes the following exception in github processing:

```
Error: Code Scanning could not process the submitted SARIF file:
parsing restricted subset of SARIF data has failed: parse error: expected string near offset 2910 of '7.2'
```

This PR adds the quotation marks.

### DIFF
--- a/core/src/main/resources/templates/sarifReport.vsl
+++ b/core/src/main/resources/templates/sarifReport.vsl
@@ -41,7 +41,7 @@ For more information see [How dependency-check works](https://jeremylong.github.
                 #end
                 #if($rule.cvssv3BaseSeverity)
                     "cvssv3_baseScore": $rule.cvssv3BaseScore,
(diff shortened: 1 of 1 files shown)


aikebah (approved):
Thanks for having the PR in before I could start working on it

aikebah:
To bad it got noticed by you after @jeremylong already had been working hard to get 8.0.1 out, it will have to go into the next patch release

mrclrchtr:
I also had the stackoverflow problems, that's why I could only test 8.0.1.

Sorry for the inconvenience and thanks for merging. 😊

PR #5350, opened 2023-01-19
fix: reduce log size

jeremylong:
Replace debug logging with trace logging for least necessary log statements that generate a ton of log messages.


aikebah (approved):
Makes sense to set these to trace... but makes me wondering... do we have a way to still trigger trace logging should we need the detailed info in future? (If not.... I'd say we would need a separate enhancement ticket to add such a flag, but this should still move forward to make the debug logs more manageable in size)

jeremylong:
I figured most of the time we would need these items *we* would be testing locally...

PR #5351, opened 2023-01-19
fix: add missing setting configuration for cli

jeremylong:
The configuration option exists - it was just never used when configuring the settings.


aikebah (approved):
Was about to request documentation... but even the documentation is there

PR #5364, opened 2023-01-21
fix: update CSV duplicate description column

### LINKED ISSUE #5355: --format CSV duplicate fieldname "description"
**Describe the bug**
there seem to be new fields in the csv export:  VendorProject | Product | Name | DateAdded | Description | RequiredAction | DueDate | Notes
My biggest problem is the description field. There is already a description field. Can you make the field **name** **unique**?
Also the new fields seem to be always empty, is this logical?

there are also 2 fields with a space in the fieldname (this is not a new issue). CPE Confidence and Evidence Count...   Other fields have an underscore, perhaps fix these also?

The problem occurs using version 8.0.1 of the commandline tool

**Log file**
there is no error, so log file not really useful

**To Reproduce**
--format CSV 

**Expected behavior**
each field should have an unique name 

**Additional context**

jeremylong:
- update known exploited vulnerability short description column name
- resolves #5355


aikebah (approved):
LGTM... wanted to go and work on it... but you already did

jeremylong:
I need to do a better job about assigning myself to issues i'm working on....

Issue #4969, opened 2022-10-20
Issues since version 6.3.0


StefanHirche:
Hi,
I was using version 6.2.2 and wanted to upgrade to 7.2 but I got this issue

```
Unable to resolve system scoped dependency: p2.eclipse.plugin:org.apache.xerces:eclipse-plugin:2.12.1.v20210115-0812:system
exception: org.owasp.dependencycheck.exception.DependencyNotFoundException: Unable to resolve system scoped dependency: p2.eclipse.plugin:org.apache.xerces:eclipse-plugin:2.12.1.v20210115-0812:system

org.owasp.dependencycheck.maven.BaseDependencyCheckMojo.collectMavenDependencies(BaseDependencyCheckMojo.java:1292)
org.owasp.dependencycheck.maven.BaseDependencyCheckMojo.collectDependencies(BaseDependencyCheckMojo.java:1437)
org.owasp.dependencycheck.maven.BaseDependencyCheckMojo.scanArtifacts(BaseDependencyCheckMojo.java:1098)
org.owasp.dependencycheck.maven.AggregateMojo.scanDependencies(AggregateMojo.java:73)
org.owasp.dependencycheck.maven.BaseDependencyCheckMojo.runCheck(BaseDependencyCheckMojo.java:1689)
org.owasp.dependencycheck.maven.BaseDependencyCheckMojo.execute(BaseDependencyCheckMojo.java:950)
org.apache.maven.plugin.DefaultBuildPluginManager.executeMojo(DefaultBuildPluginManager.java:137)
org.apache.maven.lifecycle.internal.MojoExecutor.execute(MojoExecutor.java:210)
org.apache.maven.lifecycle.internal.MojoExecutor.execute(MojoExecutor.java:156)
org.apache.maven.lifecycle.internal.MojoExecutor.execute(MojoExecutor.java:148)
org.apache.maven.lifecycle.internal.LifecycleModuleBuilder.buildProject(LifecycleModuleBuilder.java:117)
org.apache.maven.lifecycle.internal.LifecycleModuleBuilder.buildProject(LifecycleModuleBuilder.java:81)
org.apache.maven.lifecycle.internal.builder.singlethreaded.SingleThreadedBuilder.build(SingleThreadedBuilder.java:56)
org.apache.maven.lifecycle.internal.LifecycleStarter.execute(LifecycleStarter.java:128)
org.apache.maven.DefaultMaven.doExecute(DefaultMaven.java:305)
[...]

PR #5366, opened 2023-01-22
fix: Support maven extensions like Tycho adding system-scoped dependencies without a systemPath

aikebah:
## Description of Change

When regular system-scoped dependencies don't yield a resolved system-scoped artifact attempt resolving the dependencies to make maven extension hook into the artifact resolution and provide us with the appropriate dependencies.

No, adding an integrationtest with Tycho pomless 2.7.5. would require the build to run on JDK 11 or later, the updated Tycho 3 version would require JDK17 or later.
Manually verified that the scan runs to completion with the reproduction sample referred to in issue #4969 (both original Tycho 2.7.5 and the updated Tycho 3.0.0 versions)

### DIFF
--- a/maven/src/main/java/org/owasp/dependencycheck/maven/BaseDependencyCheckMojo.java
+++ b/maven/src/main/java/org/owasp/dependencycheck/maven/BaseDependencyCheckMojo.java
@@ -2680,6 +2680,24 @@ public abstract class BaseDependencyCheckMojo extends AbstractMojo implements Ma
                     }
                 }
             }
+            Throwable ignored = null;
+            if (!isResolved) {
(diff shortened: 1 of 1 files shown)


jeremylong (approved):
LGTM

jeremylong:
Any chance we can add project using Tycho to the maven/src/it tests? I'm fine if no - the changes look fine.

aikebah:
> Any chance we can add project using Tycho to the maven/src/it tests? I'm fine if no - the changes look fine.

Only if you know a way to trigger running it on Java 11 or 17 instead of Java 8

jeremylong:
good enough - thanks!

PR #5371, opened 2023-01-23
fix: ArchiveAnalyzer should throw AnalysisException

### LINKED ISSUE #5357: --zipExtensions does not always work
**Describe the bug**
I use --zipExtensions "mpk" for checking .js files in the zip archive mpk. This does not always work. 

**not finding the vulnerability except when the mpk is unzipped first:
widgets\RichText.mpk\RichText\ckeditor\ckeditor.js**

scanning and finding the vulnerability in zipped archive (as expected):
widgets\HTMLSnippet.mpk\HTMLSnippet\lib\jquery-3.3.1.js
widgets\HTMLSnippet.mpk\HTMLSnippet\widget\HTMLSnippet2.js

The problem occurs using version 8.0.1 of the cli tool

**Log file**
enclosed
[mylog.zip](https://github.com/jeremylong/DependencyCheck/files/10465763/mylog.zip)

**To Reproduce**
scan a mpk file with --zipExtensions "mpk"

**Expected behavior**
it should scan the zipped mpk always (not sometimes)

**Additional context**
I will provide the mpk file (in a zip)
[RichText.zip](https://github.com/jeremylong/DependencyCheck/files/10465743/RichText.zip)

I did a scan on the zipped files with --zipextensions "mpk" and the same scan on the extracted files. Results:

1. It found **184** vulnerabilities in the zipped mpk's with --zipextensions "mpk"
2. When scanning the extracted files, the check found **192** vulnerabilities. So, a difference of **8**

jeremylong:
Partially fixes #5357

Java's zip handling functions do not properly unzip mpk files. The errors may go un-noticed because the archive analyzer only printed a warning rather then throwing an analysis exception.

See https://issues.apache.org/jira/browse/COMPRESS-562

The updated code still fails for mpk files - but may work for apk files.

### DIFF
--- a/core/src/main/java/org/owasp/dependencycheck/analyzer/ArchiveAnalyzer.java
+++ b/core/src/main/java/org/owasp/dependencycheck/analyzer/ArchiveAnalyzer.java
@@ -31,6 +31,8 @@ import java.util.HashSet;
 import java.util.List;
 import java.util.Set;
 import java.util.concurrent.atomic.AtomicInteger;
+import java.util.zip.ZipEntry;
+import java.util.zip.ZipInputStream;
 import javax.annotation.concurrent.ThreadSafe;
 
 import org.apache.commons.compress.archivers.ArchiveEntry;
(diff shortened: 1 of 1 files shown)


aikebah (approved):
LGTM... maybe also remove the now unused import of the ZipArchiveInputStream

Makes sense IMO to make the analyzer break on archives that cannot be properly extracted.

PR #5394, opened 2023-01-28
chore: update data directory property for latest major version

alifahadk:
## Description of Change

*Update data directory properties for latest major release version 8*

*See also: Relevant PR for Gradle plugin https://github.com/dependency-check/dependency-check-gradle/pull/322*

*NA*


jeremylong:
This is not needed. The current database directory works.

alifahadk:
> This is not needed. The current database directory works.

Thanks for taking a look. I had overlooked the update in task.properties in ant as well.

Just out of curiosity. What drives the version here? Is it bumped only when there is a breaking change in the database schema, etc?

Thanks again!

jeremylong:
We only change the directory when there is a non-backward compatible change to the DB schema.

PR #5399, opened 2023-01-30
fix: correct maven documentation

### LINKED ISSUE #5392: Documentation says 'runtime artifacts are not bundled in distribution' ?
Hello,

The documentation of the Maven plugin (https://jeremylong.github.io/DependencyCheck/dependency-check-maven/index.html) says (Example 4):

> Create the dependency-check-report.html and skip artifacts not bundled in distribution (Provided and Runtime scope).

to demonstrate the `<skipRuntimeScope>true</skipRuntimeScope>` configuration option.

But is it true that a Maven dependency declared with scope "runtime" will not be bundled in distribution ? E.g. as far as I know, the spring-boot:repackage plugin goal bundles a JAR with runtime artifacts included.

In such case, from a security prospective, wouldn't it be better not to suggest excluding runtime scope artifacts from the Dependency Check scan ?

Thank you for your answer.

Regards,

jeremylong:
Resolves #5392

### DIFF
--- a/maven/src/site/markdown/index.md.vm
+++ b/maven/src/site/markdown/index.md.vm
@@ -126,7 +126,6 @@ Create the dependency-check-report.html and skip artifacts not bundled in distri
                 <version>${project.version}</version>
                 <configuration>
                     <skipProvidedScope>true</skipProvidedScope>
-                    <skipRuntimeScope>true</skipRuntimeScope>
                 </configuration>
                 <executions>
                     <execution>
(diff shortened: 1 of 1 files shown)


aikebah (changes requested):
Should also remove line 129, the skipRuntimeScope in the POM-snippet.

jeremylong:
You ever look back on documentation you wrote and can't believe you wrote it as it was clearly wrong?  ugh...

aikebah (approved):

aikebah:
> You ever look back on documentation you wrote and can't believe you wrote it as it was clearly wrong? ugh...

Yes, unfortunately that sounds a bit too familiar to me :)

Issue #4995, opened 2022-10-27
Python - add support for Poetry and PipEnv


agmond:
Hi,

As more and more modern Python applications start using [Poetry](https://python-poetry.org/) or [PipEnv](https://pipenv.pypa.io/en/latest/) for managing their dependencies, please support those packaging tools, in addition to the existing (experimental) support in `requirements.txt`.

**Poetry**
Poetry uses a [pyproject.toml file](https://python-poetry.org/docs/pyproject/) to declare on dependencies, and a lock file named [poetry.lock](https://python-poetry.org/docs/basic-usage/#installing-with-poetrylock).
The dependency specification can be found [here](https://python-poetry.org/docs/dependency-specification/).

Examples (taken from [Poetry repo](https://github.com/python-poetry/poetry)):
- [pyproject.toml](https://github.com/python-poetry/poetry/blob/master/pyproject.toml)
- [poetry.lock](https://github.com/python-poetry/poetry/blob/master/poetry.lock)

**PipEnv**
PipEnv uses [Pipfile & Pipfile.lock](https://pipenv.pypa.io/en/latest/basics/#example-pipfile-pipfile-lock).

Examples (taken from [PipEnv repo](https://github.com/pypa/pipenv)):
- [PipFile](https://github.com/pypa/pipenv/blob/main/Pipfile)
- [Pipfile.lock](https://github.com/pypa/pipenv/blob/main/Pipfile.lock)

jeremylong:
The poetry.lock is supported - however, the pyproject.toml is still not supported. This may cause a gap for users of pip>10 (see https://pip.pypa.io/en/stable/reference/build-system/pyproject-toml/).

jeremylong:
With #5404 I think we are about as close as we can get to fulfilling this enhancement request. We still do not support `pyproject.toml` - however, there is insufficient version information to make an analysis meaningful. Should we just add a warning if no lock file is present but we found a `pyproject.toml` file?

jeremylong:
It looks like `pip` may also support `pyproject.toml`: https://pip.pypa.io/en/stable/reference/build-system/pyproject-toml/

jeremylong:
again - sorry how long this one took me to get to. I think we have covered this entirely with the 8.1.0 release (or at least the best we can).

PR #5404, opened 2023-01-31
feat: support Pipenv - Pipfile.lock

jeremylong:
Adds support for Pipfile.lock per #4995.

This currently uses the exact same enable/disable settings as the PipfileAnalyzer - and will supersede the PipfileAnalyzer if bother a `Pipfile` and `Pipfile.lock` exist.


aikebah:
Looks like autocorrect messed up your PR message (pipefile / pipelock / bother)

Code looks good to me. 

One remark: I think it would be good to add some remark in the javadoc of both analyzers that they operate in tandem on the same activation property (with the new analyzer taking over when a lock file is present), so that when revisiting them in the future it's hinted by the docs that if one changes the other is likely to need change as well.

aikebah (approved):
LGTM

PR #5408, opened 2023-02-01
fix: improve error messages for AssemblyAnalyzer

jeremylong:
Improves the error messages when the AssemblyAnalyzer cannot be initialized.


aikebah (approved):
LGTM

PR #5421, opened 2023-02-05
fix: bad version string interpolation

### LINKED ISSUE #5418: Broken link in HTML report
Hello,

The Maven package spring-retry appears in the Dependency Check report with the version number not being substituted:

image

Which causes the hyperlink to be broken:  [pkg:maven/org.springframework.retry/spring-retry@%24%7Brevision%7D](https://ossindex.sonatype.org/component/pkg:maven/org.springframework.retry/spring-retry@%24%7Brevision%7D?utm_source=dependency-check&utm_medium=integration&utm_content=8.0.2)

The original POM file can be found here: https://repo1.maven.org/maven2/org/springframework/retry/spring-retry/1.3.4/spring-retry-1.3.4.pom

Thank you and have a nice day.

Regards,

Maven plugin 8.0.2

jeremylong:
- Resolves https://github.com/jeremylong/DependencyCheck/issues/5418 by returning a null value when variable interpolation fails when evaluating a pom.xml
- also includes a fix for a missed TODO when constructing the includedBy data

### DIFF
--- a/core/src/main/java/org/owasp/dependencycheck/analyzer/JarAnalyzer.java
+++ b/core/src/main/java/org/owasp/dependencycheck/analyzer/JarAnalyzer.java
@@ -1295,7 +1295,7 @@ public class JarAnalyzer extends AbstractFileTypeAnalyzer {
      * @return the correct value which may be null
      */
     private static String intepolationFailCheck(String value) {
-        if (value != null && value.startsWith("${") && value.endsWith("}")) {
+        if (value != null && value.contains("${")) {
             return null;
         }
         return value;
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/owasp/dependencycheck/analyzer/JarAnalyzer.java
```diff
@@ -1288,6 +1288,19 @@ private boolean isPackage(String key, String value) {
...
     }
 
+    /**
+     * Returns null if the value starts with `${` and ends with `}`.
+     *
+     * @param value the value to check
+     * @return the correct value which may be null
+     */
+    private static String intepolationFailCheck(String value) {
+        if (value != null && value.startsWith("${") && value.endsWith("}")) {
```
aikebah: I think even just using value.contains("${") may be considered to account for patterns where e.g. artifactid would be set using a variable for a framework/technology version ('the library is a bridge for version x of the framework'/'the library is designed for API version y of the  related technology') and that variable was not oroperly resolved.

aikebah (approved):

core/src/main/java/org/owasp/dependencycheck/analyzer/JarAnalyzer.java
jeremylong:
```suggestion
        if (value != null && value.contains("${")) {
```

PR #5470, opened 2023-02-17
fix: do not throw error if pyproject.toml is in node_modules

### LINKED ISSUE #5464: OWASP DC 8.0.2 cli failure on node app due to Python Analyzer finding node module using poetry
**Describe the bug**
OWASP CLI scanner fails when scanning a node app when the python analyzer finds a node module using a pyproject.toml file. 

The problem occurs using version 8.0.2 of the CLI.

**Log file**
[OWASP DC 8.0.2 Fail](https://gist.github.com/brandon-pangilinan/00799f32d8fe1809998126653b15be20)

**To Reproduce**
1. Run OWASP CLI on node based app with node_gyp as a module
2. See error

**Expected behavior**
Pipeline shows that OWASP scan failed with exit code 14, and that a pyproject.toml was found, but no requirements.txt or poetry.lock file. 

**Additional context**

jeremylong:
resolves #5464


aikebah (approved):
LGTM

PR #5473, opened 2023-02-18
fix: Prefer pom.properties G/A/V over pom.xml G/A/V to resolve GAV interpolation issues

aikebah:
Improves the fix of #5418 to retain the proper pURL when a pom.properties is found.

## Description of Change

When pom.properties is found with at least the properties for G/A/V as documented by maven use the G/A/V coordinates from the pom.properties as those are expected to always contain the fully interpolated values for G/A/V whereas the pom.xml may contain property references that are resolved by Maven in the model-building.

## Have test cases been added to cover the new functionality?

no

PR #5475, opened 2023-02-19
feat: Support msbuild Directory.build.props

### LINKED ISSUE #1948: MSBuild analyzer could (should?) check for props files as well

With the new msbuild stuff (core sdk project style), dependencies can also be put into .props and .targets files in the root of the solution.

Please also scan the following files:

* .props
* .targets (e.g. Directory.build.targets)

I hope it's as easy as adding the extensions to this line:

https://github.com/jeremylong/DependencyCheck/blob/master/core/src/main/java/org/owasp/dependencycheck/analyzer/MSBuildProjectAnalyzer.java#L75

-

**Additional context**

Let me know if you need more info.

jeremylong:
- partially resolves #1948
- removes unnecessary interfaces


core/src/main/java/org/owasp/dependencycheck/analyzer/MSBuildProjectAnalyzer.java
```diff
@@ -66,7 +70,7 @@ public class MSBuildProjectAnalyzer extends AbstractFileTypeAnalyzer {
     private static final String ANALYZER_NAME = "MSBuild Project Analyzer";
 
     /**
-     * The phase in which the analyzer runs.
+     * The phase fis which the analyzer runs.
```
aikebah:
```suggestion
     * The phase in which the analyzer runs.
```

core/src/main/java/org/owasp/dependencycheck/xml/pom/Model.java
```diff
@@ -335,98 +334,28 @@ public void processProperties(Properties properties) {
         if (properties == null) {
             return;
         }
-        this.groupId = interpolateString(this.groupId, properties);
+        this.groupId = InterpolationUtil.interpolate(this.groupId, properties);
```
aikebah: Be careful when merging/rebasing... main has changes removing the G/A/V interpolation from #5473 that should be retained

core/src/test/java/org/owasp/dependencycheck/utils/InterpolationUtilTest.java
```diff
@@ -0,0 +1,46 @@
...
+    public void testInterpolate() {
+        Properties prop = new Properties();
+        prop.setProperty("key", "value");
+        prop.setProperty("nested", "nested ${key}");
+        String text = "This is a test of '${key}' '${nested}'";
+        String expResults = "This is a test of 'value' 'nested value'";
+        String results = InterpolationUtil.interpolate(text, prop);
+        assertEquals(expResults, results);
+    }
+
```
aikebah:
```suggestion

    @Test
    public void testInterpolateNonexistentErased() {
        Properties prop = new Properties();
        prop.setProperty("key", "value");
        String text = "This is a test of '${key}' and '${nothing}'";
        String expResults = "This is a test of 'value' and ''";
        String results = InterpolationUtil.interpolate(text, prop);
        assertEquals(expResults, results);
    }
```

Think it would be good to also explicitly test the erasure of non-existent property-references (as documented in the InterpolationUtil comments), to avoid them being accidentally retained as-is or replaced by `null`.

This addition will surface that the current coding is not not consistent with the documentation (as the result is currently `This is a test of 'value' and '${nothing}'`). So either the interpolation needs to be fixed, or the Javadoc comment needs to be updated to reflect the actual behaviour and my suggested addition needs to be updated to `testInterpolateNonexistentRetained` and prediction of the current actual result.

core/src/main/java/org/owasp/dependencycheck/utils/InterpolationUtil.java
```diff
@@ -0,0 +1,157 @@
...
+/**
+ *
+ * @author Jeremy Long
+ */
+public final class InterpolationUtil {
+
+    private InterpolationUtil() {
+    }
+
+    /**
```
aikebah, line 33: Documentation is not consistent with actual behaviour, see review of the corresponding testcase

core/src/test/java/org/owasp/dependencycheck/utils/InterpolationUtilTest.java
aikebah: Would also be good to add duplicates for the MSBUILD SyntaxStyle interpolation (copies of the methods, but with braces instead of curly-braces and addition of the SyntaxStyle parameter to the interpolate call

aikebah (changes requested):
Coding of interpolation is not in line with its documented behaviour (was previously already an error in interpolation method in the Model class).

core/src/main/java/org/owasp/dependencycheck/analyzer/MSBuildProjectAnalyzer.java
```diff
@@ -108,16 +112,24 @@ protected void prepareFileTypeAnalyzer(Engine engine) throws InitializationExcep
     @Override
     @SuppressWarnings("StringSplitter")
     protected void analyzeDependency(Dependency dependency, Engine engine) throws AnalysisException {
+        final File parent = dependency.getActualFile().getParentFile();
+        //TODO while we are supporting props - we still do not support Directory.Build.targets
+        final File propsProject = new File(parent, "Directory.Build.props");
```
aikebah: According to the spec we should not load both (unless there is the import) and search up all the way to the root

https://learn.microsoft.com/en-us/visualstudio/msbuild/customize-your-build?view=vs-2019#search-scope

https://learn.microsoft.com/en-us/visualstudio/msbuild/customize-your-build?view=vs-2019#use-case-multi-level-merging

core/src/main/java/org/owasp/dependencycheck/utils/InterpolationUtil.java
jeremylong: Fixed the `InterpolationUtil` to correctly return an empty string if the key is missing.

core/src/main/java/org/owasp/dependencycheck/analyzer/MSBuildProjectAnalyzer.java
jeremylong: I've attempted to implement the spec - best effort... Warnings will be logged for imports that cannot be followed.

core/src/main/java/org/owasp/dependencycheck/xml/pom/Model.java
jeremylong: I just rebased the branch from `main` and ensured your changes were kept.

core/src/main/java/org/owasp/dependencycheck/analyzer/MSBuildProjectAnalyzer.java
```diff
@@ -175,4 +187,126 @@ protected void analyzeDependency(Dependency dependency, Engine engine) throws An
...
+                    BOMInputStream bis = new BOMInputStream(fis)) {
+                //skip BOM if it exists
+                bis.getBOM();
+                entries = parser.parse(bis);
+                imports.addAll(parser.getImports());
+            } catch (FileNotFoundException ex) {
+                throw new RuntimeException(ex);
+            } catch (IOException ex) {
+                throw new RuntimeException(ex);
+            }
```
jeremylong: TODO - correct exception handling

core/src/main/java/org/owasp/dependencycheck/data/nuget/DirectoryBuildPropsParser.java
```diff
@@ -0,0 +1,103 @@
...
+                                props.put(name, value.getNodeValue().trim());
+                            }
+                        }
+                    }
+                }
+            }
+            return props;
+        } catch (ParserConfigurationException | SAXException | IOException | XPathExpressionException ex) {
+            throw new RuntimeException(ex);
+        }
```
jeremylong: TODO - correct exception handling

jeremylong:
@aikebah thank you for the review and comments. Everything should be good to go now - with the exception of the error handling. I will update that in a day or two before finalizing and merging this PR.

aikebah (approved):

aikebah:
> @aikebah thank you for the review and comments. Everything should be good to go now - with the exception of the error handling. I will update that in a day or two before finalizing and merging this PR.

YW, always keen to try and find potential issues before features are introduced. Have made a habit of reading specs and checking whether code is matching specs (though reading the specs of the MSBuild documentation leaves and itchy feeling due to their free format open-endedness so that you can only 'try your best to conform to whatever Visual Studio happens to do'). 

I prefer the formal specs where you can really determine what is conformant and what not. I think your updated naive implementation is the best we can do (which covers the only usages that I would trust to reliably work now and in the future in my projects were I to be a dotnet developer with Visual Studio solutions projects given the documentation that Microsoft has put up for it)

PR #5487, opened 2023-02-22
fix: Erroneous error-log for deprecated CLI flag usage when using properyfile based disablement of Node Audit Analyzer

### LINKED ISSUE #5483: Erroneous `disableNSP` error
When running `dependency-check` I get the following error:

```
[ERROR] The disableNSP argument has been deprecated and replaced by disableNodeAudit
[ERROR] The disableNSP argument will be removed in the next version
```

We don't supply this argument through the command-line arguments:

```
$ dependency-check -s . --enableExperimental -f JSON -f HTML --suppression dependency-suppression.xsd --exclude "**/docs/**" --exclude "Danger/" --propertyfile "dependencycheck.properties" --connectiontimeout 60000 --ossIndexUsername $SONATYPE_USERNAME --ossIndexPassword $SONATYPE_PASSWORD --ossIndexRemoteErrorWarnOnly true --bundleAuditWorkingDirectory .
```

Our properties file supplies `analyzer.node.audit.enabled=false`, which I believe is the right value.

```
# defines if the experimental and retired analyzers can be enabled
analyzer.experimental.enabled=true
analyzer.retired.enabled=false

# ios specific
analyzer.cocoapods.enabled=true
analyzer.ruby.gemspec.enabled=true
analyzer.swift.package.manager.enabled=true
analyzer.swift.package.resolved.enabled=true

# others
analyzer.archive.enabled=true
analyzer.assembly.enabled=false
analyzer.autoconf.enabled=false
analyzer.bundle.audit.enabled=true
analyzer.cmake.enabled=false
analyzer.composer.lock.enabled=false
analyzer.cpanfile.enabled=false
analyzer.golang.dep.enabled=false
analyzer.golang.mod.enabled=false
analyzer.jar.enabled=false
analyzer.node.audit.enabled=false
analyzer.node.package.enabled=false
analyzer.nugetconf.enabled=false
analyzer.nuspec.enabled=false
analyzer.openssl.enabled=false
analyzer.pe.enabled=false
analyzer.pip.enabled=false
analyzer.pipfile.enabled=false
analyzer.pnpm.audit.enabled=false
analyzer.python.distribution.enabled=false
analyzer.python.package.enabled=false
analyzer.retirejs.enabled=false
analyzer.yarn.audit.enabled=false
```

Is this a bug, or an issue with our configuration?

aikebah:
## Description of Change

- Removal of the erroneous code, as the deprecation was already back in the v3.3.3 release.
- Refactor the method-name that can be seen as the root-cause of existence to prevent regressions in case of future CLI-flag deprecations.
- Removed the TODO-comment that has been lurking in the code since v3.3.3

In my view, given that we are 5 majors ahead of v3 it would be safe to remove in a bugfix-release, so scheduling it for 8.1.1

no

### DIFF
--- a/cli/src/main/java/org/owasp/dependencycheck/App.java
+++ b/cli/src/main/java/org/owasp/dependencycheck/App.java
@@ -558,7 +558,6 @@ public class App {
                 !cli.isDisabled(CliParser.ARGUMENT.DISABLE_DART, Settings.KEYS.ANALYZER_DART_ENABLED));
         settings.setBoolean(Settings.KEYS.ANALYZER_NODE_PACKAGE_ENABLED,
                 !cli.isDisabled(CliParser.ARGUMENT.DISABLE_NODE_JS, Settings.KEYS.ANALYZER_NODE_PACKAGE_ENABLED));
-        //TODO next major - remove the deprecated check in isNodeAuditDisabled
         settings.setBoolean(Settings.KEYS.ANALYZER_NODE_AUDIT_ENABLED,
                 !cli.isNodeAuditDisabled());
         settings.setBoolean(Settings.KEYS.ANALYZER_YARN_AUDIT_ENABLED,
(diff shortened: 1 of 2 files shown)


jeremylong (approved):
LGTM

PR #5509, opened 2023-02-27
fix: allow hosted suppressions file to be disabled

### LINKED ISSUE #5508: How to disable hostedSuppressions
I am looking to disable `hostedSuppressions` when using the Gradle plugin `8.0.1` since the URL is not accessible via our proxy. Initially I tried adding this in my Gradle config

```
hostedSuppressions {
     enabled = false
}
```
This did not work. I get the error below:
`Could not set unknown property 'enabled' for object of type org.owasp.dependencycheck.gradle.extension.HostedSuppressionsExtension`

I then noticed documentation states 
`hostedSuppressions | enabled | The number of hours to wait before checking for new updates of the hosted suppressions file`

http://jeremylong.github.io/DependencyCheck/dependency-check-gradle/configuration.html

Is this correct and intentional? And what can I do to disable this feature?

jeremylong:
Needed to resolve https://github.com/jeremylong/DependencyCheck/issues/5508

We should allow the hosted suppressions file usage to be disabled.


aikebah:
I think the documentation should also be updated for gradle plugin update task? I would at least expect an impact to that task as well (https://github.com/jeremylong/DependencyCheck/blob/main/src/site/markdown/dependency-check-gradle/configuration-update.md)

other than that LGTM

PR #5512, opened 2023-02-27
fix: Ensure retrievePomProperties always returns non-null

### LINKED ISSUE #5511: xmlbeans-5.1.1.jar causes NullPointerException in setGAVFromPomDotProperties in 8.1.1
**Describe the bug**
Dependency check fails on xmlbeans-5.1.1.jar. I can also reproduce it with some internal artifacts.

Issue appeared after upgrading to 8.1.1. using the command line version. I assume it is caused by #5473.

**Log file**
```
[WARN] An unexpected error occurred during analysis of '/tmp/dctemp0837e096-b355-48c9-95a1-163f1d9d837e/check3825733863783876208tmp/94/WEB-INF/lib/xmlbeans-5.1.1.jar' (Jar Analyzer): Cannot invoke "java.util.Properties.containsKey(Object)" because "pomProperties" is null
[ERROR] 
java.lang.NullPointerException: Cannot invoke "java.util.Properties.containsKey(Object)" because "pomProperties" is null
        at org.owasp.dependencycheck.xml.pom.Model.setGAVFromPomDotProperties(Model.java:399)
        at org.owasp.dependencycheck.analyzer.JarAnalyzer.analyzePOM(JarAnalyzer.java:450)
        at org.owasp.dependencycheck.analyzer.JarAnalyzer.analyzeDependency(JarAnalyzer.java:330)
```

aikebah:
Fixes #5511

## Description of Change

Update code to prevent null-valued object to be returned

no, validated with a local build using `pkg:maven/io.micronaut/micronaut-inject@3.8.5` that the NPE is resolved, intend to look for either detection in an existing regression-test or addition of one in order to not have regressions in the future, but consider the bug's disruptive disruptive potential enough to first fix it and work on getting a proper test in later.

### DIFF
--- a/core/src/main/java/org/owasp/dependencycheck/analyzer/JarAnalyzer.java
+++ b/core/src/main/java/org/owasp/dependencycheck/analyzer/JarAnalyzer.java
@@ -512,7 +512,7 @@ public class JarAnalyzer extends AbstractFileTypeAnalyzer {
      * @return a Properties object or null if no pom.properties was found
      */
     private Properties retrievePomProperties(String path, final JarFile jar) {
-        Properties pomProperties = new Properties();
+        final Properties pomProperties = new Properties();
         final String propPath = path.substring(0, path.length() - 7) + "pom.properties";
         final ZipEntry propEntry = jar.getEntry(propPath);
         if (propEntry != null) {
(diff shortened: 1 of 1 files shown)


jeremylong:
guessing we should push another release?

jeremylong (approved):

schdief:
> guessing we should push another release?

yes please, this breaks our pipelines

FloEberle (approved):

aikebah:
> guessing we should push another release?

Yes, think so too. Unfortunate that it was found only after release.

PR #5557, opened 2023-03-15
fix: casing issue in SQL

### LINKED ISSUE #5553: Wrong column name in dbStatements.properties
**Describe the bug**
Using OWASP Check with a central database as described in https://jeremylong.github.io/DependencyCheck/data/database.html leads to an SQLException.

The problem occurs using version 8.1.2-release of the ANT-Task

**Log file**
no log needed, please see "Additional Context"

**To Reproduce**

1. Create needed QWASP database in your favarite DB-System (not h2, bug was found using MSSQL)
2. Create ANT-Script with call to <dependency-check> Task, configured with a dedicated database server
3. Run Ant Task (may need a lot of time due to the database will be filled initially)
4. "Clean up orphans" will cause the SQLException

**Expected behavior**
Run without SQLException

**Additional context**
The causing bug is simply a typo in resources/data/dbStatements.properties in line 19:
`CLEANUP_ORPHANS=DELETE FROM cpeEntry WHERE id not in (SELECT CPEEntryId FROM software);`
contains a wrong column "**CPEEntryId**" for table software. The column seems to be case sensitive in MSSQL and should be "**cpeEntryId**". Please correct to 
`CLEANUP_ORPHANS=DELETE FROM cpeEntry WHERE id not in (SELECT cpeEntryId FROM software);`

jeremylong:
resolves #5553

### DIFF
--- a/core/src/main/resources/data/dbStatements.properties
+++ b/core/src/main/resources/data/dbStatements.properties
@@ -16,7 +16,7 @@
 # Copyright (c) 2018 The OWASP Foundation. All Rights Reserved.
 
 DELETE_VULNERABILITY=DELETE FROM vulnerability WHERE cve = ?
-CLEANUP_ORPHANS=DELETE FROM cpeEntry WHERE id not in (SELECT CPEEntryId FROM software);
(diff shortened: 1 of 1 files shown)


aikebah (approved):
LGTM

PR #5565, opened 2023-03-17
docs: Update arguments.md

ghost:
The --ossIndexRemoteErrorWarnOnly parameter needs a true/false as an argument according to https://github.com/jeremylong/DependencyCheck/blob/main/cli/src/main/java/org/owasp/dependencycheck/CliParser.java#L400

This is a change to the documentation to reflect that.

*no*


jeremylong (approved):

jeremylong:
Thanks!

Issue #5451, opened 2023-02-14
Github Workflow - FP Issue evaluation fails on missing pkg and cannot be re-run


rseeton:
**Describe the bug**
Dependency check will detect vulnerabilities without a 'pkg' value.  
When reporting FP issues and no pkg is provided the issues workflow (evaluate false positives) will fail due to the missing pkg information.  Subsequently adding the pkg information to the FP issue description will trigger a rerun of the evaluate false positives workflow... unfortunately this rerun will fail (not sure where the parsing fails)

**Version of dependency-check used**
Attempting to report false positives from any package (specificially 8.0.2 & 8.1.0) under the github.com issues tab.  The failure appears to lie in the issues workflow.

**Log file**
See https://github.com/jeremylong/DependencyCheck/issues/5449 

**To Reproduce**
1. Create FP Issue with empty/missing/invalid Package details 
2. The evaluate false positive workfow process will fail and attach 'pending more information' label
3. Edit the problem description and assign a valid Package string
4. The evaluate false positive workfow process continues to fail

The same details (with the updated package detail) can be assigned to clean/new FP Issue and this will cleanly trigger the evaluate false positive workflow

**Expected behavior**
Ideally, if we fix the problem description, the evaluation process would be able to run to completion.

**Additional context**

jeremylong:
I've seen this as well... but I haven't been able to figure out how to fix this...

rseeton:
@jeremylong  - I'll take a crack at this and let you know what I find.

Issue #5450, opened 2023-02-14
[FP]: Release 8.1.0 is throwing a lot of old Springboot Framework CVEs against spring-retry


rseeton:
### Package URl

pkg:maven/org.springframework.retry/spring-retry@1.3.3

### CPE

`cpe:2.3:a:pivotal_software:spring_framework:1.3.3:*:*:*:*:*:*:*`

### CVE

CVE-2018-1270

### ODC Integration

{"label"=>"CLI"}

### ODC Version

8.1.0

### Description

This may be an issue with 8.1.0, as we don't see this with the 8.0.2 package and the relevant CVE has not changed June 2022.

There are a number of other/relate issues reported against the spring-retry-1.3.3.jar in our package.

(Retrying after failed processing of initial ticket)

aikebah:
Appears to be caused by #5421

8.0.2 CLI report still has an ill-formed package-URL 
`pkg:maven/org.springframework.retry/spring-retry@%24%7Brevision%7D`

which makes the 'spring libraries that are not spring-framework' suppression to step in and prevent the FP up to 8.0.2 CLI

8.1.0 no longer has a package-url

PR #5567, opened 2023-03-18
fix: #5451 - Correction to console.log definition for reprocessing

rseeton:
## Description of Change

When processing a 'False Positive' issue report, the system fails when an invalid Package URI is provided (this is correct behaviour).  This failure occurs in the _Parse Package URL_ step.
However attempts to correct the problem (e.g. by providing a value Package URI) would fail in the _Remove Labels_ step.

This was traced to the console.log call.  The console.log call has been extended to call 'listLabelsOnIssue'.  An alternative would be to remove the console.log call in this step entirely.

no 

(My local fork does not appear to allow the creation of Issues.  Tests can be conducted with the existing Issue #5450

### DIFF
--- a/.github/workflows/false-positive-ops.yml
+++ b/.github/workflows/false-positive-ops.yml
@@ -18,13 +18,20 @@ jobs:
         uses: actions/github-script@v6.4.0
         with:
           script: |
-              console.log(github.event.issue.labels);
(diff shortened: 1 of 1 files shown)


jeremylong (approved):
LGTM - Thanks!

Issue #5583, opened 2023-03-22
Building Dependency-Check without running tests does not work


Marcono1234:
**Describe the bug**
The README of this project suggests the following to command to build the project without running any tests:
```
mvn -s settings.xml install -Dmaven.test.skip=true
```
([README section link](https://github.com/jeremylong/DependencyCheck#building-without-running-tests))

However, that command fails for the "Dependency-Check Ant Task" module:
```
...
[INFO] -------------------< org.owasp:dependency-check-ant >-------------------
[INFO] Building Dependency-Check Ant Task 8.2.1-SNAPSHOT                  [5/7]
[INFO] --------------------------------[ jar ]---------------------------------
[INFO] ------------------------------------------------------------------------
[INFO] Reactor Summary for Dependency-Check 8.2.1-SNAPSHOT:
[INFO] 
[INFO] Dependency-Check ................................... SUCCESS [  6.542 s]
[INFO] Dependency-Check Utils ............................. SUCCESS [  8.800 s]
[INFO] Dependency-Check Core .............................. SUCCESS [ 11.010 s]
[INFO] Dependency-Check Command Line ...................... SUCCESS [  4.587 s]
[INFO] Dependency-Check Ant Task .......................... FAILURE [  0.038 s]
[INFO] Dependency-Check Maven Plugin ...................... SKIPPED
[INFO] Dependency-Check Plugin Archetype .................. SKIPPED
[INFO] ------------------------------------------------------------------------
[INFO] BUILD FAILURE
[INFO] ------------------------------------------------------------------------
[INFO] Total time:  36.797 s
[INFO] Finished at: 2023-03-22T20:46:46Z
[INFO] ------------------------------------------------------------------------
Error:  Failed to execute goal on project dependency-check-ant: Could not resolve dependencies for project org.owasp:dependency-check-ant:jar:8.2.1-SNAPSHOT: Could not find artifact org.owasp:dependency-check-core:jar:tests:8.2.1-SNAPSHOT -> [Help 1]
```

My assumption is that `-Dmaven.test.skip=true` also skips creation of a test artifact, which is however required by later steps.
To verify that this is not caused by a mistake I made with my local setup, I let GitHub actions run the same command and it fails with the same error ([log](https://github.com/Marcono1234/DependencyCheck/actions/runs/4494328354/jobs/7906670683#step:9:583)).

**Version of dependency-check used**
`main` (9ca6da02496797e930653e236d48bb6aba79da5d)

[...]

PR #5624, opened 2023-03-30
docs: Suggest `mvn verify` instead of `mvn install` in building instructions

Marcono1234:
## Fixes Issue #
_none_

## Description of Change

Suggests usage of `mvn verify` instead of `mvn install` in the building instructions in the README. This avoids placing the artifacts in the local Maven repository.

See also https://github.com/jeremylong/DependencyCheck/issues/5583#issuecomment-1489320710

*no*

### DIFF
--- a/README.md
+++ b/README.md
@@ -143,13 +143,13 @@ git clone --depth 1 https://github.com/jeremylong/DependencyCheck.git
 
 On *nix
 ```
-$ mvn -s settings.xml install
+$ mvn -s settings.xml verify
 $ ./cli/target/release/bin/dependency-check.sh -h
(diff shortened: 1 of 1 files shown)


aikebah:
When building the project for usage of the produced snapshots `mvn install` is the most appropriate. Only usage of a custom-build CLI would require no more than the `verify` goal.

Marcono1234:
Good point; `mvn install` is probably not so problematic then.

Also I cannot reproduce the case I mentioned in https://github.com/jeremylong/DependencyCheck/issues/5583#issuecomment-1489320710 anymore that Maven did not pick up changes to files, so I assume I made a mistake while building the project. Sorry for the trouble.

PR #5696, opened 2023-05-04
feat: enable local proxy configuration in maven plugin configuration

mischa-n:
## Description of Change

CICD-SEC-03 recommends:
1. ... packages are pulled through an internal proxy rather than directly from the internet ...
2. ... pull packages from internal repositories ...

We have solved (2), trust our developers, they trust the developers of the third party libraries in use, and those trust the developers of dependent fourth party libraries ...
Nevertheless, we would like to go one step further and completely disable access to the Internet in our CI/CD pipeline.
Therefore, we need to configure a proxy for the dependency-check-maven plugin.

Currently, you can define a proxy to use in the dependency-check-maven plugin via the Maven settings or the system property `http.proxyHost`.
However, in both cases, the configured proxy is used not only for loading data for analysis, but potentially for dependency resolution as well:
* https://maven.apache.org/guides/mini/guide-proxies.html "... configure a proxy to use for some or all of your HTTP requests with Maven ..."
* https://www.baeldung.com/maven-behind-proxy

With the proposed change, a local proxy can be configured for the dependency-check-maven plugin, which is then used exclusively for loading the analysis data:

```
<configuration>
	...
	<proxy>
		<host>myproxy</host>
		<port>1234</port>
	</proxy>
	...
</configuration>
```

*no*

### DIFF
--- a/maven/src/main/java/org/owasp/dependencycheck/maven/BaseDependencyCheckMojo.java
+++ b/maven/src/main/java/org/owasp/dependencycheck/maven/BaseDependencyCheckMojo.java
@@ -2204,8 +2204,8 @@ public abstract class BaseDependencyCheckMojo extends AbstractMojo implements Ma
         else if ( this.proxy != null && this.proxy.host != null) {
             settings.setString(Settings.KEYS.PROXY_SERVER, this.proxy.host);
             settings.setString(Settings.KEYS.PROXY_PORT, Integer.toString(this.proxy.port));
-            settings.setStringIfNotNull(Settings.KEYS.PROXY_USERNAME, this.proxy.username);
-            settings.setStringIfNotNull(Settings.KEYS.PROXY_PASSWORD, this.proxy.password);
+            // user name and password from <server> entry settings.xml
+            configureServerCredentials(this.proxy.serverId, Settings.KEYS.PROXY_USERNAME, Settings.KEYS.PROXY_PASSWORD);
         }
         
         final String[] suppressions = determineSuppressions();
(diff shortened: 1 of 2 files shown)


maven/src/main/java/org/owasp/dependencycheck/maven/BaseDependencyCheckMojo.java
```diff
@@ -2197,6 +2200,14 @@ protected void populateSettings() {
                 settings.setString(Settings.KEYS.PROXY_NON_PROXY_HOSTS, System.getProperty("http.nonProxyHosts"));
             }
         }
+        // or use configured <proxy>
+        else if ( this.proxy != null && this.proxy.host != null) {
+            settings.setString(Settings.KEYS.PROXY_SERVER, this.proxy.host);
```
jeremylong, line 2205: Wouldn't you also want to be able to store these values in the settings.xml so you don't have to store the credentials in clear text? See how the method `configureServerCredentials()` is used in for other configurations.

jeremylong (changes requested):
Any place we are accepting credentials we should allow this to be retrieved from the settings.xml.

maven/src/main/java/org/owasp/dependencycheck/maven/BaseDependencyCheckMojo.java
mischa-n: We have a slightly different setup: the settings.xml is centrally specified, so it is used by all projects. In some places there are actually credentials, but these only refer to environment variables, e.g. `<password>${env.xyz}</password>`. `xyz` is then stored (encrypted) in the CI system.

So in our case, the password would not be stored in plain text in the POM either, but would also refer to an environment variable.


But I will try to implement the requirement anyway.

mischa-n: Should be solved now.

mischa-n:
@jeremylong : is there anything else I can do here to get the PR accepted?

jeremylong (approved):

jeremylong:
Sorry about the delay - thanks for the PR!

PR #5768, opened 2023-06-12
build: Release 8.3.1

jeremylong:
re-release 8.3.0 as 8.3.1.


Dhanxy:
hi team,

One question, version 8.3.1 or 8.30
 doesn't work with gradle? I am trying to edit my code but an error is displayed.

I would also like to ask why 2 versions were released on the same day?

Issue #5779, opened 2023-06-19
[FP]: CVE-2023-35116


AB-xdev:
### Package URl

pkg:maven/com.fasterxml.jackson.core/jackson-databind@2.15.2

### CPE

cpe:2.3:a:fasterxml:jackson-databind:2.15.2:*:*:*:*:*:*:*

### CVE

CVE-2023-35116

### ODC Integration

{"label"=>"Maven Plugin"}

### ODC Version

8.2.1

### Description

As mentioned in https://github.com/FasterXML/jackson-databind/issues/3972 CVE-2023-35116 is a FP.

The CVE is reported for all versions of the dependency.

The CVE is detected through https://central.sonatype.com/artifact/com.fasterxml.jackson.core/jackson-databind/2.15.2/versions.

aikebah:
The finding by DC is not a false positive. The CVE is getting disputed in the tickets. Resolution should be with the vulnerability sources - OSSIndex and NVD (delisting it as improperly issued) and/or the CVE-issuer (to get it fully withdrawn).

Whether or not it is a vulnerability of the library is open for interpretation. From the ticket it is clear that chances for exploitability of it can be considered very low.

As an attacker you would need some other attack vector to coerce an application into the creation of a cyclic Map that you know to be subsequently serialized by Jackson. When you manage to do so you would be able to break the application on a StackOverflow, which could be classified as a vulnerability against the availability of the application.

Note: I agree with cowtowncoder that for similar effects within hashCode Map implementations in the JDK should also be considered vulnerable if the authorities in charge decide to not withdraw the CVE for this issue. Especially since the cyclic nature of the Map is explicitly allowed (though discouraged) in Map's Javadoc, with no mention whatsoever on the fatal consequences for hashCode ('no longer well-defined' is something else then 'calling the method is guaranteed to throw a StackOverflowException'): 

> While it is permissible for a map to contain itself as a value, extreme caution is advised: the equals and hashCode methods are no longer well defined on such a map

CristianDuta:
> The finding by DC is not a false positive. The CVE is getting disputed in the tickets. Resolution should be with the vulnerability sources - OSSIndex and NVD (delisting it as improperly issued) and/or the CVE-issuer (to get it fully withdrawn).
> 
> Whether or not it is a vulnerability of the library is open for interpretation. From the ticket it is clear that chances for exploitability of it can be considered very low.
> 
[...]

PR #5780, opened 2023-06-19
fix: #5779 FP

CristianDuta:
## Fixes Issue #

#5779

## Description of Change

Suppress `CVE-2023-35116`  for `jackson-databind` as it is a false positive.


aikebah:
The finding by DC is not a false positive. The CVE is getting disputed in the tickets. Resolution should be with the vulnerability sources - OSSIndex and NVD.
Whether or not it is a vulnerability of the library is open for interpretation. From the ticket it is clear that chances for exploitability of it can be considered very low.

As an attacker you would need some other attack vector to coerce an application into the creation of a cyclic Map that you know to be subsequently serialized by Jackson. When you manage to do so you would be able to break the application on a StackOverflow, which could be classified as a vulnerability against the availability of the application.

Note: I tend to agree with cowtowncoder that for similar behaviour Map's hashCode should also be considered vulnerable if the authorities in charge decide to not withdraw the CVE for this issue. Especially since the cyclic nature of the Map is explicitly allowed (though discouraged) in Map's Javadoc, with no mention whatsoever on the fatal consequences for hashCode ('no longer well-defined' is something else then 'calling the method is guaranteed to throw a StackOverflowException'): 

> While it is permissible for a map to contain itself as a value, extreme caution is advised: the equals and hashCode methods are no longer well defined on such a map

aikebah:
Already delisted by Sonatype OSSIndex

PR #5808, opened 2023-07-04
fix: use curl with -L to follow github redirect

krimdomu:
## Description of Change

This only adds the `-L` option to curl to follow the download redirect.

*no*


aikebah (approved):
Thanks for the PR

PR #5818, opened 2023-07-10
fix: Hint Analyzer should run before VersionFilter Analyzer

aikebah:
## Description of Change

Ensure that HintAnalyzer runs before VersionFilterAnalyzer as HintAnalyzer may add versions that are not yet present in the evidences. VersionFilterAnalyzer may otherwise wrongly set an explicit version for the dependency if the collected evidences converge to a single version but the hints add a new and different version.

Also added an implementation note on the AnalysisPhase to document which analyzers are bound to which phase.

Introduces 2 additional POST_INFORMATION_COLLECTION phases so that DependencyMergingAnalyzer is guaranteed to run before HintAnalyzer which in its turn is guaranteed to run before VersionFilterAnalyzer

no


jeremylong (approved):
LGTM

PR #5849, opened 2023-07-31
feat: Add support for Nexus v3 to NexusAnalyzer

aikebah:
## Description of Change

Implement support for Nexus V3 REST API to search for artifact by SHA-1 hash in addition to the current Nexus V2 API support.

The NexusAnalyzer now attempts to use Nexus v2 API if the standard Nexus v2 API prefix (`service/local`) is found in the Nexus Analyzer URL. Otherwise it will attempt to use the Nexus v3 API.

Existing test-case for NexusV2 has been cloned into a copy for V3, including the `@Ignore` of all testcases by default. Locally they have been used to test against private Nexus Repository OSS V2 and Nexus Repository OSS V3 instances to validate that both the Nexus V2 and Nexus V3 implementations are passing the tests.


jeremylong (approved):
LGTM

PR #5900, opened 2023-08-27
fix: Do not filter out evidences added by hints

### LINKED ISSUE #5894: Adding version through through hints no longer shows up as evidence in the report
**Describe the bug**
Looking under the evidence section in the report would show versions added through hints before version 8.4.0 of dependency check. Using the maven plugin (haven't tested with others so they might be affected).
I don't know if the issue is just in the report and if the added version is considered correctly when scanning or not.

The problem occurs using version 8.4.0 of the maven plugin

**Log file**

**To Reproduce**

Can be reproduced with the following pom, running `mvn dependency-check:check`:
```
<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
  <modelVersion>4.0.0</modelVersion>
  <groupId>com.example</groupId>
  <artifactId>foo</artifactId>
  <version>0.1-SNAPSHOT</version>
  <packaging>pom</packaging>

  <dependencies>
    <dependency>
      <groupId>org.hibernate</groupId>
      <artifactId>hibernate-validator</artifactId>
      <version>5.0.2.Final</version>
    </dependency>
  </dependencies>

  <build>
    <plugins>
      <plugin>
        <groupId>org.owasp</groupId>
        <version>8.4.0</version>
      </plugin>
    </plugins>
  </build>
</project>
```
hibernate-validator has a version added from the default hints, but it doesn't show up in the report. Running the same pom but with version 8.3.1 instead shows the version under evidence in the report.

**Expected behavior**
The report should show the added version under evidence.

**Additional context**
I assume that this is related to #5812 and #5818

aikebah:
## Description of Change

Add a marker to the evidences to indicate an evidence came from a hints file (as the evidence source is fully user-controllable the source is not usable to detect hint-origin) and avoid removal of those hints when filtering 'inaccurate' evidences in the VersionFilterAnalyzer.

Also removes the now obsolete hint for a historical Hibernate-validator false-negative that is already properly resolved since the switch to JSON datafeeds.

no

### DIFF
--- a/core/src/main/java/org/owasp/dependencycheck/analyzer/HintAnalyzer.java
+++ b/core/src/main/java/org/owasp/dependencycheck/analyzer/HintAnalyzer.java
@@ -198,7 +198,7 @@ public class HintAnalyzer extends AbstractAnalyzer {
             for (VendorDuplicatingHintRule dhr : vendorHints) {
                 if (dhr.getValue().equalsIgnoreCase(e.getValue())) {
                     dependency.addEvidence(EvidenceType.VENDOR, new Evidence(e.getSource() + " (hint)",
-                            e.getName(), dhr.getDuplicate(), e.getConfidence()));
+                            e.getName(), dhr.getDuplicate(), e.getConfidence(), true));
                 }
             }
         }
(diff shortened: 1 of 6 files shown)


jeremylong (approved):
LGTM

Issue #5919, opened 2023-09-03
Report format compatible with the GitLab Dependency Check format


niklasfi:
**Is your feature request related to a problem? Please describe.**
It is good practice to have a dependency scanner as part of a ci/cd pipeline. GitLab even has a handy vulnerability overview in merge requests, if you do so. Unfortunately, the dependency scanner supported by GitLab [1] is very much lacking in Java version support. In general, only LTS versions are supported, and even they only arrive half a year after the Java release. Right now, I only have two choices: use the current Java version or have a properly working dependency scanner in my pipeline.

![image](https://github.com/jeremylong/DependencyCheck/assets/299626/805be1f6-ad95-499d-97c0-52b49ac11745)

**Describe the solution you'd like**
GitLab's vulnerability feature is not restricted to using GitLab's scanner. It's possible to provide your own scanner [2]. My proposal to solve the dilemma above is to add a new report format conforming to GitLab's report schema [3].

**Describe alternatives you've considered**
I have previously implemented a "translator" from DependencyCheck into GitLab's format that ran after DependencyCheck and generated an output file in the desired format. This however feels clunky, when the power of velocities templating language is already at our hands when generating the report file.

**Additional context**
I have implemented an initial version but I'm missing some additional fields. I will shortly add a pull request to this issue for that.

---
**References**
[1] gemnasium, which I think uses DependencyCheck inside
[2] documentation: https://docs.gitlab.com/ee/user/application_security/dependency_scanning/#reports-json-format
[3] schema definition: https://gitlab.com/gitlab-org/security-products/security-report-schemas/-/raw/v15.0.6/dist/dependency-scanning-report-format.json?ref_type=tags

aikebah:
Gemnasium uses its own analyzer that hooks into Gitlab's own vulnerability database

https://gitlab.com/gitlab-org/security-products/analyzers/gemnasium

aikebah:
And from the looks of the documentation at https://docs.gitlab.com/ee/user/application_security/dependency_scanning/#working-around-missing-support-for-certain-languages-or-package-managers

you cannot have it process a JSON report created by another tool, but only have a tool create a supported dependencies input format so that gemnasium can do its work even if your build tool's dependency configuration format is not supported

aikebah:
So my gut feel is that your approach will turn out not to work

niklasfi:
[...]

Issue #5824, opened 2023-07-12
feature request: add a markdown summary


efenderbosch-atg:
Generate a summary markdown file that can be included in build results instead of the entire result document.

As a workaround, this can be done w/ some utilities, but would be nice to have as a built in feature.

```bash
pandoc build/reports/dependency-check-jenkins.html -f html -t gfm | \
  sed -n '/## Summary/,/## Dependencies/p' | \
  grep -v '## Dependencies' >> $GITHUB_STEP_SUMMARY
```

aikebah:
A cleaner work-around (or even solution for your use case?) might be to use your own custom report template (a somewhat hidden, but available feature which, since version 7.4.4, PR #5197, can be used in addition to standard built-in report formats)

You can feed your own template by providing the full path to a velocity template (must be a path to a file that can be given to `java.io.File#File(java.lang.String)` constructor yielding `true` for `java.io.File#isFile()`) or the name of a custom template that you provide on the classpath. It allows for exactly the amount of detail that you would like to have as a github-flavored-markdown formatted report.

https://github.com/jeremylong/DependencyCheck/blob/6d1d84b4fd423754dde772914b3a55557a116623/core/src/main/java/org/owasp/dependencycheck/reporting/ReportGenerator.java#L299-L308

https://github.com/jeremylong/DependencyCheck/blob/6d1d84b4fd423754dde772914b3a55557a116623/core/src/main/java/org/owasp/dependencycheck/reporting/ReportGenerator.java#L436-L446

efasel:
example VSL file:
```
#**
 See VSL templates in https://github.com/jeremylong/DependencyCheck/tree/main/core/src/main/resources/templates for information about available values
 *###
#set($double_hash = '##')
#set($triple_hash = '###')
# Dependency-Check Report

[How to read the report](https://jeremylong.github.io/DependencyCheck/general/thereport.html) | [Suppressing false positives](https://jeremylong.github.io/DependencyCheck/general/suppression.html)

$double_hash Project: $applicationName

* application: $applicationName
#if($applicationVersion)
* application version: $applicationVersion
#end
* NVD API Last Checked: ${properties.getMetaData().get("NVD API Last Checked")}
* NVD API Last Modified: ${properties.getMetaData().get("NVD API Last Modified")}
* scan date: $scanDateXML
* DependencyCheck version: $version
#if($groupID)
* groupID: $groupID
#end
#if($artifactID)
* artifactID: $artifactID
#end

[...]

PR #5920, opened 2023-09-03
feat: gitlab dependency scanner report format #5919

niklasfi:
## Description of Change

As described in #5919, I am working on adding a new report format to DependencyCheck that can be directly fed to GitLab to be used as a dependency scanner in ci/cd-pipelines.

no

### DIFF
--- a/core/src/main/resources/templates/gitlabReport.vsl
+++ b/core/src/main/resources/templates/gitlabReport.vsl
@@ -54,19 +54,26 @@
                     #end
                     ##    ((List<Dependency>)context.get("dependencies")).get(5).getVulnerabilities().stream().collect(Collectors.toList()).get(0)
                     {
-                        "id": "$enc.json($vulnerability.name)", ##todo
+                        "id": "$enc.json($vulnerability.name)",
                         "identifiers": [
-                            #foreach( $ref in $vulnerability.getReferences(true) )
-                                {
-                                    "type": "$enc.json($ref.source)",
-                                    "name": "$enc.json($ref.name)",
-                                    "value": "$enc.json($ref.name)",
+                            {
+                                "type": "$enc.json($vulnerability.getSource().name())"
+                                #if( $vulnerability.getSource().name().equals("NVD") )
+                                    , "name": "$enc.json($vulnerability.name)"
+                                #elseif( $vulnerability.getSource().name().equals("NPM") )
+                                    , "name": "$enc.json($vulnerability.name) (NPM)"
+                                #else
+                                    , "name": "$enc.json($vulnerability.name)"
+                                #end
+                                , "value": "$enc.json($dependency.Sha1sum)"
 
-                                    ## optional properties
-                                    "url": "$enc.json($ref.url)"
-                                }
-                                #if( $foreach.hasNext ),#end
-                            #end
+                                ## optional properties
+                                #if( $vulnerability.getSource().name().equals("NVD") )
+                                    , "url": "https://web.nvd.nist.gov/view/vuln/detail?vulnId=$enc.url($vulnerability.name)"
+                                #elseif( $vulnerability.getSource().name().equals("NPM") )
+                                    , "url": "https://github.com/advisories/$enc.url($vulnerability.name)"
+                                #end
+                            }
                         ],
                         "location": {
                             "file": "$enc.json($dependency.filePath)",
@@ -97,7 +104,17 @@
                         ## we have to include this field to be compliant with the format spec
                         "solution": "" ## todo: not in our dataset?
 
-                        ## "links": [], --> not implemented
+                        "links": [
+                            #foreach( $ref in $vulnerability.getReferences(true) )
+                                {
+                                    "name": "$enc.json($ref.name)",
+
+                                    ## optional properties
+                                    "url": "$enc.json($ref.url)"
+                                }
+                                #if( $foreach.hasNext ),#end
+                            #end
+                        ],
                         ## "details": [], --> not implemented
                         ## "tracking": {}, --> not implemented
                         ## "flags": [], --> not implemented.
--- a/core/src/main/resources/templates/gitlabReport.vsl
+++ b/core/src/main/resources/templates/gitlabReport.vsl
@@ -1,0 +1,137 @@
+{
+    "version": "15.0.6",
+    "schema": "https://gitlab.com/gitlab-org/security-products/security-report-schemas/-/raw/v15.0.6/dist/dependency-scanning-report-format.json?ref_type=tags", ##todo
+    "scan": {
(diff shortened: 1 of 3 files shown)


niklasfi:
Hi, I would gladly receive some feedback on my pull request.

1. Are you interested in adding a GitLab report format to DependencyCheck?
2. Do you agree the best way to implement it is by adding a template and a new report format?
3. Do you have any pointers on (some of) the missing fields?
4. Do you think it's still a good idea to merge this even if some fields of the target format are not available right now?
5. Would you accept pull requests adding the additionally needed information to the template engine's context?
6. Are additional tests needed? If so: what do you think is a good test template to start with? The other report formats probably have tests as well, right?

core/src/main/resources/templates/gitlabReport.vsl
```diff
@@ -0,0 +1,137 @@
...
+                        },
+
+                        ## optional properties
+                        "name": "$enc.json($vulnerability.name)",
+                        "description": "$enc.json($vulnerability.description)",
+                        "severity": "$enc.json($vulnerability.cvssV3.getBaseSeverity())",
+                        ## we have to include this field to be compliant with the format spec
+                        "solution": "" ## todo: not in our dataset?
+
+                        ## "links": [], --> not implemented
```
aikebah: This would be the proper place to put the references that you coded in the identifiers

core/src/main/resources/templates/gitlabReport.vsl
```diff
@@ -0,0 +1,137 @@
...
+                                ##]
+                            }
+                        },
+
+                        ## optional properties
+                        "name": "$enc.json($vulnerability.name)",
+                        "description": "$enc.json($vulnerability.description)",
+                        "severity": "$enc.json($vulnerability.cvssV3.getBaseSeverity())",
+                        ## we have to include this field to be compliant with the format spec
+                        "solution": "" ## todo: not in our dataset?
```
aikebah: Solution indeed not available in our dataset; in most cases solution would be 'update to patched version', but cases may exist every now and then where a vulnerability is valid 'for all current versions' as there is either no fix yet, or e.g. the maintainer refuses to change the code to 'secure by default' and only indicates that their library is not intended to be used with untrusted data.. 
However I think the format spec allows to leave this property completely out:
```json
        "required": [
          "id",
          "identifiers",
          "location"
        ],
```
So not sure why you think it would have to be included to be compliant with the spec
only id, identifiers and location are required in the vulnerabilities array items

core/src/main/resources/templates/gitlabReport.vsl
```diff
@@ -0,0 +1,137 @@
...
+                        ## "iid": number, --> not implemtend
+                        ##"direct": false, --> not implemeten
+                        ##"dependency_path": [] --> not implemented
+                    }
+                    #if( $foreach.hasNext ),#end
+                #end
+            ]
+            ## no optional properties
+        }
+    ],
```
aikebah, line 147: In my view for this case we should do a better job if we claim to support the GitLab dependency_scanner reporting, rather than blindly lie that we obtained all the dependency information from a maven pom.xml.

core/src/main/resources/templates/gitlabReport.vsl
```diff
@@ -0,0 +1,137 @@
...
+                            #foreach( $ref in $vulnerability.getReferences(true) )
+                                {
+                                    "type": "$enc.json($ref.source)",
+                                    "name": "$enc.json($ref.name)",
+                                    "value": "$enc.json($ref.name)",
+
+                                    ## optional properties
+                                    "url": "$enc.json($ref.url)"
+                                }
+                                #if( $foreach.hasNext ),#end
```
aikebah: This code should move to the 'links' section. The identifiers list would hold the information similar to what's in the vulnerability section header in the HTML report:
https://github.com/jeremylong/DependencyCheck/blob/3871517289d2cb0cf2c75cbe6ec229d280da8b8d/core/src/main/resources/templates/htmlReport.vsl#L993-L999

aikebah (commented):
Some preliminary feedback. Based on the various comments you already put in I think it needs some further investigation into resolving the various issues with the GitLab report format versus the information we currently have available to make it so mature that I would vote for inclusion.

Nevertheless I'm in favor of the initiative, and even if we can't reach to an acceptable level of support the efforts to try and achieve as much as reasonbly possible are useful for anyone wanting to embed DependencyCheck into GitLab dependency-scanning. 

If not as a formally supported format people could always use a partially completed and syntactically working template as their own report format (using the VSL as a custom-format report - see https://github.com/jeremylong/DependencyCheck/issues/5824#issuecomment-1632834729 for some details on that approach)

jeremylong (changes requested):
Please see the comments from @aikebah.

core/src/main/resources/templates/gitlabReport.vsl
niklasfi: yup, moved it there

core/src/main/resources/templates/gitlabReport.vsl
niklasfi: sorry, this was an oversight. You are right, that it is optional, and as such, I've commented it out for now.

core/src/main/resources/templates/gitlabReport.vsl
niklasfi: updated in latest commit

niklasfi:
hi @jeremylong @aikebah sorry for taking so long to respond to your well thought out remarks on the pull request. I have implemented the changes as you suggested except for the `dependency_files` section. There I don't really know how to best proceed.

jeremylong:
@niklasfi thanks! Supporting this format shows that we might be missing a field on the dependency object: `source`. Where the source would be something like: `pom.xml`, `poetry.lock`, `path/to/some.jar`. We could then add dependencies to the array grouped on source. I think this information would be useful for other reasons as well. I'll file an enhancement request.

niklasfi:
fyi we have set up a fork of this project on our private gitlab and are now using the generated SNAPSHOT builds in our piples. I can report it's working as is, with the known restrictions (source file is kind of a lie). In the process of setting this up, I found a couple more bugs. That is where the extra commits come from.

This is what it looks like on our internal gitlab now (for the ant subfolder of this repo).

image

jeremylong:
Sorry for the delays - I will include this in the 9.0.0 release.

core/src/main/resources/templates/gitlabReport.vsl
jeremylong: @aikebah I agree - but I think I'm going to merge this one and then fix it once we have better source information (see #5989).

jeremylong (approved):

Issue #5925, opened 2023-09-06
[FP]: form-filler-addon has been reported wrongly 


ZheSun88:
### Package URl

pkg:maven/com.vaadin.flow.ai/form-filler-addon@1.0.0

### CPE

cpe:2.3:a:vaadin:flow:1.0.0:*:*:*:*:*:*

### CVE

CVE-2019-25027, CVE-2021-31412, CVE-2018-25007, CVE-2021-31404

### ODC Integration

{"label"=>"Maven Plugin"}

### ODC Version

8.4.0

### Description

This version of form-filler-addon is the new product by Vaadin, it should not be included in the reported CVE. as it is a prerelease version, the artifact is in our prerelease nexus instead of maven central.

ZheSun88:
I made a PR that combine this fix with previous ones.

PR #5927, opened 2023-09-06
fix(fp): fixes FP #5925

ZheSun88:


aikebah (approved):
LGTM thanks for the PR

Issue #5976, opened 2023-10-06
versionCheckEnabled doesn't seemt to work?


Dekari:
**Describe the bug**
Running a dependency-check against a pom.xml that has 

`<versionCheckEnabled>true</versionCheckEnabled>
`
doesn't report that new version is available.

**Version of dependency-check used**

<groupId>org.owasp</groupId>
<artifactId>dependency-check-maven</artifactId>
<version>7.4.3</version>

**Log file**

A log line something like:
"[WARNING] A new version of dependency-check is available ..." doesn't seem to be there.

**To Reproduce**

run "mvn org.owasp:dependency-check-maven:check"
on any pom.xml having the plugin version 7.4.3 and versionCheckEnabled = true  

**Expected behavior**
A warning message that newer version of the plugin to be typed on mvn output.

**Additional context**
-

chadlwilson:
This check only is done if you have auto update of NVD feeds enabled, but since this is quite old/stale I’ll close this now as there have been many changes in the last couple of years. If you’re still having issues please share your config and we’ll look into it.

PR #5978, opened 2023-10-08
feat: Utilize NVD API

jeremylong:
## Description of Change

Utilize the NVD API to retrieve the CVE data. This can also utilize the NVD API data feed that can be created using the [vulnz-cli](https://github.com/jeremylong/Open-Vulnerability-Project/tree/main/vulnz#creating-the-cache).


jeremylong:
In order for these tests to pass we will likely need to store an NVD API key and use this during tests.

jeremylong:
This PR is getting closer to being finalized. Left to complete:

1. Utilize API Key during regression testing in GH Actions
2. Document the new configuration.
3. Lots of testing

.github/workflows/build.yml
```diff
@@ -63,6 +63,7 @@ jobs:
         env:
           MAVEN_USERNAME: ${{ secrets.OSSRH_USERNAME }}
           MAVEN_PASSWORD: ${{ secrets.OSSRH_TOKEN }}
+          NVD_API_KEY: ${{ secrets.NVD_API_KEY }}
         run: mvn -s settings.xml -Prelease clean package verify source:jar javadoc:jar gpg:sign deploy -DreleaseTesting --no-transfer-progress --batch-mode -Dgpg.passphrase=${{ secrets.OSSRH_GPG_SECRET_KEY_PASSWORD }}
       - name: SARIF Multitool
         uses: microsoft/sarif-actions@v0.1
```
jeremylong, line 69: We trust the microsoft action...

jeremylong:
Update - the PR is getting closer. There is still something to fix with how the last updated checks are being done. The release testing ends up with multiple update sequences (which is the main reason the action is failing).

core/src/main/java/org/owasp/dependencycheck/data/update/NvdApiDataSource.java
```diff
@@ -0,0 +1,513 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * Copyright (c) 2023 Jeremy Long. All Rights Reserved.
+ */
+package org.owasp.dependencycheck.data.update;
+
+import com.fasterxml.jackson.core.JsonProcessingException;
```
jeremylong:
```suggestion
```

core/src/main/java/org/owasp/dependencycheck/data/update/NvdApiDataSource.java
```diff
@@ -0,0 +1,513 @@
...
+                    if (ctr == 0) {
+                        LOGGER.info(String.format("NVD API has %,d records in this update", max));
+                    }
+                    if (items != null && !items.isEmpty()) {
+                        final Future<NvdApiProcessor> f = processingExecutorService.submit(new NvdApiProcessor(cveDb, items));
+                        submitted.add(f);
+                        errorCount = 0;
+                        ctr += 1;
+                        if ((ctr % 10) == 0) {
+                            double percent = (double) (ctr * 2000) / max * 100;
```
jeremylong:
```suggestion
                            final double percent = (double) (ctr * 2000) / max * 100;
```

core/src/main/java/org/owasp/dependencycheck/data/update/NvdApiDataSource.java
```diff
@@ -0,0 +1,513 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * Copyright (c) 2023 Jeremy Long. All Rights Reserved.
+ */
+package org.owasp.dependencycheck.data.update;
+
+import com.fasterxml.jackson.core.JsonProcessingException;
+import io.github.jeremylong.openvulnerability.client.nvd.DefCveItem;
+import io.github.jeremylong.openvulnerability.client.nvd.NvdApiException;
```
jeremylong:
```suggestion
```

jeremylong:
The only two items I know of left are documenting the upgrade process in the `README.md` and improving the error handling in the `NvdApiDataSource` (around line 341).

core/src/main/java/org/owasp/dependencycheck/analyzer/OssIndexAnalyzer.java
```diff
@@ -318,34 +318,61 @@ private Vulnerability transform(final ComponentReport report, final ComponentRep
...
+                    if (tmp != null) {
+                        availabilityImpact = CvssV2Data.CiaType.fromValue(tmp);
+                    }
+                    final String severity = Cvss2Severity.of((float) cvssScore).name().toUpperCase();
+                    final CvssV2Data cvssData = new CvssV2Data("2.0", source.getCvssVector(), accessVector,
+                            accessComplexity, authentication, confidentialityImpact,
+                            integrityImpact, availabilityImpact, cvssScore,
+                            severity, null, null, null, null, null, null, null, null, null, null);
+                    final CvssV2 cvssV2 = new CvssV2(null, null, cvssData, severity, null, null, null, null, null, null, null);
+                    result.setCvssV2(cvssV2);
```
aikebah, line 367: Just wondering, wouldn't it be more logical to use CVSSUtil#vectorToCvssV2(String, Double) for this branch?

core/src/main/java/org/owasp/dependencycheck/data/nvdcve/CveDB.java
```diff
@@ -946,25 +991,35 @@ private int updateOrInsertVulnerability(DefCveItem cve, String description) {
...
+            if (cve.getCve().getMetrics() != null && cve.getCve().getMetrics().getCvssMetricV31() != null) {
+                optCvssv31 = cve.getCve().getMetrics().getCvssMetricV31().stream().sorted(Comparator.comparing(CvssV3::getType)).findFirst();
+            }
+
+            CvssV3 cvssv3 = null;
+            if (optCvssv31 != null && optCvssv31.isPresent()) {
+                cvssv3 = optCvssv31.get();
+            } else if (optCvssv30 != null && optCvssv30.isPresent()) {
+                cvssv3 = optCvssv30.get();
+            }
```
aikebah:
```suggestion
            Optional<CvssV3> optCvssv30 = Optional.empty();
            if (cve.getCve().getMetrics() != null && cve.getCve().getMetrics().getCvssMetricV30() != null) {
                optCvssv30 = cve.getCve().getMetrics().getCvssMetricV30().stream().sorted(Comparator.comparing(CvssV3::getType)).findFirst();
            }
            Optional<CvssV3> optCvssv31 = Optional.empty();
            if (cve.getCve().getMetrics() != null && cve.getCve().getMetrics().getCvssMetricV31() != null) {
                optCvssv31 = cve.getCve().getMetrics().getCvssMetricV31().stream().sorted(Comparator.comparing(CvssV3::getType)).findFirst();
            }

            CvssV3 cvssv3 = null;
            if (optCvssv31.isPresent()) {
                cvssv3 = optCvssv31.get();
            } else if (optCvssv30.isPresent()) {
                cvssv3 = optCvssv30.get();
            }
```
Cleaner code to initialize Optionals to the empty Optional

core/src/main/java/org/owasp/dependencycheck/data/nvdcve/CveDB.java
```diff
@@ -1167,14 +1223,29 @@ private void updateVulnerabilityInsertSoftware(int vulnerabilityId, String cveId
...
-                    insertReference.setString(4, r.getRefsource());
+                    Optional<String> name = null;
+                    if (r.getTags() != null) {
+                        name = r.getTags().stream().sorted().findFirst();
+                    }
+                    if (name != null && name.isPresent()) {
+                        insertReference.setString(2, name.get());
+                    } else {
+                        insertReference.setNull(2, java.sql.Types.VARCHAR);
+                    }
```
aikebah:
```suggestion
                    Optional<String> name = Optional.empty();
                    if (r.getTags() != null) {
                        name = r.getTags().stream().sorted().findFirst();
                    }
                    if (name.isPresent()) {
                        insertReference.setString(2, name.get());
                    } else {
                        insertReference.setNull(2, java.sql.Types.VARCHAR);
                    }
```
Cleaner code to initialize Optional as empty instead of null

core/src/main/java/org/owasp/dependencycheck/data/nvdcve/CveDB.java
```diff
@@ -1528,6 +1591,134 @@ private void addNullableStringParameter(PreparedStatement ps, int pos, String va
...
+        } else {
+            ps.setString(i, value.value());
+        }
+    }
+
+    private void setUpdateColumn(PreparedStatement ps, int i, CvssV3Data.Version value) throws SQLException {
+        if (value == null) {
+            ps.setNull(i, java.sql.Types.NULL);
+        } else {
+            ps.setString(i, value.value());
```
aikebah, line 1718: java.sql.Types.NULL is
>The constant in the Java programming language that identifies the generic SQL value NULL.

Both the javadoc of `setNull` and the text and the code example of the [JDBC 4 JSR spec](https://jcp.org/en/jsr/detail?id=221) in paragraph 13.2.2.4 indicate that the actual SQL datatype of the target parameter is to be used as the second parameter. The code example listed there for setting a String parameter to JDBC NULL:
> ps.setNull(2, java.sql.Types.VARCHAR);

The current code works for e.g. h2, because h2 completely ignores the type variable, it just sets the value to ValueNull.INSTANCE, h2's internal representation of the database NULL value.

As to why the JDBC API designers decided to include `NULL` in the `java.sql.Types` class rather than adding it somewhere else is something likely only they will know. In my view it is an anomaly in that class that leads to mistakes like this.

core/src/main/java/org/owasp/dependencycheck/data/nvdcve/DatabaseProperties.java
```diff
@@ -46,26 +45,37 @@ public class DatabaseProperties {
...
+    public static final String MODIFIED = "Modified";
+
     /**
      * Stores the last updated time for each of the NVD CVE files. These
      * timestamps should be updated if we process the modified file within 7
      * days of the last update.
      */
     public static final String LAST_UPDATED_BASE = "NVD CVE ";
+
+    // TODO DELETE END---------------------------------------------------------
```
aikebah: Shouldn't the TODO be processed by deleting from START to END?

core/src/main/java/org/owasp/dependencycheck/data/update/EngineVersionCheck.java
```diff
@@ -123,14 +123,14 @@ public boolean update(Engine engine) throws UpdateException {
...
             /*
              * Only update if auto-update is enabled, the engine check is
-             * enabled, and the NVD CVE URLs have not been modified (i.e. the
-             * user has not configured them to point to an internal source).
+             * enabled, and the NVD DataFeed is being used (i.e. the user
+             * is likely on a private network). This check is not really needed
+             * so we are okay skipping it.
              */
-            if (enabled && autoupdate && original != null && original.equals(current)) {
+            if (enabled && autoupdate && datafeed != null) {
```
aikebah, line 133: I was already doubting this condition in the context of #5976.

Perfectly fine to keep it for this PR as is... but wanted to raise the awareness to the outstanding issue that is rooted at these conditions, which are now modified, but in a way that would leave the issue unresolved.

The major version bump required by this PR could be a good moment to also make only the update_version_check_enabled flag govern execution of the version check?

core/src/main/java/org/owasp/dependencycheck/reporting/SarifRule.java
```diff
@@ -564,17 +600,17 @@ public void setCvssv2IntegrityImpact(String cvssv2IntegrityImpact) {
...
+    public String getCvssv2ConfidentialityImpact() {
+        return cvssv2ConfidentialityImpact;
     }
 
     /**
      * Set the value of CVSS2 Confidential Impact.
      *
-     * @param cvssv2ConfidentialImpact new value of CVSS2 Confidential Impact
+     * @param cvssv2ConfidentialityImpact new value of CVSS2 Confidential Impact
      */
```
aikebah, line 611: Would be good to also update the remaining occurences in documentation from Confidential to Confidentiality

core/src/main/java/org/owasp/dependencycheck/utils/DateUtil.java
```diff
@@ -80,6 +82,22 @@ public static boolean withinDateRange(long date, long compareTo, int dayRange) {
...
+     * @param compareTo the date to compare against
+     * @param dayRange the number of days allowed
+     * @return if the date is within the dayRange of compareTo
+     */
+    public static boolean withinDateRange(ZonedDateTime date, ZonedDateTime compareTo, int dayRange) {
+        //seconds = dayRange x 24 hours/day x 60 min/hour x 60 sec/min
+        final long seconds = dayRange * 24L * 60L * 60L;
+        final Duration duration = Duration.between(date, compareTo);
+        final long secondsDifference = duration.getSeconds();
+        return secondsDifference < seconds;
```
aikebah:
```suggestion
        return secondsDifference >=0 && secondsDifference < seconds;
```
I think it would be 'not within range' if `compareTo` is before `date`?

Could potentially also be rewritten to use the comparison- and time-travel methods of ZonedDateTime as in: 

```java
        final ZonedDateTime rangeEnd = date.plusDays(dayRange);
        return compareTo.isAfter(date) && compareTo.isBefore(rangeEnd);
```

core/src/main/java/org/owasp/dependencycheck/xml/suppression/SuppressionRule.java
```diff
@@ -534,13 +525,14 @@ public void process(Dependency dependency) {
                     }
                 }
                 if (!remove) {
-                    for (float cvss : this.cvssBelow) {
-                        if (v.getCvssV2() != null && v.getCvssV2().getScore() < cvss) {
+                    for (Double cvss : this.cvssBelow) {
+                        //TODO validate this comparison
```
aikebah, line 529: I wonder if it would be more appropriate (until and unless we extend suppressions to explicit cvssv2Below / cvssv3Below) to first attempt to check CVSS v3 and only evaluate cvss v2 after that if v3 score did not trigger suppression?
A new major version would allow for changing the reference-version of CVSS that we target with the cvssBelow suppression rule and my impression is that CVSSv3 nowadays is in the lead, so it would be more logical to use CVSSv3 when present and only check CVSSv2 score when v3 failed to exceed the suppression threshold.

Not necessarily something to process for this PR, but food for thought before releasing 9.0.0.

The TODO itself should in the end be gone I think.... it needs to either have been adressed, or moved into a github issue identifying what still needs to be validated here.

core/src/test/java/org/owasp/dependencycheck/data/nvdcve/DatabasePropertiesIT.java
```diff
@@ -61,21 +60,6 @@ public void testIsEmpty() throws Exception {
...
-        NvdCveInfo updatedValue = new NvdCveInfo("test","url",1337L);
-        String key = "test";
-        long expected = 1337L;
-        DatabaseProperties instance = cveDb.getDatabaseProperties();
-        instance.save(updatedValue);
-        instance = cveDb.reloadProperties();
-        long results = Long.parseLong(instance.getProperty("NVD CVE " + key));
-        assertEquals(expected, results);
-    }
-
```
aikebah: Shouldn't the test be retained, rewritten to update and validate another database property?

core/src/test/java/org/owasp/dependencycheck/utils/CvssUtilTest.java
```diff
@@ -0,0 +1,183 @@
...
+        expResult = "UNKNOWN";
+        result = CvssUtil.cvssV2ScoreToSeverity(score);
+        assertEquals(expResult, result);
+    }
+
+    /**
+     * Test of cvssV3ScoreToSeverity method, of class CvssUtil.
+     */
+    @Test
+    public void testCvssVeScoreToSeverity() {
```
aikebah:
```suggestion
    public void testCvssV3ScoreToSeverity() {
```

maven/src/main/java/org/owasp/dependencycheck/maven/ProxyConfig.java
```diff
@@ -1,17 +1,95 @@
...
+    /**
+     * The proxy host.
+     */
+    private String host;
+    /**
+     * The proxy port.
+     */
+    private int port = 8080;
+    /**
+     * ID od server in Maven settings.xml, &lt;username&gt; and &lt;password&gt; will be
```
aikebah:
```suggestion
     * ID of server in Maven settings.xml, &lt;username&gt; and &lt;password&gt; will be
```
While at it better to also fix this existing typo

src/site/markdown/data/mirrornvd.md
```diff
@@ -2,30 +2,15 @@ Mirroring External Resources
...
-   must be specified (see the configuration for the specific dependency-check client used):
-   <ul>
-     <li>cveUrlModified</li>
-     <li>cveUrlBase</li>
-   </ul>
-</li>
-</ol>
+The Open Vulnerability Project's [vuln CLI](https://github.com/jeremylong/Open-Vulnerability-Project/tree/main/vulnz#caching-the-nvd-cve-data)
+can be used to create an offline copy of the data obtained from the NVD API.
+The configure dependency-check to use the NVD Datafeed URL.
```
aikebah:
```suggestion
Then configure dependency-check to use the NVD Datafeed URL.
```
Typo fix

utils/src/main/java/org/owasp/dependencycheck/utils/Settings.java
```diff
@@ -161,59 +161,47 @@ public static final class KEYS {
          * be imported.
          */
         public static final String CVE_CPE_STARTS_WITH_FILTER = "cve.cpe.startswith.filter";
+
+        //TODO NVD Begin New
```
aikebah: Remove the stale TODO
```suggestion
```

aikebah (commented):
Filing the comments from a first review-round, still need to take a close look at the core new classes pertaining to the new NVD API
* core/src/main/java/org/owasp/dependencycheck/data/update/NvdApiDataSource.java
* core/src/main/java/org/owasp/dependencycheck/data/update/nvd/api/DownloadTask.java
* core/src/main/java/org/owasp/dependencycheck/data/update/nvd/api/NvdApiProcessor.java
* core/src/main/java/org/owasp/dependencycheck/utils/CvssUtil.java

core/src/main/java/org/owasp/dependencycheck/data/nvdcve/CveDB.java
jeremylong: So you are saying this should be: `ps.setNull(i, java.sql.Types.VARCHAR);`. Thanks - I've likely done this in other places as well.

core/src/main/java/org/owasp/dependencycheck/xml/suppression/SuppressionRule.java
jeremylong: Good point. However, 9.0.0 is going to need to be released fairly soon as I'd like to have a little testing in the wild done before the NVD datafeeds shutdown on Dec 15th.

core/src/main/java/org/owasp/dependencycheck/data/nvdcve/CveDB.java
jeremylong: The calls to `setNull()` have been corrected.

core/src/main/java/org/owasp/dependencycheck/analyzer/OssIndexAnalyzer.java
jeremylong: While we could - it would require more string parsing generate the map and as such be slower. I get the reasoning - but the code is written, this works, and I think it is slightly faster.

core/src/main/java/org/owasp/dependencycheck/reporting/SarifRule.java
jeremylong: See https://github.com/jeremylong/DependencyCheck/pull/5978/commits/99280ab3f0c142066b20597744c79ffe18c0ed5c

core/src/main/java/org/owasp/dependencycheck/data/update/nvd/api/DownloadTask.java
```diff
@@ -0,0 +1,121 @@
...
+     *
+     * @param url the file to download
+     * @param processor the processor service to submit the downloaded files to
+     * @param cveDB the CVE DB to use to store the vulnerability data
+     * @param settings a reference to the global settings object; this is
+     * necessary so that when the thread is started the dependencies have a
+     * correct reference to the global settings.
+     * @throws UpdateException thrown if temporary files could not be created
+     */
+    public DownloadTask(String url, ExecutorService processor, CveDB cveDB, Settings settings) throws UpdateException {
```
aikebah:
```suggestion
    public DownloadTask(String url, ExecutorService processor, CveDB cveDB, Settings settings) {
```

UpdateException is never thrown

core/src/main/java/org/owasp/dependencycheck/data/update/NvdApiDataSource.java
```diff
@@ -0,0 +1,541 @@
...
+        try {
+            delay = settings.getLong(Settings.KEYS.NVD_API_DELAY);
+        } catch (InvalidSettingException ex) {
+            LOGGER.debug("Invalid setting `NVD_API_DELAY`?");
+        }
+        if (delay > 0) {
+            builder.withDelay(delay);
+        }
+
+        //TODO consider using CVE_CPE_STARTS_WITH_FILTER
```
aikebah: Isn't this TODO already covered by lines 283-286?

core/src/main/java/org/owasp/dependencycheck/data/update/NvdApiDataSource.java
```diff
@@ -0,0 +1,541 @@
...
+        ExecutorService processingExecutorService = null;
+        try {
+            processingExecutorService = Executors.newFixedThreadPool(PROCESSING_THREAD_POOL_SIZE);
+            final List<Future<NvdApiProcessor>> submitted = new ArrayList<>();
+            int errorCount = 0;
```
aikebah: errorCount is unused, only initialized to 0 here and reset to 0 at line 315. Gut feel is this is a left-over from a retry-max-n-times-on-error mechanism and should be removed.

core/src/main/java/org/owasp/dependencycheck/data/update/NvdApiDataSource.java
```diff
@@ -0,0 +1,541 @@
...
+                    if (ctr == 0) {
+                        LOGGER.info(String.format("NVD API has %,d records in this update", max));
+                    }
+                    if (items != null && !items.isEmpty()) {
+                        final Future<NvdApiProcessor> f = processingExecutorService.submit(new NvdApiProcessor(cveDb, items));
+                        submitted.add(f);
+                        errorCount = 0;
+                        ctr += 1;
+                        if ((ctr % 10) == 0) {
+                            final double percent = (double) (ctr * 2000) / max * 100;
```
aikebah: The magic number 2000 is dependent on the (external to dependencyCheck) defaults of the API client. I think a cleaner approach would be to explicitly code the resultsPerPage when building the client (still at 2000 results per page) and use a single static variable for both the explicitly requesting the resultsPerPage and computing this percentage.
As a nitpicking: the percentage could grow over 100 if there total number of records in the output is slightly (`<2000`) below a multiple of 20,000 (last page results in `(ctr % 10) == 0`, but the last page has less than 2,000 items), but I would be totally fine with that (worst case would be 111.1% `(20,000/18,001)*100` for 18,001 records in the results)

core/src/main/java/org/owasp/dependencycheck/data/update/NvdApiDataSource.java
```diff
@@ -0,0 +1,541 @@
...
+        try {
+            dbProperties = cveDb.getDatabaseProperties();
+            if (checkUpdate()) {
+                String url = settings.getString(Settings.KEYS.NVD_API_DATAFEED_URL);
+                if (url == null) {
+                    throw new UpdateException("NVD API Cache URL was not provided");
+                }
+                if (!url.endsWith("/")) {
+                    url += "/";
+                }
```
aikebah:
```suggestion
                final String url;
                if (!nvdDataFeedUrl.endsWith("/")) {
                    url = nvdDataFeedUrl + "/";
                } else {
                    url = nvdDataFeedUrl;
                }
```
The method parameter nvdDataFeedUrl is never used, instead the URL is re-requested from settings here. In my view keeping the parameter and removing the new lookup in this method would yield the most readable code.

core/src/main/java/org/owasp/dependencycheck/data/update/NvdApiDataSource.java
```diff
@@ -0,0 +1,541 @@
...
+        }
+        return lastCall;
+    }
+
+    private void processFuture(final Set<Future<NvdApiProcessor>> processFutures) throws UpdateException {
+        //complete processing
+        for (Future<NvdApiProcessor> future : processFutures) {
+            try {
+                final NvdApiProcessor task = future.get();
+                //TODO - do we need to check for errors?
```
aikebah: I tend to think that's a yes, otherwise we would register timestamps as if all is up-to-date to that point while some of the data processing has not successfully completed and only a purge+update would allow to recover from that?

core/src/main/java/org/owasp/dependencycheck/data/update/NvdApiDataSource.java
```diff
@@ -0,0 +1,541 @@
...
+            LOGGER.debug("Thread was interrupted during download", ex);
+            Thread.currentThread().interrupt();
+            throw new UpdateException("The download was interrupted", ex);
+        } catch (ExecutionException ex) {
+            LOGGER.debug("Thread was interrupted during download execution", ex);
+            throw new UpdateException("The execution of the download was interrupted", ex);
+        }
+    }
+
+    private boolean processApi() throws UpdateException {
```
aikebah, line 265: Method always returns true. To signal 'no updates done' I think it should return false in the case that 0 records have been downloaded from the API.
If my minds model of the code is correct this could be achieved by an insertion in-between line 332 and 333

```java
            max = submitted.size();
            final boolean updated = max > 0;
            ctr = 0;
```

 and update of the return statement at line 353
```java
            return updated;
```

core/src/main/java/org/owasp/dependencycheck/data/update/NvdApiDataSource.java
```diff
@@ -0,0 +1,541 @@
...
+     * downloaded
+     */
+    protected final Properties getRemoteCacheProperties(String url) throws UpdateException {
+        try {
+            final URL u = new URL(url + "cache.properties");
+            final Downloader d = new Downloader(settings);
+            final String content = d.fetchContent(u, true, Settings.KEYS.NVD_API_DATAFEED_USER, Settings.KEYS.NVD_API_DATAFEED_PASSWORD);
+            final Properties properties = new Properties();
+            properties.load(new StringReader(content));
+            return new Properties();
```
aikebah:
```suggestion
            return properties;
```
I presume you intended to return the retrieved/loaded properties here.

core/src/main/java/org/owasp/dependencycheck/utils/DateUtil.java
jeremylong: I used the isAfter/isBefore. See https://github.com/jeremylong/DependencyCheck/pull/5978/commits/0cacc48abd6ee0a2ef0c581ee27401d1410f3d92

core/src/main/java/org/owasp/dependencycheck/data/update/NvdApiDataSource.java
```diff
@@ -0,0 +1,541 @@
...
+     */
+    protected final Map<String, String> getUpdatesNeeded(String url, Properties cacheProperties, ZonedDateTime now) throws UpdateException {
+        LOGGER.debug("starting getUpdatesNeeded() ...");
+        final Map<String, String> updates = new HashMap<>();
+        if (dbProperties != null && !dbProperties.isEmpty()) {
+            final int startYear = settings.getInt(Settings.KEYS.NVD_API_DATAFEED_START_YEAR, 2002);
+            // for establishing the current year use the timezone where the new year starts first
+            // as from that moment on CNAs might start assigning CVEs with the new year depending
+            // on the CNA's timezone
+            final int endYear = now.getYear();
```
aikebah:
```suggestion
            final int startYear = settings.getInt(Settings.KEYS.NVD_API_DATAFEED_START_YEAR, 2002);
            // for establishing the current year use the timezone where the new year starts first
            // as from that moment on CNAs might start assigning CVEs with the new year depending
            // on the CNA's timezone
            final int endYear = now.withZoneSameInstant(ZoneId.of("UTC+14:00")).getYear();
```
Taking the comment into account we should re-zone `now` to the Pacific/Kiritimati timezone. In order to be safe for any timezone database the approach is taken of doing it with an explicit UTC offset instead.

core/src/main/java/org/owasp/dependencycheck/data/update/NvdApiDataSource.java
```diff
@@ -0,0 +1,541 @@
...
+            }
+        }
+        if (updates.size() > 3) {
+            LOGGER.info("NVD API Cache requires several updates; this could take a couple of minutes.");
+        }
+        return updates;
+    }
+
+    /**
+     * Downloads the NVD CVE Meta file properties.
```
aikebah:
```suggestion
     * Downloads the metadata properties of the NVD API cache.
```

aikebah (commented):
Some remarks from reviewing the new sources

core/src/main/java/org/owasp/dependencycheck/xml/suppression/SuppressionRule.java
aikebah: Sounds fine to me to go ahead and release this without change. Just wanted to raise it as a potential idea for your consideration. After filing it I also realized that there is likely more to change throughout the code and documentation when we want to treat CVSSv3 instead of CVSSv2 as the primary citizen and try to map anything else towards it on a best-effort basis.

aikebah:
@jeremylong My review is now finished. Thanks for all your hard work on getting DependencyCheck NVD API ready.

core/src/main/java/org/owasp/dependencycheck/data/update/NvdApiDataSource.java
jeremylong: yes, thank you. I have not done enough testing of the data-feed mode. This would have caused a bit of confusion.

core/src/main/java/org/owasp/dependencycheck/data/update/NvdApiDataSource.java
jeremylong: yup - I added that and forgot to remove the comment.

core/src/main/java/org/owasp/dependencycheck/data/update/NvdApiDataSource.java
jeremylong: Yes, left over code from the data-feed being unstable. I've removed it.

core/src/main/java/org/owasp/dependencycheck/data/update/NvdApiDataSource.java
jeremylong: fixed.

core/src/main/java/org/owasp/dependencycheck/data/update/NvdApiDataSource.java
jeremylong: I believe that comment was left over code I had commented out from a copy/paste. If there is an exception during processing it will present as `ExecutionException`; which would result in an `UpdateException`.

core/src/main/java/org/owasp/dependencycheck/data/update/NvdApiDataSource.java
jeremylong: fixed.

aikebah (approved):
LGTM, assuming that no tests were broken by the latest review patches

s3nl:
I'd like to test the new method of scanning offline. I compiled the development version of Dependency Check 9 according to these steps:
https://github.com/jeremylong/DependencyCheck/tree/scratch/nvdapi#development-usage
I was able to successfully scan a `.jar` online with:
`./cli/target/release/bin/dependency-check.sh --scan ./cli/target/release/lib/h2-2.1.214.jar`

This generated the DB and created the following folder structure:
image

My plan is to zip the required files inside the `release` dictionary, upload this zip to our intranet and have the servers which need to scan without internet access unzip this file prior to scanning. A nightly update of this zip file would ensure the latest CVEs are used. The fact that these files are inside a `release` is a bit strange to me. Before version 9 all I needed was `dependency-check-data/odc.mv.db`.

Would that work? Or would I have to use `vulnz` to generate the DB as is suggested by the description of this PR:
> This can also utilize the NVD API data feed that can be created using the [vulnz-cli](https://github.com/jeremylong/Open-Vulnerability-Project/tree/main/vulnz#creating-the-cache).

If I really have to use vulnz, then how would I post-process the generated DB and where should I put it in order for Dependency Check to use it?

Edit:
After extracting `dependency-check-9.0.0-SNAPSHOT-release.zip` and running 
`./bin/dependency-check.sh --project "Download vulnerability database" --data ./dependency-check-data --updateonly --nvdApiKey myapikey` from the extracted folder I now have `dependency-check-data/odc.mv.db` like I was used to. Am I right that not much has changed (besides the NVD API key) in this regard and I can continue to create and distribute a zip file of `dependency-check-data/odc.mv.db` like I was used to with the current 8.x.x version?

jeremylong:
Yes, you can continue to create and distribute the archive. In general, this is one of the better solutions to the offline problem.

humblekofe:
Dear Jeremy,
first of all a million thanks to you for all your hard work for this great tool!
I'm really looking forward for this to be released, in the meantime I tried to combine the scratch/nvdapi branch with the vulnz version 5.0.1 to create an offline mirror as described in the documentation.

I observed a few things that may be caused by my stupidity (maybe I'm trying to put things together in a way that they are not supposed to?) but I have looked in the documentation and code and couldn't find a solution.
If it's not asked too much, maybe you could point me into the right direction on what I am doing wrong?
As you're certainly very busy, I can also just wait for the final release and try again :-)

1) Vulnz cache files: I'm calling the vulnz.jar directly like described in https://github.com/jeremylong/Open-Vulnerability-Project/tree/main/vulnz#creating-the-cache and the files are stored with filenames like "nvdcve-2023.json.gz" (even without the additional gzip command)
When I configure the nvdDatafeedUrl to point to the jetty serving those files and run the check, it tries to download files named like "nvdcve--2023.json.gz" with two dashes. The reason is that the cache.properties on the server contains the prefix "nvdcve-" and NvdApiDataSource class additionally appends a dash after the prefix. If my understanding is correct, one or the other would need to be changed?
2) gzipped or not: After my attempt to remove the additional dashes in the NvdApiDataSource, the Download started but soon failed because illegal characters were encountered while downloading. As a quick try I decompressed a gz-file and placed it on the mirrorserver with the same filename "nvdcve-2023.json.gz" and the download succeeded.
So how is this supposed to work? Should the files downloaded from the nvdDatafeedUrl be compressed? As their name is hardcoded in the NvdApiDataSource to be "...json.gz" that is what I had thought but for some reason it didn't work
3) I have then patched the NvdApiDataSource to download ".json" files and decompressed all *.json.gz on the mirror. All Downloads were successful but I ran in to an NPE afterwards... and now I'm completely confused if I'm doing the right thing here.

Thanks again and best wishes,
Felix

jeremylong:
@humblekofe there were bugs in the nvd data feed implementation; commit https://github.com/jeremylong/DependencyCheck/pull/5978/commits/3df710e9e2369ac24d554daf11ab863765c4da39 should have resolved them all.

humblekofe:
> @humblekofe there were bugs in the nvd data feed implementation; commit [3df710e](https://github.com/jeremylong/DependencyCheck/commit/3df710e9e2369ac24d554daf11ab863765c4da39) should have resolved them all.

You are amazing, thank you for your quick response and fix!
I tried it out and it worked perfectly after one tiny additional modification in NvdApiDataSource.processDatafeed(...): the download URL for the *.json.gz files now contains two slashes (e.g. ".../cache//nvdcve-2005.json.gz" because there is a "/" appended in line 127-129.
If this is removed plus on the other hand added to the download of the cache.properties in row 130 (or 535 as it is required there), the cache update works.

jeremylong:
@humblekofe  what was the exact command line you used? I'm trying to understand the comment you made about the double slashes because in my testing everything is correct as is.

humblekofe:
I'm using the maven plugin and have the following configuration:
`<nvdDatafeedUrl>http://internal-server:8081/nvd-api-mirror/nvdcve-{0}.json.gz</nvdDatafeedUrl>`
which results in
org.owasp.dependencycheck.utils.DownloadFailedException: Download failed, unable to retrieve 'http://internal-server:8081/nvd-api-mirror//nvdcve-2018.json.gz'; Error downloading file http://internal-server:8081/nvd-api-mirror//nvdcve-2018.json.gz; unable to connect.

My understanding is that in NvdApiDataSource.java
- Line 123 cuts off the trailing slash from "url" but line 128 adds it again
- Line 122 creates the pattern but including the leading slash resulting in pattern="/nvdcve-{0}.json.gz"
- both combined in Line 498, 592 etc. resulting in double slashes.
Using nvdDataFeedUrl.substring(lio+1); in Line 122 fixes it for me.

I hope that helps.

humblekofe:
@jeremylong : I figured out why I'm having this problem and you may not experience this:
I'm serving the cache files with a jetty which rejects URLs with double slashes (see https://stackoverflow.com/questions/74395500/jetty-returns-400-when-uri-has-2-slashes-in-a-row-while-tomcat-doesnt-care). I don't know how other web-servers behave but if I use a Tomcat, the owasp check can successfully download the files although the requested URLs contain the double forward slash.
Double slashes are usually considered as a URI violation and I would consider them as a minor bug in owasp dep-check but the workaround with using a tomcat works.

jeremylong:
@humblekofe what I missed is that line 122 was:

```java
pattern = nvdDataFeedUrl.substring(lio);
```

But should be:

```java
pattern = nvdDataFeedUrl.substring(lio + 1);
```

I'll submit a PR shortly to fix this before publishing.

jeremylong:
You can also specify the data feed url like this:
```xml
<nvdDatafeedUrl>http://internal-server:8081/nvd-api-mirror/</nvdDatafeedUrl>
```

jeremylong:
@humblekofe see https://github.com/jeremylong/DependencyCheck/pull/6096

humblekofe:
@jeremylong thanks a lot for the fix and also for the hint regarding the alternative configuration of nvdDatafeedUrl that will wirk with 9.0.0 as well.

Issue #5989, opened 2023-10-14
Add Dependency Source


jeremylong:
Currently, ODC primarily tracks the file path of the dependency - not necessarily the "source". If scanning a pom.xml we do not provide the path to the pom. We do list the project/scope - but that doesn't necessarily get back to the specific pom.xml. This can cause confusion when trying to generate reports - see #5920.

ytzzh0726:
May I try to do this?
Last year, I worked on a project that utilized the dependencycheck.
At that time, in order to support Maven and Gradle source code, I added some logic to retrieve dependency information by executing the "mvn dependency:tree" and "gradle dependencies -b %s --configuration runtimeClasspath" commands.
 I combined this information with dependencycheck to analyze vulnerabilities in the dependencies. 
However, the results were not very satisfactory because this implementation relied too much on the network, resulting in slow and unreliable execution. 

Therefore, I later wanted to improve this implementation by parsing the pom.xml file instead of executing commands to retrieve dependencies. 

This year, I have been working on a low-code project and haven't continued with last year's project. However, upon seeing this requirement, I think I can give it a try.

jeremylong:
Sorry for the delay - if you want to give this a try please do! We love PRs.

regedit0726:
OK,I'll do it this week.

jeremylong:
While the PR may have some merit. The intent of this was to complete something like this: 

https://github.com/jeremylong/DependencyCheck/compare/scratch/add-source

Where we would need to call `dependency. addSourceReferences(source)` everywhere a new dependency was generated from a package manager. Pypi, nuget, lock files, maven, etc. We do not always reflect the source of the dependency - especially in the Java ecosystem where this is likely the most useful.

regedit0726:
So I need to add the path of parent pom to a dependency,but the sourceReferences of parent should not be added to the dependency,right?

jeremylong:
if we have a JAR file that was included because of the pom - we should put the pom in the sourceReference. While it would be great if we could figure out the exact line and in which pom (i.e. did it come from a parent pom) - for now we can just indicate that the source was the pom.xml that was being directly analyzed.

PR #6051, opened 2023-11-13
feat: Add maven Source analyzer (#5989)

regedit0726:
Fixes Issue #5989

 Description of Change

Add Maven source code analyzer,
Support Maven 3 version. Check if Maven is installed and if it is Maven 3 version by executing the 'mvn -version' command. If not installed or not Maven 3 version, the analyzer will not be enabled.
Then execute the 'mvn help:effective-settings' command to obtain the Maven local repository path.
By finding the pom.xml in the source code directory and executing the 'mvn dependency:list' command on it, analyze the result to obtain the local jar file paths of the dependencies. Further analysis of the dependencies can be done using ArchiveAnalyzer, JarAnalyzer, CentralAnalyzer, and other analyzers.

 Have test cases been added to cover the new functionality?

yes


jeremylong:
I think there is an interesting possibility with this - but this is not what I was trying to describe in the related issue. Due to timelines this PR may sit open for a bit before I can integrate this idea.

Issue #5572, opened 2023-03-21
gradle with http/https proxy - ignores no_proxy hosts


nauni77:
Hello.
We have a HTTP/ HTTPS proxy, which needs to be used. Now we configured a local proxy location for CVE data (proxy inside our company). But the none proxy hosts are not working. That means, if I want to use the local CVE proxy I have to disable the HTTP/ HTTPS proxy at gradle settings. If I enable the proxy settings at gradle.properties it will end with a timeout - because to proxy is used. But the server is listed at `systemProp.https.nonProxyHosts`.

But because of the disabling of the proxy settings - the other services can't be reached at the internet. During the build I see some timeouts, but at the end it fails because of the exception:

```
Cannot drop "VULNERABILITY" because "FKCWEENTRY" depends on it; SQL statement:
DROP TABLE IF EXISTS vulnerability
```

After disabling all the other services and avoid the failing of these requests the database shows no problem anymore.

Here is my snipped of the configuration:
```
    dependencyCheck {
        failOnError = false
        autoUpdate = true
        cveValidForHours = 1
        format = 'ALL'
        failBuildOnCVSS = 11
        data {
            directory = rootProject.file('nvddata').absolutePath
        }
        cve {
            urlBase = 'https://repo.intra.mycompany.de/artifactory/remote-ext-http-owasp/feeds/json/cve/1.1/nvdcve-1.1-%d.json.gz'
            urlModified  = 'https://repo.intra.mycompany.de/artifactory/remote-ext-http-owasp/feeds/json/cve/1.1/nvdcve-1.1-modified.json.gz'
        }
        suppressionFile = 'owasp.suppression.xml'

        analyzers {
            knownExploitedEnabled = false
            ossIndex {
                enabled = false
	     }
	     retirejs {
                enabled = false
             }
             hostedSuppressions {
                 enabled = false
             }
        }
    }
```
The important part is the section of the analyzers. Because of all the `enabled = false` there is no request to the external services.

In my eyes this is a bug and the settings from `.gradle/gradle.properties` should work correct and only the request to the cve files from `repo.intra.mycompany.de` should be transferred direct. But all other services (not named at the no_proxy variable) should be use the proxy.

Snippet from `~/.gradle/gradle.properties`:
```
systemProp.http.proxyHost=proxy.mycompany.de
systemProp.http.proxyPort=8080
[...]

PR #6054, opened 2023-11-13
fix: read proxy.nonproxyhosts as an array

PaddyDrury:
## Fixes Issue #

 #5572

## Description of Change

The proxy.nonproxyhosts setting is stored as an array but it is read as a string which corrupts the values. Treat it as an array instead.

no


aikebah:
@jeremylong 

I tend to think that the issue should be fixed in the gradle plugin WDYT?

All the other integrations set it as a String. The gradle plugin is the anomaly that uses an array for it.

PaddyDrury:
> @jeremylong
> 
> I tend to think that the issue should be fixed in the gradle plugin WDYT?
> 
> All the other integrations set it as a String. The gradle plugin is the anomaly that uses an array for it.

I hadn't appreciated this, being a user of only the gradle plugin. Happy to raise a PR to that repository instead.

PaddyDrury:
Closing this and will raise a PR to the gradle plugin repository.

PR #6094, opened 2023-11-21
fix: correct import

### LINKED ISSUE #6085: Concerning the Use of org.h2.util's IOUtils.copy(InputStream in, OutputStream out) method in JarAnalyzer
IOUtils.copy(InputStream in, OutputStream out) from org.h2.util is used by JarAnalyzer's extractPom(String path, JarFile jar) method. IOUtils.copy(InputStream in, OutputStream out) from org.apache.commons.io can be used in its place. This will eliminate the h2 need when utilising JarAnalyzer alone in the EVIDENCE_COLLECTION mode.

jeremylong:
resolves #6085

### DIFF
--- a/core/src/main/java/org/owasp/dependencycheck/analyzer/JarAnalyzer.java
+++ b/core/src/main/java/org/owasp/dependencycheck/analyzer/JarAnalyzer.java
@@ -54,7 +54,7 @@ import java.util.zip.ZipEntry;
 
 import org.apache.commons.io.FilenameUtils;
 import org.apache.commons.lang3.StringUtils;
-import org.h2.util.IOUtils;
+import org.apache.commons.io.IOUtils;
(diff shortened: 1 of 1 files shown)


aikebah:
Don't know (haven't checked) if there are other uses of IOUtils in this class, but for copying files in java there is nowadays also the Java nio API with [Files.copy(Path source, Path target, CopyOption... options](https://docs.oracle.com/javase/8/docs/api/java/nio/file/Files.html#copy-java.nio.file.Path-java.nio.file.Path-java.nio.file.CopyOption...-), so we may even be able to reduce dependencies even further.

Have a busy time, at the moment, so only every now and a  quick glance over the open items.

The change as such would also be fine and an improvement over the current code.

aikebah (approved):

PR #6109, opened 2023-11-22
fix: Use the correct key for NVD API-Key from Maven Settings serverId

aikebah:
## Description of Change

Fixup of the Settings-key that gets set for the nvdApiServerId configuration in the maven plugin

no, locally tested using snapshot build


jeremylong (approved):

jeremylong (approved):
Thanks for fixing my copy/paste mistake!

PR #6110, opened 2023-11-22
docs: update NVD notice

### LINKED ISSUE #6105: Does Display NVD API Attribution Notice
As of [version 9.0.0](https://github.com/jeremylong/DependencyCheck/releases/tag/v9.0.0), Dependency-Check is subject to the [Terms of Use of the NVD API](https://nvd.nist.gov/developers/terms-of-use). Regarding attribution, the Terms of Use states the following.

> Services which utilize or access the NVD API are asked to display the following notice prominently within the application: "This product uses the NVD API but is not endorsed or certified by the NVD." You may use the NVD name in order to identify the source of API content subject to these rules. You may not use the NVD name, to imply endorsement of any product, service, or entity, not-for-profit, commercial or otherwise.

While Dependency-Check itself may not be a service that utilizes or accesses the NVD API (I'm not a lawyer), my opinion is that software that makes use of Dependency-Check is most likely considered a service under the Terms of Use. Dependency-Check should automatically display the attribution notice at runtime so that consumers meet the NVD API Terms of Use.

jeremylong:
resolves #6105

### DIFF
--- a/README.md
+++ b/README.md
@@ -320,6 +320,8 @@ Permission to modify and redistribute is granted under the terms of the Apache 2
 
 Dependency-Check makes use of several other open source libraries. Please see the [NOTICE.txt][notices] file for more information.
 
+This product uses the NVD API but is not endorsed or certified by the NVD.
+
 Copyright (c) 2012-2023 Jeremy Long. All Rights Reserved.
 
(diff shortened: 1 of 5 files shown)


A-Fitz-Nelnet:
Thanks @jeremylong!

PR #6112, opened 2023-11-22
fix: Configure nvd.api.delay default

aikebah:
Configure nvd.api.delay to take the coded defaults and raise log in case of actual misconfiguration to WARN level

## Description of Change

Add a default for nvd.api.key with value 0 to use the coded API-key present/not-present defaults and raise the log-level in case of Long parsing issues to WARN level as there was an invalid user-configured api delay

no


jeremylong (approved):
LGTM

jeremylong:
The IT will fail. I'm working on a work around for the release testing.

PR #6131, opened 2023-11-24
chore: Revert "build(deps): bump com.h2database:h2 from 2.1.214 to 2.2.224"

jeremylong:
Reverts jeremylong/DependencyCheck#5940

Unfortunately, due to issue with Gradle the H2 dependency cannot be upgraded. we can reintroduce this PR once https://github.com/gradle/gradle/issues/27156 is resolved.


bmeier-pros:
@jeremylong - given this reversion, is the note about the database breaking changes mentioned in the documentation still relevant?

jeremylong:
While technically it won't **break** - it will basically act as if you've purged due to differences between the data feed and the API you will end up having to re-download the entire NVD dataset.

PR #6154, opened 2023-11-27
fix: correct meta data in report after switching the NVD API

### LINKED ISSUE #6148: Report change in dependencyCheck 9.0.0
I recently upgraded dependencyCheck to version 9.0.0 and noticed a change in the generated report. In previous versions, the report included information about "NVD CVE Checked" and "NVD CVE Modified," but it seems that these details are no longer present in the 9.0.0 version. I would like to understand if this change is due to an incomplete database on my end or if there are other reasons behind it.

DCreports

jeremylong:
resolves #6148


marcelhdl (approved):
LGTM

PR #6177, opened 2023-12-01
fix: remove virtual match string on NVD API Request

jeremylong:
The virtualMatchString is a heavy operation on the API; for performance reasons we are removing this filter


jeremylong:
fixes #6149 (hopefully)

mattbond (approved):

PR #6178, opened 2023-12-01
fix: bump logback-core

jeremylong:
resolves https://github.com/jeremylong/DependencyCheck/security/dependabot/930


cwautogen:
Hi @jeremylong , any idea when the logback-core bump will be merged? Trivy scan is flagging dependency-check because of it.

jeremylong:
1. https://github.com/jeremylong/DependencyCheck/security/dependabot/934 - we do not use the logback receiver component.

Would publishing a VEX explaining the project is not vulnerable help? We have other things to focus on now as bumping logback is going to take some re-working a bit of code due to needing to upgrade logback and slf4j at the same time and dealing with breaking changes.

cwautogen:
No dramas, I can a publish a VEX on this one. Thanks for your prompt response. Cheers.

PR #6215, opened 2023-12-04
fix: exemple 5 with new nvdDatafeedUrl parameter

xaviou:
Old configuration with cveUrlBase and cveUrlModified does not work anymore in 9.x


jeremylong (approved):
LGTM

jeremylong:
Thanks for the PR!

PR #6218, opened 2023-12-04
fix: correct NVD cache last checked logic

### LINKED ISSUE #6209: --nvdValidForHours does not work
**Describe the bug**
The --nvdValidForHours seems to not work as expected. The log says:
`DEBUG - Last NVD API was at 2023-11-28T14:57:24Z, and now 2023-12-04T12:43:31.049266Z[UTC] is within 86400 s.`
But that is wrong, the november 18th  is not withing 24 hours of december 4th.
I had a look at the code and I think I know why. In the `NvdApiDataSource` there is this code:
```         
final Duration duration = Duration.between(now, lastChecked);
final long difference = duration.getSeconds();
proceed = difference > validForSeconds;
```
The `Duration.between` method returns a negative value if the end is before the start. Since it is called with `now` as the first parameter (start), the end is always before the start, so the number is negative. A negative value is never greater than any positive number, so proceed is always false. 
 
The problem occurs using version 9.0.2 of the cli (cli, gradle plugin, maven plugin, etc.)

**Log file**
`DEBUG - Last NVD API was at 2023-11-28T14:57:24Z, and now 2023-12-04T12:43:31.049266Z[UTC] is within 86400 s.`

**To Reproduce**
1. Download an nvd-cache into a folder
2. Run the cli tool with that folder configured
3. Wait a day
4. Update the nvd-cache
5. Run the cli tool with the updated ncd-cache and the --log option to see the debug output
6. Check the debug log for the message that tells you that it doesn't want to check the API because it is within the configured seconds

**Expected behavior**
The parameter `nvdValidForHours` should work as described in the documentation

jeremylong:
resolves #6209

### DIFF
--- a/core/src/main/java/org/owasp/dependencycheck/data/update/NvdApiDataSource.java
+++ b/core/src/main/java/org/owasp/dependencycheck/data/update/NvdApiDataSource.java
@@ -441,7 +441,7 @@ public class NvdApiDataSource implements CachedWebDataSource {
             final ZonedDateTime lastChecked = dbProperties.getTimestamp(DatabaseProperties.NVD_CACHE_LAST_CHECKED);
             if (lastChecked == null) {
                 final ZonedDateTime now = ZonedDateTime.now(ZoneId.of("UTC"));
-                final Duration duration = Duration.between(now, lastChecked);
+                final Duration duration = Duration.between(lastChecked, now);
                 final long difference = duration.getSeconds();
                 proceed = difference > validForSeconds;
                 if (!proceed) {


core/src/main/java/org/owasp/dependencycheck/data/update/NvdApiDataSource.java
```diff
@@ -441,7 +441,7 @@ private boolean checkUpdate() throws UpdateException {
             final ZonedDateTime lastChecked = dbProperties.getTimestamp(DatabaseProperties.NVD_CACHE_LAST_CHECKED);
             if (lastChecked == null) {
```
janosvitok, line 442: This should be lastChecked != null, otherwise this happens (I did not test the fix, just guessing where the problem is):
```
Caused by: java.lang.NullPointerException: Cannot invoke "java.time.temporal.TemporalAccessor.query(java.time.temporal.TemporalQuery)" because "temporal" is null
        at org.owasp.dependencycheck.data.update.NvdApiDataSource.checkUpdate(NvdApiDataSource.java:443)
        at org.owasp.dependencycheck.data.update.NvdApiDataSource.processDatafeed(NvdApiDataSource.java:117)
        at org.owasp.dependencycheck.data.update.NvdApiDataSource.update(NvdApiDataSource.java:108)
        at org.owasp.dependencycheck.Engine.doUpdates(Engine.java:906)
```

jeremylong: doh - thank you for catching this.  i would have been publishing another path release if you hadn't alerted me.  seriously - thank you!

jeremylong: see https://github.com/jeremylong/DependencyCheck/pull/6232

PR #6140, opened 2023-11-26
fix(fp): FP per issue #6138 and #6139

chadlwilson:
- fixes #6138 
- fixes #6139

## Description of Change

The FP automation doesn't work for these Maven Central-stored RubyGems, but seems more sensible to fix these properly since they seem to have started with `9.0.x` where all specific Ruby Gems are being incorrectly detected as rubygems itself.

Merging to the `generatedSuppressions` branch itself to avoid this needing a "release" - similar to other FP suppressions.

## Have test cases been added to cover the new functionality?

no

PR #6220, opened 2023-12-05
docs: Clarify H2 database caching strategy

chadlwilson:
## Description of Change

- Fixes a few typos for `repository` for RetireJS
- Attempts to clarify the intent of the first H2 DB caching strategy
  - ~I understand this is about caching JUST the H2 DB alone, not the entire data directory~
  - Theoretically, I think you could use the single node approach caching either the H2 DB alone OR the entire data directory, so it's possible there needs to be some rewording or re-titling of the strategies here
  - It's possible there are actually three strategies here?
    1. Caching H2 DB alone with single node update
    2. Caching entire data dir with single node update
    2. Caching entire data dir with multiple/all node update (all nodes can upload archive)
    
Is the intent of the two strategies to primarily focus on 
1. single node vs multiple node update, OR
2. ~to focus on whether the H2 DB vs. the entire data dir including cache is archived~ (and other files such as `publishedSuppressions.xml`, `jsrepository.json` etc)

?

I'll help reword/adjust to meet your intention :-)

### DIFF
--- a/src/site/markdown/data/cacheh2.md
+++ b/src/site/markdown/data/cacheh2.md
@@ -2,31 +2,49 @@ Caching ODC's H2 Database
 =========================================
 
 Many users of dependency-check ensure that ODC runs as fast as possible by caching
-the `data` directory (or in some cases just the H2 database). Where the `data`
-directory exists is different for each integration (cli, maven, gradle, etc.).
-However, each integration allows users to configure the location of the data directory.
+the entire `data` directory included the H2 database (`odc.mv.db`). The location of the `data`
+directory is different for each integration (cli, maven, gradle, etc.), however each
+allows users to configure this location.
 
-Within the data directory there is a cache directory that contains temporary caches
+Within the `data` directory there is a `cache` directory that contains temporary caches
 of data requested that is not stored in the database and is generally build specific
-- but can be re-used. There are two primary stratigies used:
+- but can be re-used.
 
-1. Cache the H2 database
+There are two primary strategies used:
+
+1. Single node database updater with multiple node "readers" 
 
 Use a single node to build the database using the integration in update only mode
 (e.g., `--updateOnly` for the cli) and specify the data directory location (see 
-the configuration documentation for each integrgations configuration). The data
-directory is then archived. Subsequent nodes that perform scanning will then
-download the archived database and configure the scan to occur and in general,
-the node would be configured with `--noupdate` (or the releated configuration to
-disable the updates in each configuration). The database is generally updated daily
-in this use case - but could be designed with a more frequent update.
+the configuration documentation for each integration's configuration).
 
-2. Cache the H2 database and the cache
+The `data` directory is then archived somewhere accessible to all nodes.
+Subsequent nodes that perform scanning will download the archived database before 
+scanning. These "reader" nodes would be configured with `--noupdate` (or the related 
+configuration to disable the updates in each integration) so they are not reliant
+on outgoing calls.
+
+The cached `data` directory (and H2 database) is generally updated by the single 
+node/process daily in this use case - but could be designed with a more frequent update.
+
+2. Multiple node database updaters collaborating on a common cache location
 
 Some users have a slightly modified version of the above caching strategy. Instead
 of only having a single update node - they allow all nodes to update. However,
-the data directory is zipped and stored in an common location. Each node will execute
-a scan (with updates enabled) and if succesful the updated data directory is zipped
-and uploaded to the common location. This has the small advantage of being updated
-faster and will store the cache between executions which can improve the performance
-on some builds.
+the entire `data` directory is zipped and stored in a common location, including the H2
+database, `cache`, and in some cases cached data from multiple upstream sources.
+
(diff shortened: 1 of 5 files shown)


jeremylong (approved):
LGTM

jeremylong:
I was trying to focus on single node update vs multiple node updating. If you scan with autoupdate off and you don't cache the retire js JSON file, etc. there could be problems.

chadlwilson:
OK, in that case do we want to specifically recommend that users are best to archive the entire `data` directory (including `cache` and `oss_cache` subfolders) in both single node and multi-node strategies (to keep it simple for them)?

Or are there circumstances in which you would *not* want to include the cache directories (perhaps when they become "build specific" and might actually break ODC in some way?) so we can possibly give users some guidance as to how to make the decision on what to include in the archive. 

Zipping an entire directory is obviously easier than filtering it, unless we can give more specific guidance on what you might need to filter without constraining the future development too much by creating an implicit "contract" on how the `data` dir is used and structured.

jeremylong:
likely just zipping the entire data directory is best. If you use a single update node - there won't be anything in the cache if the only thing it is doing is updating the db...

chadlwilson:
OK, thanks. How does that now read to you? Hope I have covered your intent.

src/site/markdown/data/cacheh2.md
```diff
@@ -2,31 +2,49 @@ Caching ODC's H2 Database
...
+needing to allow all nodes to update the common cache, and thus requiring some degree of
+consistency in how they configure ODC.
+
+Additional Notes
+----------------
+
+The `data` directory may also contain cached data from other upstream sources, dependent 
+on which analyzers are enabled. Ensuring that file modification times are retained during 
+archiving and un-archiving will make these safe to cache, which is especially important in
+a multi-node update strategy.
```
aarongoldenthal, line 50: Is there additional information available on which files are sensitive to timestamps? I'm using a strategy that copies the `data/` folder from the last container build (but updates the timestamps), then updates the ODC data from there, which seems to be working properly, but want to make sure I'm not missing something.

Looking at the `/usr/share/dependency-check/data/` directory in the `owasp/dependency-check` container, for example:

- The `odc.mv.db` database updates even if the timestamp is the current time, so it's not using that timestamp for the 30 minute check.
- The `nvdcache/` directory has cache files by year, but also a `nvdcve-1.1-modified.meta` file that details when that cache was created (it looks like the timestamp of the last full rebuild, not the date on the files or the database itself).
- I assume the timestamp is used to determine whether to update the `jsrepository.json` or `publishedSuppressions.xml` files, but there are CLI arguments to override and always download (e.g. `--retireJsForceUpdate` and `--hostedSuppressionsForceUpdate`), which can be used to force the update for a periodic re-build.
- That leaves the `cache/` directory with files `CENTRAL.data`, `CENTRAL.key`, `NODEAUDIT.data`, `NODEAUDIT.key`, `POM.data`, and `POM.key`, but all have no content.

chadlwilson: This comment I added from my own personal experience, so still to be validated by the maintainers. Thanks for calling it out. Your analysis is consistent with mine.

I found with the 2nd strategy, when I first introduced my own cache restoration the `publishedSuppressions.xml` (which is a local cache of the so-called dynamic suppressions (from [here](https://jeremylong.github.io/DependencyCheck/suppressions/publishedSuppressions.xml)) was not being updated. I inferred that the code was using the file modified time to decide if it needed to re-fetch from the source, which I have just confirmed now from the code at https://github.com/jeremylong/DependencyCheck/blob/4c44019554dbb873d44da14f13a30da35ad16d91/core/src/main/java/org/owasp/dependencycheck/data/update/HostedSuppressionsDataSource.java#L101C9-L116

The `jsrepository.json` also seems to rely on file [lastModified](https://github.com/jeremylong/DependencyCheck/blob/4c44019554dbb873d44da14f13a30da35ad16d91/core/src/main/java/org/owasp/dependencycheck/data/update/RetireJSDataSource.java#L108-L123).

You can see the [Gradle hijinx I had to perform to fix this issue with default Copy task here](https://github.com/gocd/gocd/blob/b7d52977ff856b56333428a19484b4fdbbce9311/build.gradle#L724-L736) and it was a gotcha that would have been easy to miss. Fortunately, I had recently had a suppression added and merged in #6140 so I was expecting it to resolve and when it did not, it exposed the `lastModified` issues. Sigh. [Double sigh](https://github.com/gradle/gradle/issues/1252).

I used loose language because there are a number of sources and didn't want to assume I knew all possible analyzers and data sources, and how new sources in future would behave.

jeremylong: Would it be better if we did not use the last modified time stamp? We could put a properties file along side the suppression file and retireJS json files to track the retrieved times. thoughts?

jeremylong: I'm going to merge this as is so that the documentation gets updated with the release. However, I do plan on revisiting this and expanding the documentation. If storing the last modified date in a separate file would be beneficial - please open an enhancement request and I can make that happen.

jeremylong (approved):
LGTM

PR #6224, opened 2023-12-05
docs: Update list of supported report formats

TobiX:
## Description of Change

This updates all lists of supported report formats in documentation and Javadoc to the content of `ReportGenerator.Formats`. If there are other spurious changes to the lists: I just took one list, added the missing formats and copied that everywhere, overwriting whatever is there. In Javadoc comments where it was reasonable to assume that the user might have a Javdoc browser available, I replaced the list with a link to the enum.

As a future improvement it might be a good idea to define a canonical order and always list reports in that order (maybe the same as the enum?).

Additionally, I moved one log message around, so the list of formats is discoverable by requesting "ALL" formats: Each log line now shows which format is written.

*no*


jeremylong (approved):
LGTM

PR #6225, opened 2023-12-05
fix: check valid for hours for NVD API

### LINKED ISSUE #6222: Where did cveValidForHours go with 9.x? (gradle)
Checked the release notes and could not find that i has been removed. I get

```
 Could not set unknown property 'cveValidForHours' for extension 'dependencyCheck' of type org.owasp.dependencycheck.gradle.extension.DependencyCheckExtension
```

by setting

```
dependencyCheck {
    cveValidForHours = 24
}
```

This used to work in 8.x and as far as i check the docs https://jeremylong.github.io/DependencyCheck/dependency-check-gradle/configuration.html i find 

```
dependencyCheck {
    autoUpdate=false
    cveValidForHours=1
    format='ALL'
}
```

I see there is `nvd.validForHours` and `hostedSuppressions.validForHours` which might be the way to now replace the value above? Could someone clarify? Thank you!

jeremylong:
resolves #6222

### DIFF
--- a/core/src/main/java/org/owasp/dependencycheck/data/update/NvdApiDataSource.java
+++ b/core/src/main/java/org/owasp/dependencycheck/data/update/NvdApiDataSource.java
@@ -272,7 +272,7 @@ public class NvdApiDataSource implements CachedWebDataSource {
             final Duration duration = Duration.between(lastChecked, now);
             final long difference = duration.getSeconds();
             if (difference < validForSeconds) {
-                LOGGER.info("Skipping the NVD API Update as it was completed within the last 30 minutes");
+                LOGGER.info("Skipping the NVD API Update as it was completed within the last {} minutes", validForSeconds/60);
                 return false;
             }
         }
--- a/core/src/main/java/org/owasp/dependencycheck/data/update/NvdApiDataSource.java
+++ b/core/src/main/java/org/owasp/dependencycheck/data/update/NvdApiDataSource.java
@@ -264,11 +264,16 @@ public class NvdApiDataSource implements CachedWebDataSource {
 
     private boolean processApi() throws UpdateException {
         final ZonedDateTime lastChecked = dbProperties.getTimestamp(DatabaseProperties.NVD_API_LAST_CHECKED);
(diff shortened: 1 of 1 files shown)


core/src/main/java/org/owasp/dependencycheck/data/update/NvdApiDataSource.java
```diff
@@ -264,11 +264,16 @@ private void processDownload(Future<Future<NvdApiProcessor>> future, final Set<F
...
-            if (thirtyMinutesAgo.isBefore(lastChecked)) {
+        final int validForHours = settings.getInt(Settings.KEYS.NVD_API_VALID_FOR_HOURS, 0);
+        if (cveDb.dataExists() && lastChecked != null && validForHours>0) {
+            // ms Valid = valid (hours) x 60 min/hour x 60 sec/min x 1000 ms/sec
+            final long validForSeconds = validForHours * 60L * 60L;
+            final ZonedDateTime now = ZonedDateTime.now(ZoneId.of("UTC"));
+            final Duration duration = Duration.between(lastChecked, now);
+            final long difference = duration.getSeconds();
+            if (difference < validForSeconds) {
                 LOGGER.info("Skipping the NVD API Update as it was completed within the last 30 minutes");
```
strangelookingnerd: @jeremylong Isn't the log output misleading here? It should not say 30 minutes, should it?

strangelookingnerd (changes requested):

core/src/main/java/org/owasp/dependencycheck/data/update/NvdApiDataSource.java
jeremylong:
```suggestion
                LOGGER.info("Skipping the NVD API Update as it was completed within the last {} minutes", validForSeconds/60);
```

jeremylong: thanks for pointing this out!

strangelookingnerd (approved):

strangelookingnerd (approved):

PR #6285, opened 2023-12-12
fix: set correct property for nonProxyHosts

weyhmueller:
## Description of Change

While there are different system properties for http and https proxies, there is only one shared property for proxy exclusions: `http.nonProxyHosts`

Setting `https.nonProxyHosts` has no effect on any recent JDK

*no*


jeremylong (approved):

jeremylong:
Thank you for catching my copy/paste mistake. I knew the non-proxy host property is shared...

PR #6298, opened 2023-12-14
fix: Examples using cveValidForHours which no longer exists fixes #6222

### LINKED ISSUE #6222: Where did cveValidForHours go with 9.x? (gradle)

EugenMayer:


jeremylong (approved):
Thanks!

PR #6299, opened 2023-12-14
fix: more efficient memory usage

jeremylong:


jeremylong:
We can chaulk the codeql failures up to codeql being dumb.

PR #6307, opened 2023-12-15
fix: mask nvd.api.key in logs

jeremylong:
resolves GHSA-qqhq-8r2c-c3f5

### DIFF
--- a/core/src/main/resources/dependencycheck.properties
+++ b/core/src/main/resources/dependencycheck.properties
@@ -27,7 +27,7 @@ data.version=5.4
 odc.analysis.timeout=180
 
 # define which settings are masked when logged
-odc.settings.mask=.*password.*,.*token.*
+odc.settings.mask=.*password.*,.*token.*,.*api.key.*
 
 data.connection_string=jdbc:h2:file:%s;AUTOCOMMIT=ON;CACHE_SIZE=65536;RETENTION_TIME=1000;MAX_COMPACT_TIME=10000;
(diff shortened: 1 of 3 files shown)


hott-box:
[GHSA-qqhq-8r2c-c3f5](https://github.com/jeremylong/DependencyCheck/security/advisories/GHSA-qqhq-8r2c-c3f5) isn't resolved; 9.0.6 still logs nvdApiKey; see the advisory for more details.

jeremylong:
@hott-box the documentation is being updated indicating that if specific configuration options are used maven debug logging could expose them. See https://github.com/jeremylong/DependencyCheck/pull/6315.

This is how maven works - the configuration is presented in the debug output.

jeremylong:
@hott-box if you have put the credential clear text in the build file - it is already exposed. Allowing maven to write this back to the debug logging is not making it any less exposed.

zodac-personal:
@jeremylong , FYI, while I don't really see this as a huge issue myself, I am passing in the API credential through environment variables, so it is possible for it to be exposed only through the debug logs.

jeremylong:
@zodac it all depends on how you are passing it in via env variable - and whether or not you are using a CI env that will mask the secrets printed to the console. GH Secrets as an ENV variable would likely mask the possible exposure; the same with Jenkins withCredential.

jeremylong:
and I 100% agree - this is not a very impactful issue if the secret exposure occurs.

PR #6316, opened 2023-12-16
fix: skip pyproject.toml unless it contains `tool.poetry`

### LINKED ISSUE #6303: pyproject.toml should not automatically activate poetry scanner
**Describe the bug**
The scanner incorrectly activates the poetry analyzer in the presence of a pyproject.toml file, regardless of its contents. While poetry does indeed use this file, so do a number of other tools, including setuptools and pytest. The scanner should only interpret a pyproject.toml as being a poetry file if it includes a `[tool.poetry]` TOML section.

Docker image docker.io/owasp/dependency-check:9.0.4

**Log file**
Output excerpt follows
```
[WARN] An error occurred while analyzing '/tmp/dctempc399499e-0516-4ec5-92a8-f494256212a5/check11208990173254915996tmp/2/redacted/pyproject.toml' (Poetry Analyzer).
...
[ERROR] Python `pyproject.toml` found and there is not a `poetry.lock` or `requirements.txt` - analysis will be incomplete
```

**To Reproduce**
Example non-poetry pyproject.toml (setuptools)
```
[project]
name = "redacted"
version = "0.1.0"
description = ""
authors = [
    {name = "redacted", email = "redacted"},
]
license = {text = "Proprietary"}
readme = "README.md"

[project.optional-dependencies]
test = [
    "pytest",
]
```

**Expected behavior**
This file not to be treated as poetry, and for the poetry analyzer to not activate.

**Additional context**
See [here](https://python-poetry.org/docs/pyproject/) for details on the `[tool.poetry]` toml section.

See [here](https://setuptools.pypa.io/en/latest/userguide/pyproject_config.html) and [here](https://docs.pytest.org/en/7.3.x/reference/customize.html#pyproject-toml) for information on an example of tools other than poetry using this file.

A possible fix would be [here](https://github.com/jeremylong/DependencyCheck/blob/15e641309b1973c52c505380414c12d936c0f62b/core/src/main/java/org/owasp/dependencycheck/analyzer/PoetryAnalyzer.java#L150C13-L150C27) to first parse the TOML file, and check if `result.getTables("tool.poetry")` returns non-null, and to disregard otherwise.

I believe this to be the cause of https://github.com/jeremylong/DependencyCheck/issues/5971

jeremylong:
resolves #6303

### DIFF
--- a/core/src/main/java/org/owasp/dependencycheck/analyzer/PoetryAnalyzer.java
+++ b/core/src/main/java/org/owasp/dependencycheck/analyzer/PoetryAnalyzer.java
@@ -155,6 +155,10 @@ public class PoetryAnalyzer extends AbstractFileTypeAnalyzer {
         }
 
         final Toml result = new Toml().read(dependency.getActualFile());
+        if (PYPROJECT_TOML.equals(dependency.getActualFile().getName()) && result.getTables("tool.poetry")==null) {
+            LOGGER.debug("skipping {} as it does not contain `tool.poetry`", dependency.getDisplayFileName());
+            return;
+        }
         final List<Toml> projectsLocks = result.getTables("package");
         if (projectsLocks == null) {
             return;
--- a/core/src/main/java/org/owasp/dependencycheck/analyzer/PoetryAnalyzer.java
+++ b/core/src/main/java/org/owasp/dependencycheck/analyzer/PoetryAnalyzer.java
@@ -155,7 +155,7 @@ public class PoetryAnalyzer extends AbstractFileTypeAnalyzer {
         }
 
         final Toml result = new Toml().read(dependency.getActualFile());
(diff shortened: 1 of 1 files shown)


core/src/main/java/org/owasp/dependencycheck/analyzer/PoetryAnalyzer.java
```diff
@@ -155,6 +155,10 @@ protected void analyzeDependency(Dependency dependency, Engine engine) throws An
         }
 
         final Toml result = new Toml().read(dependency.getActualFile());
+        if (PYPROJECT_TOML.equals(dependency.getActualFile().getName()) && result.getTables("tool.poetry")==null) {
```
jeremylong:
```suggestion
        if (PYPROJECT_TOML.equals(dependency.getActualFile().getName()) && result.getTables("tool.poetry") == null) {
```

PR #6321, opened 2023-12-18
fix: improve memory usage on NVD update

markitovtr1:
## Fixes Issue #

Fix high memory usage when updating NVD database.

Before:
image

After:
image

## Description of Change
Just moved BufferedInputStream to wrap GzipInputStream and not the other way around.

## Test description
I did not test this extensively. I lost some hours finding a way to run this, so sorry if I did not do this as it should.

1. Download NVD cache generated from vulnz and create a local HTTP server to serve it (remove Download latency from equation)
2. Load project via Eclipse
3. Add two run configurations on CLI module pointing to local HTTP server for data feed: one with `--purge` and other with `--updateonly`
4. Monitor memory usage when running `--updateonly` after database is purged

## How I got to that point of error?
On `NvdApiDatasource`, I saw that there were two ExecutorServices when using Datafeed, so first I limited Download to 1 thread and saw no difference in memory usage and then I limited `processingExecutorService` to 1 and saw a huge difference.

## Further improvements
I believe first gunzipping to a temporary file and reading from it would save a lot of memory. Because GzipInputStream has an internal buffer, as Gzip decompression works in blocks, and JSON usually has a high compression rate (around 80 to 95% in some cases, but in this case I saw around 90%), there is a lot of gunzipped content in memory probably. Not sure if BufferedInputStream is enough though, so I did not bother to test that.

No

### DIFF
--- a/core/src/main/java/org/owasp/dependencycheck/data/update/nvd/api/CveApiJson20CveItemSource.java
+++ b/core/src/main/java/org/owasp/dependencycheck/data/update/nvd/api/CveApiJson20CveItemSource.java
@@ -44,14 +44,14 @@ public class CveApiJson20CveItemSource implements CveItemSource<DefCveItem> {
         mapper = new ObjectMapper();
         mapper.registerModule(new JavaTimeModule());
         inputStream = jsonFile.getName().endsWith(".gz") ?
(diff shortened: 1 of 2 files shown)


jeremylong (approved):
LGTM

PR #6359, opened 2024-01-04
docs: Document default for CLI --data

marcelstoer:
## Fixes Issue #

n/a

## Description of Change

It took me a while to figure out what the default data dir for the CLI app is.

n/a


jeremylong:
This is not always correct.  The path to the data directory is relative to the location of the install directory:

https://github.com/jeremylong/DependencyCheck/blob/361e255562af79fc5326151e9721d9ee3029bf6d/cli/src/main/java/org/owasp/dependencycheck/App.java#L451-L463C73

https://github.com/jeremylong/DependencyCheck/blob/361e255562af79fc5326151e9721d9ee3029bf6d/cli/src/main/conf/unixBinTemplate#L18

I could unzip the CLI anywhere on the system - it will have:

```
bin/dependency-check.sh
```

When executed the data directory would be generated right next to the bin directory:

```
bin/dependency-check.sh
data/
```

marcelstoer:
Hhmm...I see. The fact that my data dir is `/usr/local/var/dependencycheck` is due to the fact that I installed it with brew?

```
❯ which dependency-check
/usr/local/bin/dependency-check
❯ ls -al /usr/local/bin/dependency-check
lrwxr-xr-x  1 marcelstoer  admin  53 Jan  4 15:22 /usr/local/bin/dependency-check -> ../Cellar/dependency-check/9.0.7/bin/dependency-check
```

jeremylong:
Correct

marcelstoer:
Shall I update the PR and try to summarize what I learned through these comments? Or did you not document the defaults on purpose in which case I'll close it.

jeremylong:
Please, update the PR. The documentation can always improve and with 9.x
more people will lokely need to figure out where the default location is.

jeremylong (approved):

jeremylong:
Thanks!

Issue #6374, opened 2024-01-09
non-empty dctemp directory does not get deleted


sellersj:
**Describe the bug**
Pretty sure that the change from moving from commons-io to pure java made it so the temp directories do not get deleted in FileUtils.delete(file)
https://github.com/jeremylong/DependencyCheck/pull/6343/files#diff-85b91249daadcc4a913e9b171dc8095e85c76e359d182046f8df56220df78278
This cause out an out disk space on my build box where I had 100-250 GB of free space. Each maven module creates about 135 MB of space.

**Version of dependency-check used**
The problem occurs using version 9.0.8 of the maven plugin. The issue did not exist in 9.0.7

**Log file**
When reporting errors, 99% of the time log file output is required. Please post the log file as a [gist](https://gist.github.com/) and provide a link in the new issue.
n/a in this case I think

**To Reproduce**
1. Run a build with the maven plugin
2. Go to the temp directory
3. See it filled with folders that start with "dctemp"

**Expected behavior**
That the temp files are correctly cleared out

**Additional context**
n/a

sellersj:
I rolled back the change. The temp space had taken up about 160G of space.

I think that the fix would be change the delete call to something like this
```
Files.walk(file.toPath())
  .sorted(Comparator.reverseOrder())
  .map(Path::toFile)
  .forEach(File::delete);
```

sellersj:
I tested 9.0.9-SNAPSHOT with a couple of hundred projects on a CI server and the disk usage didn't change much.

 I can't tie this issue to a milestone, but that's where it was fixed.
 
 Closing ticket.

PR #6375, opened 2024-01-09
fix: for #6374 to delete non-empty directories

sellersj:
## Description of Change
Walk the files to delete using just java.

yes

### DIFF
--- a/utils/src/main/java/org/owasp/dependencycheck/utils/FileUtils.java
+++ b/utils/src/main/java/org/owasp/dependencycheck/utils/FileUtils.java
@@ -25,15 +25,15 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.net.URL;
 import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.Comparator;
 import java.util.UUID;
-
-import org.jetbrains.annotations.NotNull;
-import org.jetbrains.annotations.Nullable;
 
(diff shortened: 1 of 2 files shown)


jeremylong (approved):
LGTM

jeremylong:
Thank you for reporting and **fixing** the issue. Really appreciate this.

utils/src/main/java/org/owasp/dependencycheck/utils/FileUtils.java
```diff
@@ -91,7 +91,10 @@ public static boolean delete(@Nullable File file) {
         }
 
         try {
-            Files.delete(file.toPath());
+            Files.walk(file.toPath())
```
christian-ertl, line 94: The stream returned by Files.walk should be closed.

jeremylong: Interesting... most streams don't need to be closed like this; as documented in the Stream javadoc. Hadn't thought about this possibly being tied to an IO channel. Turns out this isn't uncommon: https://sourcegraph.com/search?q=context:global+content:%22try+%28Stream%3CPath%3E+paths+%3D+Files.walk%28%22+lang:java&patternType=standard&sm=1&groupBy=repo

Fixed with https://github.com/jeremylong/DependencyCheck/pull/6382

Thanks!

PR #6384, opened 2024-01-11
docs: fix links to releases in CHANGELOG.md

herrlock:
This fixes the links to the releases in CHANGELOG.md
The releases from 9.0.1 linked to the user `jereong` instead of `jeremylong`, which probably was not intentional.


jeremylong (approved):

jeremylong:
Thank you - not sure. how I messed that up.

PR #6454, opened 2024-02-08
feat: Allow to pass NVD API key via environment variable

### LINKED ISSUE #6443: Maven Plugin: Allow to read NVD API Key from environment variable
Currently the NVD API key can be passed as direct plugin parameter [`nvdApiKey`](https://jeremylong.github.io/DependencyCheck/dependency-check-maven/check-mojo.html#nvdApiKey) or as server id to be extracted from the `settings.xml` ([`nvdApiServerId`](https://jeremylong.github.io/DependencyCheck/dependency-check-maven/check-mojo.html#nvdApiServerId)). The former has the disadvantage that it may be appear in the Maven logs and the latter has the disadvantage that it requires you to duplicate the credentials for CI/CD (where usually the settings.xml is ephemeral and therefore the canonical storage is somewhere else).

As environment variables are the common means for passing secrets (at least for container environments like e.g. used in GitHub Actions) it should also be supported to extract the NVD API key from a configured environment variable.

kwin:
Clarify precedence of API key plugin parameters in javadoc This closes #6443

no

### DIFF
--- a/maven/src/main/java/org/owasp/dependencycheck/maven/BaseDependencyCheckMojo.java
+++ b/maven/src/main/java/org/owasp/dependencycheck/maven/BaseDependencyCheckMojo.java
@@ -910,7 +910,9 @@ public abstract class BaseDependencyCheckMojo extends AbstractMojo implements Ma
     @Parameter(property = "serverId")
     private String serverId;
     /**
-     * The NVD API Key. The `nvdApiServerId` should be used instead otherwise maven debug logging could expose the API Key. See GHSA-qqhq-8r2c-c3f5.
+     * The NVD API Key. The parameters {@link #nvdApiKeyEnvironmentVariable} or {@link #nvdApiServerId} should be used instead otherwise 
+     * Maven debug logging could expose the API Key. See GHSA-qqhq-8r2c-c3f5.
+     * This takes precedence over {@link #nvdApiServerId} and {@link #nvdApiKeyEnvironmentVariable}.
      */
     @SuppressWarnings("CanBeFinal")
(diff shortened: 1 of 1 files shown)


jeremylong:
It would likely be good to include a default env variable name - and if it exists use it.

kwin:
@jeremylong Do you have a good idea for a default environment variable name?

The problem is https://github.com/eclipse/sisu.plexus/issues/29 here, i.e. it is not easy to invalidate the default value. In order to deliberately not use the environment variable name with the default value one would need to provide an invalid environment name (as empty strings do not overwrite the default value). Also with the default value it is no longer clear what the intention of the user is (i.e. when to emit warnings/error in case of unassigned environment variables).

Given all that I would rather not give a default value.

jeremylong:
Regarding the default environment variable name - why would it be a problem if there was a default? Do you think random machines using ODC would have an environment variable: `NVD_API_KEY`  setup that isn't an NVD API KEY?

kwin:
> Do you think random machines using ODC would have an environment variable: NVD_API_KEY setup that isn't an NVD API KEY?

Unlikely but may happen. My concern is more that it becomes impossible to emit a proper warning in case something is not setup as it should be. With this PR the following sources are considered
1. explicit API key
2. API key from environment variable
3. API key in `settings.xml` in configured server id

Each option is only considered if the accoding plugin parameter is set.
Once you give 2 a default, that would always need to be considered and it is unclear under which circumstances a warning should be emitted if the API key cannot be found somewhere and what message it should contain...

jeremylong (approved):

PR #6484, opened 2024-02-23
docs: document the default data directory

chkpnt:
## Fixes Issue #

Answers the question "Where is the database located?".

## Description of Change

Update the docs corresponding to [DataExtension.groovy](https://github.com/dependency-check/dependency-check-gradle/blob/main/src/main/groovy/org/owasp/dependencycheck/gradle/extension/DataExtension.groovy#L30). But shouldn't it be "9.0" nowadays?

There is no new functionality, so no.


src/site/markdown/dependency-check-gradle/configuration-update.md
```diff
@@ -70,7 +70,7 @@ nvd          | datafeedUrl       | The URL for the NVD API Data feed that can be
 nvd          | datafeedUser      | Credentials used for basic authentication for the NVD API Data feed.                                         | &nbsp;                                                              |
 nvd          | datafeedPassword  | Credentials used for basic authentication for the NVD API Data feed.                                         | &nbsp;                                                              |
 nvd          | validForHours     | The number of hours to wait before checking for new updates from the NVD. The default is 4 hours.            | 4                                                                   |
-data         | directory         | Sets the data directory to hold SQL CVEs contents. This should generally not be changed.                     | &nbsp;                                                              |
+data         | directory         | Sets the data directory to hold SQL CVEs contents. This should generally not be changed.                     | ~/.gradle/dependency-check-data/7.0                                 |
```
jeremylong: The 7.0 could change on version updates (generally major version updates). I'd rather not have it listed here in case I miss updating it in multiple places. Just pointing to this location should be good enough for most.

```suggestion
data         | directory         | Sets the data directory to hold SQL CVEs contents. This should generally not be changed.                     | ~/.gradle/dependency-check-data/                                   |
```

jeremylong (changes requested):

chkpnt:
~~wip... I've to squash.~~ done

jeremylong (approved):

Issue #6527, opened 2024-03-18
Support maven_install.json v2


efabens:
**Is your feature request related to a problem? Please describe.**
As of version [5.1](https://github.com/bazelbuild/rules_jvm_external/releases/tag/5.1) of `rules_jvm_external`, the `maven_install.json` schema was updated for `v2`. The current implementation in dependency check is compatible with `v0.1.0` but `v2`.

**Describe the solution you'd like**
The `PinnedMavenInstallAnalyzer` should support both `maven_install.json` versions.

**Describe alternatives you've considered**
- Roll back the version of `rules_jvm_external` to pre 5.1. That version is over a year old and a major version behind.
- Do container scanning later in the CI process, which would be slower and potentially less reliable

**Additional context**
Bazel continues to grow in popularity, it seems important to support updates.

efabens:
@jeremylong Thanks for getting that PR merged in so quickly! I see it is marked as part of `9.1.0` release. Any sense of when that might be coming down the pike. I am weighing cost/benefit of waiting for the official release vs building my own image pipeline and then using the official image post release.

Ball park of days vs weeks vs months is totally sufficient. 

Thanks for managing this project, it is super appreciated!

jeremylong:
likely within 2 weeks..

PR #6528, opened 2024-03-18
feat: Add v2 support for maven_install.json

efabens:
## Fixes Issue #

https://github.com/jeremylong/DependencyCheck/issues/6527

## Description of Change

- Implement v2 schema support for `maven_install.json`.
- Maintains backwards compatibility with v0.1.0

yes
Added a v2 testfile from Bazel's mainline root directory. It is large (but smaller than the ones in the repos I am currently working with), I would rather not manually construct a smaller one but can if there is a strong preference.


jeremylong (approved):
LGTM

Issue #5971, opened 2023-10-02
Exception thrown when no `poetry.lock` in other site-package library with `pyproject.toml` not in root project directory


ashoofly:
**Describe the bug**
OWASP dependency-check version 8.4.0 fails with exception "Python `pyproject.toml` found and there is not a `poetry.lock` or `requirements.txt` - analysis will be incomplete" when there is `pyproject.toml` in a site-package library. In this case, it was the `ddtrace` library.

**Version of dependency-check used**
The problem occurs using version 1.1.0 of the Github Action: https://github.com/marketplace/actions/dependency-check,
which uses version 8.4.0 of dependency check. 

**Log file**
```
[INFO] Analysis Started
[INFO] Finished Archive Analyzer (0 seconds)
[INFO] Finished File Name Analyzer (0 seconds)
[INFO] Finished Assembly Analyzer (1 seconds)
[INFO] Finished Python Distribution Analyzer (0 seconds)
[INFO] Finished Python Package Analyzer (6 seconds)
[WARN] An error occurred while analyzing '/github/workspace/.venv/lib/python3.11/site-packages/ddtrace/appsec/iast/_taint_tracking/_vendor/pybind11/tools/pyproject.toml' (Poetry Analyzer).
[INFO] Finished Poetry Analyzer (0 seconds)
[INFO] Finished Autoconf Analyzer (0 seconds)
[INFO] Finished CMake Analyzer (0 seconds)
[INFO] Finished Ruby Gemspec Analyzer (0 seconds)
[INFO] Finished Ruby Bundler Analyzer (0 seconds)
[INFO] Finished PE Analyzer (0 seconds)
[INFO] Finished Dependency Merging Analyzer (0 seconds)
[INFO] Finished Hint Analyzer (0 seconds)
[INFO] Finished Version Filter Analyzer (0 seconds)
[INFO] Created CPE Index (2 seconds)
[INFO] Finished NPM CPE Analyzer (2 seconds)
[INFO] Created CPE Index (2 seconds)
[INFO] Finished CPE Analyzer (4 seconds)
[INFO] Finished False Positive Analyzer (0 seconds)
[INFO] Finished NVD CVE Analyzer (0 seconds)
[INFO] Finished RetireJS Analyzer (0 seconds)
[INFO] Finished Sonatype OSS Index Analyzer (2 seconds)
[INFO] Finished Vulnerability Suppression Analyzer (0 seconds)
[...]

PR #6681, opened 2024-05-22
fix: skip pyproject.toml unless it contains `tool.poetry` before ensuring lockfiles

maxgip:
## Description of Change

Even after [this PR](https://github.com/jeremylong/DependencyCheck/pull/6316) was merged I am encountering the issue found in #5971. When scanning a dependency with a `pyproject.toml` file that does not use Poetry, an exception is still thrown.

The changes in this PR check if "tool.poetry" exists in `pyproject.toml` _before_ checking that a lockfile (poetry.lock, requirements.txt) exist so that we can return gracefully if Poetry is not present instead of throwing an exception.

yes, but am open to adding more if needed


jeremylong (approved):

jeremylong:
Thanks for the PR!

Issue #6437, opened 2024-02-01
Dependency check does not scan the package-lock.json for Typescript projects even if it says it does when the node_modules haven't been installed.


IBolken:
**Describe the bug**
I run dependency check with `sh dependency-check.sh --scan "/home/haxxor/dep-test/package-lock.json"`

When I look at what gets printed in my terminal I see this

> [WARN] Analyzing `/home/haxxor/dep-test/package-lock.json` - however, the node_modules directory does not exist. Please run `npm install` prior to running dependency-check

When I then look at my report I find that it has scanned 0 dependencies.

Now when I install all my node_modules and run the exact same command I find that it scans 1399 dependencies of which 419 are unique. 

**Version of dependency-check used**
CLI 9.0.9

**To Reproduce**
1. Clone a Typescript project
2. Run the tool and specify the package-lock.json
3. Install the node_modules and run the tool again specifying the package-lock.json again
4. Compare both outputs

**Expected behavior**
Dependency check should scan dependencies even if the node_modules aren't installed.

almaz045:
We faced the same problem. I thought that it was enough for DC to see package.json and package-lock.json as described in the documentation, but it turns out that it needs node_modules. But nobody commits node_modules to the repository, for this they came up with package-lock.json

jeremylong:
Sorry for the delayed response - there are two nodejs analyzers. One should be disabled as it either needs to be thrown out or completely re-written. The warning message is from that analyzer. You can simply run:

```
sh dependency-check.sh --disableNodeJS --scan "/home/haxxor/dep-test/package-lock.json"
```

Yes - the option say nodeJS, but it is disabling the node package analyzer. The node audit analyzer will still run. Note that the node audit analyzer is equivalent to just running `node audit`.

s1nke:
@jeremylong How should these settings be configured in maven plugin correctly so it works with only providing package-lock.json? I would want that node packages are scanned too.

mik-the-deutsch-dev:
> You can simply run:
> 
> ```
> sh dependency-check.sh --disableNodeJS --scan "/home/haxxor/dep-test/package-lock.json"
> ```
> 

Unfortunately, while the warning is gone, 0 dependencies are detected. I attached a sample project in ZIP: 
[dep-check-js-test.zip](https://github.com/jeremylong/DependencyCheck/files/15106991/dep-check-js-test.zip)

[...]

PR #6690, opened 2024-05-27
fix: change request if lockfile is file v3

spielerf:
## Fixes Issue #
 #6437

## Description of Change
changed request to npm, when package-lock.json used in V3. The "dependencies" block seems to be ignored from npm api, so changed block back to "requires" like in V2. 

No


jeremylong (approved):
LGTM

Issue #6743, opened 2024-06-28
[FP]: junit jupiter being falsely matched as CPE `1e:platform`


chadlwilson:
### Package URl

pkg:maven/org.junit.jupiter/junit-jupiter-engine@5.10.3

### CPE

`cpe:2.3:a:1e:platform:5.10.3:*:*:*:*:*:*:*`

### CVE

CVE-2023-45161, CVE-2023-45163, CVE-2023-5964

### ODC Integration

{"label"=>"Gradle Plugin"}

### ODC Version

9.2.0

### Description

1E Platform is something completely different: https://www.1e.com/platform/

There are actually three jars that are being detected as FPs now:

```
pkg:maven/org.junit.jupiter/junit-jupiter-engine@5.10.3
pkg:maven/org.junit.platform/junit-platform-commons@1.10.3
pkg:maven/org.junit.platform/junit-platform-engine@1.10.3 
```

The FPs weren't reporting with junit `5.10.2` which implies the version number is part of the heuristic for CPE matching here :-(

Could submit 3 different FPs for the automation, but not sure if that's the best way forward here?

chadlwilson:
Perhaps this would be better here.

```xml
  <suppress>
    <notes><![CDATA[
    FP per issue #6743
    ]]></notes>
    <packageUrl regex="true">^pkg:maven/org\.junit\..*/junit-.*@.*$</packageUrl>
    <cpe>cpe:/a:1e:platform</cpe>
  </suppress>
```

aikebah:
Resolved via the PR, followed by FP publishing pipeline of another FP

PR #6744, opened 2024-06-28
fix(fp): FP per issue #6743

chadlwilson:
## Description of Change

Adds more consolidated multi-package fix for the same false positive on https://github.com/jeremylong/DependencyCheck/issues/6743 which affects multiple JUnit artifacts.

no


aikebah:
Thanks for the PR... now I only need to find another FP to trigger the publishing of this suppression

Issue #2046, opened 2019-07-04
Suppression doesn't work with gradle plugin


emartynov:
**Describe the bug**
I updated the gradle plugin to the latest. Here is our configuration:

```
  apply plugin: 'org.owasp.dependencycheck'

  dependencyCheck {
    // Fail the build when there is a finding with "medium" severity.
    failBuildOnCVSS = 4
    // Only check the compile classpath and ignore findings in other configurations.
    scanConfigurations = ['debugCompileClasspath', 'acceptanceCompileClasspath', 'productionCompileClasspath']
    suppressionFiles = ["${rootDir}/extra/owasp.suppression.xml"]
  }
```

After the first run, there are some findings that are false positive for our case. I used html `suppress` button to generate XML and added to owasp suppression:

```
  <suppress>
    <notes><![CDATA[
   Suppress jackson that is dependency for Safened.
   It is not used since Safened communication are happening from our BE
   ]]></notes>
    <packageUrl regex="true">^pkg:maven/com\.fasterxml\.jackson\.core/jackson-databind@.*$
    </packageUrl>
    <cve>CVE-2019-12814</cve>
  </suppress>

  <suppress>
    <notes><![CDATA[
   Suppress jackson that is dependency for Safened.
   It is not used since Safened communication are happening from our BE
   ]]></notes>
    <packageUrl regex="true">^pkg:maven/com\.fasterxml\.jackson\.core/jackson-databind@.*$
    </packageUrl>
    <cpe>cpe:/a:fasterxml:jackson</cpe>
  </suppress>

  <suppress>
    <notes><![CDATA[
   Suppress jackson that is dependency for Safened.
   It is not used since Safened communication are happening from our BE
   ]]></notes>
    <packageUrl regex="true">^pkg:maven/com\.fasterxml\.jackson\.core/jackson-databind@.*$
    </packageUrl>
    <cpe>cpe:/a:fasterxml:jackson-databind</cpe>
  </suppress>
```

Next run is still generating these errors:

```
> Task :common-category:dependencyCheckAggregate
Generating report for project common-category
Found 1 vulnerabilities in project common-category


One or more dependencies were identified with known vulnerabilities in common-category:

[...]

PR #6749, opened 2024-06-30
fix: avoid escaping unnecessary chars in HTML report suppression regexes

chadlwilson:
## Description of Change

Bit of a nitpick: when the HTML report generates an example suppression, it unnecessarily escapes hyphens when constructing a regex inside HTML.

e.g. `pkg:maven/org.eclipse.jetty.websocket/websocket-jetty-server`

becomes

```xml
   <packageUrl regex="true">^pkg:maven/org\.eclipse\.jetty\.websocket/websocket\-jetty\-server@.*$</packageUrl>
```

This is a little more annoying to read and maintain than it needs to be :-) This would be better:

```xml
   <packageUrl regex="true">^pkg:maven/org\.eclipse\.jetty\.websocket/websocket-jetty-server@.*$</packageUrl>
```

Within regexes, `-` only needs to be escaped within a character class, and only if not the first character in the character class (ironically the old code relied on this). Since we are not putting elements into character classes and the the relevant `[` is escaped we don't need to escape this, and it reads better to not escape.

Similar
- `#` has no special meaning in regex, not sure why it was being escaped
- `,` has no special meaning outside a repetition qualifier (already `{}` are escaped), so similarly does not need to be escaped

Technically the escaping is a bit janky right now as it is also escaping some things for the purpose of using JavaScript to construct XML so _technically_ it needs to `escapeForXml` afterwards, or the `packageUrl` = `pkg:hello-world-</packageUrl>` will get you into broken XML trouble, but I guess that is not a major concern :-)

I guess escaping the whitespace escaping with `\s` is one of these XML-related workarounds based on #2046 so have left it around.

no

### DIFF
--- a/core/src/main/resources/templates/htmlReport.vsl
+++ b/core/src/main/resources/templates/htmlReport.vsl
@@ -152,7 +152,7 @@ Copyright (c) 2012 Jeremy Long. All Rights Reserved.
                 }
             }
             function escapeRegExp(text) {
-              return text.replace(/[-[\]{}()*+?.,\\^$|#\s]/g, '\\$&');
+              return text.replace(/[[\]{}()*+?.\\^$|\s]/g, '\\$&');
             }
             function setCopyText(name, matchType, matchValue, suppressType, suppressVal) {
                 xml =  '<suppress>\n';


jeremylong (approved):
lgtm

Issue #6812, opened 2024-07-06
[FP]: Eclipse jetty toolchain libs should also be excluded from eclipse:jetty cpe


lread:
### Package URl

pkg:maven/org.eclipse.jetty.toolchain/jetty-jakarta-websocket-api@2.0.0

### CPE

`cpe:2.3:a:eclipse:jetty:2.0.0:*:*:*:*:*:*:*`

### ODC Integration

None

### ODC Version

10.0.2

### Description

I listed a specific package URI and cpe because this is what I understand the issue form to require, but I think we should instead tweak an existing suppression.

In #6799, the eclipse jetty tool chain libs were excluded from `cpe:/a:jetty:jetty`, but I think they also need to be excluded from `cpe:/a:eclipse:jetty`.

And with that, hopefully, the game of whack-a-mole will be complete!

chadlwilson:
Thanks! I guess this can be closed now :-)

lread:
Sweet! 

Fixed by #6813

PR #6813, opened 2024-07-06
fix(fp): FP per issue #6812

lread:
## Description of Change

Jetty toolchain lib version is not tied to eclipse/jetty version.

no


lread:
I think this makes sense, but is my first PR to your repo, so please lemme know if you need any adjustments!

aikebah (approved):

aikebah:
Makes sense to be added in this way indeed. Will only become active after we've processed another 'regular' FP report as publishing automation is part of that flow.

PR #6843, opened 2024-07-15
feat: Enable configuration of a lower resultsPerPage on NVD API

aikebah:
## Description of Change

Add configurability of resultsPerPage parameter towards the NVD API to values lower than the default of 2000 in response to indicators that some environments are restricted to lower limits.

no


jeremylong (approved):
LGTM

jeremylong:
In addition to the configuration change, I made the OVC auto-adust if the NVD provides a response saying the resultsPerPAge is too high.

PR #6899, opened 2024-08-15
fix: do not report over 100% download complete

### LINKED ISSUE #6897: NVD API has 8.115 records in this update: 10.000/8.115 (123%)
**Describe the bug**
This isn't an error, but looks strange. After some days I run the check manualy.

```
[INFO] Checking for updates
[INFO] NVD API has 8.115 records in this update
[INFO] Downloaded 10.000/8.115 (123%)
[INFO] Downloaded 8.115/8.115 (100%)
```

```
<groupId>org.owasp</groupId>
<version>10.0.3</version>
```

jeremylong:
resolves #6897

### DIFF
--- a/core/src/main/java/org/owasp/dependencycheck/data/update/NvdApiDataSource.java
+++ b/core/src/main/java/org/owasp/dependencycheck/data/update/NvdApiDataSource.java
@@ -367,7 +367,9 @@ public class NvdApiDataSource implements CachedWebDataSource {
                         if ((ctr % 5) == 0) {
                             //TODO get results per page from the API as it could adjust automatically
                             final double percent = (double) (ctr * resultsPerPage) / max * 100;
-                            LOGGER.info(String.format("Downloaded %,d/%,d (%.0f%%)", ctr * resultsPerPage, max, percent));
+                            if (percent < 100) {
+                                LOGGER.info(String.format("Downloaded %,d/%,d (%.0f%%)", ctr * resultsPerPage, max, percent));
+                            }
                         }
                     }
                     final ZonedDateTime last = api.getLastUpdated();


core/src/main/java/org/owasp/dependencycheck/data/update/NvdApiDataSource.java
```diff
@@ -367,7 +367,9 @@ private boolean processApi() throws UpdateException {
                         if ((ctr % 5) == 0) {
                             //TODO get results per page from the API as it could adjust automatically
                             final double percent = (double) (ctr * resultsPerPage) / max * 100;
-                            LOGGER.info(String.format("Downloaded %,d/%,d (%.0f%%)", ctr * resultsPerPage, max, percent));
+                            if (percent < 100) {
+                                LOGGER.info(String.format("Downloaded %,d/%,d (%.0f%%)", ctr * resultsPerPage, max, percent));
+                            }
```
aikebah, line 372: Maybe still write the completion log entry as well, but use just the actual counts and a hardcoded 100% ?

```suggestion
                            if (percent < 100) {
                                LOGGER.info(String.format("Downloaded %,d/%,d (%.0f%%)", ctr * resultsPerPage, max, percent));
                            } else {
                                LOGGER.info(String.format("Downloaded %,d/%,d (%.0f%%)", max, max, 100));
                            }
```

aikebah (commented):
LGTM, but could be further improved as per my suggestion

core/src/main/java/org/owasp/dependencycheck/data/update/NvdApiDataSource.java
jeremylong: The 100% complete is already written out here: https://github.com/jeremylong/DependencyCheck/pull/6899/files#diff-871e872a0687fe3f7ff7af515affbfc9c90ef55c0322f2e9c3a83602065e1715R398

aikebah: ah right...

PR #6916, opened 2024-08-24
build(deps): exclude unused dependency

### LINKED ISSUE #6911: Also exclude `jcl-over-slf4j` from the `org.sonatype.ossindex:ossindex-service-client` dependency
`org.sonatype.ossindex:ossindex-service-client` depends on `org.apache.httpcomponents:httpclient` but excludes `commons-logging:commons-logging` and instead badly depends on `jcl-over-slf4j` explicitly while not even using the `commons-logging` API itself.

This easily causes problems (most often silently unless you have some protection) if any other dependency (like other dependencies of this project) depend on `commons-logging`: sonatype/ossindex-public#51.

As you explicitly exclude `org.apache.httpcomponents:httpclient` from your `org.sonatype.ossindex:ossindex-service-client` dependency, I suggest you also exclude the `jcl-over-slf4j` dependency as with that exclude it is totally pointless anyway, at least until that upstream bug got fixed and you depend on a newer version with that fix.

jeremylong:
resolves #6911

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -1308,6 +1308,10 @@ Copyright (c) 2012 - Jeremy Long
                         <groupId>com.google.guava</groupId>
                         <artifactId>guava</artifactId>
                     </exclusion>
+                    <exclusion>
+                        <groupId>org.slf4j</groupId>
+                        <artifactId>jcl-over-slf4j</artifactId>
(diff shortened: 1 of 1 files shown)


aikebah (approved):
LGTM

PR #6917, opened 2024-08-25
fix: improve regex

jeremylong:
resolves https://github.com/jeremylong/DependencyCheck/security/code-scanning/14

Improves the regex so it does not use `+` and instead cap the version number at 100 digits.


aikebah (approved):
LGTM

Though 100 digits feels a excessive for a version number, it would do the job of ReDOS prevention and still allow any reasonable previously matching version number to still match the RegEx.

jeremylong:
I agree, 100 is excessive - but I've been burnt in the past by thinking there is no way we would see data of X characters.

PR #6932, opened 2024-08-27
build(deps): Update minimum version of Maven to 3.6.3

### LINKED ISSUE #6928: Require maven 3.6.3 or above for the Maven plugin
The Apache maven project has EOLed Maven 3.6.3 in April 2023 and have their plugin policy updated to have plugins require maven 3.6.3 as the minimal maven version.

https://maven.apache.org/developers/compatibility-plan.html

Currently the OWASP DependencyCheck maven-plugin lists the minimum required version of Maven as 3.1

Update the maven-plugin to also follow the maven policy and abandon supporting compatibility with older Maven versions.

Keeping compatibility in with older maven versions as long as nothing breaks while even the Maven community has moved on to only support maven versions 3.6.3 and later in their native plugin ecosystem.

**Additional context**
Any software developer taking security seriously would've upgraded their maven stack to at least 3.8.1 which ensures by default that http (more susceptible to MITM attacks) may never be used to access artifact repositories, so the target audience of DependencyCheck should typically already be beyond the new minimal maven version (especially since the new minimal maven version is also already an EOLed version)

aikebah:
## Description of Change

Bump the required minimum maven version to use for the dependency-check-maven-plugin to 3.6.3 (in line with the current Apache Maven policy for new plugin releases).
As there is no direct benefit, and a small chance that some users might still be on lower Maven versions where this would constitute a breaking change, I think it's best to await the 11.0.0 snapshots to merge this in.

not applicable

### DIFF
--- a/maven/pom.xml
+++ b/maven/pom.xml
@@ -37,7 +37,7 @@ Copyright (c) 2013 Jeremy Long. All Rights Reserved.
         <tag>v6.4.1</tag>
     </scm>
     <prerequisites>
-        <maven>3.1.0</maven>
+        <maven>3.6.3</maven>
     </prerequisites>
     <build>
         <resources>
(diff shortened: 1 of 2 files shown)


jeremylong (approved):
LGTM

PR #6936, opened 2024-08-30
fix: store timestamps locally for local resources

### LINKED ISSUE #6434: up to date check for file based data sources broken
**Describe the bug**
There is a chance that I do not see the logic behind but what I see:

* the LAST_CHECKED date information of retirejs and hosted suppression is stored in the central database
* the actual `publishedSuppressions.xml` or `jsrepository.json` is stored in the local file system (.m2)

this leads to inconsistent data. Where the database property can state that the file is up to date but locally the copy might be outdated or even none existing.

Also the fact that such an update performs a write operation to the database leads to below warning since the usual scan user (dcuser) does not have write permission to this shared database. But this depends on the setup.

    [WARNING] Unable to save property 'retirejs.checked' with a value of '12345678' to the database

This was introduced in https://github.com/jeremylong/DependencyCheck/issues/6260. I've put a comment on #6399 which is related but not the same.

The problem occurs using version 9.0.7 & 9.0.9ff

**Expected behavior**
Up to date information for local files must be stored locally.

**Additional context**
We use a central (PostgreSQL) database as mirror for the NVD data.  A regular job executes ` org.owasp:dependency-check-maven:9.0.9:update-only` using a privileged database user. The actual scans run independently on distributed systems where the less privileged `dcuser` is used to access the database.

jeremylong:
## Fixes Issue https://github.com/jeremylong/DependencyCheck/issues/6434

Store last updated timestamps locally for local, file-based resources.

### DIFF
--- a/core/src/main/java/org/owasp/dependencycheck/data/update/LocalDataSource.java
+++ b/core/src/main/java/org/owasp/dependencycheck/data/update/LocalDataSource.java
@@ -71,6 +71,9 @@ public abstract class LocalDataSource implements CachedWebDataSource {
                 lastUpdatedOn = Integer.parseInt(props.getProperty("LAST_UPDATED", "0"));
             } catch (IOException | NumberFormatException ex) {
                 LOGGER.debug("error reading timestamp file", ex);
+            }
+            if (lastUpdatedOn <= 0) {
+                //fall back on conversion from file last modified to storing in the db.
                 lastUpdatedOn = repo.lastModified();
             }
         }
--- a/core/src/main/java/org/owasp/dependencycheck/data/update/LocalDataSource.java
+++ b/core/src/main/java/org/owasp/dependencycheck/data/update/LocalDataSource.java
@@ -56,15 +56,12 @@ public abstract class LocalDataSource implements CachedWebDataSource {
 
     /**
      * Retrieves the last updated date from the local file system (in a file
-     * next to the repo file). If this fails, the database properties are
-     * checked to see if the timestamp can be obtained there.
+     * next to the repo file).
      *
      * @param repo the local file data source
-     * @param dbProperties the database properties
-     * @param dbPropertyName the property key for the timestamp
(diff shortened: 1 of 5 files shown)


core/src/main/java/org/owasp/dependencycheck/data/update/LocalDataSource.java
```diff
@@ -0,0 +1,87 @@
...
+                lastUpdatedOn = Integer.parseInt(props.getProperty("LAST_UPDATED", "0"));
+            } catch (IOException | NumberFormatException ex) {
+                LOGGER.debug("error reading timestamp file", ex);
+                lastUpdatedOn = repo.lastModified();
+            }
+        } else {
+            //fall back for conversion from storing in the DB to storing locally
+            if (dbProperties != null) {
+                lastUpdatedOn = dbProperties.getPropertyInSeconds(dbPropertyName);
+            }
```
aikebah: Given its origin (issue on DB has a timestamp, but local file is outdated) I don't think the plugin should do anything with the DB timestamp of the resource. If the properties file is not stored side-by-side with the cached file it should simply assume the file was retrieved too long ago and update it (return 0 for the lastUpdatedOn).

IMO the corresponding db property should be removed-if-present and its value should be ignored (because it's not trusted to validly reflect the update timestamp for the local filesystem cached resource)

jeremylong: I was thinking that this would make the transition easier - but these files are relatively small and easy to re-download. I'll update the code to remove the DB dependency.

aikebah (approved):
LGTM, though (on this second read)... it might be worth considering refactoring the

```
            if (lastUpdatedOn <= 0) {
                //fall back on conversion from file last modified to storing in the db.
                lastUpdatedOn = repo.lastModified();
            }
```
into the getLastUpdated method of the LocalDataSource rather than repeating it in each subclass as a fallback when not getting a lastUpdated back from the call.

PR #6950, opened 2024-09-07
build: Remove the animal-sniffer, propagate java version to plugin-archetype

aikebah:
## Description of Change

As a result of a first PR that was using the Java 11 API I found out that the Java 11 migration PR had missed the removal of the no longer needed animal-sniffer plugin.

Now that we've upgraded to target Java 11 as the minimal baseline and use the javac `--release` option (triggered by the `maven.compiler.release` property) the compiler already takes care of ensuring compliance with the Java public API of the target version.

Also added population of the `maven.compiler.release` property to the pom that gets generated from the dependency-check-plugin archetype.

N/A

### DIFF
--- a/archetype/pom.xml
+++ b/archetype/pom.xml
@@ -38,13 +38,6 @@ Copyright (c) 2017 Jeremy Long. All Rights Reserved.
     <build>
         <plugins>
             <plugin>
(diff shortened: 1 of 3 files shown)


jeremylong (approved):
LGTM

PR #6951, opened 2024-09-08
build: Update Checkstyle configuration and Suppression DTD references

aikebah:
## Description of Change

Discovered that the DTD locations were outdated http locations. Checkstyle project has them now published on checkstyle.org served with https. Also found that there have been some newer suppressions DTDs, so while updating the location and public ID to comply with current Checkstyle documentation also updated the suppressions file to point to the current version (which has been modified to allow more flexible filtering: `files` is no longer required and at least one of `checks` (previously required), `message` (new) or `id` (new) is required to be present as the item to filter)

N/A


jeremylong (approved):
LGTM

PR #7016, opened 2024-10-08
fix(fp): Improve suppression for Glassfish Server false positives

### LINKED ISSUE #7015: [FP]: glassfish project implementations (JAXB etc) mistaken for glassfish server
### Package URl

pkg:maven/org.glassfish.jaxb/jaxb-core@4.0.5

### CPE

`cpe:2.3:a:eclipse:glassfish:4.0.5:*:*:*:*:*:*:*`

### CVE

CVE-2024-9329

### ODC Integration

{"label"=>"Gradle Plugin"}

### ODC Version

10.0.4

### Description

https://nvd.nist.gov/vuln/detail/CVE-2024-9329 has been reported on Glassfish server however is being reported on all manner of projects maintained separately under `org.glassfish.x` groups in Maven Central.

I'll attempt to submit a custom generatedSuppression to turn https://github.com/jeremylong/DependencyCheck/issues/6626 into something more generic. (this CVE had a lower bound affected version so didn't lead to too many false positives)

chadlwilson:
## Fixes Issue #

- fixes #7015

## Description of Change

Changes existing suppression to more generic, to exclude all org.glassfish maven artifacts other than `org.glassfish.main` where the main Glassfish Server is published.

N/A

### DIFF
--- a/generatedSuppressions.xml
+++ b/generatedSuppressions.xml
@@ -1556,9 +1556,9 @@
 </suppress>
 <suppress base="true">
    <notes><![CDATA[
-   FP per issue #6626
+   hand-curated better suppression for FP per issues #6626 and #7015
    ]]></notes>
-   <packageUrl regex="true">^pkg:maven/org\.glassfish\.expressly/expressly@.*$</packageUrl>
+   <packageUrl regex="true">^pkg:maven/org\.glassfish\.(?!main).*$</packageUrl>
    <cpe>cpe:/a:eclipse:glassfish</cpe>
 </suppress>
 <suppress base="true">


jeremylong (approved):

Lars5678:
@chadlwilson Is your RegEx `^pkg:maven/org\.glassfish\.(?!main).*$` correct?

In my case it doesn't recognize `pkg:maven/org.glassfish/jakarta.json@2.0.1`. 
If I change the RegEx to `^pkg:maven/org\.glassfish(?!\.main).*$` , it works. Please validate again!

chadlwilson:
Sorry? It depends how you define "correct" - I didn't attempt to fix "your case".

I did not attempt to suppress artifacts directly under `org.glassfish` as that would require reviewing if older components of the Glassfish server were published there which is not easy to do. Remember it's very important to avoid false negatives too.

It's probably OK, and I'll submit another PR to update it. Perhaps you are new to open source, but would suggest in future interactions that you take a bit of time to understand how to interact with other folks appropriately. I am just a contributor trying to improve things; please don't make requests/demands of other people.

Lars5678:
I'm sorry. I really like that you provided a solution. And it's also clear to me that you didn't want to solve my case with that. I definitely didn't want to come across as negative. If I have expressed myself incorrectly, I am very sorry. I just wanted to note that I have another case that is not covered by your solution and just ask to check if the RegEx may need to be adjusted before rolling it out. If you think it's not correct then I can handle it.

chadlwilson:
Created another PR at #7024. Normally these suppressions are rolled out dynamically without a release, but manual PRs to the `generatedSuppressions` outside the automated workflow don't get released automatically.

Lars5678:
Many thanks!

Issue #7030, opened 2024-10-10
DependencyCheck 10.0.4 tries to lock database directory even when autoUpdate is set to false


guidoschreuder:
**Describe the bug**
Switching from version 10.0.2 to 10.0.4 jobs on Jenkins are getting stuck because DependencyCheck 10.0.4 tries to lock database directory even when autoUpdate is set to false.

The dataDirectory is mounted on a read-only filesystem, as we centrally update this cache to keep the NVD API from causing problems by throttling. This has worked well for us, but now this is broken

**Version of dependency-check used**
10.0.4

**Log file**
/bin/mvn  -X org.owasp:dependency-check-maven:10.0.4:aggregate  -DdataDirectory=/owasp-cache/owasp-api-data -DautoUpdate=false
[snip]
[DEBUG] Settings.getDataFile() - file: '/owasp-cache/owasp-api-data'
[DEBUG] Sleeping thread main (a39cd3c744de8a843d0095cb26149de3) for 15 seconds because an exclusive lock on the database could not be obtained (2024-10-10 14:58:23.759)
[DEBUG] Sleeping thread main (a39cd3c744de8a843d0095cb26149de3) for 15 seconds because an exclusive lock on the database could not be obtained (2024-10-10 14:58:38.761)
[DEBUG] Sleeping thread main (a39cd3c744de8a843d0095cb26149de3) for 15 seconds because an exclusive lock on the database could not be obtained (2024-10-10 14:58:53.762)
[DEBUG] Sleeping thread main (a39cd3c744de8a843d0095cb26149de3) for 15 seconds because an exclusive lock on the database could not be obtained (2024-10-10 14:59:08.763)
[DEBUG] Sleeping thread main (a39cd3c744de8a843d0095cb26149de3) for 15 seconds because an exclusive lock on the database could not be obtained (2024-10-10 14:59:23.765)
[DEBUG] Sleeping thread main (a39cd3c744de8a843d0095cb26149de3) for 15 seconds because an exclusive lock on the database could not be obtained (2024-10-10 14:59:38.774)
[DEBUG] Sleeping thread main (a39cd3c744de8a843d0095cb26149de3) for 15 seconds because an exclusive lock on the database could not be obtained (2024-10-10 14:59:53.775)
[DEBUG] Sleeping thread main (a39cd3c744de8a843d0095cb26149de3) for 15 seconds because an exclusive lock on the database could not be obtained (2024-10-10 15:00:08.777)
[etc]

[...]

PR #7031, opened 2024-10-10
fix: #7030 use temp-dir instead of data-dir for lock files

guidoschreuder:
## Fixes Issue #
#7030 

## Description of Change

Use temp directory for lockfile instead of data directory

no

### DIFF
--- a/core/src/main/java/org/owasp/dependencycheck/utils/WriteLock.java
+++ b/core/src/main/java/org/owasp/dependencycheck/utils/WriteLock.java
@@ -143,63 +143,59 @@ public class WriteLock implements AutoCloseable {
         if (!isLockable) {
             return;
         }
-        try {
-            final File dir = settings.getDataDirectory();
-            lockFile = new File(dir, lockFileName);
-            checkState();
-            int ctr = 0;
-            do {
-                try {
-                    if (!lockFile.exists() && lockFile.createNewFile()) {
-                        file = new RandomAccessFile(lockFile, "rw");
-                        lock = file.getChannel().lock();
-                        file.writeBytes(magic);
-                        file.getChannel().force(true);
-                        Thread.sleep(20);
-                        file.seek(0);
-                        final String current = file.readLine();
-                        if (current != null && !current.equals(magic)) {
-                            lock.close();
-                            lock = null;
-                            LOGGER.debug("Another process obtained a lock first ({})", Thread.currentThread().getName());
-                        } else {
-                            addShutdownHook();
-                            final Timestamp timestamp = new Timestamp(System.currentTimeMillis());
-                            LOGGER.debug("Lock file created ({}) {} @ {}", Thread.currentThread().getName(), magic, timestamp);
-                        }
-                    }
-                } catch (InterruptedException ex) {
-                    Thread.currentThread().interrupt();
-                    LOGGER.trace("Expected error as another thread has likely locked the file", ex);
-                } catch (IOException ex) {
-                    LOGGER.trace("Expected error as another thread has likely locked the file", ex);
-                } finally {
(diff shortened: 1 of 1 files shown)


aikebah:
https://github.com/jeremylong/DependencyCheck/blob/585016a461edd96b2e0a06055266c75a927d5389/utils/src/main/java/org/owasp/dependencycheck/utils/Settings.java#L1269

Your 'fix' is breaking the exclusive locking meant to ensure that 'during update' the files will not be used by another parallel execution of ODC that uses the same datadirectory.

Tempdirectory will be a unique folder per ODC process.

guidoschreuder:
@aikebah thanks for the response, i've converted the PR to draft, since it was kind of low effort, end-of-day and couldn't get the tests running properly locally

But, i see, would it be acceptable to use system temp directory for lock files?

guidoschreuder:
OK, the tests do look better now.

The once that are failing are also failing for me locally on a clean checkout of this repo, so i'm not sure what to do with that

aikebah:
@guidoschreuder system temp folder for the locks would, for the same reasons, typically only work for single system CI environments. Containerized with volume-mounted datafolder and multiserver with network-mounted datafolder would also need the lock folder to be shared with the same granularity so that all instances that share the same datafolder also share the same locks folder.

The only change I can imagine to support your setup while retaining the intended locking semantics is to add a configuration possibility for a separate data-locks folder (that would need to be shared between builds in the exact same granularity as the datafolder itself and that would typically default, if not configured explicitly, to the datafolder itself)

guidoschreuder:
@aikebah thanks for the clarification, i totally get your point

that solution would work for me, so, i could continue on that if that would be accepted

 (just for background) we really would never experience any problems without locking, what we do is, roughly:
* we have a single, exclusive, task that runs periodically
* we maintain multiple versions of the data directory e.g. `/owasp_cache/d1`, `/owasp_cache/d2`, etc.
* before updating we copy the latest directory to a new directory and then run update over that
* when successful we `ln -snf /owasp_cache/cache /owasp_cache/d3`
* and then rotate old directories away

aikebah:
When I spotted your issue in the tracker an approach along those lines would be what I had in mind to resolve it, so if you code a PR along those lines I would be happy to accept it.

Be aware though that there are also the various dynamic caches (unless you disable their caching): for centralAnalyzer, NodeAudit and the OSSIndexAnalyzer.

Depending on the frequency of your builds and the amount of overlap in dependencies across projects you might have a benefit by hosting those on a shared, writeable volume/mount to have the builds profit from a lower number of dependencies to lookup at the remote APIs for those.

These dynamic caches are currently stored (if writeable, otherwise caching is automatically disabled) inside two subfolders (`oss_cache` and `cache`) underneath the data folder; I think these locations are also not configurable at the moment so if the caching would be valuable to you  you might want to extend configurability to also allow these two cache folders  to point to (shared across your builds) writeable locations outside of your read-only mounted data folder.

jeremylong:
Closing the PR as a better approach would be figuring out why a lock is generated when auto-update is disabled.

aikebah:
@jeremylong lock is taken by the engine to guard that the copy-to-a-temp-folder of the H2 database file for readOnly access copies from a stable and consistent state - https://github.com/jeremylong/DependencyCheck/blob/116a63fdba7706cec2cf6a33d8002c9d53a16f8f/core/src/main/java/org/owasp/dependencycheck/Engine.java#L1002

The strategy to copy over to a temp location for read-only acces dates back all the way to version 3 and was introduced by https://github.com/jeremylong/DependencyCheck/commit/04dc5f84912e4cc2f48aced0007e6c4ffecfb278

PR #7033, opened 2024-10-11
fix: Fixup parameter sequence for Dowloader credentials

aikebah:
## Description of Change

While working on (still work in progress) also using HTTPClient for access to Nexus I encountered an authentication issue due to an improper sequence of arguments for some of the tryAdd...Credentials methods that was overlooked in the changes of PR #6949. Two occurrences had the arguments accidentally coded as the sequence urlkey, userkey, passwordkey instead of the userkey, urlkey, passwordkey required by the `addUserPasswordCreds` method

no


jeremylong (approved):
LGTM

PR #7036, opened 2024-10-13
chore: Update test db schema

jeremylong:
Resolves an issue with failing test cases in the Ant task. The numerous schema updates hit a [bug in H2](https://github.com/h2database/h2database/issues/4048#issuecomment-2379953514).

Additionally:
- removes old un-needed upgrade SQL scripts
- removes un-necessary copy-resources execution in the ant task.


aikebah (approved):
LGTM

Issue #7072, opened 2024-10-22
Maven plugin 11.0.0 does not respect nonProxyHosts


stklcode:
**Describe the bug**
After updating the Maven Plugin from 10.0.4 to 11.0.0 the proxy logic seems to be broken.

Scenario:
* internal cache for the NVD CVE datafeed at `https://cache.example.com/nvdcve-{0}.json.gz`
* proxy for external connections that does not forward internal traffic configured in Maven _settings.xml_

So we have to rely on `nonProxyHosts` to properly route the connections.

With 11.0.0 we receive 503 responses from the proxy (the cache is online and available).
Disabling the global proxy makes these calls work again, but `www.cisa.gov` and `jeremylong.github.io` fail with connection errors (as expected without proxy)

With 10.0.4 it works as expected. I suspect some missing link from the HTTP client switch, but I didn't yet look deeper into it.

**Version of dependency-check used**
The problem occurs using version 11.0.0 of the maven plugin.

**To Reproduce**
1. Clear the local database, if any
  ```
  mvn dependency-check:purge
  ```
2. Have a local cache, e.g.
  ```xml
   <nvdDatafeedUrl>https://cache.example.com/nvdcve-{0}.json.gz</nvdDatafeedUrl>
  ```
3. Have a selective proxy configured in maven _settings.xml_
  ```xml
  <proxies>
    <proxy>
      <id>example-proxy</id>
      <active>true</active>
      <protocol>http</protocol>
      <host>proxy.example.com</host>
      <port>8080</port>
      <nonProxyHosts>*.example.com</nonProxyHosts>
    </proxy>
  </proxies>
  ```
4. Run dependency check via Maven
  ```
  mvn dependency-check:check
  ```

Same behavior without the Maven settings and explicit proxy properties

1. Clear the local database, if any
2. Have a local cache
3. Run dependency check via Maven with proxy properties
  ```
  mvn -Dhttp.proxyHost=proxy.example.com \
      -Dhttp.proxyPort=8080 \
      -Dhttp.nonProxyHosts="*.example.com" \
      dependency-check:check
  ```

**Expected behavior**
Local NVD CVE database should be updated from NVD cache, and external CISA resources and the checks run as usual.

**Additional context**
Tested with Java 17.0.12 and 21.0.4 and Apache Maven 3.9.9
No additional proxy-relatved environment variables or properties set.

miniupnp:
I have the same problem...

miniupnp:
[...]

PR #7074, opened 2024-10-22
fix: do not set legacy proxy from maven or env (#7072)

stklcode:
## Fixes Issue #

#7072 

## Description of Change

The Apache HTTPClient based downloader supports http(s).proxy* properties, so we do not need to use legacy logic. In legeacy mode http.nonProxyHosts is not honored, so setting both leads to issues due to missing proxy selectors.

Omit populating legacy properties resolves this issue. In addition we have to move the password decryption from Maven settings up, so it actually works.

no

### DIFF
--- a/maven/src/main/java/org/owasp/dependencycheck/maven/BaseDependencyCheckMojo.java
+++ b/maven/src/main/java/org/owasp/dependencycheck/maven/BaseDependencyCheckMojo.java
@@ -2212,45 +2212,18 @@ public abstract class BaseDependencyCheckMojo extends AbstractMojo implements Ma
                 if (mavenProxy.getUsername() != null && !mavenProxy.getUsername().isEmpty()) {
                     System.setProperty("https.proxyUser", mavenProxy.getUsername());
                 }
-                if (mavenProxy.getPassword() != null && !mavenProxy.getPassword().isEmpty()) {
-                    System.setProperty("https.proxyPassword", mavenProxy.getPassword());
+                String password = mavenProxy.getPassword();
+                if (password != null && !password.isEmpty()) {
+                    try {
+                        password = decryptPasswordFromSettings(password);
+                    } catch (SecDispatcherException ex) {
+                        password = handleSecDispatcherException("proxy", mavenProxy.getId(), password, ex);
+                    }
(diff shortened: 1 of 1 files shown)


maven/src/main/java/org/owasp/dependencycheck/maven/BaseDependencyCheckMojo.java
```diff
@@ -2212,46 +2212,17 @@ protected void populateSettings() {
...
-                if (settings.getBoolean(Settings.KEYS.PROXY_DISABLE_SCHEMAS, true)) {
-                    System.setProperty("jdk.http.auth.tunneling.disabledSchemes", "");
-                }
-                try {
-                    password = decryptPasswordFromSettings(password);
-                } catch (SecDispatcherException ex) {
-                    password = handleSecDispatcherException("proxy", mavenProxy.getId(), password, ex);
+                    System.setProperty("https.proxyPassword", password);
                 }
             }
```
aikebah, line 2227: Your change breaks configuring the nonProxyHosts in the maven settings.xml
I think it should be restored (would've added it as a suggestion for the line where setting it in the legacy settings is deleted, but suggestions on deleted code are not supported by Github)

```suggestion
                final String nonProxyHosts = mavenProxy.getNonProxyHosts();
                if (nonProxyHosts != null && !nonProxyHosts.isEmpty()) {
                    System.setProperty("http.nonProxyHosts", nonProxyHosts);
                }
            }
```

aikebah (changes requested):

maven/src/main/java/org/owasp/dependencycheck/maven/BaseDependencyCheckMojo.java
stklcode: Oh dear, given this is actually the use case I wanted to fix first place :see_no_evil: The line got lost with the last minute password change. Good to have a second pair of eyes on a new day.

Restored the original lines.

jeremylong (approved):
LGTM

aikebah (approved):

aikebah:
@stklcode Thanks for the PR

PR #7077, opened 2024-10-23
fix(regression): Make the Downloader honour the proxy.nonproxyhosts ODC Setting

aikebah:
## Description of Change
As an extension to the fix for the Maven case covered by #7074 ensure that when the Downloader configures the proxy from the "legacy" ODC Settings properties the `proxy.nonproxyhosts` is still honoured.

no, used manual testing with the CLI, a custom properties file and a local dockerized squid proxy to verify that the proxy-route was only taken for Downloader actions targeting hosts that don't match the proxy.nonproxyhosts settings


jeremylong (approved):
LGTM

PR #7130, opened 2024-11-04
docs: improve false positive issue templat

jeremylong:
attempt to obtain bug reports that can be processed by the automation. In the future, the automation should be improved to accept multiple CPE/CVEs.


aikebah (approved):
LGTM

Being able to support multiple  CPEs would be a great addition in future indeed, especially for various 'open source + vendor' mixed products that have a CPE for the OSS edition and one for the vendor commercial edition.

PR #7137, opened 2024-11-07
fix(fp): Consolidate duplicate jetcd suppression and ensure considered base suppressions

chadlwilson:
## Description of Change

The change in #7117 failed to mark the suppressions as `base` which makes them show up incorrectly as unused suppressions in user runs.

```
Suppression Rule had zero matches: SuppressionRule{packageUrl=PropertyType{value=^pkg:maven/io\.etcd/jetcd-[a-z]*@.*$, regex=true, caseSensitive=false},cpe={PropertyType{value=cpe:/a:redhat:etcd, regex=false, caseSensitive=false},PropertyType{value=cpe:/a:etcd:etcd, regex=false, caseSensitive=false},}}
Suppression Rule had zero matches: SuppressionRule{packageUrl=PropertyType{value=^pkg:maven/io\.etcd/jetcd-grpc@.*$, regex=true, caseSensitive=false},cpe={PropertyType{value=cpe:/a:grpc:grpc, regex=false, caseSensitive=false},}}
```

Additionally consolidates a duplicate earlier suppression and uses `.*` to handle other dependencies like `jetcd-blah-blah`. 

FYI @joannakotula :-)

## Have test cases been added to cover the new functionality?

N/A

PR #7138, opened 2024-11-07
chore: cleanup base suppression

jeremylong:
PR #7137 cleanse-up older suppressions in the generated suppression file that have already been copied into the base suppression file.


aikebah (approved):
LGTM

PR #7147, opened 2024-11-09
fix: expose flag to disable version check

### LINKED ISSUE #7142: Please add option to disable version-check to CLI
I noticed that DC v11 was running a lot slower (~5x) than v10 on our servers, so I dug into this a little. I figured this was because of our proxy (which is usually the culprit of our issues) and it appears it was.
Seeing we already pretty much mirror all input for DC, I was a little confused why/where it would be trying to connect to, but in the logs I then found the culprit: the version check.

I think the issue of it getting very slow was caused by switching to the usage of Apache HTTP-client in [Commit b78ab24](https://github.com/jeremylong/DependencyCheck/commit/b78ab24120df2d2017c57081ab6288d6833c8ecf).

For Maven I then found the Parameter `versionCheckEnabled`, which made our builds even faster than before! I guess we never noticed/felt that DC took 'long' to run, until the update to v11...
So, my request is to add an option to disable the version-check in the CLI as well. We are not using ant and gradle, but I guess it might be useful there too.

jeremylong:
resolves #7142


aikebah (approved):
LGTM

PR #7150, opened 2024-11-11
chore: add chad as a FP approver

jeremylong:
@chadlwilson unless you object - we can add you as a false positive approver. You don't have to approve them if you do not want to. But you've been extremely helpful with your PRs and reviews.


chadlwilson:
I'm happy to help if I am trusted to be of assistance 🫡

PR #7204, opened 2024-12-01
feat: report on CVSS v4

jeremylong:
Updates reporting to utilize CvssV4.


jeremylong:
@aikebah should this be a feature or a breaking change (i.e. add in version `11.2.0` or `12.0.0`).

aikebah:
@jeremylong Given that the reports are frequently post-processed I recon it would be wiser to treat it as a breaking change.

jeremylong:
I'll try and publish 12.0 in the next day or two depending on other commitments I have.

PR #7255, opened 2024-12-15
fix: Make the HTTP-Client use pre-emptive authentication

### LINKED ISSUE #7108: Authentication failure after upgrade to 11.0.0 with Maven
**Describe the bug**
Since version 11.0.0 Apache Httpclient is used for web requests instead of plain java.
Configuration properties `suppressionFileUser` and `suppressionFilePassword` are no longer respected. (and the alternative configuration using `suppressionFileServerId` is neither respected).
By intercepting the http request the owasp dependency-check maven plugin tries to make, we can verify that it skips authentication, even when authentication is provided by the user!

The problem occurs using version 11.0.0 of the maven plugin

**Log file**
```
org.owasp.dependencycheck.exception.ExceptionCollection: One or more exceptions occurred during analysis:
        InitializationException: Warn initializing the suppression analyzer: Failed to load http://localhost:8087/suppressions.xml, caused by Unable to fetch the configured suppression file. 
                caused by SuppressionParseException: Failed to load http://localhost:8087/suppressions.xml, caused by Unable to fetch the configured suppression file.
```

**To Reproduce**
1. Configure maven dependency-check plugin with the following:
```xml
<configuration>
    <suppressionFiles>
        <suppressionFile>http://localhost:8087/suppressions.xml</suppressionFile>
    </suppressionFiles>
    <suppressionFileUser>my-username</suppressionFileUser>
    <suppressionFilePassword>my-secret-password</suppressionFilePassword>
</configuration>
```
2. On linux run netcat listening on port 8087 to intercept traffic and see what authentication maven dependency-check plugin comes up with:   `nc -l 8087`
3. Run mvn dependency-check:check
4. Check your netcat interceptor and you will see something along the lines of: 
```
GET /suppressions.xml HTTP/1.1
Accept-Encoding: gzip, x-gzip, deflate
Host: localhost:8087
Connection: keep-alive
User-Agent: Apache-HttpClient/5.4 (Java/21.0.2)
Upgrade: TLS/1.2
Connection: Upgrade
```
5. We can see that the `Authorization` request header is missing!!

**Expected behavior**
For step 4 in the reproduction steps above, we expect to see the Authorization header present on the request... e.g.:
```
GET /suppressions.xml HTTP/1.1
Authorization: Basic ........
Accept-Encoding: gzip, x-gzip, deflate
Host: localhost:8087
Connection: keep-alive
User-Agent: Apache-HttpClient/5.4 (Java/21.0.2)
Upgrade: TLS/1.2
Connection: Upgrade
```

**Other comments**

- If you run the reproduction steps again, but this time switch the plugin version from 11.0.0 to version 10.0.4 and you will see the `Authorization` header is present in the requests!

### LINKED ISSUE #7253: Extend HTTPClient usage for downloads to NexusSearch
Confirmed... only NexusSearch and ArtifactorySearch are pending. As both are typically solutions within the enterprise datacenter not requiring proxy access I think it would be fine to finish up this part and leave the remainder phase-out of URLConnectionFactory for a later feature-release in the 11.x series.

_Originally posted by @aikebah in https://github.com/jeremylong/DependencyCheck/issues/6949#issuecomment-2356535422_

Follow-up issue to #6949 to use HTTPClient for NexusSearch in order to honor the standard proxy configuration also for those (for the (expected rare) cases in which a Nexus repository is only reachable via a proxy)

aikebah:
Make the HTTP-Client use pre-emptive authentication for configured server credentials and extend HTTPClient usage to Nexus search

Fixes #7108
Fixes #7253

## Description of Change

Use pre-emptive authentication for the servers for which credentials are configured.

And add usage of the Apache HTTPClient for the web requests from Nexus search.

## Related issues

Fixes #7108 

Fixes #7253 

no;
Nexus search locally manually tested against private instances of Nexus v2 and Nexus v3 including access via a local squid proxy container.

### DIFF
--- a/core/src/main/java/org/owasp/dependencycheck/data/nexus/NexusV2Search.java
+++ b/core/src/main/java/org/owasp/dependencycheck/data/nexus/NexusV2Search.java
@@ -19,25 +19,29 @@ package org.owasp.dependencycheck.data.nexus;
 
 import java.io.FileNotFoundException;
 import java.io.IOException;
-import java.net.HttpURLConnection;
 import java.net.MalformedURLException;
+import java.net.URISyntaxException;
 import java.net.URL;
-import java.nio.charset.StandardCharsets;
-import java.util.Base64;
+import java.util.List;
 import javax.annotation.concurrent.ThreadSafe;
-import javax.xml.parsers.DocumentBuilder;
-import javax.xml.parsers.ParserConfigurationException;
 import javax.xml.xpath.XPath;
 import javax.xml.xpath.XPathExpressionException;
 import javax.xml.xpath.XPathFactory;
+
+import org.apache.hc.client5.http.HttpResponseException;
+import org.apache.hc.core5.http.ContentType;
+import org.apache.hc.core5.http.HttpHeaders;
+import org.apache.hc.core5.http.message.BasicHeader;
+import org.owasp.dependencycheck.utils.DownloadFailedException;
+import org.owasp.dependencycheck.utils.Downloader;
+import org.owasp.dependencycheck.utils.ResourceNotFoundException;
 import org.owasp.dependencycheck.utils.Settings;
 
-import org.owasp.dependencycheck.utils.URLConnectionFactory;
-import org.owasp.dependencycheck.utils.XmlUtils;
+import org.owasp.dependencycheck.utils.ToXMLDocumentResponseHandler;
+import org.owasp.dependencycheck.utils.TooManyRequestsException;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.w3c.dom.Document;
-import org.xml.sax.SAXException;
 
 /**
  * Class of methods to search Nexus repositories.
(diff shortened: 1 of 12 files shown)


jeremylong (dismissed):
LGTM

aikebah:
Marked draft pending inclusion of improvements from #7268 (adding configurability for hosted suppressions and support for Bearer Authentication + exposure of credential configuration for hostedSuppressions)

aikebah:
Might be better to do the remainder in a separate PR, marking this one as ready for review/merge. PR for extending auth to various integrations and enabling bearer auth to follow in a separate PR.

core/src/main/java/org/owasp/dependencycheck/data/nexus/NexusV2Search.java
```diff
@@ -94,118 +98,76 @@ public MavenArtifact searchSha1(String sha1) throws IOException {
...
-                    final Document doc = builder.parse(conn.getInputStream());
-                    final XPath xpath = XPathFactory.newInstance().newXPath();
-                    final String groupId = xpath
-                            .evaluate(
-                                    "/org.sonatype.nexus.rest.model.NexusArtifact/groupId",
-                                    doc);
-                    final String artifactId = xpath.evaluate(
-                            "/org.sonatype.nexus.rest.model.NexusArtifact/artifactId",
+        try {
+            // JSON would be more elegant, but there's not currently a dependency
```
jeremylong, line 102: Not that you have to change this - but we do have this dependency:

https://github.com/jeremylong/DependencyCheck/blob/486ca9435844990c040146decf37fc1d95d23524/pom.xml#L1231-L1235

jeremylong (approved):
LGTM

jeremylong:
@aikebah see my one comment - but you don't have to change anything. I'm fine if you just want to merge the PR.

core/src/main/java/org/owasp/dependencycheck/data/nexus/NexusV2Search.java
aikebah: Yeah, it was very tempting to change its behavior while sifting through the code, but thought it would be better to first cleanly fix authentication and leave the TODO-ish comment in as is for another time.

PR #7293, opened 2025-01-08
feat: ArtifactoryAnalyzer updated to use the HTTPClient5-based Downloader and skip unusable results

### LINKED ISSUE #5868: Custom Artifactory Setup : ERROR - Cannot extract the Maven information from the path retrieved in Artifactory
**Describe the bug**
We have setup Dependency-Check to use our Artifactory through a property file using the following options:

_analyzer.artifactory.enabled=true
analyzer.artifactory.url=artifactoy-instance_

During certain scans we got such kind of error:

`2023-08-08 10:34:43,354 org.owasp.dependencycheck.App:213
ERROR - Cannot extract the Maven information from the path retrieved in Artifactory /download/postgresql-42.6.0.jar
2023-08-08 10:34:43,354 org.owasp.dependencycheck.App:214
DEBUG - unexpected error
java.lang.IllegalStateException: Cannot extract the Maven information from the path retrieved in Artifactory /download/postgresql-42.6.0.jar
	at org.owasp.dependencycheck.data.artifactory.ArtifactorySearch.processResponse(ArtifactorySearch.java:223)
	at org.owasp.dependencycheck.data.artifactory.ArtifactorySearch.search(ArtifactorySearch.java:143)
	at org.owasp.dependencycheck.analyzer.ArtifactoryAnalyzer.analyzeDependency(ArtifactoryAnalyzer.java:199)
	at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:317)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1144)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:642)
	at java.base/java.lang.Thread.run(Thread.java:1589)
2023-08-08 10:34:43,354 org.owasp.dependencycheck.utils.Settings:895
`

Into our Artifactory, the content related to this file is the following:

_{
    "results": [
        {
            "uri": "<Artifactory-Instance>/api/storage/postgresql.org-cache/download/postgresql-42.6.0.jar"
        },
        {
            "uri": "<Artifactory-Instance>/api/storage/repo1.mvn-cache/org/postgresql/postgresql/42.6.0/postgresql-42.6.0.jar"
        }
    ]
}_

It seems that the system is taking the first element found in the list and error appears because of the pattern format that does not match with what is expected. 
We can solve this problem if we remove the reference of the first element into our Artifactory. 

The problem occurs using version 8.3.1 through CLI 

**Expected behavior**
I would like to have a way to handle this error. 
One solution I see is to add an option to include a list of repositories where the search of the file is made into Artifactory. 
With the API used, https://jfrog.com/help/r/jfrog-rest-apis/checksum-search this seems to be possible. 
Or a way to configure this generated error as a warning instead an error?

**Additional context**
I am using a property file to setup the default configuration. Ideally it would be good as well to get an additional option from the command line to offer the possibility to disable the usage of Artifactory.  
Right now, it is only possible to enable it.

### LINKED ISSUE #7254: Extend HTTPClient usage for downloads to ArtifactorySearch
Confirmed... only NexusSearch and ArtifactorySearch are pending. As both are typically solutions within the enterprise datacenter not requiring proxy access I think it would be fine to finish up this part and leave the remainder phase-out of URLConnectionFactory for a later feature-release in the 11.x series.

_Originally posted by @aikebah in https://github.com/jeremylong/DependencyCheck/issues/6949#issuecomment-2356535422_
            

Follow-up issue to #6949 to use HTTPClient for ArtifactorySearch in order to honor the standard proxy configuration also for the (expected rare) cases in which an Artifactory repository is only reachable via a proxy

aikebah:
## Description of Change

Make ArtifactoryAnalyzer use the Downloader class to finalize the migration of HTTP-traffic to use the Apache HTTPClient5 finalizing the work still left after #6949 

Make ArtifactoryAnalyzer skip results that do not fit the expectations (matching hashes and a maven-like G/A/V path structure) and attempt to find further matches in the remaining search results instead of throwing an exception on a first unexpected result entry format.

## Related issues

Fixes #5868 
Fixes #7254 

yes a testcase was added for the fix of #5868

### DIFF
--- a/core/src/main/java/org/owasp/dependencycheck/data/artifactory/ArtifactorySearch.java
+++ b/core/src/main/java/org/owasp/dependencycheck/data/artifactory/ArtifactorySearch.java
@@ -19,33 +19,26 @@ package org.owasp.dependencycheck.data.artifactory;
 
 import java.io.FileNotFoundException;
 import java.io.IOException;
-import java.io.InputStreamReader;
-import java.net.HttpURLConnection;
 import java.net.MalformedURLException;
+import java.net.URISyntaxException;
 import java.net.URL;
 import java.nio.charset.StandardCharsets;
-import java.util.ArrayList;
-import java.util.Base64;
 import java.util.List;
 import java.util.UUID;
-import java.util.regex.Matcher;
-import java.util.regex.Pattern;
 
 import javax.annotation.concurrent.ThreadSafe;
 
+import org.apache.hc.core5.http.message.BasicHeader;
 import org.owasp.dependencycheck.data.nexus.MavenArtifact;
 import org.owasp.dependencycheck.dependency.Dependency;
 import org.owasp.dependencycheck.utils.Checksum;
-import org.owasp.dependencycheck.utils.InvalidSettingException;
+import org.owasp.dependencycheck.utils.Downloader;
+import org.owasp.dependencycheck.utils.ResourceNotFoundException;
 import org.owasp.dependencycheck.utils.Settings;
-import org.owasp.dependencycheck.utils.URLConnectionFactory;
+import org.owasp.dependencycheck.utils.TooManyRequestsException;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-import com.fasterxml.jackson.core.JsonParser;
-import com.fasterxml.jackson.databind.DeserializationFeature;
-import com.fasterxml.jackson.databind.ObjectMapper;
-import com.fasterxml.jackson.databind.ObjectReader;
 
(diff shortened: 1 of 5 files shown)


aikebah:
Marked as a draft PR awaiting results from the community on live-testing a snapshot build against an Artifactory Pro instance.

jeremylong (dismissed):
LGTM

jeremylong (approved):

PR #7294, opened 2025-01-08
build(deps): Sanitize dependencies based on dependency:analyze-report results

aikebah:
## Description of Change

Processed the various dependency-analysis reports for missing / superfluous dependencies, so that now the poms also reflect the various used-undeclared usages of transitive dependencies.

Includes a code-change to upgrade the last remnants of commons-collections to commons-collections4.

The only items I left 'unfixed' are the unused cases for the various compile/codescan annotations libraries and junit/hamcrest.

## Related issues

N/A

N/A


jeremylong (dismissed):
LGTM

jeremylong (approved):

Issue #4894, opened 2022-09-30
Support yarn berry in Analyzer


Purii:
**Is your feature request related to a problem? Please describe.**
Using [`yarn berry`](https://github.com/yarnpkg/berry) with workspaces doesn't work right now. Would be great to also support this updated version of yarn.

Tried with the latest CLI version: `./dependency-check.sh --out . --scan ...`
Repository to reproduce: https://github.com/Purii/yarn-monorepo-example
Result: 
```sh
[INFO] Analysis Started
[INFO] Finished Archive Analyzer (0 seconds)
[INFO] Finished File Name Analyzer (0 seconds)
[WARN] No lock file exists - this will result in false negatives; please run `npm install --package-lock`
[WARN] Analyzing `/Users/../yarn-monorepo-example/packages/package-b/package.json` - however, the node_modules directory does not exist. Please run `npm install` prior to running dependency-check
[WARN] No lock file exists - this will result in false negatives; please run `npm install --package-lock`
[WARN] Analyzing `/Users/../yarn-monorepo-example/packages/package-a/package.json` - however, the node_modules directory does not exist. Please run `npm install` prior to running dependency-check
[INFO] Finished Node.js Package Analyzer (0 seconds)
[INFO] Finished Dependency Merging Analyzer (0 seconds)
[INFO] Finished Version Filter Analyzer (0 seconds)
[INFO] Finished Hint Analyzer (0 seconds)
[INFO] Created CPE Index (0 seconds)
[INFO] Finished CPE Analyzer (0 seconds)
[INFO] Finished False Positive Analyzer (0 seconds)
[INFO] Finished NVD CVE Analyzer (0 seconds)
[WARN] An unexpected error occurred during analysis of '/Users/../yarn-monorepo-example/yarn.lock' (Yarn Audit Analyzer): No value present
[ERROR] 
java.util.NoSuchElementException: No value present
	at java.base/java.util.Optional.get(Optional.java:143)
	at org.owasp.dependencycheck.analyzer.YarnAuditAnalyzer.fetchYarnAuditJson(YarnAuditAnalyzer.java:241)
	at org.owasp.dependencycheck.analyzer.YarnAuditAnalyzer.analyzePackage(YarnAuditAnalyzer.java:281)
	at org.owasp.dependencycheck.analyzer.YarnAuditAnalyzer.analyzeDependency(YarnAuditAnalyzer.java:106)
[...]

PR #7319, opened 2025-01-15
feat: Implement Yarn Berry Analyser

segovia:
I have implemented an Analyzer to address the issue that DependencyCheck does not work with projects that use yarn berry (https://github.com/jeremylong/DependencyCheck/issues/4894). This is a first draft and I would be happy to implement tests and make any other modifications if there is interest to merge this to main.


marcelstoer:
Thank you so much for this! This really simplifies our lives with multi-language projects driven by a Maven build (Java backend, JavaScript frontend). Btw, it might also fix #4215.

I think this can and should be improved to reduce future maintenance costs. Think about what happens when Yarn releases the next breaking change (like classic --> Berry)?

I would like to see a Yarn analyzer abstraction with some sort of auto-detection mechanism. IMO users should only have to set the `yarnAuditAnalyzerEnabled` without having to worry about the implementation. `yarnAuditAnalyzerEnabled` and `yarnBerryAuditAnalyzerEnabled` are actually mutually exclusive.

Can you extract a `YarnAuditAnalyzer` superclass with two subclasses for classic/v1 and Berry? Or, if they don't share enough, extract a `YarnAuditAnalyzer` interface with two implementations? They would need to be accompanied with a matching factory that introspects the lock file (or something else?) to instantiate the matching implementation.

/CC @chadlwilson @jeremylong

jeremylong:
@marcelstoer completely agree - your suggestions are spot on.

segovia:
With the updated PR, I have refactored `YarnAuditAnalyzer` into: `AbstractYarnAuditAnalyzer`, `YarnClassicAuditAnalyzer` and `YarnBerryAuditAnalyzer`, where the first contains shared analyzer code, the second is basically the same as the original `YarnAuditAnalyzer` and the third contains the Yarn Berry implementation.
As suggested, both analyzers share settings and are enabled or disabled with `yarnAuditAnalyzerEnabled`. They check the major version of yarn to determine if the Yarn Classic Analyzer should be used or the Berry, so they will run mutually exclusively.

core/src/main/java/org/owasp/dependencycheck/analyzer/AbstractYarnAuditAnalyzer.java
```diff
@@ -71,23 +59,42 @@ public class YarnAuditAnalyzer extends AbstractNpmAnalyzer {
...
+    protected String getYarnMajorVersion() {
+        if (StringUtils.isBlank(yarnVersion)) {
+            throw new IllegalArgumentException("Version string cannot be null or empty");
+        }
+        String[] parts = yarnVersion.split("\\.");
+        if (parts.length == 0) {
+            throw new IllegalArgumentException("Invalid version string format");
+        }
+        return parts[0];
+    }
```
jeremylong: Consider:
```suggestion
    protected int getYarnMajorVersion() {
        if (StringUtils.isBlank(yarnVersion)) {
            throw new IllegalArgumentException("Version string cannot be null or empty");
        }
        try {
            Semver semver = new Semver(yarnVersion);
            return semver.getMajor();
        } catch (SemverException e) {
            throw new IllegalArgumentException("Invalid version string format", e);
        }
    }
```

core/src/main/java/org/owasp/dependencycheck/analyzer/AbstractYarnAuditAnalyzer.java
```diff
@@ -19,45 +19,33 @@
...
 import org.owasp.dependencycheck.analyzer.exception.UnexpectedAnalysisException;
 import org.owasp.dependencycheck.data.nodeaudit.Advisory;
-import org.owasp.dependencycheck.data.nodeaudit.NpmPayloadBuilder;
 import org.owasp.dependencycheck.dependency.Dependency;
 import org.owasp.dependencycheck.exception.InitializationException;
 import org.owasp.dependencycheck.utils.FileFilterBuilder;
 import org.owasp.dependencycheck.utils.Settings;
-import org.owasp.dependencycheck.utils.URLConnectionFailureException;
 import org.owasp.dependencycheck.utils.processing.ProcessReader;
 import org.slf4j.Logger;
```
jeremylong:
```suggestion
import org.semver4j.Semver;
import org.semver4j.SemverException;
import org.slf4j.Logger;
```

core/src/main/java/org/owasp/dependencycheck/analyzer/YarnBerryAuditAnalyzer.java
```diff
@@ -0,0 +1,173 @@
...
+     */
+    @Override
+    protected void prepareFileTypeAnalyzer(Engine engine) throws InitializationException {
+        super.prepareFileTypeAnalyzer(engine);
+        if (!isEnabled()) {
+            LOGGER.debug("{} Analyzer is disabled skipping yarn major version check", getName());
+            return;
+        }
+        var yarnMajorVersion = getYarnMajorVersion();
+        if ("1".equals(yarnMajorVersion)) {
```
jeremylong:
```suggestion
        if (yarnMajorVersion == 1) {
```

core/src/main/java/org/owasp/dependencycheck/analyzer/YarnClassicAuditAnalyzer.java
```diff
@@ -0,0 +1,201 @@
...
+     */
+    @Override
+    protected void prepareFileTypeAnalyzer(Engine engine) throws InitializationException {
+        super.prepareFileTypeAnalyzer(engine);
+        if (!isEnabled()) {
+            LOGGER.debug("{} Analyzer is disabled skipping yarn major version check", getName());
+            return;
+        }
+        var yarnMajorVersion = getYarnMajorVersion();
+        if (!"1".equals(yarnMajorVersion)) {
```
jeremylong:
```suggestion
        if (yarnMajorVersion > 1) {
```

core/src/main/java/org/owasp/dependencycheck/analyzer/AbstractYarnAuditAnalyzer.java
```diff
@@ -71,23 +59,42 @@ public class YarnAuditAnalyzer extends AbstractNpmAnalyzer {
...
+
+
+    protected String getYarnVersion() {
+        return yarnVersion;
+    }
+
+    /**
+     * Extracts the major version from a version string.
+     *
+     * @return the major version (e.g., "4" from "4.2.1")
```
jeremylong:
```suggestion
     * @return the major version (e.g., `4` from "4.2.1")
```

jeremylong (commented):
LGTM - I left some suggestions around the yarn version number. I prefer to use something like `semver4j` as we already have it on the classpath.

jeremylong (changes requested):
It looks like the build failed:

```
 YarnBerryAuditAnalyzerIT.testAnalyzePackageYarn:46 More than 1 dependency should be identified
 ```

Do we need to update the workflow itself to include yarn berry?

**Comment by chadlwilson** [comment]
> Do we need to update the workflow itself to include yarn berry?

FWIW the canonical way to run and install yarn 2+ is via corepack and source controlling yarn itself alongside a package.json which defines the version number.

https://yarnpkg.com/getting-started/install

Although you can use yarn 1 to switch dynamically to yarn 2+.projects if you have to.

Ideally the tests should test a project that has package.json etc preconfigured for yarn usage and 'corepack enable' run on the npm install. This has the advantage that specific yarn versions are source controlled alongside the tests in the `.yarn` folder (canonically).

File: core/src/main/java/org/owasp/dependencycheck/analyzer/YarnBerryAuditAnalyzer.java (hunk 3 above)
Comment (marcelstoer) [reply core/src/main/java/org/owasp/dependencycheck/analyzer/YarnBerryAuditAnalyzer.java]: Changing the comparison isn't addressing the real issue IMO.

I'm with all those who see magic numbers or magic strings as an anti-pattern. The suggested fix usually is to use (properly) named constants. In this case, however, that would only improve the legibility of the code but would not improve its long-term stability.

At this point, all we know is that the Yarn classic analyzer supports Yarn v1 and that the Yarn Berry analyzer supports Yarn v2. We are not in a position to make any promises as for a potential future Yarn v3. The Berry analyzer may or may not support it. To play it safe, I'd like to see an ok-list approach.

Suggestion:
- Declare an `abstract supportedNodeMajorVersions()` method in `AbstractYarnAuditAnalyzer` that returns an array of `int`s or a set of `Int`s.
- Let both sub classes implement it to the best of their knowledge. Hence, both would currently return a set containing a single value.
- Add a `boolean isYarnVersionSupported()` method to `AbstractYarnAuditAnalyzer` that compares  `supportedNodeVersions()` against the return value of `getYarnMajorVersion()` (major contained in list of supported). Nice side effect: `getYarnMajorVersion()` can likely be made `private`.
- The `prepareFileTypeAnalyzer()` implementations would thus be shielded from all that complexity as they would only need to do an `if (!isYarnVersionSupported())`-then-disable-analyzer.

This idea could be spun even further but that would require more refactoring.

File: core/src/main/java/org/owasp/dependencycheck/analyzer/YarnClassicAuditAnalyzer.java (hunk 4 above)
Comment (marcelstoer) [reply core/src/main/java/org/owasp/dependencycheck/analyzer/YarnClassicAuditAnalyzer.java]: Ditto

**Comment by marcelstoer** [comment]
@segovia thanks for the updated version. This is much more maintainable now!

File: core/src/main/java/org/owasp/dependencycheck/analyzer/YarnBerryAuditAnalyzer.java (hunk 3 above)
Comment (chadlwilson) [reply core/src/main/java/org/owasp/dependencycheck/analyzer/YarnBerryAuditAnalyzer.java]: Yarn v3 and v4 already exist, and are considered part of the 'berry' family. Current version is 4.6.0. It's reasonable to consider this analyzer will be required indefinitely.

Is this a theoretical objection, or are you familiar with modern Yarn and its compat/history?

We don't lock other analyzers against software versions that might be released in future. Given the yarn analyzer has been broken for years now with respect to current yarn I don't think adding more maintenance overhead to specifically check new versions is necessary?

Comment (marcelstoer) [reply core/src/main/java/org/owasp/dependencycheck/analyzer/YarnBerryAuditAnalyzer.java]: As you might have guessed, I'm not at all familiar with anything related to Yarn. Also, I don't consider my comment to be an objection but merely a suggestion. When I scanned the PR initially, the magic numbers stuck out immediately.

> It's reasonable to consider this analyzer will be required indefinitely.

If that's true, then my suggestion doesn't make much sense. Thanks for pointing that out.

Comment (chadlwilson) [reply core/src/main/java/org/owasp/dependencycheck/analyzer/YarnBerryAuditAnalyzer.java]: Yup, no worries - appreciate the thoughtful input - even if theoretical :-) Magic numbers can indeed be a bit gross and it's reasonable to debate about Yarn's compatibility guarantees for this stuff over time.

It does occur to me to contemplate whether the "berry" analyzer as it is here works the same back through yarn 2 / 3 /4 as I cannot recall off the top of my head when it was introduced or whether format has changed.

Comment (jeremylong) [reply core/src/main/java/org/owasp/dependencycheck/analyzer/YarnBerryAuditAnalyzer.java]: Converting the comparison to use `int` will make it easier to implement a new analyzer if a successor to yarn berry has a different format.

**Comment by segovia** [comment]
Thank you all for your feedback and suggestions. With my latest update, I incorporated semver4j and named the magic number 1 as YARN_CLASSIC_MAJOR_VERSION.

Concerning the integration test, I suggest putting the latest yarn berry .cjs file somewhere. In my commit, I put it in core/src/test/resources/yarn-berry-audit/.yarn/releases/yarn-4.6.0.cjs, but if you don't want the yarn .cjs to be part of the commit, it could be placed anywhere else. I am not familiar with the container running the tests, but if we add it somewhere specific there, I can reference it in the `.yarnrc.yml` file.

File: core/src/main/java/org/owasp/dependencycheck/analyzer/YarnClassicAuditAnalyzer.java [hunk 6]
```diff
@@ -0,0 +1,202 @@
...
+        if (!isEnabled()) {
+            LOGGER.debug("{} Analyzer is disabled skipping yarn major version check", getName());
+            return;
+        }
+        var yarnMajorVersion = getYarnMajorVersion();
+
+        if (YARN_CLASSIC_MAJOR_VERSION == yarnMajorVersion) {
+            this.setEnabled(false);
+            LOGGER.debug("The {} has been disabled. This analyzer is not applicable for major version {}.", getName(), yarnMajorVersion);
+        }
```
jeremylong: This was probably my error when moving from `String` -> `int` - I think the comparison was wrong. In 'Classic' mode we want to disable if the yarn Major version is greater than the classic major version.

```suggestion
        if (YARN_CLASSIC_MAJOR_VERSION > yarnMajorVersion) {
            this.setEnabled(false);
            LOGGER.debug("The {} has been disabled. This analyzer is not applicable for major version {}.", getName(), yarnMajorVersion);
        }
```

core/src/main/java/org/owasp/dependencycheck/analyzer/YarnBerryAuditAnalyzer.java
```diff
@@ -0,0 +1,175 @@
...
+            return;
+        }
+        var yarnMajorVersion = getYarnMajorVersion();
+
+        // we assume that any major version that is not the classic version, is the berry version.
+        // at the time of writing, the classic version is 1 and berry versions are 2, 3 and 4.
+        if (YARN_CLASSIC_MAJOR_VERSION < yarnMajorVersion) {
+            this.setEnabled(false);
+            LOGGER.debug("The {} has been disabled. This analyzer is not applicable for major version {}.", getName(), yarnMajorVersion);
+        }
```
jeremylong: Again, I think this might have been my mistake when moving from `String` -> `int`. In the berry analyzer we want to disable the analyzer if the yarn major version is equal to the classic major version:

```suggestion
        if (YARN_CLASSIC_MAJOR_VERSION <= yarnMajorVersion) {
            this.setEnabled(false);
            LOGGER.debug("The {} has been disabled. This analyzer is not applicable for major version {}.", getName(), yarnMajorVersion);
        }
```

jeremylong (commented):
@segovia sorry about the back-and-forth on this one. Can you double-check the comparisons for the `String` -> `'int` conversion?

core/src/main/java/org/owasp/dependencycheck/analyzer/YarnClassicAuditAnalyzer.java
segovia: Yes, good catch. I also should have noticed. I think you have switched berry and classic with your suggestions but I think I got it right now with the PR.

core/src/main/java/org/owasp/dependencycheck/analyzer/YarnBerryAuditAnalyzer.java
segovia: Same

segovia:
I pushed an update in an attempt to fix the tests. Unfortunately, I am not able to run all the tests locally. In case it fails again, is there some documentation on how to run the tests locally? Or do you have any tips?

jeremylong:
I've been trying to figure out why the test case is failing - locally I've just been using variations of this:

```
mvn clean verify -Dtest=org.owasp.dependencycheck.analyzer.YarnBerryAuditAnalyzerIT  -Dsurefire.failIfNoSpecifiedTests=false
```

In order for the tests to run on core, utils has to have been packaged in the execution.

jeremylong:
Looking at why the tests were failing and the fix - I don't think the enable/disable function is going to work correctly in the wild. I know how to fix it - but it will take a bit of refactoring. Instead of having two separate analyzers a single analyzer should use the correct mechanism to `fetchYarnAdvisories` vs `fetchYarnAuditJson`. You'll notice in these we set the `ProcessBuilder.directory` so if we run `--version` from there we should get the correct version so we'll know which command to run.

jeremylong:
@segovia do you want me to do the refactoring - or do you want to?

segovia:
@jeremylong  I've taken your feedback into consideration. Now we determine which method to use based on the dependency's directory. Hopefully, this is what you were thinking. In any case, if any further modifications are required, please feel free to adapt.

jeremylong:
@segovia thank you for all your work on this. It might be another day or two before I can do a final review and get this merged.

jeremylong (approved):
LGTM

jeremylong:
@segovia thanks for contributing!!!!

segovia:
Thanks for the feedback and merging!

PR #7322, opened 2025-01-16
docs: Fix OSS Index Maven config documentation

marcelstoer:
```
❯ grep -ir "property = \"ossindex" *
maven/src/main/java/org/owasp/dependencycheck/maven/BaseDependencyCheckMojo.java:    @Parameter(property = "ossindexAnalyzerEnabled")
maven/src/main/java/org/owasp/dependencycheck/maven/BaseDependencyCheckMojo.java:    @Parameter(property = "ossindexAnalyzerUseCache")
maven/src/main/java/org/owasp/dependencycheck/maven/BaseDependencyCheckMojo.java:    @Parameter(property = "ossindexAnalyzerUrl")
maven/src/main/java/org/owasp/dependencycheck/maven/BaseDependencyCheckMojo.java:    @Parameter(property = "ossIndexServerId")
maven/src/main/java/org/owasp/dependencycheck/maven/BaseDependencyCheckMojo.java:    @Parameter(property = "ossIndexWarnOnlyOnRemoteErrors")
```

There are exactly two properties where "OSS Index" is correctly camel-cased. Yet, one of them was incorrect in the documentation.

I suggest you consider renaming the other three properties and their documentation with the next major version as fixing this is a breaking change.

**Background story**

I wanted to use a registered OSS account as we hit their rate limit today. At some point during the testing phase we were freed from the rate limit but I didn't know whether that was because I had configured user/pw in the `settings.xml` or whether time solved it for us. I ran Maven in debug mode to see wether the ODC output would give me any hints whether it used the configured OSS server credentials or not. I didn't find anything. I then started mitmproxy and inspected the requests against OSS - only to find that they didn't have an `Authorization` header. That's when I started digging why my `-DossindexServerId=` was ignored.

Is there a simpler method available to verify that remote server credentials are applied?


jeremylong (approved):
LGTM

jeremylong:
We likely need to add some logging indicating that credentials are being used for a particular connection.

PR #7328, opened 2025-01-19
docs(gradle): Hint to users as to how to see the unused rules with the current gradle implementation

chadlwilson:
## Description of Change

Currently due to the way the Gradle plugin is written when `failBuildOnUnusedSuppressionRule` (and some other analyzers) are used you just get `Analysis Failed` which is not much fun.

Short of improving the way exceptions are logged with the plugin, or better integrating logging with Gradle lifecycle - this at least tells folks how to see the unused rules, e.g

```
$ ./gradlew dependencyCheckAggregate --warn
Suppression Rule had zero matches: SuppressionRule{cpe={PropertyType{value=cpe:/a:json-c:json-c, regex=false, caseSensitive=false},}}
There are 1 unused suppression rule(s): check logs.
An error occurred while analyzing '/Users/chad/.gradle/caches/modules-2/files-2.1/org.jetbrains/annotations/26.0.1/c7ce3cdeda3d18909368dfe5977332dfad326c6d/annotations-26.0.1.jar' (Unused Suppression Rule Analyzer).
Region [NODEAUDIT] : Not alive and dispose was called, filename: NODEAUDIT
Region [CENTRAL] : Not alive and dispose was called, filename: CENTRAL
Region [POM] : Not alive and dispose was called, filename: POM

[Incubating] Problems report is available at: file:///Users/chad/Projects/community/gocd/gocd/target/reports/problems/problems-report.html

FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':dependencyCheckAggregate'.
> Analysis failed.
```

N/A


jeremylong (approved):
LGTM

PR #7340, opened 2025-01-22
build: remove repo.spring.io from workflows

chadlwilson:
The FP workflows are failing due to this repo being unauthorised for access. Not sure how it was working before unless there was an (expired?) authentication token somewhere, as anonymous access I believe was blocked some time ago.

Do we need this here, or is Maven Central good enough?


jeremylong (approved):
LGTM

jeremylong:
The repo is unnecessary - I think I was trying to add as many alternatives as possible in case the FP came from another public repo. If this is no longer public we should remove it.

PR #7364, opened 2025-01-29
fix: Add missing used dependency and remove force-usage for dependency-plugin for a now explicitly used dependency

### LINKED ISSUE #7362: [BUG]: site plugin fails
When trying to release 12.0.2 I ran into an issue with the maven-site:

```
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-site-plugin:3.21.0:site (default-site) on project dependency-check-core: Error generating maven-dependency-plugin:3.8.1:analyze-report report: Cannot analyze dependencies: Trying to force use of dependencies which are declared but already detected as used: [ch.qos.logback:logback-classic] -> [Help 1]
```

See the error on the attempted release: https://github.com/jeremylong/DependencyCheck/actions/runs/13030928514/job/36350430369

aikebah:
## Description of Change

Removes the force-usage configuration that makes maven-dependency-plugin error up on an unneeded forced usage

Adds another dependency that (in addition to the previously 'forced-in-use' dependency is also newly used by the test-code and was reported as 'used undeclared dependency'

## Related issues

fixes #7362 

N/A

### DIFF
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -229,6 +229,11 @@ Copyright (c) 2012 Jeremy Long. All Rights Reserved.
             <scope>test</scope>
         </dependency>
         <dependency>
+            <groupId>ch.qos.logback</groupId>
+            <artifactId>logback-core</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
             <groupId>org.owasp</groupId>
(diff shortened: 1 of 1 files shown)


nhumblot (approved):

jeremylong:
@aikebah thank you!

Issue #7338, opened 2025-01-21
dependencycheck 12.0.0 randomly fails with comparison method violates its general contract


johnou:
**Describe the bug**

From time to time dependencycheck fails with java.lang.IllegalArgumentException: Comparison method violates its general contract

**Version of dependency-check used**

org.owasp.dependencycheck 12.0.0
gradle 7.6.4

**Log file**

https://gist.github.com/johnou/ab6a6d9a48ae98ff91450b904b1522ec

**Expected behavior**

No failure

**Additional context**

Multi module gradle project with nvd api key defined.

```
dependencyCheck {
    analyzers.retirejs.enabled = false
    analyzers.assemblyEnabled = false
    nvd.apiKey = System.getenv("NVD_API_KEY") ?: ""
}

subprojects {
    apply plugin: 'io.spring.dependency-management'
    apply plugin: 'idea'
    apply plugin: 'org.owasp.dependencycheck'
    apply plugin: 'jacoco'


    dependencyCheck {
        suppressionFile = "${rootDir}/dependency-check-suppress.xml"
        format = 'ALL'
        analyzers {
            assemblyEnabled = false
            retirejs {
                enabled = false
            }
        }
        nvd.apiKey = System.getenv("NVD_API_KEY") ?: ""
    }

...
}
```

chadlwilson:
```
Caused by: org.apache.velocity.exception.MethodInvocationException: Invocation of method 'getVulnerableSoftware' in  class org.owasp.dependencycheck.dependency.Vulnerability threw exception java.lang.IllegalArgumentException: Comparison method violates its general contract! at templates/xmlReport.vsl[line 423, column 23]
```

Issue seems to be with the lines below in the XML report - are you able to figure out which dependency/vulnerability the problem is with, possibly be enabling info or debug logging and seeing what it was doing when it failed?

https://github.com/jeremylong/DependencyCheck/blob/bebb4317cf49e7e082f5aa6cd943729289f003d4/core/src/main/resources/templates/xmlReport.vsl#L423-L431

Workaround is possibly to disable the XML report if you don't need it.

jeremylong:
It is possible the report is even half-written and you can see where it stopped. Knowing the vulnerability would be very helpful.

aikebah:
Issue has nothing to do with the report template, it has to do with a bug that's been hiding in Vulnerability for 3 yrs waiting to be discovered.

```
[...]

PR #7366, opened 2025-01-30
fix: Improve thread safety issue #7338

johnou:
## Description of Change

Improve thread safety to avoid possible issues with timsort.

no

### DIFF
--- a/core/src/main/java/org/owasp/dependencycheck/dependency/Vulnerability.java
+++ b/core/src/main/java/org/owasp/dependencycheck/dependency/Vulnerability.java
@@ -92,7 +92,7 @@ public class Vulnerability implements Serializable, Comparable<Vulnerability> {
     /**
      * References for this vulnerability.
      */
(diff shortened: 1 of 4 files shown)


core/src/main/java/org/owasp/dependencycheck/dependency/Vulnerability.java
```diff
@@ -196,7 +196,9 @@ public void setDescription(String description) {
      * @return the value of references
      */
     public Set<Reference> getReferences() {
-        return references;
+        synchronized (references) {
+            return Collections.unmodifiableSet(new HashSet<>(references));
```
johnou, line 200: lmk if you think unmodifiableSet is too heavy handed, but I think modifying anything through the getter is an anti-pattern and out-right dangerous, especially when there are already add methods.

johnou:
to avoid gc thrashing this approach makes more sense https://github.com/jeremylong/DependencyCheck/pull/7367

Issue #7235, opened 2024-12-09
Move repository to the dependency-check organization


jeremylong:
To better support a growing team of contributors we plan to move the project from a personal repository to the https://github.com/dependency-check organization. The gradle plugin has been there along with other related projects.

This means the documentation pages will move to a new URL. I believe GH puts in redirects when the project moves. The possible issues:

1. Many users have automation to check the `current` release and download the latest from this repo. Even if redirects exist it could break the automation
2. The schemas all point to the personal repo:
           ```
           targetNamespace="https://jeremylong.github.io/DependencyCheck/dependency-check.4.1.xsd"
           xmlns:dc="https://jeremylong.github.io/DependencyCheck/dependency-check.4.1.xsd">
           ```

@aikebah any concerns with this move? Can you think of any other issues?

aikebah:
On item 2) Strictly spoken the schema's don't point to your personal repo. An xmlns is simply a resource identifier to uniquely identify a schema, but many people tend to get that wrong when URL-conforming URIs are used, so I think it might be beneficial to even update those (would require a new major if we do so as people need to update the namespace of their filter/hint files to migrate them to the new namespace)

For retrieval location specification there is the schemaLocation attribute in the XML Schema Instance namespace (https://www.w3.org/TR/xmlschema-1/#schema-loc)


I think the main part besides ensuring all our docs are updated for the new location is ensuring that the various integrations out there as far as we know about them (from memory I know azure devops, jenkins plugin, sonarqube plugin) get aligned on the project location change-over so that they can also publish an update once the change-over is completed with a release at the new location so as to not rely on Github forwarding them.

Can't think of further items to take into account (apart from me joining up on the dependency-check org on github in an appropriate role), but have never had to deal with a GH project move, so there might be some gotchas we only find out about when the move is done.

marcelstoer:
https://docs.github.com/en/repositories/creating-and-managing-repositories/transferring-a-repository might have more hints, I only scanned it briefly.

The term "move" hasn't been clearly defined yet. You can transfer the GitHub project as a whole or can selectively move assets (eg. Git repository). 

[...]

PR #7373, opened 2025-02-04
chore: update links in preparation for repository move

jeremylong:
To be merged just prior to transferring the repository.

See #7235


aikebah:
@jeremylong you're still working on this, or is it only draft to prevent accidental merge as it needs to wait till just-before-move? There are still a significant amount of github URLs in the current branch in need of change to make the code properly reflect the new home after the move.

Let me know if I can help on getting those updated.

jeremylong:
It was a draft for both reasons - I hadn't finished updating things (I ran out of time) and I'm keeping this as a draft until the move happens.

As of now - I believe all of the references have been updated. If you know of anything I missed please speak up.

aikebah:
@jeremylong I've processed most items in my PR towards the scratch-branch.... an item I wasn't sure about was the coverity config. I can imagine that the jeremylong/DependencyCheck there is a reference to a Coverity project and not a reference to the github project as such. If not I think there would be an additional change required in the coverity_scan.sh (the value for env variable COVERITY_SCAN_PROJECT_NAME) - however your update to coverity.yml appears to suggest the change would be needed as in there you've modified the project parameter various times

PR #7378, opened 2025-02-06
chore: update more URLs

aikebah:
## Description of Change

I think these are still missing in your scratch branch.


aikebah:
@jeremylong I think these are still missing in the changes you prepared

aikebah:
created it as a PR towards your scratch branch, so that you can review and then incorporate them on your branch

jeremylong (approved):

Issue #7411, opened 2025-02-17
Dependency check fails: Error updating the NVD Data / SAFETY


ccudennec-otto:
Hi @jeremylong !

We're running into the described issue and I assume that this is not temporal but a bug?

**Describe the bug**
Running `./gradlew dependencyCheckAnalyze` fails with the following error message:

```
Verifying dependencies for project albatros_product_service
Checking for updates and analyzing dependencies for vulnerabilities
Error updating the NVD Data
org.owasp.dependencycheck.data.update.exception.UpdateException: Error updating the NVD Data
	at org.owasp.dependencycheck.data.update.NvdApiDataSource.processApi(NvdApiDataSource.java:399)
	at org.owasp.dependencycheck.data.update.NvdApiDataSource.update(NvdApiDataSource.java:117)
	at org.owasp.dependencycheck.Engine.doUpdates(Engine.java:903)
	at org.owasp.dependencycheck.Engine.initializeAndUpdateDatabase(Engine.java:708)
	at org.owasp.dependencycheck.Engine.analyzeDependencies(Engine.java:634)
	at org.owasp.dependencycheck.gradle.tasks.AbstractAnalyze.analyze(AbstractAnalyze.groovy:103)
...
Caused by: io.github.jeremylong.openvulnerability.client.nvd.NvdApiException: Failed to parse NVD data
	at io.github.jeremylong.openvulnerability.client.nvd.NvdCveClient._next(NvdCveClient.java:389)
	at io.github.jeremylong.openvulnerability.client.nvd.NvdCveClient.next(NvdCveClient.java:357)
	at org.owasp.dependencycheck.data.update.NvdApiDataSource.processApi(NvdApiDataSource.java:355)
	... 129 more
Caused by: com.fasterxml.jackson.databind.exc.ValueInstantiationException: Cannot construct instance of `io.github.jeremylong.openvulnerability.client.nvd.CvssV4Data$ModifiedCiaType`, problem: SAFETY
[...]

PR #7412, opened 2025-02-17
chore: Amend the bug template

marcelstoer:
## Description of Change
It annoys me to observe how many users can't be bothered to at least check the issues list before reporting their own. I accept that this is a loosing battle but giving extra hints to people doesn't do any harm.

## Related issues
#7411, just one of many examples

not applicable


jeremylong (approved):
LGTM - thank you for adding this.

PR #7427, opened 2025-02-18
fix: CVE-2021-26291

jpmartins-ca:
## Description of Change

Update maven-core api dependency

## Related issues

https://github.com/dependency-check/DependencyCheck/issues/7156
https://github.com/dependency-check/DependencyCheck/issues/7426

no

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -125,7 +125,7 @@ Copyright (c) 2012 - Jeremy Long
         <slf4j.version>1.7.36</slf4j.version>
         <logback.version>1.2.13</logback.version>
         
-        <maven.api.version>3.6.3</maven.api.version>
+        <maven.api.version>3.9.9</maven.api.version>
         <reporting.checkstyle-plugin.version>3.6.0</reporting.checkstyle-plugin.version>
         <reporting.checkstyle.tool.version>9.3</reporting.checkstyle.tool.version>
         <doxia-base.version>2.0.0</doxia-base.version>


jpmartins-ca:
Hello everyone. Hello @jeremylong. Is it possible to do this change? 

Looks like a simple update on maven-core version, and it should clear up the CVE-2021-26291.

I strongly suspect this is the cause for usage "false positive" error:

maven-core-3.6.3.jar (pkg:maven/org.apache.maven/maven-core@3.6.3, cpe:2.3:a:apache:maven:3.6.3:*:*:*:*:*:*:*): CVE-2021-26291(9.1)

Thank you.

aikebah:
Maven API versions are deliberately on 3.6.3, it's the API compatibility we intend to maintain. All maven libraries are expected to be provided. Using later APIs would risk using features added after maven 3.6.3.

jpmartins-ca:
Hello @aikebah,

Thank you for your reply.

We are considering forcing/supplying the Maven core API 3.9.9 at runtime (as I have seen done, or at least attempted, in multiple examples). Do the risks you mentioned apply only during dependency-check development, or could they also affect dependency-check scan/runtime?

Where can I read more about the features introduced after maven core 3.6.3 that might pose risks? Additionally, how do these risks apply to dependency-check?

(I did a maven install with the 3.9.9, and after installing requirements to run the tests, all the tests passed)

Have a nice day.

aikebah:
The risks of `CVE-2021-26291` only apply to a maven run, not to dependency-check itself.

DependencyCheck does not try do perform all the magic of Maven.

In a maven build maven will honor any additional `<repositories>` defined in a pom file and attempt to download artifacts from there (if not overruled by a mirror setting)

DependencyCheck does no such thing. 

In case ODC is run as a maven plugin the maven plugin bits will interact with the API of Maven to have maven resolve/download all project (transitive) dependencies so it delegates all of that magic to the internals of the Maven version that runs the plugin.

All other variants of DependencyCheck read some selected items directly from a pom.xml file when that is discovered as part of the scan, but will never deal with repositories.

So the risk of `CVE-2021-26291` is during builds of maven projects in environments that have neither upgraded their maven to 3.8.1 or later nor use a catch-all mirror in the settings to replace any repository-url configured anywhere by their own authoritive mirror-repository.

Features introduced in maven you would have to harvest from the apache maven project. You'd have to do a diff across the 3.6.3 version and the current version of the APIs to see what was newly introduced since 3.6.3.

By depending on the 3.6.3 we make sure that ODC does not call any newly introduced method without having to resort to closely following what was introduced when. On 3.6.3 API libraries the post-3.6.3 newly introduced methods in the API will not be available, so our code would not compile if we tried using them. That way we can be sure that we do not use features introduced after 3.6.3 while being fairly sure that we can still run on latest maven 3.x as the maven community has a good track-record of not breaking API within a major version.

PR #7429, opened 2025-02-18
build(deps): Update mysql and postgres drivers of the docker image 

lloyd-theophilus:
## Related issues
Compatibly issue

No test cases need as I only change the driver versions to the latest
*yes/no*

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -7,8 +7,8 @@ RUN "$JAVA_HOME/bin/jlink" --compress=2 --module-path /opt/java/openjdk/jmods --
 FROM mcr.microsoft.com/dotnet/runtime:8.0-alpine3.18
 
 ARG VERSION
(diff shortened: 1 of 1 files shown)


jeremylong (approved):
LGTM

jeremylong:
Build failed - https://github.com/dependency-check/DependencyCheck/actions/runs/13413379465

reverting this PR...

aikebah:
@jeremylong Connector-J is for current versions at
`https://dev.mysql.com/get/Downloads/Connector-J/mysql-connector-j-<version>.tar.gz`

instead of the previous

`https://dev.mysql.com/get/Downloads/Connector-J/mysql-connector-java-<version>.tar.gz`

For Postgresql I'm not sure whether the space after the equal-sign would be included in the URL, if it is retained that would also break.

But first some sleep and then another day in the office for me. If you don't get to it I might prepare a proper PR tomorrow evening.

Issue #7430, opened 2025-02-19
Sonatype Index error


lydacious:
Hello,

We always keep having to retry dependency check jobs as it almost never runs on first try. Sometimes it takes 2 retries, sometimes 3. The error we get is:
```Verifying dependencies for project ms-user
Checking for updates and analyzing dependencies for vulnerabilities
WARNING: A restricted method in java.lang.foreign.Linker has been called
WARNING: java.lang.foreign.Linker::downcallHandle has been called by the unnamed module
WARNING: Use --enable-native-access=ALL-UNNAMED to avoid a warning for this module
Java vector incubator module is not readable. For optimal vector performance, pass '--add-modules jdk.incubator.vector' to enable Vector API.
An error occurred while analyzing '/root/.gradle/caches/modules-2/files-2.1/org.projectlombok/lombok/1.18.36/5a30490a6e14977d97d9c73c924c1f1b5311ea95/lombok-1.18.36.jar' (Sonatype OSS Index Analyzer).
An error occurred while analyzing '/root/.gradle/caches/modules-2/files-2.1/org.springframework.boot/spring-boot-starter-web/3.4.1/ff7227fc62338e0f6eba3f9f94c12eb952d4da95/spring-boot-starter-web-3.4.1.jar' (Sonatype OSS Index Analyzer).
An error occurred while analyzing '/root/.gradle/caches/modules-2/files-2.1/org.apache.httpcomponents.core5/httpcore5/5.3.1/eaf64237945d7d0f301d48420e8bdb7f565a7b0e/httpcore5-5.3.1.jar' (Sonatype OSS Index Analyzer).
An error occurred while analyzing '/root/.gradle/caches/modules-2/files-2.1/org.apache.httpcomponents.core5/httpcore5-h2/5.3.1/760c34db3ba41b0ffa07e956bc308d3a12356915/httpcore5-h2-5.3.1.jar' (Sonatype OSS Index Analyzer).
An error occurred while analyzing '/root/.gradle/caches/modules-2/files-2.1/org.apache.httpcomponents.client5/httpclient5/5.4.1/ce913081e592ee8eeee35c4e577d7dce13cba7a4/httpclient5-5.4.1.jar' (Sonatype OSS Index Analyzer).
[...]

PR #7435, opened 2025-02-19
chore(logging): increase log-verbosity on analysis exceptions

aikebah:
Helps in analyzing cases like #7430 without having to resort to debug-logging when a build breaks before ODC writes the report with all the analysis exceptions

## Description of Change

Add the log message of the analysis exception so that the standard job-output can be helpful in case of analysis exceptions when no report is produced at the end.

Changes logs from e.g.

`[WARN] An error occurred while analyzing '/Users/aikebah/Projects/github.com/dependency-check/DependencyCheck/core/target/test-classes/python-poetry-toml/pyproject.toml' (Poetry Analyzer).`

to

`[WARN] An error occurred while analyzing '/Users/aikebah/Projects/github.com/dependency-check/DependencyCheck/core/target/test-classes/python-poetry-toml/pyproject.toml' (Poetry Analyzer): Python `pyproject.toml` found and there is not a `poetry.lock` or `requirements.txt` - analysis will be incomplete`

## Related issues

- relates to #7430 

no

### DIFF
--- a/core/src/main/java/org/owasp/dependencycheck/AnalysisTask.java
+++ b/core/src/main/java/org/owasp/dependencycheck/AnalysisTask.java
@@ -87,7 +87,7 @@ public class AnalysisTask implements Callable<Void> {
             try {
                 analyzer.analyze(dependency, engine);
             } catch (AnalysisException ex) {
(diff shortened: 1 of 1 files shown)


jeremylong (approved):
lgtm

PR #7469, opened 2025-02-25
fix: Also apply muteNoisyLoggers to UpdateMojo

qsiebers:
## Description of Change

The `update-only` mojo also has some log spamming that can be disabled by using `BaseDependencyCheckMojo.muteNoisyLoggers`

## Related issues

None as far as I know or found

Manually tested, not sure if there is a test for log spam


jeremylong (approved):
LGTM

PR #7295, opened 2025-01-11
fix: add product evidence as vendor to reduce FN

jeremylong:
## Description of Change

Adds product evidence to the vendor for several analyzers that did not already do this.

## Related issues

- resolves #7269

Issue #2972, opened 2020-11-24
Composer packages with hyphen not resolved to correct CPE


jackbentley:
**Describe the bug**
We currently have `php-ampqlib` as a composer dependency in our project at version `2.6.3`. When running the dependency check, lots of vulnerabilities are erroneously listed against the package.

This is because the CPE assigned is `cpe:2.3:a:php:php:2.6.3:*:*:*:*:*:*:*` which is for the main PHP library. Which means we end up with hundreds of vulns that aren't even related.

**Version of dependency-check used**
6.0.3

**Log file**
_currently don't have but can be provided_

**To Reproduce**
Steps to reproduce the behaviour:
1. Require `php-ampqlib` as a dependency in a composer project
2. Run DependencyCheck on the project

**Expected behaviour**
Only vulns related to the package should be shown - not those of PHP entirely

**Additional context**
N/A

jackbentley:
I've got a log file but its >1GB in size, so won't be uploading it. However, searching the log file reveals these interesting snippets:

```
2020-11-24 16:44:43,018 org.owasp.dependencycheck.analyzer.CPEAnalyzer:261
DEBUG - vendor search: {php-amqplib=1}
2020-11-24 16:44:43,018 org.owasp.dependencycheck.analyzer.CPEAnalyzer:263
DEBUG - product search: {php-amqplib=1}
2020-11-24 16:44:43,018 org.owasp.dependencycheck.data.cpe.AbstractMemoryIndex:273
DEBUG - product:(php\-amqplib) AND vendor:(php\-amqplib)
2020-11-24 16:44:43,039 org.owasp.dependencycheck.analyzer.CPEAnalyzer:281
DEBUG - identified vendor/product: php/php
```

Note a package by the name of `php-font-lib` is also searched. But the vendor is not the same name:
```
2020-11-24 16:44:43,015 org.owasp.dependencycheck.analyzer.CPEAnalyzer:261
DEBUG - vendor search: {phenx=1}
2020-11-24 16:44:43,015 org.owasp.dependencycheck.analyzer.CPEAnalyzer:263
DEBUG - product search: {php-font-lib=1}
2020-11-24 16:44:43,015 org.owasp.dependencycheck.data.cpe.AbstractMemoryIndex:273
DEBUG - product:(php\-font\-lib) AND vendor:(phenx)
2020-11-24 16:44:43,015 org.owasp.dependencycheck.analyzer.CPEAnalyzer:261
[...]

PR #7486, opened 2025-02-28
fix: Avoid FPs when Composer product name has php

### LINKED ISSUE #7444: Significant Increase in Loose CPE Matches After Upgrading from v11.1.0 to v12.1.0
After upgrading from DependencyTrack v11.1.0 to v12.1.0, we noticed a sharp increase in CPE matches, leading to a flood of tickets for failures and suppression requests.

Upon further investigation, it appears that very loose CPE matches are causing this spike. For example, we’re seeing matches for any package with "Redis" in the name, even when they are unrelated. 

Example Packages matched:
```
pkg:nuget/Microsoft.AspNetCore.DataProtection.Redis@0.3.3
pkg:nuget/Hangfire.Redis.StackExchange@1.9.4
pkg:nuget/EasyCaching.Redis@1.9.2
pkg:nuget/EFCoreSecondLevelCacheInterceptor.StackExchange.Redis@5.0.0
pkg:nuget/StackExchange.Redis@2.8.24
```
The CPE match shown in the HTML report typically looks like this:
```
cpe:2.3:a:redis:redis:5.0.0:::::::* (Confidence: Low)
```

We have reviewed the [changelog](https://github.com/dependency-check/DependencyCheck/blob/main/CHANGELOG.md), but we haven’t found any changes that directly explain this behavior.

Could you provide any insight into what might have changed in v12.1.0 regarding CPE matching? Is there a way to adjust the matching logic or fine-tune the confidence threshold to avoid these false positives?

Thanks in advance for your help!

sigv:
## Description of Change

Previously, only PHP package's vendor (product URL's namespace) was considered as evidence. As of DependencyCheck v12, specifically change from #7295, the product (name) is also being considered as evidence.

This results in new false positives, as noticed in #7444. PHP Composer checks are affected, for example, considering package `pkg:composer/phpunit/php-timer@6.0.0` as `cpe:2.3:a:php:php:6.0:*:*:*:*:*:*:*` resulting in 17 CVEs (including Critical).

This commit adds two new suppression rules: one for php as the prefix, and one for php as the suffix. Both can be observed in the wild. Additionally, underscore is sometimes used instead of hyphen, and should be respected. Furthermore, there is `symfony/polyfill-php83` which adds number suffix, which should also be suppressed as it currently maps to the base `cpe:/a:php:php`.

## Related issues

- relates to #7444
- relates to #7295
- relates to #2972

no

### DIFF
--- a/core/src/main/resources/dependencycheck-base-suppression.xml
+++ b/core/src/main/resources/dependencycheck-base-suppression.xml
@@ -361,8 +361,26 @@
     <suppress base="true">
         <notes><![CDATA[
         FP per #2972
+        hyphenated PHP library vendor names
         ]]></notes>
         <packageUrl regex="true">^pkg:composer/php\-.*$</packageUrl>
+        <cpe>cpe:/a:php:php</cpe>
+    </suppress>
+    <suppress base="true">
+        <notes><![CDATA[
+        FP per #2972 + #7444
+        hyphenated PHP library product names (prefix)
+        ]]></notes>
+        <packageUrl regex="true">^pkg:composer/[^/]+/php[\-_].*$</packageUrl>
+        <cpe>cpe:/a:php:php</cpe>
+    </suppress>
+    <suppress base="true">
+        <notes><![CDATA[
+        FP per #2972 + #7444
+        hyphenated PHP library product names (suffix)
+        including number suffix, e.g., `symfony/polyfill-php80`
+        ]]></notes>
+        <packageUrl regex="true">^pkg:composer/[^/]+/.*[\-_]php[0-9]*@.*$</packageUrl>
         <cpe>cpe:/a:php:php</cpe>
     </suppress>
     <suppress base="true">


sigv:
I wonder if `-php` suffix should also be added as a suppression, while we are at it. Some packages don't add `php` as prefix, but instead add it at the end...

Then there's Symfony's polyfill, which for example is `pkg:composer/symfony/polyfill-php80@1.30.0` and gets considered as `cpe:2.3:a:php:php:1.30.0:*:*:*:*:*:*:*` reporting 319 CVEs (including Critical, and under active exploitation). The PHP v1.30 detection is a fairly obvious false positive. There are `pkg:composer/symfony/polyfill-php72@1.30.0` and `pkg:composer/symfony/polyfill-php83@1.30.0` so the URL expression could be generalized as `^pkg:composer/[^/]+/.*\-php[0-9]*@.*$`. 🤔

aikebah:
Thanks for the PR

aikebah (approved):

PR #7511, opened 2025-03-07
docs: correct gradle documentation

jeremylong:
centralEnabled should be documented as defaulting to true.


aikebah:
Should also be done for maven plugin docs

jeremylong:
Maven has `analyzer.central.enabled=false`.

aikebah:
> Maven has `analyzer.central.enabled=false`.

ah yes... I forgot to check the mojo.properties

aikebah (approved):

PR #7513, opened 2025-03-08
fix: disables automatic content decompression

### LINKED ISSUE #7475: java.util.zip.ZipException: Not in GZIP format when using a ndv datafeed mirror created with vulnz
**Precondition**

**Describe the bug**
Downloading nvd datafeed from a mirror created with vulnz fail with the following error : 
java.util.zip.ZipException: Not in GZIP format

In detail : 
I created a bitbucket page that mirror the nvd feed. The mirror is created with the vulnz docker image (steps: create cache with vulnz, git commit and push content, mirror is served by bitbucket page).
I then added nvdDataFeedUrl in the another project pom.xml pointing to the page.
I am able to check the urls are indeed correct, as I can download the json.gz with my browser.
dependency-check-maven (and I suppose others too) fail with the error above

if I check the http header I get from the server when I used my browser, I can see that 
content-encoding is gzip
content-type is application/json

My best guess is that both client and server agree on handling gzip and the client ends up with an uncompressed json and then try to uncompress it again

The problem occurs using version 12.1.0 of the maven plugin (docker: maven:3.9.6-eclipse-temurin-17-alpine)

**Log file**
https://gist.github.com/JoKoT3/5a077a3a5d2d50c3abc17eb04c96f35d

**To Reproduce**
1. create a mirror with vulnz
2. serve said mirror with HTTP/2 with gzip content-encoding (bitbucket pages appear to use cloudfront)
3. set the URL for the mirror in nvdDataFeedUrl 
4. run dependency-check in a way that the database has to be populated
5. see error

**Expected behavior**
Fix may be to implement a catch on the java.util.zip.ZipException and try to load it as an uncompressed json

**Additional context**
I can provide my personnal mirror URL in private for testing purposes

jeremylong:
resolves #7475

See https://hc.apache.org/httpcomponents-client-4.5.x/current/httpclient/apidocs/org/apache/http/impl/client/HttpClientBuilder.html#disableContentCompression()

### DIFF
--- a/utils/src/main/java/org/owasp/dependencycheck/utils/Downloader.java
+++ b/utils/src/main/java/org/owasp/dependencycheck/utils/Downloader.java
@@ -144,10 +144,12 @@ public final class Downloader {
         //TODO: ensure proper closure and eviction policy
         httpClientBuilder = HttpClientBuilder.create()
                 .useSystemProperties()
+                .disableContentCompression()
                 .setConnectionManager(connectionManager)
                 .setConnectionManagerShared(true);
         httpClientBuilderExplicitNoproxy = HttpClientBuilder.create()
                 .useSystemProperties()
+                .disableContentCompression()
                 .setConnectionManager(connectionManager)
                 .setConnectionManagerShared(true)
                 .setProxySelector(new ProxySelector() {


aikebah:
Don't think this would be the correct fix. HTTP compression https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Encoding is supposed to be used for compress-at-server-decompress-at-client to preserve network bandwidth. It could be used for any file.

> If the original media is already encoded (e.g., as a .zip file), this information is not included in the Content-Encoding header.

When `.gz` files are served up by a server with a Content-Encoding gzip header set they should be double-compressed.

I would be curious to know as to whether the bitbucket mirror of the source issue is really hosting the gz files.

aikebah:
@jeremylong I got confused by the PR title. Your change would not disable automatic content decompression, but would prevent the entire use of content compression.

It would increase network traffic overall for any resource that would normally use content compression on the conmection.

In my view the root of the issue is still a (proxy?) configuration error that serves an explicit request for a (binary) (.json.).gz file wrongly as a .json file served with gzip content-encoding when a client indicates that it knows how to handle gzip content-encoding.

aikebah:
As an example it would increase data-traffic for the KEV json from 127_750 to 1_154_748 bytes

Results with some added tracing of the headers with and without the fix shows that apparently KEV serving is optimized for gzipped content-encoding as it switches to a `Transfer-Encoding: chunked` streaming serving of the file when requested for the resource without having gzip compression available as an accepted encoding:

With your fix:
```
[INFO] Updating CISA Known Exploited Vulnerability list: https://www.cisa.gov/sites/default/files/feeds/known_exploited_vulnerabilities.json
Sending header - Host: www.cisa.gov
Sending header - Connection: keep-alive
Sending header - User-Agent: Apache-HttpClient/5.4.2 (Java/21.0.6)
Protocol version: HTTP/1.1
Received header - Content-Type: application/json
Received header - Server: Apache
Received header - X-Content-Type-Options: nosniff
Received header - Last-Modified: Fri, 04 Apr 2025 18:55:10 GMT
Received header - ETag: "119ebc-631f86d00a468"
Received header - Cache-Control: max-age=2803
Received header - Expires: Sun, 06 Apr 2025 13:10:24 GMT
Received header - Date: Sun, 06 Apr 2025 12:23:41 GMT
Received header - Transfer-Encoding: chunked
Received header - Connection: keep-alive
Received header - Connection: Transfer-Encoding
Received header - Strict-Transport-Security: max-age=31536000 ; includeSubDomains
```

Without your fix (`.disableContentCompression()` commented out):
```
[INFO] Updating CISA Known Exploited Vulnerability list: https://www.cisa.gov/sites/default/files/feeds/known_exploited_vulnerabilities.json
Sending header - Accept-Encoding: gzip, x-gzip, deflate
Sending header - Host: www.cisa.gov
Sending header - Connection: keep-alive
Sending header - User-Agent: Apache-HttpClient/5.4.2 (Java/21.0.6)
Protocol version: HTTP/1.1
Received header - Content-Type: application/json
Received header - Server: Apache
Received header - X-Content-Type-Options: nosniff
Received header - Last-Modified: Fri, 04 Apr 2025 18:55:10 GMT
Received header - ETag: "119ebc-631f86d00a468"
Received header - Accept-Ranges: bytes
Received header - Content-Encoding: gzip
Received header - Content-Length: 127750
Received header - Cache-Control: max-age=3069
Received header - Expires: Sun, 06 Apr 2025 13:10:24 GMT
Received header - Date: Sun, 06 Apr 2025 12:19:15 GMT
Received header - Connection: keep-alive
Received header - Vary: Accept-Encoding
Received header - Strict-Transport-Security: max-age=31536000 ; includeSubDomains
```

PR #7521, opened 2025-03-12
docs: drop redundant text in the Internet Access Required section

brlin-tw:
## Description of Change

This pull request fixes a trivial writing error in the introductory text of the [Internet Access Required – dependency-check-maven](https://dependency-check.github.io/DependencyCheck/data/index.html) page.

Please take a look at the diff for the problem.

## Related issues

N/A

N/A

### DIFF
--- a/src/site/markdown/data/index.md
+++ b/src/site/markdown/data/index.md
@@ -1,6 +1,6 @@
 # Internet Access Required
 
-Dependency-check requires access, by default, to several externally
+Dependency-check, by default, requires access to several externally
 hosted resources.
 
 ## The NVD Database
--- a/src/site/markdown/data/index.md
+++ b/src/site/markdown/data/index.md
@@ -1,6 +1,6 @@
 # Internet Access Required
 
-Dependency-check requires access, by default, requires access to several externally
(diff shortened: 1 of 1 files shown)


aikebah:
I should mention I'm not a native English speaker, but I do get the impression that you dropped the wrong duplicate for arriving at non-redundant language.

@jeremylong would the proposed new text be correct to native speakers or should we drop the initial `requires access` and leave the one that was removed in place?

src/site/markdown/data/index.md
```diff
@@ -1,6 +1,6 @@
 # Internet Access Required
 
-Dependency-check requires access, by default, requires access to several externally
+Dependency-check requires access, by default, to several externally
```
jeremylong:
```suggestion
Dependency-check, by default, requires access to several externally
```

jeremylong: @aikebah I'm an American - so my English isn't very good either ;)

jeremylong (approved):

brlin-tw:
@aikebah Thanks for the correction!

PR #7523, opened 2025-03-12
fix: Ignore require-bundle MANIFEST.MF entry for evidence

### LINKED ISSUE #7518: Manifest entries that describe dependencies should not be used as evidence for CPEs
**Precondition**
- [ x] I checked the issues list for existing open or closed reports of the same problem.

**Describe the bug**
A java library that contains this OSGI entry in the /META-INF/MANIFEST.MF file

`Require-Bundle: com.google.guava,org.eclipse.xtext,org.eclipse.xtext.generator;resolution:=optional,org.apache.commons.logging;....`

gets Evidence entries in the html report like:
_Product or Vendor | Manifest | require-bundle com.google.guava ..._
and therefore is detected as
_cpe:2.3:a:google:guava:6.28.0:snapshot:*:*:*:*:*:*  (Confidence:Low)_
which is a false positive.

The problem occurs using version 12.1.0 of the maven plugin

**Expected behavior**
Java Manifest entries that describe dependencies like "Require-Bundle" should not be used as evidence for CPEs to avoid false positives.

aikebah:
Fixes #7518

## Description of Change

Add the require-bundle key to the Manifest keys that are ignored during evidence-gathering

## Related issues

- fixes #7518 

no

### DIFF
--- a/core/src/main/java/org/owasp/dependencycheck/analyzer/JarAnalyzer.java
+++ b/core/src/main/java/org/owasp/dependencycheck/analyzer/JarAnalyzer.java
@@ -160,7 +160,8 @@ public class JarAnalyzer extends AbstractFileTypeAnalyzer {
             "dstamp",
             "eclipse-sourcereferences",
             "kotlin-version",
-            "require-capability");
+            "require-capability",
+            "require-bundle");
     /**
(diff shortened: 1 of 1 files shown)


jeremylong (approved):
LGTM

PR #7530, opened 2025-03-15
fix: Use unscored severity only in absence of any CVSS baseScore

### LINKED ISSUE #7528: Build fails at CVSS 9.8 even when failOnCVSS is set to 10.0 for cve-2021-23369
Hi,

I set failOnCVSS to 10.0 in my CI/CD pipeline, but still fails on a vulnerability with a CVSS score of 9.8.

Example from the logs:

```
[ERROR] 

One or more dependencies were identified with vulnerabilities that have a CVSS score greater than or equal to '10.0': 
-------
yarn.lock?ansi-html (pkg:npm/ansi-html@0.0.7): GHSA-whgm-jr23-g3j9(7.5)

yarn.lock?async (pkg:npm/async@2.4.1): GHSA-fwr7-v2mv-hh25(7.800000190734863)
....................
=> "many more yarn.lock info"....
--------
handlebars-4.0.5.js (pkg:javascript/handlebars@4.0.5): Versions of `handlebars` prior to 3.0.8 or 4.5.2 are vulnerable to Arbitrary Code Execution. The package's lookup helper fails to properly validate templates, allowing attackers to submit templates that execute arbitrary JavaScript in the system. It can be used to run arbitrary code in a server processing Handlebars templates or on a victim's browser (effectively serving as Cross-Site Scripting).
``` 
I even updated the handlebars package to the patched version, but I’m still getting these logs.

🔹 Why is the build failing at CVSS 9.8 when the threshold is set to 10.0?
🔹 Could this be a rounding issue or a problem with float precision?

By the way, I'm using the official Docker image owasp/dependency-check-action.

Any insights would be greatly appreciated. Thanks! 😊

aikebah:
fixes #7528

## Description of Change

Only use the fabricated high-watermark guestimated CVSSv2 score for breaking the build when there is not any datasource that has published a CVSS basescore for the vulnerability.

- fixes #7528 

no

### DIFF
--- a/ant/src/main/java/org/owasp/dependencycheck/taskdefs/Check.java
+++ b/ant/src/main/java/org/owasp/dependencycheck/taskdefs/Check.java
@@ -1524,12 +1524,23 @@ public class Check extends Update {
         for (Dependency d : dependencies) {
             boolean addName = true;
             for (Vulnerability v : d.getVulnerabilities()) {
-                if ((v.getCvssV2() != null && v.getCvssV2().getCvssData().getBaseScore() >= failBuildOnCVSS)
-                        || (v.getCvssV3() != null && v.getCvssV3().getCvssData().getBaseScore() >= failBuildOnCVSS)
-                        || (v.getCvssV4() != null && v.getCvssV4().getCvssData().getBaseScore() >= failBuildOnCVSS)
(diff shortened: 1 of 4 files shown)


jeremylong (approved):
LGTM

PR #7544, opened 2025-03-22
fix: prevent rogue base suppression files

jeremylong:
## Description of Change

PR #7541 showed the maintainers that it was possible for a rogue library to add a suppression file

No, existing test cases appropriately load the suppression file. A negative test case was not added.

### DIFF
--- a/core/src/main/java/org/owasp/dependencycheck/analyzer/AbstractSuppressionAnalyzer.java
+++ b/core/src/main/java/org/owasp/dependencycheck/analyzer/AbstractSuppressionAnalyzer.java
@@ -197,39 +197,23 @@ public abstract class AbstractSuppressionAnalyzer extends AbstractAnalyzer {
      */
     private void loadPackagedSuppressionBaseData(final SuppressionParser parser, final Engine engine) throws SuppressionParseException {
         List<SuppressionRule> ruleList = null;
-        Iterator<URL> urls = null;
-        try {
-            urls = FileUtils.getResources(BASE_SUPPRESSION_FILE);
-        } catch (IOException e) {
-            LOGGER.warn("Base suppression rules `{}}` could not be loaded; {}", BASE_SUPPRESSION_FILE, e.getMessage());
-            return;
-        }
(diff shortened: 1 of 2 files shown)


aikebah (dismissed):
LGTM

core/src/main/java/org/owasp/dependencycheck/analyzer/AbstractSuppressionAnalyzer.java
aikebah, line 1: Does the job just fine, but I think it would be more logical to make the URL of the codeSource be leading in deciding upon expecting a `file:`-URL (only in case of an unpacked ODC codebase loaded from a folder-structure) or a `jar:file:` URL (the typical to be expected case).

When retaining the current logic I think the validationBase variable should be split into its two responsibilities for code clarity: the expected jar or folder to load from (which it gets reset to on every loop iteration in the current code which could be obtained outside the loop) and the expected full location of the base-suppression-file (the final value at the end of a loop iteration)

aikebah: A naive (not check actual file/folder, just assuming a .jar is a file and any other ending of the URL is a folder) implementation of the former could be 

```java
        final String expectedBaseSuppressionURL;
        if (loc.toString().endsWith(".jar")) {
            expectedBaseSuppressionURL = "jar:" + loc + "!/" + BASE_SUPPRESSION_FILE;
        } else {
            expectedBaseSuppressionURL = loc + BASE_SUPPRESSION_FILE;
        }
```

jeremylong: The code has been updated.

aikebah:
Looking once more at the code I was thinking... why are we even bothering to go through so many loopholes when the only thing we want to load in the end is the resource we expect to find.

Can you see any reason to not simply apply a KISS principle and try to load 'the resource-URL we expect based on the codesource' rather than validating the resource-URLs discovered by java and only load when the resource-URL matches what we expect to find?

jeremylong:
Thanks for the suggestion. See the updated code.

aikebah (approved):
LGTM

PR #7546, opened 2025-03-24
fix: Avoid FPs for Symfony Contracts as framework

### LINKED ISSUE #7545: [FP]: Symfony Contracts are matched as Symfony framework
### Package URl

pkg:composer/symfony/cache-contracts@3.5.1

### CPE

cpe:2.3:a:sensiolabs:symfony:3.5.1:*:*:*:*:*:*:*

### CVE

CVE-2022-23601

### ODC Integration

{"label" => "CLI"}

### ODC Version

12.1.0

### Description

Since #7295, [Symfony Contracts](https://github.com/symfony/contracts) are being matched as [Symphony framework](https://github.com/symfony/symfony). However, they are different projects with independent versioning schemes. The reported CVE is resolved in versions 5.3.15, 5.4.4 and 6.0.4. However, Contracts project's latest version is v3.5.2.

sigv:
## Description of Change

[Symfony Contracts](https://github.com/symfony/contracts) are being matched as [Symphony framework](https://github.com/symfony/symfony). However, they are different projects with independent versioning schemes. For example, CVE-2022-23601 for Symfony is resolved in versions 5.3.15, 5.4.4 and 6.0.4. However, Contracts project's latest version is v3.5.2.

## Related issues

- fixes #7545
- relates to #7444
- caused by #7295

*no*

### DIFF
--- a/core/src/main/resources/dependencycheck-base-suppression.xml
+++ b/core/src/main/resources/dependencycheck-base-suppression.xml
@@ -392,6 +392,13 @@
     </suppress>
     <suppress base="true">
         <notes><![CDATA[
+        FP per #7545
+        ]]></notes>
+        <packageUrl regex="true">^pkg:composer/symfony/.*-contracts@.*$</packageUrl>
+        <cpe>cpe:/a:sensiolabs:symfony</cpe>
+    </suppress>
+    <suppress base="true">
+        <notes><![CDATA[
         FP per #2957
         ]]></notes>
         <packageUrl regex="true">^pkg:maven/org\.jboss\.resteasy/resteasy\-spring\-boot\-starter@.*$</packageUrl>


chadlwilson:
Pro tip in case you’re not aware: if you branch from and PR to the file within the `generatedSuppressions` branch the suppression can be made available independent of an ODC release.

jeremylong (approved):

sigv:
> Pro tip in case you’re not aware: if you branch from and PR to the file within the `generatedSuppressions` branch the suppression can be made available independent of an ODC release.

Thank you, will keep that in mind!

sigv:
@chadlwilson, does `generatedSuppressions` get pulled into `main` prior to release, to keep the bundled file up-to-date, or what is the standard flow for it?

I see [`gh-pages` branch's suppressions/publishedSuppressions.xml](https://github.com/dependency-check/DependencyCheck/blob/gh-pages/suppressions/publishedSuppressions.xml) gets updated from `generatedSuppressions` branch... but there seems to be divergence between what is in `main` and what is in `generatedSuppressions` 🤔

sigv:
Asking as there seems to be divergence between what is in `main` and what is in `generatedSuppressions` 🤔

chadlwilson:
Not via automation. It augments what’s in main for anyway that has not disabled use of hosted suppressions.

They are essentially two different data sources.

Hosted suppressions are auto-released any time a suppression is merged from the automated FP issue workflows (for which there is not one for php, but is for maven, gradle, dotnet, etc).

chadlwilson:
The problem is that the hosted suppression changes can be used by all versions, yet changes to bundled only by those on latest release. People keep reporting FPs until they update to a given version which is a pain to keep track of as a project collaborator or even as user who is trying to minimize their own suppressions (which in many ends can require time limits and discussions with internal bureaucracy).

In my personal view there should be no bundled suppressions and everything should be done in the hosted suppressions, but that’s not a discussion I’ve seen had anywhere (or one I’ve brought up).

The two data sources right now create a risk of conflicting rules and too many combinations to think about when you contemplate all the various historical bundled suppressions that exist. Makes it difficult to investigate false negatives, inefficient suppressions etc.

sigv:
This feels worth discussing as a future improvement, and I can understand the idea of shifting towards "hosted suppressions only". So I'm opening #7547 to make sure this is logged as a potential enhancement.

PR #7629, opened 2025-05-04
fix: Resolve various WCAG accessibility / css issues in the HTML report

### LINKED ISSUE #7618: Accessibility findings and duplicate CSS selector in dependency-check-report.html
SonarQube demands 3 fixes to some HTML elements of dependency-check-report.html which is produced by the current version dependency-check 12.1.1. I know sometimes it's really tedious to work on accessibility issues, but at least at our company we try to do the best to serve all possible users. Would be great if this could be fixed.

Rated as a major bug is: Add "lang" and/or "xml:lang" attributes to the `"<html>"` element at the beginning of the file.
Reason: `"<html>"` element should have a language attribute due to accessibility issues.

Rated as a minor bug is: Add a description to this `<table id="summaryTable" class="lined">` which is found towards the end of the file.
Reason: `"<table>"` tags should have a description due to accessibility issues.

Rated as a major code-smell is: duplicate selector ".hidden", actually this CSS element is specified two time
Reason: selectors should not be duplicated, possible copy-paste mistake

aikebah:
## Description of Change

* Set the html lang attribute
* Adapt the original summary text intro into one that is useable to provide the table description as an aria-labelledby and adapt the vulnerabilityDisplayToggle onClick event accordingly
* Remove the duplicated .hidden css class rule
* Increased contrast-ratio of the disclaimer text from 3.85 (not WCAG compliant) to 7.0 (WCAG AAA compliant)

Other items reported by Firefox are not taken up now, these concern the lack of proper keyboard support in the JQuery plugin that is currently used to make the summary table sortable. Resolving those would require replacement of the plugin by some other plugin that would offer full WCAG compliancy.

## Related issues

- fixes #7618 

no, some base testing performed with Firefox and Safari using MacOS VoiceOver to check that navigation works and the table description gets used as expected.

### DIFF
--- a/core/src/main/resources/templates/htmlReport.vsl
+++ b/core/src/main/resources/templates/htmlReport.vsl
@@ -211,7 +211,14 @@ Copyright (c) 2012 Jeremy Long. All Rights Reserved.
             } else {
                 dependenciesHeader.text(dependenciesHeaderAllText);
             }
-            return toggleDisplay(event.target, '.notvulnerable', 'Vulnerable Dependencies (click to show all)', 'All Dependencies (click to show less)');
+            const plainCaptionPart = $('#tablecaption-plain');
+            const plainCaptionPartAllText = 'Summary of All Dependencies';
+            if (plainCaptionPart.text() == plainCaptionPartAllText) {
+                plainCaptionPart.text('Summary of Vulnerable Dependencies');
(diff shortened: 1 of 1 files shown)


jeremylong (approved):
LGTM

PR #7631, opened 2025-05-05
feat: #7482 offer ossindex credentials for maven

### LINKED ISSUE #7482: Offer OSS Index user/pw for Maven
Related to #4533.
The CLI and Gradle clients offer to set the OSS Index user and password. For the Maven client only the indirection through server ID and `settings.xml` is currently supported.

For Maven, allow to either set user/pw _or_ server ID.

It is obvious that using the server ID is usually a safer setup. However, this assumes you have access to or control over the `settings.xml`. This is not always the case. Example: corporate CI infrastructure with Maven settings controlled by IT.

You can still include the OSS Index for ODC _safely_ in your pipeline if you do something like `mvn -U org.owasp:dependency-check-maven:$ODC_VERSION:aggregate -DossIndexUser=$OSS_INDEX_USER ... `. The actual value for `$OSS_INDEX_USER` would be stored as CI env variable with your project (e.g. in GitLab). 

If the `settings.xml` isn't read-only, the pipeline might try to alter it prior to running ODC.

**Additional context**
@jeremylong voiced potential "endorsement" for this feature here https://github.com/dependency-check/DependencyCheck/issues/4533#issuecomment-2535553786

nhumblot:
## Description of Change

This PR is a draft. Identified things that must be checked before marking it as ready:

- Tests
- Explicitly testing to prevent a new issue in the form of https://github.com/dependency-check/DependencyCheck/security/advisories/GHSA-qqhq-8r2c-c3f5

This PR aims to implement #7482 as expressed into the initial comment of the discussion.

Added change to the documentation as is:

Capture d’écran du 2025-05-05 20-33-33

## Related issues

- fixes #7482

*yes/no*: no as of now, but has to be implemented before marking the PR as ready.


aikebah:
@nhumblot Your feature implemenation should include the environmentvariable option similar to how its done for the NVD API (which was unfortunately not added to the maven doc markdown (we should fix that) but is being documented in the standard maven plugin docs at https://dependency-check.github.io/DependencyCheck/dependency-check-maven/check-mojo.html#nvdApiKeyEnvironmentVariable

Specifying the environment variable name is a safer way to pass them than directly in the command-line (typically the env variable itself would be provided by setting it as a CI secret) given that maven will always log unmasked plaintexts on the `mvn -X` executions.

Configuring the ossIndexPassword property as an env variable reference would not help, as it would get interpolated by the shell and by the time maven logs it (in the `-X`-case) it would already be a plaintext password.

Wording for the ossIndexPassword should be similar to that of the ApiKey setting nudging the users to prefer either the settings.xml or the env variablename configuration option over specifying a password directly that is guaranteed to be exposed by mvn -X https://dependency-check.github.io/DependencyCheck/dependency-check-maven/check-mojo.html#nvdApiKey

nhumblot:
Duplicate of #7640

PR #7691, opened 2025-05-29
docs: Make `Vulnerability Sources` in `Related Work` clearer

JackPGreen:
The documentation is confusing because NVD is both listed as the exclusive source _and_ an additional source "that may not be included".

Updated to clarify.


nhumblot (approved):
Hi! :wave: 

Thank you for your contribution!

jeremylong (approved):

PR #7693, opened 2025-05-30
fix: Simplify PHP framework suppression for Composer

sigv:
## Description of Change

The base suppression that is bundled for suppressing `php` in packages names should be expanded, as mentioned in [comment by @georgschoelly on #7444](https://github.com/dependency-check/DependencyCheck/issues/7444#issuecomment-2879279125).

All matches of PHP in Composer packages are safe to consider not being part of PHP framework `cpe:/a:php:php` because Composer is the package manager next level down.

This change removes three distinct suppressions, which are now covered by the extended regular expression.

## Related issues

- relates to #2972 
- relates to #7444 

*no*

### DIFF
--- a/core/src/main/resources/dependencycheck-base-suppression.xml
+++ b/core/src/main/resources/dependencycheck-base-suppression.xml
@@ -360,27 +360,11 @@
     </suppress>
     <suppress base="true">
         <notes><![CDATA[
-        FP per #2972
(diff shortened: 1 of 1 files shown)


sigv:
This is intentionally filed against `main` and not `generatedSuppressions` so that the existing three rules are merged together.

jeremylong (approved):
LGTM

PR #7705, opened 2025-06-03
fix: remove vulnerable transitive dependency - beanutils

### LINKED ISSUE #7704: CVE-2025-48734
Hi,

Are there plans to bump dependency for common-beanutils to v1.11.0 in upcoming releases?
CVE-2025-48734 has been flagged out in Dependency check v12.1.0

Trivy Scan:
   {
          "VulnerabilityID": "CVE-2025-48734",
          "PkgName": "commons-beanutils:commons-beanutils",
          "PkgPath": "usr/share/dependency-check/lib/commons-beanutils-1.9.4.jar",
          "PkgIdentifier": {
            "PURL": "pkg:maven/commons-beanutils/commons-beanutils@1.9.4",
            "UID": "e6a50b692f42c7fb"
          },
          "InstalledVersion": "1.9.4",
          "FixedVersion": "1.11.0",
          "Status": "fixed",

jeremylong:
resolves #7704

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -992,6 +992,12 @@ Copyright (c) 2012 - Jeremy Long
                 <groupId>commons-validator</groupId>
                 <artifactId>commons-validator</artifactId>
                 <version>1.9.0</version>
+                <exclusions>
+                    <exclusion>
+                        <groupId>commons-beanutils</groupId>
+                        <artifactId>commons-beanutils</artifactId>
+                    </exclusion>
+                </exclusions>
             </dependency>
             <dependency>
                 <groupId>org.apache.commons</groupId>


jeremylong:
@aikebah see any issues with this one?

aikebah:
@jeremylong Looks fine to me, but given that only RetireJS uses commons-validator, and uses it only for a URL validity check we might even fully remove commons-validator dependency as a whole and replace it by our own helper method.

For the purposes of reference URL checking I think it would suffice to do a non-null check followed by construction of a URI for non-null strings and a call to `Uri#toUrl()`and treat URISyntaxException and MalformedURLException as indicators that the string is not a valid URL (return false from the helper method) . It would not introduce new try-catch guards, as commons-validator is also constructing a URI catching URISyntaxException as one of the early parts of its (more extensive) checks on URL validity.

It would save future version-bumps on commons-validator releases and reduces the transitive dependency set at runtime (commons-beanutils itself would still be present as a test-dependency of mockserver (via velocity-tools-generic) in the dependency-check-utils module)

aikebah (approved):

PR #7771, opened 2025-06-26
chore: add marcelstoer to the FP approvers list

marcelstoer:
I'm hesitant to propose this...No harm done if you reject it.


jeremylong (approved):
LGTM

PR #7784, opened 2025-07-03
fix: Add null checking when parsing the license json in AbstractNpmAnalyzer.

### LINKED ISSUE #7783: NullPointerException on NodePackageAnalyzer license handling
**Precondition**

**Describe the bug**
Similar to https://github.com/dependency-check/DependencyCheck/issues/4293

And: https://github.com/dependency-check/DependencyCheck/issues/5318

There's an NPE in the AbstractNpmAnalyzer when the license is improperly defined: https://github.com/dependency-check/DependencyCheck/blob/main/core/src/main/java/org/owasp/dependencycheck/analyzer/AbstractNpmAnalyzer.java#L397

We use a quite old dependency hammerjs-time. This library misdefines the license block, using "name" instead of the described "type" field: https://github.com/hammerjs/hammer-time/blob/master/package.json#L18

Even older versions never used "name": https://docs.npmjs.com/cli/v6/configuring-npm/package-json#license

The problem occurs using version 12.1.3 of the maven-plugin

**Log file**
`
11:24:15  [WARNING] An unexpected error occurred during analysis of '<path>/package-lock.json' (Node.js Package Analyzer): Cannot invoke "jakarta.json.JsonString.getString()" because the return value of "org.glassfish.json.JsonObjectBuilderImpl$JsonObjectImpl.getJsonString(String)" is null
11:24:15  [ERROR] 
11:24:15  java.lang.NullPointerException: Cannot invoke "jakarta.json.JsonString.getString()" because the return value of "org.glassfish.json.JsonObjectBuilderImpl$JsonObjectImpl.getJsonString(String)" is null
11:24:15      at org.glassfish.json.JsonObjectBuilderImpl$JsonObjectImpl.getString (JsonObjectBuilderImpl.java:259)
11:24:15      at org.owasp.dependencycheck.analyzer.AbstractNpmAnalyzer.gatherEvidence (AbstractNpmAnalyzer.java:397)
11:24:15      at org.owasp.dependencycheck.analyzer.NodePackageAnalyzer.processDependencies (NodePackageAnalyzer.java:434)
11:24:15      at org.owasp.dependencycheck.analyzer.NodePackageAnalyzer.analyzeDependency (NodePackageAnalyzer.java:271)
11:24:15      at org.owasp.dependencycheck.analyzer.AbstractAnalyzer.analyze (AbstractAnalyzer.java:131)
11:24:15      at org.owasp.dependencycheck.AnalysisTask.call (AnalysisTask.java:88)
11:24:15      at org.owasp.dependencycheck.AnalysisTask.call (AnalysisTask.java:37)
11:24:15      at java.util.concurrent.FutureTask.run (FutureTask.java:317)
11:24:15      at java.util.concurrent.ThreadPoolExecutor.runWorker (ThreadPoolExecutor.java:1144)
11:24:15      at java.util.concurrent.ThreadPoolExecutor$Worker.run (ThreadPoolExecutor.java:642)
11:24:15      at java.lang.Thread.run (Thread.java:1583)
11:24:15  [INFO] Finished Node.js Package Analyzer (1 seconds)
`

**To Reproduce**
1. Create a package-lock.json that includes hammerjs library or other library which defines the license as a JSON object but doesn't set the "type" field
2. Run dependency check, will NPE since the license block is defined as a json but there's no 

**Expected behavior**
It should not NPE on this, but rather set the license to unknown.

**Additional context**
I will try to put in a fix shortly, we just need to check that the key exists before grabbing the value

floverfelt:
## Description of Change

Adds null checking when the license block is defined as a json object but the type is still missing

## Related issues

fixes: #7783 

no

### DIFF
--- a/core/src/main/java/org/owasp/dependencycheck/analyzer/AbstractNpmAnalyzer.java
+++ b/core/src/main/java/org/owasp/dependencycheck/analyzer/AbstractNpmAnalyzer.java
@@ -393,8 +393,11 @@ public abstract class AbstractNpmAnalyzer extends AbstractFileTypeAnalyzer {
                     }
                 }
                 dependency.setLicense(sb.toString());
-            } else {
-                dependency.setLicense(json.getJsonObject("license").getString("type"));
+            } else if (value instanceof JsonObject) {
+                final JsonObject object = (JsonObject) value;
+                if (object.containsKey("type") && !object.isNull("type")) {
+                    dependency.setLicense(object.getString("type"));
+                }
             }
         }
     }


jeremylong (approved):
LGTM

PR #7787, opened 2025-07-04
fix: class loading problem with fat jars (#7786)

### LINKED ISSUE #7786: Resource Loading Issues in fat jars
**Precondition**

**Describe the bug**
The initialization process fails, if the libraries (core,utils) are packed in a fat jar, like in Spring services.
The problem is the resource loading of some resources in the core lib.

ERROR: `Exception occurred initializing Vulnerability Suppression Analyzer`
following DEBUG with more information: 
* Error initializing the suppression analyzer: Unable to parse the base suppression data file
* Caused by: java.io.FileNotFoundException: nested:/app/app.jar/!BOOT-INF/lib/dependency-check-core-12.1.3.jar!/dependencycheck-base-suppression.xml (No such file or directory)

The problem occurs using version 12.1.3 of the core and utils (cli, gradle plugin, maven plugin, etc.)

**Log file**

https://gist.github.com/Thomas-Bergmann/4a272d45cd9a052c2a2e1e86e0f1533f#file-gistfile1-txt

**To Reproduce**
* it's a resource loading problem, so no steps to repeat

**Expected behavior**
* We would like to integrate your libraries in a fat jar like Spring applications doing it

**Additional context**

* we integrated the library in a service, to check former releases of our software
* our junit test are working, because these tests don't use the fat jar of the service
* many thanks for providing such a library

Thomas-Bergmann:
## Description of Change

Instead of using URL.openStream(), we propose to use the class or classloader getResourceAsStream to load resources from class path.

## Related issues

- fixes #7786

*no* - if you have any idea how to test a fat jar problem, I would be happy

### DIFF
--- a/core/src/main/java/org/owasp/dependencycheck/analyzer/AbstractSuppressionAnalyzer.java
+++ b/core/src/main/java/org/owasp/dependencycheck/analyzer/AbstractSuppressionAnalyzer.java
@@ -200,6 +200,10 @@ public abstract class AbstractSuppressionAnalyzer extends AbstractAnalyzer {
         String suppressionFileLocation = jarLocation.getFile();
         if (suppressionFileLocation.endsWith(".jar")) {
             suppressionFileLocation = "jar:file:" + suppressionFileLocation + "!/" + BASE_SUPPRESSION_FILE;
+        } else if (suppressionFileLocation.startsWith("nested:") && suppressionFileLocation.endsWith(".jar!/")) {
+            // suppressionFileLocation -> nested:/app/app.jar/!BOOT-INF/lib/dependency-check-core-<version>.jar!/
+            // goal->                 jar:nested:/app/app.jar/!BOOT-INF/lib/dependency-check-core-<version>.jar!/dependencycheck-base-suppression.xml
+            suppressionFileLocation = "jar:" + suppressionFileLocation + BASE_SUPPRESSION_FILE;
         } else {
             suppressionFileLocation = "file:" + suppressionFileLocation + BASE_SUPPRESSION_FILE;
         }


aikebah:
@Thomas-Bergmann Your current fix does not carry the core rationale behind the current loading code: to ensure that we only silently load (in addition to user-configured suppressions) project-vetted false positive suppressions that we made part of the codebase (or in rare cases: the similar named file that someone re-packaged together with our classes in one and the same jar-file).

Simply using the classloader (which we did in past) will load a resource from whichever jar-file is the first to contain a dependencycheck-base-suppression.xml resource. At the very least a similar guard as in the original proposed resourceloading in https://github.com/dependency-check/DependencyCheck/pull/7544 would be required to validate that we load the resources from the same jar that our own code originates from.

Based on your reported error I believe that original code that was simplified to direct the URL-based loading does not cover the case of your jar, which I wouldn't term as a fat-jar, but as a jar-with-jars. A fat-jar I consider a jar in which the contents of several original jars are merged together, which would perfectly survive the current code as those would use the regular file:-URL scheme. The problematic part in your scenario is the `nested:` URI scheme that your jar-with-jars capable classloader uses.

aikebah:
@Thomas-Bergmann looking at Spring's documentation of its nested classloading [1] and the reported error message I expect that it would only be needed to special-case a `nested:`-prefix detection inside the `.jar` suffix branch to compose a `jar:nested:` URL instead of a `jar:file:` URL from the CodeSource. Would you be willing to test and validate that it indeed makes things work in your setup and adapt your PR accordingly when successful?

As in: the delta would be reduced to only:

```diff
diff --git a/core/src/main/java/org/owasp/dependencycheck/analyzer/AbstractSuppressionAnalyzer.java b/core/src/main/java/org/owasp/dependencycheck/analyzer/AbstractSuppressionAnalyzer.java
index e5b928417..54d2b699e 100644
--- a/core/src/main/java/org/owasp/dependencycheck/analyzer/AbstractSuppressionAnalyzer.java
+++ b/core/src/main/java/org/owasp/dependencycheck/analyzer/AbstractSuppressionAnalyzer.java
@@ -199,7 +199,11 @@ public abstract class AbstractSuppressionAnalyzer extends AbstractAnalyzer {
         final URL jarLocation = AbstractSuppressionAnalyzer.class.getProtectionDomain().getCodeSource().getLocation();
         String suppressionFileLocation = jarLocation.getFile();
         if (suppressionFileLocation.endsWith(".jar")) {
-            suppressionFileLocation = "jar:file:" + suppressionFileLocation + "!/" + BASE_SUPPRESSION_FILE;
+            if (suppressionFileLocation.startsWith("nested:")) {
+                suppressionFileLocation = "jar:nested:" + suppressionFileLocation + "!/" + BASE_SUPPRESSION_FILE;
+            } else {
+                suppressionFileLocation = "jar:file:" + suppressionFileLocation + "!/" + BASE_SUPPRESSION_FILE;
+            }
         } else {
             suppressionFileLocation = "file:" + suppressionFileLocation + BASE_SUPPRESSION_FILE;
         }
```
[1] https://docs.spring.io/spring-boot/specification/executable-jar/jarfile-class.html

radanliev (approved):

Thomas-Bergmann:
Thx. good idea. But your proposal doesn't work directly. So I will investigate it.

aikebah (approved):
LGTM

aikebah:
@Thomas-Bergmann Thanks for the update to the PR and your local tests to get the nested URL properly working

PR #7813, opened 2025-07-13
build: remove unused code coverage

jeremylong:
I haven't looked at the code coverage reports in a VERY long time. I broke them when I was fixing the publishing by removing the wrong secret - and this is easier than recreating the secret for now...


jeremylong:
By haven't looked at the code coverage - I mean the published code coverage. I've used local reports.

Issue #7919, opened 2025-09-09
Update OSS Index documentation


framayo:
Sonatype [announced](https://ossindex.sonatype.org/doc/auth-required) that authentication will be soon required.

Update the OSS Index documentation with the steps to get an API token

PR #7920, opened 2025-09-09
fix: Update to support OSS Index Authentication Requirements

framayo:
## Description of Change

Sonatype [announced](https://ossindex.sonatype.org/doc/auth-required) that API token will be required soon.
Update the documentation to instruct users on how to get it.
image

## Related issues

relates to #7919

*no*

### DIFF
--- a/core/src/main/java/org/owasp/dependencycheck/analyzer/OssIndexAnalyzer.java
+++ b/core/src/main/java/org/owasp/dependencycheck/analyzer/OssIndexAnalyzer.java
@@ -134,7 +134,7 @@ public class OssIndexAnalyzer extends AbstractAnalyzer {
       synchronized (FETCH_MUTIX) {
         if (StringUtils.isEmpty(getSettings().getString(KEYS.ANALYZER_OSSINDEX_USER, StringUtils.EMPTY)) ||
             StringUtils.isEmpty(getSettings().getString(KEYS.ANALYZER_OSSINDEX_PASSWORD, StringUtils.EMPTY))) {
-          throw new InitializationException("Error initializing OSS Index analyzer due to missing user/password credentials.");
+          throw new InitializationException("Error initializing OSS Index analyzer due to missing user/password credentials. Authentication is now required: https://ossindex.sonatype.org/doc/auth-required");
         }
       }
     }
--- a/core/src/main/java/org/owasp/dependencycheck/analyzer/OssIndexAnalyzer.java
+++ b/core/src/main/java/org/owasp/dependencycheck/analyzer/OssIndexAnalyzer.java
@@ -37,7 +37,9 @@ import org.owasp.dependencycheck.dependency.VulnerableSoftware;
 import org.owasp.dependencycheck.dependency.VulnerableSoftwareBuilder;
 import org.owasp.dependencycheck.dependency.naming.Identifier;
 import org.owasp.dependencycheck.dependency.naming.PurlIdentifier;
+import org.owasp.dependencycheck.exception.InitializationException;
 import org.owasp.dependencycheck.utils.Settings;
+import org.owasp.dependencycheck.utils.Settings.KEYS;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import us.springett.parsers.cpe.exceptions.CpeValidationException;
@@ -125,6 +127,16 @@ public class OssIndexAnalyzer extends AbstractAnalyzer {
         synchronized (FETCH_MUTIX) {
             reports = null;
         }
+    }
+
+    @Override
+    protected void prepareAnalyzer(Engine engine) throws InitializationException {
+      synchronized (FETCH_MUTIX) {
+        if (StringUtils.isEmpty(getSettings().getString(KEYS.ANALYZER_OSSINDEX_USER, StringUtils.EMPTY)) ||
+            StringUtils.isEmpty(getSettings().getString(KEYS.ANALYZER_OSSINDEX_PASSWORD, StringUtils.EMPTY))) {
+          throw new InitializationException("Error initializing OSS Index analyzer due to missing user/password credentials.");
+        }
+      }
     }
 
     @Override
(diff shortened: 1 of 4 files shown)


jeremylong:
While the documentation is good - we may need to disable the analyzer by default unless an API key is provided...

framayo:
@jeremylong thanks for your feedback!
I've updated the PR validating if the credentials are set before analyzing.
I had to modify the enrich test to use a mock response because it was hitting `ossindex.sonatype.org`, and it would require real user and token once the authentication enforcement is in place.

core/src/main/resources/dependencycheck.properties
```diff
@@ -83,7 +83,7 @@ analyzer.central.parallel.analysis=false
 analyzer.central.use.cache=true
 central.content.url=https://search.maven.org/remotecontent?filepath=
 
-analyzer.ossindex.enabled=true
+analyzer.ossindex.enabled=false
```
jeremylong: While i do want to make this change - we do not have ways to re-enable the analyzer yet. I'm likely going to merge this PR and work on the enabling/disabling later.

```suggestion
analyzer.ossindex.enabled=true
```

core/src/main/java/org/owasp/dependencycheck/analyzer/OssIndexAnalyzer.java
```diff
@@ -127,6 +129,16 @@ protected void closeAnalyzer() throws Exception {
         }
     }
 
+    @Override
+    protected void prepareAnalyzer(Engine engine) throws InitializationException {
+      synchronized (FETCH_MUTIX) {
+        if (StringUtils.isEmpty(getSettings().getString(KEYS.ANALYZER_OSSINDEX_USER, StringUtils.EMPTY)) ||
+            StringUtils.isEmpty(getSettings().getString(KEYS.ANALYZER_OSSINDEX_PASSWORD, StringUtils.EMPTY))) {
+          throw new InitializationException("Error initializing OSS Index analyzer due to missing user/password credentials.");
```
jeremylong:
```suggestion
          throw new InitializationException("Error initializing OSS Index analyzer due to missing user/password credentials. Authentication is now required: https://ossindex.sonatype.org/doc/auth-required");
```

src/site/markdown/analyzers/oss-index-analyzer.md
```diff
@@ -8,3 +8,9 @@ identified vulnerabilities are included in the report. In addition, vulnerabilit
 found in both the NVD and OSS Index may have additional references added.
 
 This analyzer requires an Internet connection.
+
+Sonatype [announced](https://ossindex.sonatype.org/doc/auth-required) that OSS Index will require authentication.
```
jeremylong:
```suggestion
Sonatype [announced](https://ossindex.sonatype.org/doc/auth-required) that OSS Index requires authentication.
```

jeremylong (dismissed):

jeremylong:
In an email received from Sonatype - it appears the enforcement will begin on 9/22/2025:

> We are committed to making this transition as smooth as possible. Enforcement will begin on 09/22/2025. If you have any questions or concerns, please contact us at [ossindex@sonatype.com](mailto:ossindex@sonatype.org)
Thank you for your continued use of OSS Index.

jeremylong (approved):

jeremylong:
Thank you for the PR!!! Really appreciate it.

core/src/main/java/org/owasp/dependencycheck/analyzer/OssIndexAnalyzer.java
```diff
@@ -127,6 +129,16 @@ protected void closeAnalyzer() throws Exception {
         }
     }
 
+    @Override
+    protected void prepareAnalyzer(Engine engine) throws InitializationException {
+      synchronized (FETCH_MUTIX) {
+        if (StringUtils.isEmpty(getSettings().getString(KEYS.ANALYZER_OSSINDEX_USER, StringUtils.EMPTY)) ||
```
marcelstoer, line 135: Did you intentionally ignore the `ossIndexServerId` property here? Or is this method only called after user/pw were already resolved from the `settings.xml` (via server ID)?

jeremylong: Only called after this is configured in the mojo.

jeremylong: and in hindsight - this probably should have been a breaking change as most users will now get an exception...

marcelstoer: Wouldn't they have gotten an exception anyway? Either from ODC or from Sonatype?

> disable the analyzer by default unless an API key is provided

I think this would have been the best course of action (accompanied by a warning in the log).

> In an email received from Sonatype - it appears the enforcement will begin on 9/22/2025:

I didn't consider this very community friendly; received the email on the 16th...

jeremylong: If I have enough time I was going to put more work into this tomorrow and possibly one more quick release. Disable by default, enabling by either providing creds or setting enabled=true (and flipping the CLI's disable to enableOssIndex). Just not sure if I will have time.

AndreVirtimo: To disable the analyzer by default would be great. I have a lot failed builds on our CI server now. I don't want to update the configuration for all of them. Or is it possible to to that by an environment variable?

nMoncho: Yes, I'd have to agree with @AndreVirtimo here. Disabling the analyzer if the credentials are empty seems like the way to go. If users want to use OSS Index to analyze their dependencies, they setup an account, otherwise the analyzer is skipped.

We'll wait for your release @jeremylong, so we can update `sbt-dependecy-check`. Please let us know if we can help somehow.

Edit: Feel free to take a look at #7963

PR #7957, opened 2025-09-23
feat: add new suppression xsd allowing grouping of suppressions

magal000:
## Description of Change

Added new version (1.4) of suppression xsd. This allows the grouping of suppressions and metadata about the groupings. Groupings are optional. Suppression xml can be used as earlier. See [issue](https://github.com/dependency-check/DependencyCheck/issues/7898) for reasoning.

## Related issues

[Issue 7898](https://github.com/dependency-check/DependencyCheck/issues/7898)

*yes*

### DIFF
--- a/core/src/main/java/org/owasp/dependencycheck/xml/suppression/SuppressionHandler.java
+++ b/core/src/main/java/org/owasp/dependencycheck/xml/suppression/SuppressionHandler.java
@@ -20,6 +20,7 @@ package org.owasp.dependencycheck.xml.suppression;
 import java.util.ArrayList;
 import java.util.Calendar;
 import java.util.List;
+import java.util.Optional;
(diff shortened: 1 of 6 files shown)


nhumblot:
**praise:** thanks for your time on this proposal and your patience! I'm sure people will find it useful!

nhumblot:
**chore:** our [false-positives documentation](https://dependency-check.github.io/DependencyCheck/general/suppression.html) displays references to the 1.3.xsd version. The page is located in src/site/markdown/general/suppression.md`. We could upgrade the reference to 1.4 in order to have a documentation being up to date.

**suggestion(non-blocking):** it could also be interesting to add an explanation about the `suppressionGroups` and their purpose into this documentation but this could be done in an additional PR to keep changes small.

jeremylong (approved):

PR #7963, opened 2025-09-23
fix: Disable OSS Index if its credentials are missing

nMoncho:
## Description of Change

This disables the OSS Index analyzer if no credentials are provided. Feel free to drop this PR if it doesn't make sense.

## Related issues

Relates to #7920

yes

### DIFF
--- a/core/src/main/java/org/owasp/dependencycheck/analyzer/OssIndexAnalyzer.java
+++ b/core/src/main/java/org/owasp/dependencycheck/analyzer/OssIndexAnalyzer.java
@@ -134,7 +134,8 @@ public class OssIndexAnalyzer extends AbstractAnalyzer {
       synchronized (FETCH_MUTIX) {
         if (StringUtils.isEmpty(getSettings().getString(KEYS.ANALYZER_OSSINDEX_USER, StringUtils.EMPTY)) ||
             StringUtils.isEmpty(getSettings().getString(KEYS.ANALYZER_OSSINDEX_PASSWORD, StringUtils.EMPTY))) {
-          throw new InitializationException("Error initializing OSS Index analyzer due to missing user/password credentials. Authentication is now required: https://ossindex.sonatype.org/doc/auth-required");
+          LOG.warn("Disabling OSS Index analyzer due to missing user/password credentials. Authentication is now required: https://ossindex.sonatype.org/doc/auth-required");
+          setEnabled(false);
         }
       }
     }
--- a/core/src/test/java/org/owasp/dependencycheck/analyzer/OssIndexAnalyzerTest.java
+++ b/core/src/test/java/org/owasp/dependencycheck/analyzer/OssIndexAnalyzerTest.java
@@ -8,7 +8,6 @@ import org.owasp.dependencycheck.dependency.Confidence;
 import org.owasp.dependencycheck.dependency.Dependency;
 import org.owasp.dependencycheck.dependency.naming.Identifier;
 import org.owasp.dependencycheck.dependency.naming.PurlIdentifier;
-import org.owasp.dependencycheck.exception.InitializationException;
 import org.owasp.dependencycheck.utils.Settings;
 import org.owasp.dependencycheck.utils.Settings.KEYS;
 
@@ -30,7 +29,7 @@ import java.util.concurrent.Future;
 
 import static org.junit.jupiter.api.Assertions.assertDoesNotThrow;
 import static org.junit.jupiter.api.Assertions.assertEquals;
-import static org.junit.jupiter.api.Assertions.assertThrows;
+import static org.junit.jupiter.api.Assertions.assertFalse;
 import static org.junit.jupiter.api.Assertions.assertTrue;
 
 class OssIndexAnalyzerTest extends BaseTest {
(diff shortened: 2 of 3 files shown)


core/src/test/java/org/owasp/dependencycheck/analyzer/OssIndexAnalyzerTest.java
```diff
@@ -252,14 +249,14 @@ void should_analyzeDependency_fail_when_socket_error_from_sonatype() throws Exce
...
+    void should_prepareAnalyzer_disable_when_credentials_not_set() throws Exception {
         OssIndexAnalyzer analyzer = new OssIndexAnalyzer();
         Settings settings = getSettings();
         Engine engine = new Engine(settings);
         analyzer.initialize(settings);
         try {
             analyzer.prepareAnalyzer(engine);
-            assertThrows(InitializationException.class, () -> analyzer.prepareAnalyzer(engine));
+            assertFalse(analyzer.isEnabled());
         } catch (InitializationException e) {
```
r4fterman: Should this exception lead to a test failure now?

AndreVirtimo:
I hope this fix will be merged soon 🤞

core/src/test/java/org/owasp/dependencycheck/analyzer/OssIndexAnalyzerTest.java
nMoncho: No, I think this exception should be removed. I amended the PR.
Thank you for pointing this out.

core/src/main/java/org/owasp/dependencycheck/analyzer/OssIndexAnalyzer.java
```diff
@@ -134,7 +134,8 @@ protected void prepareAnalyzer(Engine engine) throws InitializationException {
       synchronized (FETCH_MUTIX) {
         if (StringUtils.isEmpty(getSettings().getString(KEYS.ANALYZER_OSSINDEX_USER, StringUtils.EMPTY)) ||
             StringUtils.isEmpty(getSettings().getString(KEYS.ANALYZER_OSSINDEX_PASSWORD, StringUtils.EMPTY))) {
-          throw new InitializationException("Error initializing OSS Index analyzer due to missing user/password credentials. Authentication is now required: https://ossindex.sonatype.org/doc/auth-required");
+          LOG.info("Disabling OSS Index analyzer due to missing user/password credentials. Authentication is now required: https://ossindex.sonatype.org/doc/auth-required");
```
marcelstoer: I feel this should be on WARN level

core/src/test/java/org/owasp/dependencycheck/analyzer/OssIndexAnalyzerTest.java
```diff
@@ -28,10 +28,7 @@
 import java.util.concurrent.Executors;
 import java.util.concurrent.Future;
 
-import static org.junit.jupiter.api.Assertions.assertDoesNotThrow;
-import static org.junit.jupiter.api.Assertions.assertEquals;
-import static org.junit.jupiter.api.Assertions.assertThrows;
-import static org.junit.jupiter.api.Assertions.assertTrue;
+import static org.junit.jupiter.api.Assertions.*;
```
marcelstoer: This should be reverted I guess

marcelstoer (changes requested):

core/src/test/java/org/owasp/dependencycheck/analyzer/OssIndexAnalyzerTest.java
```diff
@@ -252,18 +249,20 @@ void should_analyzeDependency_fail_when_socket_error_from_sonatype() throws Exce
...
-          engine.close();
-        }
+
+        // When
+        analyzer.prepareAnalyzer(engine);
+
+        // Then
+        assertFalse(analyzer.isEnabled());
+        analyzer.close();
+        engine.close();
```
r4fterman, line 267: The close() methods are omitted now if the assertFalse() fails.
If calling the close() methods is mandatory to be executed, you could use a try-finally block to ensure their execution. Otherwise this comment can be ignored.

core/src/test/java/org/owasp/dependencycheck/analyzer/OssIndexAnalyzerTest.java
nMoncho: Fixed. IntelliJ optimized the imports and I didn't look closely.

core/src/main/java/org/owasp/dependencycheck/analyzer/OssIndexAnalyzer.java
nMoncho: Done.

core/src/test/java/org/owasp/dependencycheck/analyzer/OssIndexAnalyzerTest.java
nMoncho: Not sure how to proceed here. You're right, but no other tests follows that approach if no exception is being thrown 🤷‍♂️

marcelstoer (approved):
LGTM

core/src/test/java/org/owasp/dependencycheck/analyzer/OssIndexAnalyzerTest.java
```diff
@@ -252,18 +251,20 @@ void should_analyzeDependency_fail_when_socket_error_from_sonatype() throws Exce
...
```
jeremylong: One way would be to just flip things around like this:

```suggestion
        boolean enabled = analyzer.isEnabled();
        analyzer.close();
        engine.close();
        assertFalse(enabled);
```

A better approach might be to utilize a try with resources:

```java
        OssIndexAnalyzer analyzer = new OssIndexAnalyzer();
        Settings settings = getSettings();
        try (Engine engine = new Engine(settings)) {
                analyzer.initialize(settings);   
                // When
                analyzer.prepareAnalyzer(engine);
                // Then
                assertFalse(analyzer.isEnabled());
        } finally {
                analyzer.close();
        }
```

jeremylong:
This is a better approach than what I had half completed yesterday - Thanks for the PR!

jeremylong (approved):
LGTM

M4mu5qu3:
So, using https://github.com/dependency-check/DependencyCheck/releases/tag/v12.1.6 there is no way to enforce that credentials are passed? I mean, having an error instead of auto-disable the feature? I want my dep-check to be consistent so I don't want that a job can arbitrary omit the credentials to skip the OSS index check. Is there a way?

jeremylong:
@M4mu5qu3 No, with the current implementation, there is no easy way to identify when the credentials were not supplied within the core dependency-check code.

PR #7970, opened 2025-09-24
fix: Update NVD CPE search URLs to match new search interface

### LINKED ISSUE #7962: [Bug] >300k CVEs
**Precondition**

**Describe the bug**
The link to NIST's NVD via https://nvd.nist.gov/vuln/search/results?form_type=Advanced&results_type=overview&search_type=all&cpe_vendor=cpe%3A%2F%3Aapache&cpe_product=cpe%3A%2F%3Aapache%3Astruts&cpe_version=cpe%3A%2F%3Aapache%3Astruts%3A1.2.9 showed  311.310 CVEs

Image

Image

Image

dependency-check version: 12.1.3

**Log file**
none

**To Reproduce**
Scan something e.g. with Apache Struts v1.2.x

**Expected behavior**
I passed a WAR file to dependency check which contains a defined set of Java libs. Here specifically: `./WEB-INF/lib/struts-1.2.9.jar` . So my expectation was to be precise on the software used.  

Maybe pass the product (CNA) .
**Additional context**
(not that it matters much: Results from MVN are less : 4x HIGH (trivy: 2 CVEs MEDIUM)

chadlwilson:
## Description of Change

NVD appear to have updated their search interface in a way that breaks our earlier generated deep links from reports etc.

The new interface is similar, but the URLs build CPE identifiers rather than individual components. The "applicability" mode appears to be correct (otherwise you get matches where a given vuln is running on/with a CPE but not affecting that CPE?)

Version-specific search
https://nvd.nist.gov/vuln/search#/nvd/home?sortOrder=3&sortDirection=2&cpeFilterMode=applicability&cpeName=cpe:2.3:a:vmware:spring_framework:5.0.0:*:*:*:*:*:*:*&resultType=records

Broad search
https://nvd.nist.gov/vuln/search#/nvd/home?sortOrder=3&sortDirection=2&cpeFilterMode=applicability&cpeName=cpe:2.3:a:vmware:spring_framework:*:*:*:*:*:*:*:*&resultType=records

Open questions 
- [x] should I add a default sort order? Order seems random/database order otherwise.
  - ~~Sort by identifier? descending? ascending?~~
  - Published Date? Descending ✅
- [x] Need to update URLs inside the HTML and Jenkins report templates?
  - These ones go to [URLs like this](https://web.nvd.nist.gov/view/vuln/search-results?adv_search=true&cves=on&cpe_version=cpe%3A%2Fa%3Aghostscript%3Aghostscript%3A8.64) - but they are totally broken/404ed so I am not sure if that recently happened or if they have been broken for a while.

## Related issues

- fixes #7962

yes

Example report generated via Gradle plugin locally:
[dependency-check-report.html](https://github.com/user-attachments/files/22558375/dependency-check-report.html)

### DIFF
--- a/core/src/main/java/org/owasp/dependencycheck/analyzer/CPEAnalyzer.java
+++ b/core/src/main/java/org/owasp/dependencycheck/analyzer/CPEAnalyzer.java
@@ -20,8 +20,6 @@ package org.owasp.dependencycheck.analyzer;
 import java.io.BufferedReader;
 import java.io.IOException;
 import java.io.InputStreamReader;
-import java.io.UnsupportedEncodingException;
-import java.net.URLEncoder;
 import java.nio.charset.StandardCharsets;
 import java.util.ArrayList;
 import java.util.Arrays;
@@ -111,24 +109,7 @@ public class CPEAnalyzer extends AbstractAnalyzer {
      * alpha characters.
      */
     private static final String CLEANSE_NONALPHA_RX = "[^A-Za-z]*";
-    /**
-     * UTF-8 character set name.
-     */
-    private static final String UTF8 = StandardCharsets.UTF_8.name();
-    /**
-     * The URL to search the NVD CVE data at NIST. This is used by calling:
-     * <pre>String.format(NVD_SEARCH_URL, vendor, product, version);</pre>
-     */
-    public static final String NVD_SEARCH_URL = "https://nvd.nist.gov/vuln/search/results?form_type=Advanced&"
-            + "results_type=overview&search_type=all&cpe_vendor=cpe%%3A%%2F%%3A%1$s&cpe_product=cpe%%3A%%2F%%3A%1$s%%3A%2$s&"
-            + "cpe_version=cpe%%3A%%2F%%3A%1$s%%3A%2$s%%3A%3$s";
 
-    /**
-     * The URL to search the NVD CVE data at NIST. This is used by calling:
-     * <pre>String.format(NVD_SEARCH_URL, vendor, product);</pre>
-     */
-    public static final String NVD_SEARCH_BROAD_URL = "https://nvd.nist.gov/vuln/search/results?form_type=Advanced&"
-            + "results_type=overview&search_type=all&cpe_vendor=cpe%%3A%%2F%%3A%1$s&cpe_product=cpe%%3A%%2F%%3A%1$s%%3A%2$s";
     /**
      * The CPE in memory index.
      */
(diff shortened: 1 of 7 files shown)


jeremylong:
1. We should add a sort order - published date is probably good enough, descending if possible (i.e., newest first).
2. If the links are broken in the HTML report, we likely need to fix them as well. The easiest approach would be to expose the `nvdSearchUrlFor()` methods in the `ReportTool`, as it is already added to the Velocity context.

chadlwilson:
Thanks @jeremylong - I didn’t look at the 2nd one at all, it turned out a bit messier than i thought and I ran out of time. I couldn’t work out what the urls are used for in the CpeIdentifier object if not used later to actually render the report, so it needed more digging as to the reason for the 404ing URLs and which section of the report those ones appear in.

jeremylong:
Understood, the HTML reports contain code like:

```xml
  <p>Vulnerable Software &amp; Versions:<ul>
      <li class="vs$vsctr"><a target="_blank" href="https://web.nvd.nist.gov/view/vuln/search-results?adv_search=true&cves=on&cpe_version=$enc.url($vuln.matchedVulnerableSoftware.toCpe22Uri())">$enc.html($vuln.matchedVulnerableSoftware.toString())</a></li>
  </ul></p>
```


ReportTool.java
```
...
public String getNvdSearchURL(VulnerableSoftware vs) {
     return CpeIdentifier.nvdSearchUrlFor(vs.getVendor(), vs.getProduct(), vs.getVersion());
}
...
```

Then in the reports, something like:
```xml
  <p>Vulnerable Software &amp; Versions:<ul>
      <li class="vs$vsctr"><a target="_blank" href="$rpt.getNvdSearchURL($vuln.matchedVulnerableSoftware)">$enc.html($vuln.matchedVulnerableSoftware.toString())</a></li>
  </ul></p>
```

Some of the generated URLs look like the broad pattern and others the more specific one - so we may need two methods exposed.

chadlwilson:
OK I figured out that a different mechanism was being used for the "Identifiers" vs. the "Vulnerable Software & Versions" sections. Which also had a different URL syntax. Not sure, but suspect the "Vulnerable Software & Versions" links might have been broken for a while.

I've made this change now, however needs a bit more testing.

I *think* we don't need to specifically use the broad URLs, because the search syntax just takes a CPE 2.3 formatted string anyway. If from the NVD match we know details such as the version number from the matched software (sometimes it is included, if the affected version is not a range), it should be fine to include it in the search. In most other cases including `*` should be sufficient.

This new search also allows us to search specficially for all the other pieces of the CPE that have shown up in the NVD match, not just part/vendor/product/version - so behaviour should be better, I think.

chadlwilson:
Added an example HTML report generated with this to the description.

jeremylong (approved):

PR #7977, opened 2025-09-25
fix: improve OSS Index Error Reporting

### LINKED ISSUE #7976: ossIndexRemoteErrorWarnOnly is true but maven is stopping for oss index access errors (HttpResponseException: status code: 407, reason phrase: Proxy Authorization Required)
I have setted 
`<ossIndexWarnOnlyOnRemoteErrors>true</ossIndexWarnOnlyOnRemoteErrors> `
(and also ossIndexUsername and ossIndexPassword) but the maven build stopped (_Finished: FAILURE_) with a connection ERROR, when I'm expecting only a warning.
Am I misundestanding something?

maven dependency check:
```
<groupId>org.owasp</groupId>
<version>12.1.6</version>
```

**Log file**
[ERROR] Failed to execute goal org.owasp:dependency-check-maven:12.1.6:check (default) on project xxxx: One or more exceptions occurred during dependency-check analysis: One or more exceptions occurred during analysis:
[ERROR] AnalysisException: Failed to request component-reports
[ERROR] caused by DownloadFailedException: https://ossindex.sonatype.org/api/v3/component-report - Server status: 407 - Server reason: Proxy Authorization Required
[ERROR] caused by HttpResponseException: status code: 407, reason phrase: Proxy Authorization Required
.................................
[ERROR] -> [Help 1]
[ERROR] 
[ERROR] To see the full stack trace of the errors, re-run Maven with the -e switch.
[ERROR] Re-run Maven using the -X switch to enable full debug logging.
[ERROR] 
[ERROR] For more information about the errors and possible solutions, please read the following articles:
[ERROR] [Help 1] http://cwiki.apache.org/confluence/display/MAVEN/MojoExecutionException
[ERROR] 
[ERROR] After correcting the problems, you can resume the build with the command
[ERROR]   mvn <goals> -rf   xxxxxxxx
........
Finished: FAILURE

jeremylong:
## Description of Change

ODC was not properly reporting authorization error messages. Previously, it was looking for the message to end with 401 to report an authorization issue. However, the actual error message from the client is `Failed to request component-reports: https://ossindex.sonatype.org/api/v3/component-report - Server status: 401 - Server reason: Unauthorized`. As such, the check for 401 needed to be improved.

## Related issues

Best guess this is related to https://github.com/dependency-check/DependencyCheck/issues/7975 and possibly https://github.com/dependency-check/DependencyCheck/issues/7971

Yes

### DIFF
--- a/core/src/main/java/org/owasp/dependencycheck/analyzer/OssIndexAnalyzer.java
+++ b/core/src/main/java/org/owasp/dependencycheck/analyzer/OssIndexAnalyzer.java
@@ -162,11 +162,19 @@ public class OssIndexAnalyzer extends AbstractAnalyzer {
                     final boolean warnOnly = getSettings().getBoolean(Settings.KEYS.ANALYZER_OSSINDEX_WARN_ONLY_ON_REMOTE_ERRORS, false);
                     this.setEnabled(false);
                     if (StringUtils.contains(message, "401")) {
-                        LOG.error("Invalid credentials for the OSS Index, disabling the analyzer");
-                        throw new AnalysisException("Invalid credentials provided for OSS Index", ex);
+                        if (warnOnly) {
+                            LOG.warn("Invalid credentials for theOSS Index, disabling the analyzer");
+                        } else {
+                            LOG.error("Invalid credentials for the OSS Index, disabling the analyzer");
+                            throw new AnalysisException("Invalid credentials provided for OSS Index", ex);
+                        }
                     } else if (StringUtils.contains(message, "403")) {
-                        LOG.error("OSS Index access forbidden, disabling the analyzer");
-                        throw new AnalysisException("OSS Index access forbidden", ex);
+                        if (warnOnly) {
+                            LOG.warn("OSS Index access forbidden, disabling the analyzer");
+                        } else {
+                            LOG.error("OSS Index access forbidden, disabling the analyzer");
+                            throw new AnalysisException("OSS Index access forbidden", ex);
+                        }
                     } else if (StringUtils.contains(message, "429")) {
                         if (warnOnly) {
                             LOG.warn("OSS Index rate limit exceeded, disabling the analyzer", ex);
--- a/core/src/main/java/org/owasp/dependencycheck/analyzer/OssIndexAnalyzer.java
+++ b/core/src/main/java/org/owasp/dependencycheck/analyzer/OssIndexAnalyzer.java
@@ -148,17 +148,26 @@ public class OssIndexAnalyzer extends AbstractAnalyzer {
                 try {
                     requestDelay();
                     reports = requestReports(engine.getDependencies());
-                } catch (TransportException ex) {
+                } catch (SocketTimeoutException e) {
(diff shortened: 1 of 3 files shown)


stefano-1973:
For the issues #7976 shouldn't be managed in the code also the status code 407 ? (Proxy Authorization Required error)

jeremylong:
The original code had a few paths that would end up throwing the exception without checking the warn only flag. The updated version doesn't - even though we aren't specifically checking for a 407 it won't hit a path that throws the exception if you configure warn only.

marcelstoer (approved):
LGTM

defaultbranch:
Thanks @jeremylong for keeping the tool working!

core/src/main/java/org/owasp/dependencycheck/analyzer/OssIndexAnalyzer.java
```diff
@@ -148,40 +148,45 @@ protected void analyzeDependency(final Dependency dependency, final Engine engin
...
-                    if (StringUtils.endsWith(message, "401")) {
-                        LOG.error("Invalid credentials for the OSS Index, disabling the analyzer");
-                        throw new AnalysisException("Invalid credentials provided for OSS Index", ex);
-                    } else if (StringUtils.endsWith(message, "403")) {
-                        LOG.error("OSS Index access forbidden, disabling the analyzer");
-                        throw new AnalysisException("OSS Index access forbidden", ex);
-                    } else if (StringUtils.endsWith(message, "429")) {
+                    if (StringUtils.contains(message, "401")) {
+                        if (warnOnly) {
+                            LOG.warn("Invalid credentials for theOSS Index, disabling the analyzer");
```
chadlwilson:
```suggestion
                            LOG.warn("Invalid credentials for the OSS Index, disabling the analyzer");
```

Nitpick.

core/src/main/java/org/owasp/dependencycheck/analyzer/OssIndexAnalyzer.java
```diff
@@ -148,40 +148,45 @@ protected void analyzeDependency(final Dependency dependency, final Engine engin
                 try {
                     requestDelay();
                     reports = requestReports(engine.getDependencies());
-                } catch (TransportException ex) {
+                } catch (SocketTimeoutException e) {
```
chadlwilson, line 151: I understand that the previous code also only specifically looked at `SocketTimeoutException` but it occurs to me that there are probably a number of other `IOException`s which essentially relate to an issue with the remote. I guess they will be handled by the generic `catch (Exception ex)` block, and depends how many the client wraps with `TransportException` which is a bit messy to unpack. Still this is not 'worse ' :-)

chadlwilson (approved):

core/src/main/java/org/owasp/dependencycheck/analyzer/OssIndexAnalyzer.java
jeremylong: The code was checking for a 401 inside the `TransportExceptions` - but it was not a transport exception so the code never worked. I agree, this isn't "worse" but we could do better.

chadlwilson (approved):

PR #7986, opened 2025-09-29
fix: Correct Archive Analyzer behaviour on certain tgz archives

### LINKED ISSUE #7985: (DC 12.1.6) Exception reading archive - (Archive Analyzer): Extra subfield lenght exceeds remaining bytes in extra: 21064 > 37
**Precondition**

The problem occurs using version 12.1.6 on Linux CLI.

**Describe the bug**
The console logs report the following:
`[INFO] Analysis Started

[ERROR] Exception reading archive 'project.tgz'.

[WARN] An error occurred while analyzing '/home/pipelineuser/project.tgz' (Archive Analyzer): Extra subfield lenght exceeds remaining bytes in extra: 21064 > 37

[INFO] Finished Archive Analyzer (1 seconds)`

**Other observations**

We see that DC 12.1.6 uses apache commons  -  commons-compress-1.28.0.jar

There is a bug report in Apache Compress with regards similar errors:
[GZip IOException: Extra subfield lenght exceeds remaining bytes in extra field](https://issues.apache.org/jira/browse/COMPRESS-705)

In which the report mentions _"Extra subfield lenght exceeds remaining bytes in extra: 21064 > 37"._

Given the wording in both cases, suggests it's a dependency problem _( commons-compress-1.28.0 )_.
The ticket marks it resolved at compress 1.29.0

We have previously used DC 12.1.3 and don't recall seeing this archive issue with that and it uses commons-compress-1.27.1.jar.

chadlwilson:
## Description of Change

commons-compress 1.28.0 seems to have issues with changing default behaviour on some archives: https://issues.apache.org/jira/browse/COMPRESS-705

Since the release cadence for Apache Commons stuff is a bit unpredictable, even in the face of a regression, we should probably just roll this back for now?

This reverts commit 4c33e0a5

## Related issues

- fixes #7985

no

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -158,7 +158,7 @@ Copyright (c) 2012 - Jeremy Long
         <hamcrest.version>3.0</hamcrest.version>
         <mockito.version>5.19.0</mockito.version>
         <jsoup.version>1.21.2</jsoup.version>
-        <commons-compress.version>1.28.0</commons-compress.version>
+        <commons-compress.version>1.27.1</commons-compress.version>
         <org.apache.maven.shared.file-management.version>3.2.0</org.apache.maven.shared.file-management.version>
         <maven-plugin-testing-harness.version>3.3.0</maven-plugin-testing-harness.version>
         <maven-plugin-annotations.version>3.15.1</maven-plugin-annotations.version>


marcelstoer (approved):
LGTM

I didn't find any indication in the commons-compress project as to when they plan to release the five fixes that landed on master since 1.28.0, pity.

chadlwilson:
Many of those smaller library Apache projects are basically maintained by the same solo maintainer, and they still do their (somewhat old school now) practice of voting for releases on mailing lists and stuff, I think.

jeremylong (approved):

PR #7991, opened 2025-09-30
feat: Add params to obtain content from an alternative repository

evgeniiworkst:
## Description of Change

This MR adds a new parameter for retrieving content from alternative sources.
This solves the problem where, especially recently, search.maven.org was returning a 502 error.
For example
https://search.maven.org/remotecontent?filepath=org/apache/ant/ant/1.10.14/ant-1.10.14.pom
image

## Related issues

- relates to https://github.com/dependency-check/DependencyCheck/issues/7989

*no*

### DIFF
--- a/cli/src/main/java/org/owasp/dependencycheck/App.java
+++ b/cli/src/main/java/org/owasp/dependencycheck/App.java
@@ -641,6 +641,14 @@ public class App {
                 cli.getStringArgument(CliParser.ARGUMENT.CENTRAL_PASSWORD));
         settings.setStringIfNotEmpty(Settings.KEYS.ANALYZER_CENTRAL_BEARER_TOKEN,
                 cli.getStringArgument(CliParser.ARGUMENT.CENTRAL_BEARER_TOKEN));
+        settings.setStringIfNotEmpty(Settings.KEYS.CENTRAL_CONTENT_URL,
+                cli.getStringArgument(CliParser.ARGUMENT.CENTRAL_CONTENT_URL));
+        settings.setStringIfNotEmpty(Settings.KEYS.CENTRAL_CONTENT_USER,
+                cli.getStringArgument(CliParser.ARGUMENT.CENTRAL_CONTENT_USER));
+        settings.setStringIfNotEmpty(Settings.KEYS.CENTRAL_CONTENT_PASSWORD,
+                cli.getStringArgument(CliParser.ARGUMENT.CENTRAL_CONTENT_PASSWORD));
+        settings.setStringIfNotEmpty(Settings.KEYS.CENTRAL_CONTENT_BEARER_TOKEN,
+                cli.getStringArgument(CliParser.ARGUMENT.CENTRAL_CONTENT_BEARER_TOKEN)); 
         settings.setStringIfNotEmpty(Settings.KEYS.ANALYZER_OSSINDEX_URL,
                 cli.getStringArgument(CliParser.ARGUMENT.OSSINDEX_URL));
         settings.setStringIfNotEmpty(Settings.KEYS.ANALYZER_OSSINDEX_USER,
--- a/cli/src/main/java/org/owasp/dependencycheck/CliParser.java
+++ b/cli/src/main/java/org/owasp/dependencycheck/CliParser.java
@@ -430,6 +430,15 @@ public final class CliParser {
                         "Credentials for basic auth towards the --centralUrl"))
                 .addOption(newOptionWithArg(ARGUMENT.CENTRAL_BEARER_TOKEN, "token",
                         "Token for bearer auth towards the --centralUrl"))
+                .addOption(newOptionWithArg(ARGUMENT.CENTRAL_CONTENT_URL, "url",
+                        "Alternative URL to obtain content from an alternative repository. If not set the public Sonatype Maven Central will be used."
+                        + "Use a repository endpoint, such as https://artifactory.com/artifactory/mvn-release/."))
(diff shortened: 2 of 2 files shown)


jeremylong:
Using a different endpoint won't do anything in most cases other than effectively disable the analyzer. Have you looked at https://github.com/dependency-check/DependencyCheck/issues/7686#issuecomment-3239200705?

evgeniiworkst:
@jeremylong Isn't the content supposed to return a pom.xml file? I could take the POM file from my own repository without involving the Central analyzer. Aren't these two entirely different entities?
image

jeremylong:
Sorry - I was thinking of the Central Search API.

chadlwilson:
This doesn’t feel right to me. Shouldn’t one just enable the Nexus or Artifactory analyzers instead? This seems to add complexity but I am not sure in which use case this helps?

evgeniiworkst:
@chadlwilson Analyzers for Central/Nexus/Artifactory are a slightly different topic. Yesterday, I encountered an error specifically related to uploading a pom.xml file, likely due to a 502 error in the Maven repository.
```
[WARN] Unable to download pom.xml for antlr-2.7.7.jar from Central; this could result in undetected CPE/CVEs.
[WARN] Unable to download pom.xml for asm-9.8.jar from Central; this could result in undetected CPE/CVEs.
[WARN] Unable to download pom.xml for hamcrest-3.0.jar from Central; this could result in undetected CPE/CVEs.
[WARN] Unable to download pom.xml for hamcrest-core-3.0.jar from Central; this could result in undetected CPE/CVEs.
[WARN] Unable to download pom.xml for jspecify-1.0.0.jar from Central; this could result in undetected CPE/CVEs.
[WARN] Unable to download pom.xml for junit-4.13.2.jar from Central; this could result in undetected CPE/CVEs.
```
My version of Artifactory does not support component analysis (XRay), and I understand that dependency-check relies on the results of the XRay analyzer.

chadlwilson:
I dont think it's a separate topic, no, because its an alternate pom source to Central for those scanning jars. Similar to Nexus. Don't think it needs x-ray. Have you tried it?

I still don't understand what the use case of this is. If you're not pointing it at Central where are you pointing it at exactly, and what is replacing the search?

chadlwilson:
https://github.com/dependency-check/DependencyCheck/blob/eefb8d7e849f22af294160926a6af28bccb3599b/CHANGELOG.md?plain=1#L1339

chadlwilson:
Anyway, given the underlying url setting/property already exists, I guess I don't have a huge objection, I just think the configuration interface gets more and more confusing and we end up with more and more questions from confused users the more that is exposed.

evgeniiworkst:
Okay, to avoid complicating the interface by adding new CLI parameters, let's wait for the Maven Central issues to be resolved. I just wanted to have the option to load pom.xml files from another source in case of a failure.

evgeniiworkst:
More complex scanner settings are not stored in --help parameter but are located in --advancedhelp parameter, meaning there is already a separation for deeper scanner configuration.

chadlwilson:
That is the purpose of the Artifactory and Nexus Analyzers and running your own infra as a proxy: https://github.com/dependency-check/DependencyCheck/issues/7992#issuecomment-3356301565 - these 3 analyzers are essentially all "alternate POM sources".

evgeniiworkst:
I tried connecting Artifactory to scanning several times, but it didn’t work. I couldn’t find any helpful information in the logs. Is there perhaps an example of usage somewhere?

chadlwilson:
It shouldn't be too complicated, but depends on your Artifactory setup.

https://dependency-check.github.io/DependencyCheck/dependency-check-cli/arguments.html

```shell
path/to/odc/bin/dependency-check.sh \
-s . \
--disableCentral \
--enableArtifactory \
--artifactoryUrl https://my-server/artifactory \
--artifactoryUsername $ARTIFACTORY_USER \
--artifactoryApiToken $ARTIFACTORY_TOKEN \
```

evgeniiworkst:
@chadlwilson Thank you for the tip. I was only missing the --disableCentral parameter, which wasn’t obvious to me, as the documentation for the --disableCentral parameter states that it’s necessary to disable other analyzers (Artifactory/Nexus) as well. It would also be great to specify in the --enableArtifactory parameter that --disableCentral needs to be enabled: https://dependency-check.github.io/DependencyCheck/dependency-check-cli/arguments.html.

chadlwilson:
Yes, that is misleading/wrong - also noted at https://github.com/dependency-check/DependencyCheck/issues/7988 

PR to improve the documentation is welcome :-)

jeremylong:
So this PR can be closed?

chadlwilson:
I think it's your call @jeremylong - if you think it makes sense to further expose the existing `central.content.url` setting to the CLI, we can do so. However I can't really think how this makes sense. If Maven Central Search is working but the content URLs are not - is it really valid to overwrite the content URL, or change it to a direct URL? Since you cannot replace the Maven Central Search API with your own, it seems you cannot safely override the content URL either.

From the other Maven/Gradle plugins I guess it's already possible to override this property using Java system properties.

Having said all this, if you'd rather leave this code untouched until Sonatype work out the future of the Maven Central search API, that might make as much sense. For all we know, a future implementation will return fully qualified POM URLs to follow anyway.

evgeniiworkst:
@chadlwilson can you see my PR for docs https://github.com/dependency-check/DependencyCheck/pull/7999

Issue #7992, opened 2025-09-30
[WARN] Unable to download pom.xml for *jar from Central; this could result in undetected CPE/CVEs.


proo4509:
**Precondition**
- [x ] I checked the issues list for existing open or closed reports of the same problem.

**Describe the bug**
Since today I see this warning/error:
Unable to download pom.xml *.jar from Central
this occurs for all jars and takes for ever (about 10 minutes for each jar)

I did had to change yesterday my config and added -retireJSURL "https://github.com/RetireJS/retire.js/blob/master/repository/jsrepository.json"
-ossIndexUsername  AND --ossIndexPassword

**Version of dependency-check used**
The problem occurs using version 12.1.0 of the  CLI but have been using this versions since it came out. No problems until today.

**Log file**
When reporting errors, 99% of the time log file output is required. Please post the log file as a [gist](https://gist.github.com/) and provide a link in the new issue.
I will do that when my scan is finished (extremely slow)

**To Reproduce**
running dependency-check-12.1.0\dependency-check\bin\dependency-check.bat 

**Expected behavior**

**Additional context**

readonlyuser1:
also https://github.com/dependency-check/DependencyCheck/issues/7989

chadlwilson:
Perhaps Maven Central was having issues, although nothing noted on https://status.maven.org/ just yet

You can temporarily disable the analyzer.

xtermi2:
We see the same issue. Our CI Pipeline was canceled after 200 Minutes. In this time I see 27 Warnings regarding `Unable to download pom.xml ...`. This means there is a timeout of ~9.5 Minutes! 

Is there a option to configure the timeout of this analyzer? If not, there should be one, so my pipeline will succeed and the dependency check is maybe not perfect because of missing information, but better than nothing!

chadlwilson:
If you shorten the timeout, it will simply make it fail faster - but dont see how that helps your pipeline succeed?

If you don't mind the (potential) missing information, disable the analyzer?

petergphillips:
I raised a similar issue about the timeouts for the gradle plugin in https://github.com/dependency-check/dependency-check-gradle/issues/468.

Our issue is that we use four github action runners to process all of our security jobs.  This is needed as each one caches the NVD feed locally.  When each one takes ~3h to process a single job then jobs are left waiting and never run.  I don't mind the jobs failing if search.maven.org is down, but would prefer them to fail after a few minutes instead of a few hours.

[...]

Issue #7686, opened 2025-05-27
[RFE]: Possibility to configure URL to "Maven central"


alixwar:
**Is your feature request related to a problem? Please describe.**
This is the issue that I'm facing: https://github.com/dependency-check/DependencyCheck/issues/5291

In my case (using version 12.1.0 of the gradle plugin):

```
Caused by: java.io.IOException: https://search.maven.org/solrsearch/select?q=1:8110ba5ec44be61aa96f8bd8d1c57eed0efdb7b9&wt=xml - Server status: 502 - Server reason: Bad Gateway
	at org.owasp.dependencycheck.utils.Downloader.fetchAndHandle(Downloader.java:724)
	at org.owasp.dependencycheck.utils.Downloader.fetchAndHandle(Downloader.java:685)
	at org.owasp.dependencycheck.data.central.CentralSearch.searchSha1(CentralSearch.java:162)
	at org.owasp.dependencycheck.analyzer.CentralAnalyzer.fetchMavenArtifacts(CentralAnalyzer.java:339)
	... 8 more

...

An error occurred while analyzing '/Users/xxxxx/.gradle/caches/modules-2/files-2.1/org.jetbrains.kotlin/kotlin-scripting-compiler-embeddable/2.1.20/efd094224781affbc9aa8c60c5cae926517e028f/kotlin-scripting-compiler-embeddable-2.1.20.jar' (Central Analyzer).
Could not connect to Central search. Analysis failed.
java.io.IOException: Finally failed connecting to Central search. Giving up after 7 tries.
```

**Describe the solution you'd like**
I'd like to be able to configure the plugin to use my own mirrored URL to Maven Central so that I don't have to rely on uptime for this public repository (bonus: less latency).

**Describe alternatives you've considered**
No alternatives

Nonobis:
Yes maven is down for a second times in 1 week .... https://status.maven.org/ i am currently integrating DependencyCheck in our build process ... for now it's only me ... but i can't have 10 developer blocked for a day....

jmsdc:
I am facing the same issue with the dependency check analysis in failure :  read timeout

The `search.maven.org` server is often down
Maven server status: https://status.maven.org/

```plaintext
Checking for updates and analyzing dependencies for vulnerabilities
Could not connect to Central search. Analysis failed.
java.io.IOException: Finally failed connecting to Central search. Giving up after 7 tries.
[...]

PR #7993, opened 2025-10-01
fix: disable central analyzer after failures

jeremylong:
## Description of Change

Due to the stability of Cental Search - this PR reduces the retry count and will disable the analyzer after failures. 

## Related issues

- #7992
- #7686

### DIFF
--- a/core/src/main/java/org/owasp/dependencycheck/analyzer/CentralAnalyzer.java
+++ b/core/src/main/java/org/owasp/dependencycheck/analyzer/CentralAnalyzer.java
@@ -289,12 +289,10 @@ public class CentralAnalyzer extends AbstractFileTypeAnalyzer {
                         } else {
                             LOGGER.warn("Unable to download pom.xml for {} from Central; "
                                     + "this could result in undetected CPE/CVEs.", dependency.getFileName());
-                            if (lastException != null) {
-                                setEnabled(false);
-                                LOGGER.warn("Disabling the Central Analyzer due to repeated download failures; Central Search " +
-                                        "may be down see https://status.maven.org/\n Note that this could result in both false " +
-                                        "positives and false negatives", lastException);
-                            }
+                            setEnabled(false);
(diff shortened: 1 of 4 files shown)


chadlwilson:
Is 2 failed batches in a row (after retries) what you normally abort on?

is it affected by parallel processing? (E.g each thread would need 2 consecutive batch failures?)

chadlwilson (approved):

jeremylong:
Parallel processing would just have a few searches running at the same time - if one failed, it would prevent other searches from starting up.

core/src/main/java/org/owasp/dependencycheck/analyzer/CentralAnalyzer.java
```diff
@@ -287,6 +289,12 @@ public void analyzeDependency(Dependency dependency, Engine engine) throws Analy
                         } else {
                             LOGGER.warn("Unable to download pom.xml for {} from Central; "
                                     + "this could result in undetected CPE/CVEs.", dependency.getFileName());
+                            if (lastException != null) {
```
aikebah: IntelliJ code inspection tells me that this null-check is useless, we only hit the else-branch in cases where lastException will be set to a non-null value.

core/src/main/java/org/owasp/dependencycheck/analyzer/CentralAnalyzer.java
```diff
@@ -287,6 +289,12 @@ public void analyzeDependency(Dependency dependency, Engine engine) throws Analy
...
+                                setEnabled(false);
+                                LOGGER.warn("Disabling the Central Analyzer due to repeated download failures; Central Search " +
```
aikebah: According to the current checkstyle rules the plus-sign should be coded on the new line, not eol (there are a few more cases to fix for that as well)

aikebah (dismissed):
Given the current level of (in)stability indeed a better choice to disable it on errors.

Only a few style-related review remarks you might want to address before merging.

PR #7999, opened 2025-10-02
docs: Documentation artifactory settings fix

evgeniiworkst:
## Description of Change

Extended description of CLI parameters

## Related issues
https://github.com/dependency-check/DependencyCheck/issues/7988
https://github.com/dependency-check/DependencyCheck/pull/7991#issuecomment-3359747956

no


marcelstoer:
I guess my personal preference would be for that comment to somehow express that it's recommended that you use any of the supported repositories (Central _or_ custom Artifactory/Nexus). If you run Artifactory in proxy mode, it should be safe to disable the Central analyzer.

chadlwilson:
I have a related PR at #8000 for Nexus - I could cherry-pick this commit over there and make some edits there - or you could send PRs to that branch if you want to collaborate somewhere?

I agree it'd be good if 
- all 3 analyzer pages mention the other two analyzers
- all 3 `enabled` flags for each plugin mention the other analyzers.

chadlwilson (approved):
Generally we probably need to sync the documentation across the other docs for ant, and maven/gradle as well, but this is better than current, so no objections.

evgeniiworkst:
@marcelstoer The issue is that when using --enableArtifactory, it is mandatory to also use the --disableCentral parameter. Otherwise, ODC will still attempt to connect to Maven Central, and Artifactory will not work. Artifactory does not function as an addition to Central but only as a complete replacement for it. Therefore, in my comment, I wrote that it must(!) be disabled, as otherwise it simply won’t work. In your comment, however, it is only suggested that this is safe and won’t disrupt functionality. Regarding the note "It is recommended to use any of the repositories (Central or custom Artifactory/Nexus)," I completely agree with you, but I’m not sure where this important addition could be placed. Under the --disableCentral parameter? Perhaps in a comment, you could provide a sample text of how you personally envision this?

evgeniiworkst:
@chadjvw If possible, I would separate all these types of analyzers into a distinct category or somehow mark them among the other parameters. Then, it would be possible to add a header to them stating that it is recommended to use any of the supported repositories (Central or custom Artifactory/Nexus). Additionally, the table of contents could immediately include information for all analyzers.

marcelstoer:
> The issue is that when using --enableArtifactory, it is mandatory to also use the --disableCentral parameter. Otherwise, ODC will still attempt to connect to Maven Central, and Artifactory will not work. Artifactory does not function as an addition to Central but only as a complete replacement for it.

Ok, I wasn't aware of this constraint. Could you point to reference to back this up? We have been running ODC with `--enableArtifactory` for the longest time with the Central analyzer still enabled.

chadlwilson:
@marcelstoer i think it’s probably just the ordering/preference of the analyzers, and so if you have both enabled it tries central first and then fails/gets stuck before getting to artifactory or nexus.

https://github.com/dependency-check/DependencyCheck/blob/c566904f10a193bb696f33f9a5912b3204187aa7/core/src/main/resources/META-INF/services/org.owasp.dependencycheck.analyzer.Analyzer#L13

If it succeeds on central the POM would already be there and the artifactory search would be a no-op.

evgeniiworkst:
@marcelstoer @chadlwilson  Yes, it’s likely triggering this way due to the priority of the analyzers. I’m currently testing on my project. Right now, Maven Central is returning a 502 error and isn’t working. Meanwhile, ODC is getting stuck hard at the [INFO] Finished Jar Analyzer (0 seconds) stage. So, I have to interrupt the execution and read the logs.
with params: --enableArtifactory
```
cat log | grep Downloading
DEBUG - Downloading https://search.maven.org/remotecontent?filepath=antlr/antlr/2.7.7/antlr-2.7.7.pom
DEBUG - Downloading https://search.maven.org/remotecontent?filepath=antlr/antlr/2.7.7/antlr-2.7.7.pom
DEBUG - Downloading https://search.maven.org/remotecontent?filepath=antlr/antlr/2.7.7/antlr-2.7.7.pom
```
with params: --enableArtifactory --disableCentral (The scanner completes the operation successfully.)
```
cat log | grep Downloading
DEBUG - Downloading https://artifactory.xx.xx:443/artifactory/mvn-remote-cache/org/ow2/asm/asm/9.8/asm-9.8.pom
DEBUG - Downloading https://artifactory.xx.xx:443/artifactory/mvn-remote-cache/org/jspecify/jspecify/1.0.0/jspecify-1.0.0.pom
DEBUG - Downloading https://artifactory.xx.xx:443/artifactory/mvn-remote-cache/org/hamcrest/hamcrest/3.0/hamcrest-3.0.pom
DEBUG - Downloading https://artifactory.xx.xx:443/artifactory/mvn-remote-cache/antlr/antlr/2.7.7/antlr-2.7.7.pom
DEBUG - Downloading https://artifactory.xx.xx:443/artifactory/mvn-remote-cache/junit/junit/4.13.2/junit-4.13.2.pom
... etc.

```
Without the --disableCentral parameter, my work is paralyzed.

chadlwilson:
I don't think that's a "constraint" though. It seems.pretty consistent and intuitive that if an analyzer is stuck then your whole analysis will be stuck and you need to disable it to use an alternative approach.

Anyway, the next version will disable Central automatically on repeated failures.

evgeniiworkst:
@chadlwilson Then, all the more reason to describe it more clearly in the documentation: when enableArtifactory is used, it adds it with a priority of 2, while, for example, enableNexus adds it with a priority of 3, and all these parameters can work together. Upon first reading, I thought that when one is enabled, the others are disabled. Yes, it would be a good solution if the analyzer could switch to the next option if the first one is unavailable. In that case, it turns out that disableCentral is not mandatory (only in the current situation with Maven Central as a workaround).

chadlwilson:
Sure. You know how it is with open source though - this is one of the most heavily used security tooling projects in the world by large companies, governments, and others…. yet extremely few contributors around 1 primary maintainer (who has a full time job elsewhere) - and infinitely more complaints (often from folks who hardly use GitHub) than solutions. especially low interest when it comes to developer distaste for writing and refactoring docs! :)

https://github.com/dependency-check/DependencyCheck/graphs/contributors

So - many thanks for creating a couple of PRs!

evgeniiworkst:
@chadlwilson  Then I think I need to change the comment from "To use Artifactory, you will need to disable the central analyzer by adding the --disableCentral parameter." To "If you want to use only(!) Artifactory without Central, you should disable it with the --disableCentral parameter."
Or do nothing and wait for 1) Maven Central to be fixed 2)adding switch to another option after failed attempts :)

jeremylong (dismissed):
LGTM

PR #8000, opened 2025-10-02
docs: Clarify Nexus Analyzer requirements and usage

chadlwilson:
## Description of Change

- Doesn't require Pro any longer
- Remove "superseded by" language, as there are a number of reasons you might want to use Nexus instead of Central, especially as capability is now available in the OSS version.

## Related issues

Discussed in https://github.com/dependency-check/DependencyCheck/issues/7686#issuecomment-3360461481

N/A


jeremylong (approved):
LGTM

Issue #7998, opened 2025-10-02
[FP]: CVE-2023-46669 and CVE-2024-52976 on co.elastic.apm:apm-agent-api


aggeboe:
### Package URl

pkg:maven/co.elastic.apm/apm-agent-api@1.51.0

### CPE

`cpe:2.3:a:elastic:elastic_agent:1.51.0:*:*:*:*:*:*:*`

### ODC Integration

{"label" => "Gradle Plugin"}

### ODC Version

12.1.6

### Description

apm-agent-api identified as Elastic Agent (https://github.com/elastic/elastic-agent)
Relates to #7996

chadlwilson:
approved

OrangeDog:
Would be better to match the whole `co.elastic.apm` group in a single rule, none of which are `elastic:apm_agent`, but most hit this FP.

chadlwilson:
@OrangeDog Feel free to submit a PR to the `generatedSuppressions` branch to refactor the suppression to that. The automation isn't smart enough to do that, but takes a few seconds rather than multiple minutes.

Issue #7996, opened 2025-10-02
[FP]: CVE-2023-46669 on co.elastic.apm:apm-agent-api


aggeboe:
### Package URl

pkg:maven/co.elastic.apm/apm-agent-api@1.51.0

### CPE

`cpe:2.3:a:elastic:apm_agent:1.51.0:*:*:*:*:*:*:*`

### CVE

CVE-2023-46669

### ODC Integration

{"label" => "Gradle Plugin"}

### ODC Version

12.1.6

### Description

apm-agent-api identified as Elastic Agent (https://github.com/elastic/elastic-agent)

chadlwilson:
approved

chadlwilson:
Although I have approved this (maybe too early), you might want to check your NVD data. The CVEs you link are actually marked against `cpe:2.3:a:elastic:elastic_agent:*:*:*:*:*:*:*:*` which is not causing an FP - not the CPE you mention here.

Are you operating on stale NVD data?

chadlwilson:
Actually, I am more confused - both CPEs are returned, however the CVEs aren't being found in the test run here which is unusual. Will worry about that later - probably not your problem.

aggeboe:
@chadlwilson FYI - the vulnerabilities reported by dependency-check

`apm-agent-api-1.51.0.jar (pkg:maven/co.elastic.apm/apm-agent-api@1.51.0, cpe:2.3:a:elastic:apm_agent:1.51.0:*:*:*:*:*:*:*, cpe:2.3:a:elastic:elastic_agent:1.51.0:*:*:*:*:*:*:*) : CVE-2024-52976, CVE-2023-46669`

Regarding stale NVD data. We're using the open-vulnerability-data-mirror. Thanks, will check with the ops team.

chadlwilson:
OK, we will probably also need to also suppress the `cpe:2.3:a:elastic:elastic_agent` CPE as well or you'll still get the FPs recorded against that. When filing the ticket here you probably chose the wrong CPE of the ones matched (which I should have checked, my bad).

aggeboe:
You're right - verified with local supression. I'll create a new FP issue (#7998) for the elastic_agent CPE

OrangeDog:
This suppression was incorrect. I am fixing it.

Issue #3876, opened 2021-12-14
False Positive on elastic-apm-agent


OrangeDog:
apm-agent-attach-1.28.1.jar\META-INF/maven/co.elastic.apm/apm-agent-common/pom.xml (`pkg:maven/co.elastic.apm/apm-agent-common@1.28.1`, `cpe:2.3:a:elastic:apm_agent:1.28.1:*:*:*:*:*:*:*`) : CVE-2019-7617
apm-agent-attach-1.28.1.jar (`pkg:maven/co.elastic.apm/apm-agent-attach@1.28.1`, `cpe:2.3:a:elastic:apm_agent:1.28.1:*:*:*:*:*:*:*`) : CVE-2019-7617
elastic-apm-agent-1.28.1.jar (`pkg:maven/co.elastic.apm/elastic-apm-agent@1.28.1`, `cpe:2.3:a:elastic:apm_agent:1.28.1:*:*:*:*:*:*:*`) : CVE-2019-7617

```xml
<dependency>
  <groupId>co.elastic.apm</groupId>
  <artifactId>apm-agent-attach</artifactId>
  <version>1.28.1</version>
</dependency>
```

CVE-2019-7617 is for the Python agent, not the Java agent.
The NVD entry specifies this: `cpe:2.3:a:elastic:apm_agent:*:*:*:*:*:python:*:*`

aikebah:
Would need enhancement in DependencyCheck to take `target_sw` CPE attribute into account

bremme:
I just ran into this issue using version 1.52.1 of the Java APM agent as well.

chadlwilson:
Was workaround fixed by #8005 , bigger picture issue needs fixing in #7139

Issue #7139, opened 2024-11-07
fix: use CPE target_sw to populate ecosystem to reduce FP


jeremylong:
The NVD has some CVEs such as https://nvd.nist.gov/vuln/detail/CVE-2022-31115, that use the target_sw to specify the runtime (e.g. ruby, java, etc.). This can be used to populate the ecosystem used to reduce false positives in the ODC analysis. See `org.owasp.dependencycheck.data.nvd.ecosystem.CveEcosystemMapper`.

OrangeDog:
This would fix e.g. #3460, #3876, #6851, and allow a general cleanup of bundled suppressions.

PR #8005, opened 2025-10-06
fix(fp): cleanup suppressions for Elastic APM Java

OrangeDog:
## Description of Change

Generated suppressions were variously wrong and not broad enough.

## Related issues

#3876
#7996
#7998 

N/A

### DIFF
--- a/generatedSuppressions.xml
+++ b/generatedSuppressions.xml
@@ -2487,15 +2487,8 @@ only pkg:maven/org.clojure:clojure@.* is the CPE cpe:/a:clojure:clojure
 </suppress>
 <suppress base="true">
    <notes><![CDATA[
-   FP per issue #7996
-   ]]></notes>
-   <packageUrl regex="true">^pkg:maven/co\.elastic\.apm/apm-agent-api@.*$</packageUrl>
-   <cpe>cpe:/a:elastic:apm_agent</cpe>
-</suppress>
-<suppress base="true">
-   <notes><![CDATA[
    FP per issue #7998
    ]]></notes>
-   <packageUrl regex="true">^pkg:maven/co\.elastic\.apm/apm-agent-api@.*$</packageUrl>
+   <packageUrl regex="true">^pkg:maven/co.elastic.apm/.*</packageUrl>
    <cpe>cpe:/a:elastic:elastic_agent</cpe>
 </suppress>
--- a/generatedSuppressions.xml
+++ b/generatedSuppressions.xml
@@ -2487,8 +2487,9 @@ only pkg:maven/org.clojure:clojure@.* is the CPE cpe:/a:clojure:clojure
 </suppress>
 <suppress base="true">
    <notes><![CDATA[
-   FP per issue #7998
+   FP per issues #3876, #7998
    ]]></notes>
-   <packageUrl regex="true">^pkg:maven/co.elastic.apm/.*</packageUrl>
+   <packageUrl regex="true">^pkg:maven/co\.elastic\.apm/.*</packageUrl>
    <cpe>cpe:/a:elastic:elastic_agent</cpe>
+   <cve>CVE-2019-7617</cve>
 </suppress>


generatedSuppressions.xml
```diff
@@ -2487,15 +2487,9 @@ only pkg:maven/org.clojure:clojure@.* is the CPE cpe:/a:clojure:clojure
 </suppress>
 <suppress base="true">
    <notes><![CDATA[
-   FP per issue #7996
+   FP per issues #3876, #7998
    ]]></notes>
-   <packageUrl regex="true">^pkg:maven/co\.elastic\.apm/apm-agent-api@.*$</packageUrl>
-   <cpe>cpe:/a:elastic:apm_agent</cpe>
```
chadlwilson, line 2493: Why removing this?

In my assessment in #7996  it wasn't "incorrect", it was just unnecessary for the given CVE at the time.

That package URL does not correspond to this CPE anyway (CPE: actual agent vs package: API classes -  so not the actual package itself) so the suppression doesnt really do any harm and is still accurate.

generatedSuppressions.xml
```diff
@@ -2487,15 +2487,9 @@ only pkg:maven/org.clojure:clojure@.* is the CPE cpe:/a:clojure:clojure
...
-<suppress base="true">
-   <notes><![CDATA[
-   FP per issue #7998
-   ]]></notes>
-   <packageUrl regex="true">^pkg:maven/co\.elastic\.apm/apm-agent-api@.*$</packageUrl>
+   <packageUrl regex="true">^pkg:maven/co\.elastic\.apm/.*</packageUrl>
    <cpe>cpe:/a:elastic:elastic_agent</cpe>
-</suppress>
\ No newline at end of file
+   <cve>CVE-2019-7617</cve>
```
chadlwilson, line 2494: I don't have a strong personal feeling but I'm not clear about when we do and do not generally suppress individual CVEs now, as it risks getting unworkable - even when they are due to a weakness in the tool as noted in #3876 and all the many others with "target sw" problems.

But I do note #7736 was done recently, which has a similar "tool cant tell" root cause.

Otherwise *technically* we should also suppress https://nvd.nist.gov/vuln/detail/CVE-2021-22133 to be consistent (a golang vuln on the same CPE).

@aikebah @jeremylong Do we have a policy somewhere on when we will consider suppressing individual CVEs? If we are doing this in some "tool weakness cases", I'd fix a number of the FP reports with individual CVE suppressions (the GRPC client ones are rife) and raise clear overarching issues for the underlying weakness so we dont have all these old FP Report / Enhancement tickets all over the place.

chadlwilson (approved):

generatedSuppressions.xml
OrangeDog: As far as I can see, NVD would assign the same CPE to this specific package.

Either this causes false negatives or it does nothing. In both cases it should be removed.

generatedSuppressions.xml
OrangeDog: > clear overarching issues for the underlying weakness

In this case: #7139

generatedSuppressions.xml
chadlwilson: I don't think you're correct, or I'd have reverted it myself.

Sorry, is there a reason for the attitude coming across in your messages? Did I do something to aggravate you?

OrangeDog: What attitude? I'm just answering the question.

I did some paging through search results and it turns out they assigned `elastic:apm_java_agent` for previous CVEs here. So this suppression would be good for the whole group, but because that CPE is apparently only used for non-Java products (meaning the specific CVE suppression isn't needed).

Is there a bundled hints file somewhere to try to fix the false negative?

OrangeDog: i.e. I'm now using this locally, but you'd probably dispute the broadness of the hint. I don't know how one should resolve trying to predict what NVD will do with a hypothetical future CVE.

```xml
<suppress>
    <packageUrl regex="true">^pkg:maven/co.elastic.apm/.*</packageUrl>
    <cpe>cpe:/a:elastic:apm_agent</cpe>
    <cpe>cpe:/a:elastic:elastic_agent</cpe>
</suppress>
```
```xml
<hint>
    <given><evidence type="product" source="pom" name="groupid" value="co.elastic.apm" confidence="HIGHEST"/></given>
    <add><evidence type="product" source="hint analyzer" name="product" value="apm_java_agent" confidence="HIGHEST"/></add>
</hint>
```

chadlwilson: > What attitude? I'm just answering the question

Alright. Personally I don’t see a need to be so matter-of-fact in the way you claim your own correctness as if there is no alternative viewpoint to your own (such as asserting a false dichotomy logical fallacy above) including on https://github.com/dependency-check/DependencyCheck/issues/7996#issuecomment-3371019300

To me that comes across as ‘attitude’, and I generally choose to work with people in OSS who are keen to show some grace in how they interact with those trying to help. I’ll be on my way and someone else can adjudicate this in any case.

OrangeDog: I specifically said "As far as I can see" because it's only what I thought it was and maybe someone else knows better...

generatedSuppressions.xml
jeremylong: @aikebah we do not have a policy - but if it makes sence we can. I believe we have done this in the past for MySQL.

jeremylong (approved):

OrangeDog:
@jeremylong wait, I don't think this is right any more. Please see the discussion.

PR #8007, opened 2025-10-06
fix(fp): cleanup suppressions for Elastic APM Java

OrangeDog:
## Description of Change

#8005 was merged mid-discussion. After closer investigation of the NVD data, this appears to be the better solution.

### DIFF
--- a/generatedSuppressions.xml
+++ b/generatedSuppressions.xml
@@ -2487,15 +2487,8 @@ only pkg:maven/org.clojure:clojure@.* is the CPE cpe:/a:clojure:clojure
 </suppress>
 <suppress base="true">
    <notes><![CDATA[
-   FP per issue #7996
-   ]]></notes>
-   <packageUrl regex="true">^pkg:maven/co\.elastic\.apm/apm-agent-api@.*$</packageUrl>
-   <cpe>cpe:/a:elastic:apm_agent</cpe>
-</suppress>
-<suppress base="true">
-   <notes><![CDATA[
    FP per issue #7998
    ]]></notes>
-   <packageUrl regex="true">^pkg:maven/co\.elastic\.apm/apm-agent-api@.*$</packageUrl>
+   <packageUrl regex="true">^pkg:maven/co.elastic.apm/.*</packageUrl>
    <cpe>cpe:/a:elastic:elastic_agent</cpe>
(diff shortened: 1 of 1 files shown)


chadlwilson (approved):

generatedSuppressions.xml
```diff
@@ -2487,13 +2487,12 @@ only pkg:maven/org.clojure:clojure@.* is the CPE cpe:/a:clojure:clojure
 </suppress>
 <suppress base="true">
    <notes><![CDATA[
-   FP per issues #3876, #7998
+   FP per issues #7996, #7998, #8005
    ]]></notes>
    <packageUrl regex="true">^pkg:maven/co\.elastic\.apm/.*</packageUrl>
+   <cpe>cpe:/a:elastic:apm_agent</cpe>
```
chadlwilson, line 2493: I’m not sure this is a good idea since https://nvd.nist.gov/vuln/detail/CVE-2021-37941 is raised against this CPE you are about to suppress (albeit with a language/target variant that ODC doesn’t support)

Presumably this applies to one of the artifacts under this package group given https://discuss.elastic.co/t/apm-java-agent-security-update/289627

NVD simply aren’t consistent at how they match CvEs to CPEs over time.

OrangeDog: Well spotted. I had looked through it but their search interface is terrible.

> NVD simply aren’t consistent at how they match CvEs to CPEs over time.

So it would seem. How very vexing of them.

chadlwilson: The new search interface is a bit weird. You have to use the ‘CPE applicability’ option. If you use CPE names, it seems to have to match exactly any parts which are populated for any vuln using the CPE. Which doesn’t work well when those extended fields might be set (or not).

https://nvd.nist.gov/vuln/search#/nvd/home?cpeFilterMode=applicability&cpeName=cpe:2.3:a:elastic:apm_agent:1.13:*:*:*:*:*:*:*&resultType=records

Vs

https://nvd.nist.gov/vuln/search#/nvd/home?cpeFilterMode=cpe&cpeName=cpe:2.3:a:elastic:apm_agent:1.13:*:*:*:*:*:*:*&resultType=records

found this out the painful way when fixing #7970

Issue #8051, opened 2025-10-15
[FP]: spdx-java-model-2_X


goneall:
### Package URl

pkg:maven/org.spdx/spdx-java-model-2_X@1.0.1

### CPE

cpe:2.3:a:x.org:x.org:1.0.1:*:*:*:*:*:*:* 

### CVE

CVE-2006-0197 

### ODC Integration

{"label" => "Maven Plugin"}

### ODC Version

12.1.8

### Description

The package URI `pkg:maven/org.spdx/spdx-java-model-2_X@1.0.1` seems to somehow been associated with [cpe:2.3:a:x.org:x.org:1.0.1:*:*:*:*:*:*:*](https://nvd.nist.gov/vuln/search#/nvd/home?sortOrder=3&sortDirection=2&cpeFilterMode=applicability&resultType=records&cpeName=cpe%3A2.3%3Aa%3Ax.org%3Ax.org%3A1.0.1%3A%2A%3A%2A%3A%2A%3A%2A%3A%2A%3A%2A%3A%2A) which is x.org software.  The spdx-java-model-2_X does not include any x.org software directly or indirectly.

chadlwilson:
This isn't showing as matching that CPE in the test run here, so adding a suppression for a false positive is unlikely to help.

<img width="1344" height="344" alt="Image" src="https://github.com/user-attachments/assets/51c0ccab-0ebc-40cb-84a0-517767b6a801" />

Are you sure this is being matched with a maven-defined artifact, or is this being scanned from inside another archive/uber-jar or something like that?

Can you perhaps share what your report looks like, and your configuration?

goneall:
Thanks @chadlwilson for the quick response.

> This isn't showing as matching that CPE in the test run here

Interesting - could I be using a stale CVE/CPE database?

Attached is a copy of the report and the configuration file.

[dependency-check-report.html](https://github.com/user-attachments/files/22933574/dependency-check-report.html)

[dependency-check-supress.xml](https://github.com/user-attachments/files/22933560/dependency-check-supress.xml)

I was able to suppress the errors successfully.

I'm running the Maven plugin dependency check which generates the FP `mvn dependency-check:check`

Here's a link to the POM file: https://github.com/spdx/tools-java/blob/master/pom.xml

Also, here's the output from the maven command - I just noticed a couple of warnings - let me know if those may impact this issue.

```
C:\Users\garys\git\tools-java>mvn dependency-check:check
[...]

Issue #4414, opened 2022-04-26
[FP]: icu4j-61.1


urld:
### Package URl

pkg:maven/com.ibm.icu/icu4j@61.1

### CPE

`cpe:2.3:a:apple:java:61.1:*:*:*:*:*:*:*`, `cpe:2.3:a:apple:java_1.6:61.1:*:*:*:*:*:*:*`, `cpe:2.3:a:unicode:international_components_for_unicode:61.1:*:*:*:*:*:*:*`, `cpe:2.3:a:unicode:unicode:61.1:*:*:*:*:*:*:*`

### CVE

CVE-2020-21913, CVE-2010-0538

### ODC Integration

{"label"=>"Maven Plugin"}

### ODC Version

7.1.0

Issue #1073, opened 2018-01-19
Duplicate CPE for ICU?


TobiX:
There seem to be different CPE IDs for the ICU project:

- [`cpe:2.3:a:icu_project:international_components_for_unicode:52.0:*:*:*:*:*:*:*`](https://nvd.nist.gov/products/cpe/detail/331302) (Excluded in response to #851) ([search](https://nvd.nist.gov/products/cpe/search/results?status=FINAL&orderBy=CPEURI&namingFormat=2.3&keyword=cpe:2.3:a:icu_project))
- [`cpe:2.3:a:icu-project:international_components_for_unicode:52.1:*:*:*:*:c\/c\+\+:*:*
`](https://nvd.nist.gov/products/cpe/detail/347895) ([search](https://nvd.nist.gov/products/cpe/search/results?status=FINAL&orderBy=CPEURI&namingFormat=2.3&keyword=cpe:2.3:a:icu-project))

When updating from dependency-check 2.1.1 to 3.1.0, every ICU4J dependency is now reclassified to the latter CPE and the dependency-check report now suggests to me:

```
<suppress>
   <notes><![CDATA[
   file name: icu4j-52.1.jar
   ]]></notes>
   <gav regex="true">^com\.ibm\.icu:icu4j:.*$</gav>
   <cpe>cpe:/a:icu-project:international_components_for_unicode</cpe>
</suppress>
```

I already wrote a mail to the NVD maintainers, so I expect that `icu_project` and `icu-project` will be merged in the database.

Related: Would it be possible to only remove the missqualification to `cpe:2.3:a:icu-project:international_components_for_unicode:*:*:*:*:*:c/c++:*:*` and keep the `cpe:2.3:a:icu-project:international_components_for_unicode:*:*:*:*:*:java:*:*` ID?

PS: dependency-check shows `c/c++` as `c%2fc%2b%2b` - can this be made more readable?

TobiX:
NVD has removed the `icu_project` CPE now.

jeremylong:
Thanks for following up with the NVD.  Can this issue be closed?

TobiX:
Not sure. The `icu_project` CPE was removed, but there are still CVEs using that CPE:

[...]

Issue #851, opened 2017-08-21
icu4j-58.2.jar mis-identified as C/C++ implementation of ICU


jdimmerman:
There are two known vulnerabilities in the C/C++ implementation of International Components for Unicode (ICU): CVE-2017-7867 and CVE-2017-7868.  
Dependency checker is correctly identifying icu4j version 58.2 as what it is, but is then associating it with these two CVEs, which is not correct as they don't affect the java implementation (insofar as the research I've done shows). 

Log attached:
[dependency-check.log.zip](https://github.com/jeremylong/DependencyCheck/files/1239963/dependency-check.log.zip)

```xml
<dependency>
    <fileName>icu4j-58.2.jar</fileName>
    <filePath>/root/.gradle/caches/modules-2/files-2.1/com.ibm.icu/icu4j/58.2/db9fd4b4c189cf1518db14c67d14a2cfcfbe59f6/icu4j-58.2.jar</filePath>
    <md5>605d8a0276a280ff6332c3bd26071180</md5>
    <sha1>db9fd4b4c189cf1518db14c67d14a2cfcfbe59f6</sha1>
    <description>
International Component for Unicode for Java (ICU4J) is a mature, widely used Java library
providing Unicode and Globalization support 
</description>
    <license>Unicode/ICU License: http://source.icu-project.org/repos/icu/trunk/icu4j/main/shared/licenses/LICENSE</license>
    <evidenceCollected>
        <evidence type="vendor" confidence="HIGHEST">
            <source>central</source>
            <name>groupid</name>
            <value>com.ibm.icu</value>
        </evidence>
        <evidence type="vendor" confidence="HIGH">
            <source>file</source>
            <name>name</name>
            <value>icu4j</value>
        </evidence>
        <evidence type="vendor" confidence="LOW">
            <source>jar</source>
            <name>package name</name>
            <value>ibm</value>
        </evidence>
        <evidence type="vendor" confidence="LOW">
            <source>jar</source>
            <name>package name</name>
            <value>icu</value>
        </evidence>
        <evidence type="vendor" confidence="LOW">
            <source>Manifest</source>
            <name>bundle-copyright</name>
            <value>Copyright 2000-2016, International Business Machines Corporation and others.  All Rights Reserved.</value>
        </evidence>
        <evidence type="vendor" confidence="MEDIUM">
            <source>manifest</source>
            <name>Bundle-Description</name>
            <value>International Components for Unicode for Java</value>
        </evidence>
[...]

PR #8062, opened 2025-10-17
fix(fp): Consolidate/update icu4j suppressions for false positives

chadlwilson:
## Description of Change

- Update the list of suppressed icu4j CVEs for java due to lack of support for the target sw ecosystem.
- Sorted it so it can be compared easily with the search
- Consolidate into single suppression across multiple java packages for ease of maintenance

Reconciled [with the NVD](https://nvd.nist.gov/vuln/search#/nvd/home?sortOrder=1&sortDirection=1&cpeFilterMode=applicability&cpeName=cpe:2.3:a:*:international_components_for_unicode:*:*:*:*:*:*:*:*&resultType=records) and checked all are C/C++.

## Related issues

- relates to earlier #851  #1073 and #4414
- noted while investigating #8051 

N/A


jeremylong (approved):
LGTM

PR #8066, opened 2025-10-19
feat: usage telemetry via scarf

jeremylong:
## Description of Change

Utilize Scarf API to better understand ODC usage.

yes

### DIFF
--- a/utils/src/main/java/org/owasp/dependencycheck/utils/scarf/TelemetryCollector.java
+++ b/utils/src/main/java/org/owasp/dependencycheck/utils/scarf/TelemetryCollector.java
@@ -1,0 +1,93 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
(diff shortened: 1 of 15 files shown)


utils/src/main/java/org/owasp/dependencycheck/utils/scarf/TelemetryCollector.java
```diff
@@ -0,0 +1,93 @@
...
+    private static final String BASE_URL = "https://dependency-check.gateway.scarf.sh/scan/";
+    private static final AtomicBoolean telemetrySubmitted = new AtomicBoolean(false);
+
+    public static void send(Settings settings) {
+        try {
+            String tool = settings.getString(Settings.KEYS.APPLICATION_NAME, "dependency-check");
+            String version = settings.getString(Settings.KEYS.APPLICATION_VERSION, "Unknown");
+            send(settings, tool, version);
+        } catch (Exception e) {
+            // Silent catch block
```
nhumblot, line 46: **issue(non-blocking):** there are 3 swallowed exceptions in this class, reducing our ability to understand the behavior of the application when an exception might occur.

nhumblot: **suggestion:** we could add a debug or trace log statement to be informed of the encountered exceptions when executing Dependency Check with a fine grained log level

nhumblot (approved):

utils/src/main/java/org/owasp/dependencycheck/utils/scarf/TelemetryCollector.java
jeremylong: Understood, I went with the option of telemetry isn't as important and if a user had an issue I doubt it would get reported as anything that occurs in this should be non-blocking and should not trickle up to the end user.

PR #8077, opened 2025-10-22
docs: fix #8076 - Error in documentation "Suppressing False Positives"

### LINKED ISSUE #8076: Error in documentation "Suppressing False Positives"
**Describe the bug**
In the documentation "Suppressing False Positives", the sample to select all CVE is wrong 
It is documented as:
`<vulnerabilityName regex="true"></vulnerabilityName>`
but it must be:
`<vulnerabilityName regex="true">.*</vulnerabilityName>`

I tested with the sample and it was not working until I updated it to use `.*`.

eballetbaz:
## Description of Change

In the documentation "Suppressing False Positives", the sample to select all CVE is wrong
It is documented as:
<vulnerabilityName regex="true"></vulnerabilityName>
but it must be:
<vulnerabilityName regex="true">.*</vulnerabilityName>

## Related issues

Fix #8076 - Error in documentation "Suppressing False Positives"

### DIFF
--- a/src/site/markdown/general/suppression.md
+++ b/src/site/markdown/general/suppression.md
@@ -82,7 +82,7 @@ HTML version of the report. The other common scenario would be to ignore all CVE
         This suppresses false positives identified on spring security.
         ]]></notes>
         <gav regex="true">org\.springframework\.security:spring.*</gav>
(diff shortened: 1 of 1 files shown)


jeremylong (approved):
LGTM

PR #8116, opened 2025-11-10
feat: package and utilize generated suppression file

jeremylong:
## Description of Change

First of three PRs around cleaning up the differences between the generated and base suppression file. This PR enhances the build to download the current generated suppression file, packages it into the core lib, and loads both the base and generated suppression files during execution.


core/src/main/java/org/owasp/dependencycheck/analyzer/AbstractSuppressionAnalyzer.java
```diff
@@ -195,17 +199,28 @@ private void loadSuppressionBaseData(final Engine engine) throws SuppressionPars
      * @throws SuppressionParseException thrown if the XML cannot be parsed.
      */
     private void loadPackagedSuppressionBaseData(final SuppressionParser parser, final Engine engine) throws SuppressionParseException {
+        loadPackagedSuppressionBaseData(BASE_SUPPRESSION_FILE, parser, engine);
+        loadPackagedSuppressionBaseData(GENERATED_SUPPRESSION_FILE, parser, engine);
```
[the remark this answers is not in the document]
jeremylong: both mechanisms are needed as the publishedSuppressions.xml may be updated after initial publication.

core/pom.xml
```diff
@@ -100,6 +100,26 @@ Copyright (c) 2012 Jeremy Long. All Rights Reserved.
             </testResource>
         </testResources>
         <plugins>
+            <plugin>
+                <groupId>org.codehaus.mojo</groupId>
+                <artifactId>exec-maven-plugin</artifactId>
+                <executions>
+                    <execution>
+                        <phase>process-resources</phase>
```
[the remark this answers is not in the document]
jeremylong:
```suggestion
                        <phase>generate-resources</phase>
```

core/pom.xml
```diff
@@ -100,6 +100,25 @@ Copyright (c) 2012 Jeremy Long. All Rights Reserved.
             </testResource>
         </testResources>
         <plugins>
+            <plugin>
+                <groupId>com.googlecode.maven-download-plugin</groupId>
+                <artifactId>download-maven-plugin</artifactId>
+                <version>1.6.3</version>
```
jeremylong:
```suggestion
                <version>1.13.0</version>
```

pom.xml
```diff
@@ -187,6 +187,11 @@ Copyright (c) 2012 - Jeremy Long
         <defaultGoal>clean install</defaultGoal>
         <pluginManagement>
             <plugins>
+                <plugin>
+                    <groupId>org.codehaus.mojo</groupId>
+                    <artifactId>exec-maven-plugin</artifactId>
+                    <version>3.6.2</version>
```
jeremylong:
```suggestion
                    <groupId>com.googlecode.maven-download-plugin</groupId>
                    <artifactId>download-maven-plugin</artifactId>
                    <version>1.13.0</version>
```

core/pom.xml
```diff
@@ -100,6 +100,25 @@ Copyright (c) 2012 Jeremy Long. All Rights Reserved.
             </testResource>
         </testResources>
         <plugins>
+            <plugin>
+                <groupId>com.googlecode.maven-download-plugin</groupId>
+                <artifactId>download-maven-plugin</artifactId>
+                <version>1.13.0</version>
```
jeremylong:
```suggestion
```

core/src/main/java/org/owasp/dependencycheck/analyzer/AbstractSuppressionAnalyzer.java
```diff
@@ -195,17 +199,28 @@ private void loadSuppressionBaseData(final Engine engine) throws SuppressionPars
...
+    }
+    /**
+     * Loads the base suppression rules packaged with the application.
+     *
+     * @param parser The suppression parser to use
```
jeremylong:
```suggestion
     * Loads the suppression rules packaged with the application.
     *
     * @param packagedFileName the name of the packaged suppression file to load
     * @param parser The suppression parser to use
```

nhumblot (dismissed):

chadlwilson:
With this approach, ODC needs to load the generatedSuppressions twice, once off packaged, and again off "hosted" suppressions within the `HostedSuppressionsDataSource` - which will largely overlap. Does it de-duplicate these in memory at least? Is this a significant source of slowness?

It's slightly more complex, but I wonder if it would be cleaner to make this self contained somewhere within the below to just pre-populate the user's cache in the data directory on disk; to keep all logic and workflow related to these suppressions self-contained, and with consistent language?

With this approach it would happen
- on first use, prior to any attempt to update
- after hosted suppressions are purged
- regardless of whether hosted suppressions updates are enabled.

https://github.com/dependency-check/DependencyCheck/blob/a9f313af4a27b0e9c08b2e534afa7328440e36d4/core/src/main/java/org/owasp/dependencycheck/analyzer/AbstractSuppressionAnalyzer.java#L260-L264

core/src/main/java/org/owasp/dependencycheck/analyzer/AbstractSuppressionAnalyzer.java
```diff
@@ -68,6 +68,10 @@ public abstract class AbstractSuppressionAnalyzer extends AbstractAnalyzer {
      * The file name of the base suppression XML file.
      */
     private static final String BASE_SUPPRESSION_FILE = "dependencycheck-base-suppression.xml";
+    /**
+     * The file name of the generated suppression XML file.
+     */
+    private static final String GENERATED_SUPPRESSION_FILE = "generated-suppressions.xml";
```
chadlwilson: If we keep this approach, perhaps we could try and align on language to make it a bit less confusing.

Right now it is referred to as
- "hosted" suppressions (main runtime terminology at runtime in ODC; but implies a remote source)
- "generated" suppressions (really refers to the automation workflows and the special branch they are tracked on)
- "published" suppressions (the file name when promoted to GitHub pages from source control, and bootstrapped here)

I wonder if referring to as `dependencycheck-published-suppression-snapshot.xml` would be better, or `dependencycheck-hosted-suppression-snapshot.xml`.

pom.xml
```diff
@@ -187,6 +187,11 @@ Copyright (c) 2012 - Jeremy Long
         <defaultGoal>clean install</defaultGoal>
         <pluginManagement>
             <plugins>
+                <plugin>
+                    <groupId>com.googlecode.maven-download-plugin</groupId>
+                    <artifactId>download-maven-plugin</artifactId>
+                    <version>1.13.0</version>
```
chadlwilson:
```suggestion
                    <groupId>io.github.download-maven-plugin</groupId>
                    <artifactId>download-maven-plugin</artifactId>
                    <version>2.0.0</version>
```

Artifact was relocated; but no other changes it seems. https://github.com/download-maven-plugin/download-maven-plugin/releases/tag/2.0.0

chadlwilson (approved):

core/pom.xml
```diff
@@ -100,6 +100,24 @@ Copyright (c) 2012 Jeremy Long. All Rights Reserved.
             </testResource>
         </testResources>
         <plugins>
+            <plugin>
+                <groupId>com.googlecode.maven-download-plugin</groupId>
```
nhumblot: **issue:** this seems to generate the following warning when performing a `mvn -s settings.xml clean install`:

```
[WARNING] 
[WARNING] Some problems were encountered while building the effective model for org.owasp:dependency-check-core:jar:12.1.10-SNAPSHOT
[WARNING] 'build.plugins.plugin.version' for com.googlecode.maven-download-plugin:download-maven-plugin is missing. @ line 103, column 21
[WARNING] 
[WARNING] It is highly recommended to fix these problems because they threaten the stability of your build.
[WARNING] 
[WARNING] For this reason, future Maven versions might no longer support building such malformed projects.
[WARNING] 
```

nhumblot:
```suggestion
                <groupId>io.github.download-maven-plugin</groupId>
```

nhumblot:
> With this approach, ODC needs to load the generatedSuppressions twice, once off packaged, and again off "hosted" suppressions within the `HostedSuppressionsDataSource` - which will largely overlap. Does it de-duplicate these in memory at least? Is this a significant source of slowness?
> 
> It's slightly more complex, but I wonder if it would be cleaner to make this self contained somewhere within the below to just pre-populate the user's cache in the data directory on disk; to keep all logic and workflow related to these suppressions self-contained, and with consistent language?
> 
> With this approach it would happen
> 
>     * on first use, prior to any attempt to update
> 
>     * after hosted suppressions are purged
> 
>     * regardless of whether hosted suppressions updates are enabled.
> 
> 
> https://github.com/dependency-check/DependencyCheck/blob/a9f313af4a27b0e9c08b2e534afa7328440e36d4/core/src/main/java/org/owasp/dependencycheck/analyzer/AbstractSuppressionAnalyzer.java#L260-L264

On my machine, a forced update took 148ms to be loaded on 4G connection with a phone.

We seem to load all elements in a `List<SuppressionRule>` and we add them through a `void addAll(List<SuppressionRule>)` call. I don't see any de-duplication so my guess is we are duplicating suppression rules between base and hosted suppressions as of now. It does not change much from now, so I may say it can be a future improvement opportunity (impact would have to be more thoroughly investigated) that could be tracked, and not a necessary change in this PR :thinking:

chadlwilson:
My concern is more about evaluation/analysis rather than loading the file.

I haven't looked at how smart the algorithm is or how parallelized, but my concern would be that there is an O(n^2) issue here, e.g if every suppression is evaluated for every single possible dependency. We also don't really have any automation to ensure the regexes aren't insanely expensive which might have a multiplicative effect.

On a bigger project with, say, 1,000 dependencies (JavaScript!!) and lots of FPs due to use of CLI and generic dependency names, that might be an issue - either slowness or excessive junk/GC production.

Aside from perf, I also think it's a bit confusing to have 3 data sources in the code to reason about, which is why I suggest using this file essentially as an initial cache value for HostedSuppressions instead (when working offline, or if stale) since it is not valid/useful to have both the packaged version and a version received remotely. I'm happy to test and suggest a PR to this one if @jeremylong wants to see what it might look like.

That'd avoid having to do any de-duplication of individual suppressions since we can generally assume there are very limited duplications between base and hosted/generated/published (especially after Jeremy's adjacent cleanup PRs).

I think that helps us as it's already quite tough figuring out why we might not be able to replicate a given FP given combinations of (ODC version, ODC variant, enabled analyzers, enabled data sources, available data sources, hosted suppressions currentness, offline status, available local tools)

jeremylong:
I like the idea of the embedded being the initial cache - especially if you don't mind putting together the PR ;).

I was also going to make sure there was a mechanism to prevent loading the same rule twice...

core/pom.xml
chadlwilson: Oops, missed this in my proposed edit earlier - thx!

jeremylong:
@chadlwilson mind taking another look at the PR after my updates?  We should only be using the hosted suppression file that gets packaged if the file was not able to be downloaded during the update process.

nhumblot:
Sorry, just came back home with 2 twins newborn to take care of. I will try to do a review at some point.

chadlwilson:
Yeah I haven't had a chance to look at this or complete my alternative.

The logic here is a bit more complex than I anticipated because of how things are modelled (data source vs load of base suppressions that happens out of band).

nhumblot (approved):

PR #8121, opened 2025-11-11
fix: correct bundle audit gem in Dockerfile

chadlwilson:
## Description of Change

The correct gem is bundler-audit, the current one is just a typo-squat wrapper gem.

## Related issues

N/A

N/A

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -24,7 +24,7 @@ ADD cli/target/dependency-check-${VERSION}-release.zip /
 RUN apk update                                                                                       && \
     apk add --no-cache --virtual .build-deps curl tar                                                && \
     apk add --no-cache git ruby ruby-rdoc npm                                                        && \
-    gem install bundler-audit                                                                         && \
+    gem install bundler-audit                                                                        && \
     bundle audit update                                                                              && \
     mkdir /opt/yarn                                                                                  && \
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -24,7 +24,7 @@ ADD cli/target/dependency-check-${VERSION}-release.zip /
 RUN apk update                                                                                       && \
     apk add --no-cache --virtual .build-deps curl tar                                                && \
     apk add --no-cache git ruby ruby-rdoc npm                                                        && \
-    gem install bundle-audit                                                                         && \
+    gem install bundler-audit                                                                         && \
```
jeremylong:
```suggestion
    gem install bundler-audit                                                                        && \
```

jeremylong (approved):

jeremylong:
actually - I'm not even sure if we need to install this anymore. I think it is included.

chadlwilson:
Nah, you still need it. It's not included with ruby or bundler - not one of the default gems https://stdgems.org/ nor standard commands https://bundler.io/v2.7/man/bundle.1.html

I install it manually as a pre ODC task for GoCD. https://github.com/gocd/gocd/blob/1fe52a36582719aa3e314d57b3d1c647950af14f/build.gradle#L783

PR #8138, opened 2025-11-18
chore: remove duplicate suppression rules from base that are in the generated branch

jeremylong:
Removes duplicate suppressions from base file that exist in the generated branch.

### DIFF
--- a/core/src/main/resources/dependencycheck-base-suppression.xml
+++ b/core/src/main/resources/dependencycheck-base-suppression.xml
@@ -4970,143 +4970,10 @@
 
     <suppress base="true">
         <notes><![CDATA[
-        FP per issue #4806
-        ]]></notes>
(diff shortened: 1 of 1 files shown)


chadlwilson (approved):
LGTM. I diffed to `generatedSuppressions` and numbers for overlapping lines looks about right. Also spot checked a number to see they are still within the `generatedSuppressions`.

chadlwilson:
Most of the stuff that is difficult to diff here is due to all the edits to the `_project` CPEs on base.

After this PR, we can perhaps remove the old ones from `generatedSuppressions`, since they have been correct in base for ~6 months via `12.1.3`; there have been quite a few reasons for people to be forced to upgrade recently?

Alternatively we could sync them across to `generatedSuppressions` and then remove from base.

PR #8202, opened 2025-12-28
fix(ant): resolve relative paths against basedir

### LINKED ISSUE #7918: Ant task does not use project-relative paths for suppression file and output directory
When using the ant task with a `reportoutputdirectory` attribute and a `<suppressionfile>` child element where the paths are relative to the project root, running the ant task from the same directory as the build.xml file works as expected. However, when running e.g. `ant -f /path/to/build.xml dependency-check` from another directory, the suppression file is not found (unless fully-qualified from within the build script itself) and the report is written to a directory relative to the CWD of the invocation and not relative to the build.xml file, which is the typical expectation for relative paths in ant build scripts.

The problem occurs using version 12.1.0 of the Apache ant plugin

**Log file**

**To Reproduce**
1. Use the following build.xml script in e.g. a `temp` directory:
```xml
<project name="test" default="dependency-check" basedir=".">
  <!-- TODO: Set this path properly -->
  <property name="dependency-check.home" value="${user.home}/packages/dependency-check-ant" />

  <target name="prep">
    <echo file="dependency-check-suppressions.xml">
&lt;suppressions xmlns="https://jeremylong.github.io/DependencyCheck/dependency-suppression.1.3.xsd"&gt;
&lt;/suppressions&gt;
    </echo>
  </target>

  <target name="dependency-check" depends="prep">
    <path id="dependency-check.path">
      <pathelement location="${dependency-check.home}/dependency-check-ant.jar"/>
      <fileset dir="${dependency-check.home}/lib">
        <include name="*.jar"/>
      </fileset>
    </path>
    <taskdef resource="dependency-check-taskdefs.properties">
      <classpath refid="dependency-check.path" />
    </taskdef>

    <property name="dependency-check.output-dir" value="dep-check-report-dir" />
    <mkdir dir="${dependency-check.output-dir}" />
    <dependency-check projectname="${ant.project.name}" reportoutputdirectory="${dependency-check.output-dir}">
      <suppressionfile path="dependency-check-suppressions.xml" />
      <fileset dir="${dependency-check.home}">
        <include name="**/*.jar"/>
      </fileset>
    </dependency-check>
  </target>
</project>
```
2. Customize the value of the dependency-check ant library home
3. Run `ant`

This should run successfully, creating a suppression file alongside the build.xml script, creating a directory called dep-check-report-dir, and placing the report in that directory.

4. `cd ..`
5. `ant -f temp/build.xml`

This will fail because the suppressions file does not exist. It _does_ exist, but it's in the temp directory, alongside the build.xml script. If you fix the reference to the suppression file to be `${basedir}/dependency-check-suppressions.xml", then the script completes, but the report is put into an unexpected place: `./dep-check-report-dir` instead of `./temp/dep-check-report-dir`.

Ant's documentation says that any non-absolute path should be relative to the project's `basedir`, which defaults to the directory containing the `build.xml` file.

It would be nice for dependency-check to resolve these relative paths in the same way, for consistency.

Artuiro-23:
## Description of Change

This change fixes how relative paths are resolved in the Dependency-Check Ant task.

Previously, when running Ant from a different working directory (for example using
`ant -f /path/to/build.xml`), relative paths provided for `suppressionFile` and
`reportOutputDirectory` were resolved against the current working directory instead
of the Ant project `basedir`.

This update ensures that all non-absolute paths are resolved relative to the
project `basedir`, which is consistent with standard Ant behavior and user
expectations.

## Related issues

- Fixes #7918
- 

No

Manual verification was performed because this change affects runtime behavior
that depends on how Ant resolves paths when invoked from different working
directories (for example using `ant -f /path/to/build.xml`). This scenario is
difficult to reliably simulate in unit tests.

Verification was done by executing Ant from outside the project directory and
confirming that suppression files and report output directories are correctly
resolved relative to the project `basedir` (the directory containing the
`build.xml` file).

### DIFF
--- a/ant/src/main/java/org/owasp/dependencycheck/taskdefs/Check.java
+++ b/ant/src/main/java/org/owasp/dependencycheck/taskdefs/Check.java
@@ -202,7 +202,7 @@ public class Check extends Update {
      * Specifies the destination directory for the generated Dependency-Check
      * report.
      */
-    private String reportOutputDirectory = ".";
+    private String reportOutputDirectory;
     /**
      * If using the JUNIT report format the junitFailOnCVSS sets the CVSS score
      * threshold that is considered a failure. The default is 0.
@@ -528,7 +528,7 @@ public class Check extends Update {
      * @param suppressionFile the suppression file to add.
      */
     public void addConfiguredSuppressionFile(final SuppressionFile suppressionFile) {
-        suppressionFiles.add(suppressionFile.getPath());
+        suppressionFiles.add(resolveRelative(suppressionFile.getPath()));
     }
 
     /**
@@ -573,13 +573,26 @@ public class Check extends Update {
         this.projectName = projectName;
     }
 
+    private String resolveRelative(String path) {
+        if (path == null) {
+            return null;
+        }
+
+        File file = new File(path);
+        if (file.isAbsolute()) {
+            return path;
+        }
+
+        return new File(getProject().getBaseDir(), path).getPath();
+    }
+
     /**
      * Set the value of reportOutputDirectory.
      *
      * @param reportOutputDirectory new value of reportOutputDirectory
      */
     public void setReportOutputDirectory(String reportOutputDirectory) {
-        this.reportOutputDirectory = reportOutputDirectory;
+        this.reportOutputDirectory = resolveRelative(reportOutputDirectory);
     }
 
     /**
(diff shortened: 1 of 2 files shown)


jeremylong (approved):

chadlwilson:
The failing test here is also the one that starting failing at 2026-01-01 and subsequently fixed on master so should be OK with a merge/rebase from master (or as-is, although I have not reviewed the code since not familiar with the ant integration)

Issue #7514, opened 2025-03-10
Support for NIST 2.0 data feeds


andham:
**Is your feature request related to a problem? Please describe.**
The current usage of NIST 2.0 APIs causes issues in some corporate environments, such as air-gapped.

**Describe the solution you'd like**
Using the NIST data feed is much easier to handle in such an environment as creating an mirror of the feeds is easier than "api-mirror". NIST has recently announced (see https://www.nist.gov/itl/nvd) that they will support 2.0 data feeds in parallell with the 2.0 APIs.
So the possibility to use the new 2.0 data feeds instead of the 2.0 APIs would we great!

**Describe alternatives you've considered**
An alternative solution could be to set up an internal centralized vuln database which we create from the data feeds, and have dep-check use that database. However, that would require centralized management instead of just creating a self-contained container image.

**Additional context**
n/a

aikebah:
*plan to support*, but it all needs to be built first by NVD. Following that it should be fairly easy to implement and support those alongside the current support of the 2.0 API I expect, but for the time being there is nothing we can do.

andham:
The new 2.0 data feeds are now available. See https://www.nist.gov/itl/nvd
The old 1.1 data feeds will be removed on Aug 20.

jeremylong:
I love requests like this. So easy to implement. Current release of ODC (with zero updates):


```bash
$ ./dependency-check.sh --updateonly    --nvdDatafeed https://nvd.nist.gov/feeds/json/cve/2.0/nvdcve-2.0-\{0\}.json.gz    
[INFO] Checking for updates
[INFO] NVD API Cache requires several updates; this could take a couple of minutes.
[INFO] Download Started for NVD Cache - https://nvd.nist.gov/feeds/json/cve/2.0/nvdcve-2.0-2012.json.gz
[INFO] Download Started for NVD Cache - https://nvd.nist.gov/feeds/json/cve/2.0/nvdcve-2.0-2011.json.gz
[INFO] Download Started for NVD Cache - https://nvd.nist.gov/feeds/json/cve/2.0/nvdcve-2.0-2010.json.gz
[INFO] Download Started for NVD Cache - https://nvd.nist.gov/feeds/json/cve/2.0/nvdcve-2.0-2009.json.gz
[INFO] Download Started for NVD Cache - https://nvd.nist.gov/feeds/json/cve/2.0/nvdcve-2.0-2008.json.gz
[...]

PR #8205, opened 2026-01-01
fix: Allow NVD data feed metadata downloads to fail on 1st Jan while logging correct errors

### LINKED ISSUE #8204: Using parameter `nvdDatafeed` stopped working (feed 2.0)
**Precondition**
- [X] I checked the issues list for existing open or closed reports of the same problem.

**Describe the bug**
We have a periodical job on our CI servers to keep the DB up-to-date. We download all data through a caching proxy, so we are using the switch `nvdDatafeed` in both the CLI and Maven Version of DC. A couple of hours ago (between 10am and 12pm CET), dependency-check stopped updating because our proxy gave it a 404 for the file `cache.properties`. If I run the command with the direct NVD url, the same error occurs. If I run the command without the feed URL, everything works fine.

The problem occurs using version 12.1.9 of **both** the cli and maven plugin.

**Log file**
[Log](https://gist.github.com/malice00/6da2e36aa37e048917c68e54fd3ee0c0)

**To Reproduce**
1. Run DC with the `nvdDatafeed`-parameter pointing to the NVD 2.0 feed

**Expected behavior**
I expect DC to be able to download & update the DB from the cached feed.

**Additional context**
I configured our proxy and DC a while back using hints from [this issue comment](https://github.com/dependency-check/DependencyCheck/issues/7514#issuecomment-2909234149)

chadlwilson:
## Description of Change

As noted in #8204 currently the code logs a misleading exception when the data feed metadata downloads fail; because it swallows the exception and logs the original `cache.properties` failure. That might be OK when using your own proxy or a cache produced by vulnz, but is misleading when using the NVD 2.0 data feeds, especially swallowing the original error (`ex` vs `ex1`).

https://github.com/dependency-check/DependencyCheck/blob/48c4595940e200bb4dffd0de361ec464e16a96cb/core/src/main/java/org/owasp/dependencycheck/data/update/NvdApiDataSource.java#L637-L638

When pointing to the NVD 2.0 data feeds, the `cache.properties` doesn't exist (I think this is a vulnz invention?), so the metadata files are always needed to be checked remotely.

Furthermore, this also changes the feed file for a given year to be considered "mandatory" ONLY once it is 2nd Jan (or later) everywhere on earth. This way the code tries its best to find metadata on January 1st, without failing - so avoids us coupling logic to when NVD happen to publish their feed (currently 03:00 Jan 1st UTC-05:00). Seeking feedback on this - this change is in a separate subsequent commit.

Minor related changes
- fixes a date-dependent test that was written to fail after 2025-12-31
- refactors the data feed parsing and URL building hijinxing off into the existing dedicated subclass
- tweaks error messages to reflect possible usage of NVD 2.0 data feeds as in #7514 (not just a "NVD API cache") 

## Related issues

- fixes #8204

yes

### DIFF
--- a/core/src/main/java/org/owasp/dependencycheck/data/update/NvdApiDataSource.java
+++ b/core/src/main/java/org/owasp/dependencycheck/data/update/NvdApiDataSource.java
@@ -33,14 +33,18 @@ import java.net.URISyntaxException;
 import java.net.URL;
 import java.text.MessageFormat;
 import java.time.Duration;
+import java.time.LocalDate;
 import java.time.ZoneId;
+import java.time.ZoneOffset;
 import java.time.ZonedDateTime;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.HashMap;
 import java.util.HashSet;
+import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 import java.util.Optional;
 import java.util.Properties;
 import java.util.Set;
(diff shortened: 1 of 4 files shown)


jeremylong (approved):
LGTM

PR #8212, opened 2026-01-03
build: break dependency so tests can be more reliable

nhumblot:
## Description of Change

Following the new year, 2 tests of the `SuppressionParserTest` class are failing, making all new pipeline jobs failing (see: https://github.com/dependency-check/DependencyCheck/actions/runs/20669128797/job/59346855943?pr=8192). 

Instead of extending the date in the suppression xml documents, this change suggests to introduce a seam, without changing the public API, in order to make the test independent of the system clock, preventing tests to eventually fail.

This change is just a refactoring, not changing the current application behavior.

## Related issues

- relates to #7957 

*no*, but existing test cases have been updated

### DIFF
--- a/core/src/main/java/org/owasp/dependencycheck/xml/suppression/SuppressionHandler.java
+++ b/core/src/main/java/org/owasp/dependencycheck/xml/suppression/SuppressionHandler.java
@@ -125,7 +125,26 @@ public class SuppressionHandler extends DefaultHandler {
 
     private Boolean groupBase = null;
     private Calendar groupUntil = null;
+    private final Calendar now;
 
+    /**
+     * Constructs a new suppression handler.
+     *
+     * @deprecated use {@link SuppressionHandler#SuppressionHandler(Calendar)}
+     */
+    @Deprecated(forRemoval = true)
+    public SuppressionHandler() {
+        this(Calendar.getInstance());
+    }
+
+    /**
+     * Constructs a new suppression handler.
+     *
+     * @param now represents current time. Used to compare suppression expiration dates set within the 'until' attribute.
+     */
+    SuppressionHandler(Calendar now) {
+        this.now = now;
+    }
 
     /**
      * Get the value of suppressionRules.
@@ -199,7 +218,7 @@ public class SuppressionHandler extends DefaultHandler {
         if (null != qName) {
             switch (qName) {
                 case SUPPRESS:
-                    if (rule.getUntil() != null && rule.getUntil().before(Calendar.getInstance())) {
+                    if (rule.getUntil() != null && rule.getUntil().before(now)) {
                         LOGGER.info("Suppression is expired for rule: {}", rule);
                     } else {
                         suppressionRules.add(rule);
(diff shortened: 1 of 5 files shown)


chadlwilson:
I fixed them in #8205 FWIW

chadlwilson:
Doesn't really need a seam IMHO given the actual expiry date is not relevant to the test.

core/src/main/java/org/owasp/dependencycheck/xml/suppression/SuppressionHandler.java
```diff
@@ -125,7 +125,26 @@ public class SuppressionHandler extends DefaultHandler {
...
+    public SuppressionHandler() {
+        this(Calendar.getInstance());
+    }
+
+    /**
+     * Constructs a new suppression handler.
+     *
+     * @param now represents current time. Used to compare suppression expiration dates set within the 'until' attribute.
+     */
+    SuppressionHandler(Calendar now) {
```
chadlwilson, line 145: Can we use something more modern like an Instant or ZonedDateTime (if TZ dependent) in API or does that require too many other changes/introduce inconsistency with existing public APIs already taking Calendars (🤮)?

chadlwilson (approved):

core/src/main/java/org/owasp/dependencycheck/xml/suppression/SuppressionHandler.java
nhumblot: `java.util.Calendar` seems to be exposed in our API in a limited amount of places, in the core, through `org.owasp.dependencycheck.xml.suppression.SuppressionRule` and `org.owasp.dependencycheck.utils.DateUtil`.

Removing these public elements will technically require a breaking change of the core at some point but we could start by providing alternative APIs and deprecating the old ones for a removal in the next major.

nhumblot:
Closing following the merge of #8205

chadlwilson:
> Closing following the merge of #8205

if you are keen to rebase, there is still some value to your change IMHO in terms of improving ability to test pieces - but your call.

core/src/main/java/org/owasp/dependencycheck/xml/suppression/SuppressionHandler.java
chadlwilson: Yeah, I am not sure how "seriously" we treat Java level visibility on these `core` APIs, especially given java's limited ability to really enforce visibility (short of the module system).

Do you know how many projects there are which depend on the core which are not official or "close" plugins (e.g jenkins, azure etc). I wonder if it is sufficient that we check we're not breaking the major ones when making changes to the core.

chadlwilson: There's some adjacent discussion at https://github.com/dependency-check/dependency-check-gradle/pull/491#issuecomment-3706796549 where I express my preference that things like the `Settings` API are not directly exposed to end users via things like the Gradle plugin - but where your opinion would probably be valuable.

PR #8248, opened 2026-01-26
fix: Suppress noisy WARN logging from Apache Lucene within Maven and Ant plugins

### LINKED ISSUE #8233: [WARNING] You are running with Java 23 or later. To make full use of the Vector API, please update Apache Lucene.
**Precondition**

**Describe the bug**
When running on JDK 23+, there's a warning that asks for Lucene to be updated:

> [WARNING] You are running with Java 23 or later. To make full use of the Vector API, please update Apache Lucene.

The problem occurs using version 12.2.0 of the maven plugin.

**Expected behavior**
No warnings during standard operation.

**Additional context**
The latest LTS, JDK 25, was released in September.

With older JDKs you get a different warning (see also #7759), which is avoided by doing what it says:

> [WARNING] Java vector incubator module is not readable. For optimal vector performance, pass '--add-modules jdk.incubator.vector' to enable Vector API.

chadlwilson:
## Description of Change

These should not be of concern to users, and already suppressed within the CLI via #8003 

## Related issues

- fixes #8233 

yes

### DIFF
--- a/cli/src/main/resources/logback.xml
+++ b/cli/src/main/resources/logback.xml
@@ -10,8 +10,9 @@
         </encoder>
     </appender>
 
+    <!-- Mirrored by configuration in muteNoisyLoggers() in plugins -->
     <logger name="org.apache.lucene" level="ERROR" />
-    <logger name="org.apache.commons.jcs3" level="FATAL" />
+    <logger name="org.apache.commons.jcs3" level="ERROR" />
     <logger name="org.apache.hc" level="ERROR" />
 
     <root level="INFO">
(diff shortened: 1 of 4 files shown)


cli/src/main/resources/logback.xml
```diff
@@ -10,8 +10,9 @@
         </encoder>
     </appender>
 
+    <!-- Mirrored by configuration in muteNoisyLoggers() in plugins -->
     <logger name="org.apache.lucene" level="ERROR" />
-    <logger name="org.apache.commons.jcs3" level="FATAL" />
+    <logger name="org.apache.commons.jcs3" level="ERROR" />
```
chadlwilson, line 15: This was a typo, but didn't have any effect because of the forced programmatic muting we have in the special jcs-to-slf4j adapter. `FATAL` isn't valid in logback/slf4j - so it was defaulting to `INFO`; but then being entirely muted.

jeremylong (approved):

cli/src/main/resources/logback.xml
jeremylong: However, this error might have led me to add the forced programmatic muting...

chadlwilson: The error was by me according to the history, but I am not sure what I was thinking 😅

PR #8254, opened 2026-01-27
docs: define schema locations in XML examples

### LINKED ISSUE #1940: Fix examples on https://jeremylong.github.io/DependencyCheck/general/suppression.html
**Describe the bug**
The example XMLs in the aforementioned page don't reference the suppression XSD correctly.

**Expected behavior**
The XSD should be used by XML editors to validate the given schema. For that the XSD needs to be referenced correctly. An example of a correct reference can be found in https://www.ibm.com/developerworks/library/x-tipsch/index.html#gettingdown.

**Additional context**
Basically each schema is referenced via a namespace. The example XMLs https://jeremylong.github.io/DependencyCheck/general/suppression.html currently only define a namespace but not the underlying schema.

### LINKED ISSUE #7627: Suppression files fail when using https://dependency-check.github.io/DependencyCheck/ namespace
**Precondition**

**Describe the bug**
Suppression files using the `https://dependency-check.github.io/DependencyCheck/dependency-suppression.1.3.xsd` namespace fails with, 
```
[WARN] Unable to parse suppression xml file '/var/folders/qh/9d43mjpj5wv47gwspj9yzdy40000gp/T/dctempd7e93310-c102-4739-bd26-3739cf967d0a/suppressiond427207b-c3f9-4e76-88f4-354d5cad3470.xml'
[WARN] org.xml.sax.SAXException: Line=2, Column=105: cvc-elt.1.a: Cannot find the declaration of element 'suppressions'.
```

CLI 12.1.1

**To Reproduce**

1. Create a suppression file following https://dependency-check.github.io/DependencyCheck/general/suppression.html but changing `https://jeremylong.github.io/DependencyCheck/dependency-suppression.1.3.xsd` to `https://dependency-check.github.io/DependencyCheck/dependency-suppression.1.3.xsd` 
E.G. 
```
<suppressions xmlns="https://dependency-check.github.io/DependencyCheck/dependency-suppression.1.3.xsd">
   <suppress>
      <notes><![CDATA[
      file name: some.jar
      ]]></notes>
      <sha1>66734244CE86857018B023A8C56AE0635C56B6A1</sha1>
      <cpe>cpe:/a:apache:struts:2.0.0</cpe>
   </suppress>
</suppressions>
```
2. Execute Dependency Check using newly created suppression file to scan any artifact
3. View following error in console log
```
[WARN] Unable to parse suppression xml file '/var/folders/qh/9d43mjpj5wv47gwspj9yzdy40000gp/T/dctempd7e93310-c102-4739-bd26-3739cf967d0a/suppressiond427207b-c3f9-4e76-88f4-354d5cad3470.xml'
[WARN] org.xml.sax.SAXException: Line=2, Column=105: cvc-elt.1.a: Cannot find the declaration of element 'suppressions'.
```

**Expected behavior**
Dependency Check supports newly hosted files from `https://dependency-check.github.io/DependencyCheck/` namespaces.

**Additional context**
This appears to be related to https://github.com/dependency-check/DependencyCheck/blob/main/core/src/main/java/org/owasp/dependencycheck/xml/suppression/SuppressionParser.java#L146 

The new suppression file is being linked to from the new documentation at https://dependency-check.github.io/DependencyCheck/general/suppression.html

chadlwilson:
## Description of Change

As discussed in #7627 and others; currently the namespaces of schemas include the old jeremylong.github.io namespace which confuses people. When ODC is not on the classpath in IDEs, this can confuse editors about where to retrieve the schema for the given namespace/IRI from. If we consistently define schema locations this may go some way to resolving this.

We can also subsequently add new versions which declare a new target namespace (the namespaces are versioned anyway), but that can be separate to this.

Smaller side changes
- aligns indentation of the generated/suggested suppressions to 4 spaces, not 3 similar to the base suppressions to satisfy my OCD and editors.
- fix minor typo in HTML report (Suppress by GAV -> Package URL)
- fix wrapping of XML in suppression modal in HTML report (now wider due to the XML headers for full doc)
- removed unused modal CSS cruft from jenkinsReport I came across at the same time
- change some minor stuff in the website SampleReport, but it really needs fully regenerating probably.

## Related issues

- fixes #7627
- fixes #1940

yes - existing tests cover

### DIFF
--- a/.github/workflows/false-positive-approvals.yml
+++ b/.github/workflows/false-positive-approvals.yml
@@ -118,7 +118,16 @@ jobs:
               if (!fs.existsSync('./suppressions')){
                   fs.mkdirSync('./suppressions');
               }
-              fs.appendFileSync('suppressions/publishedSuppressions.xml', '<?xml version="1.0" encoding="UTF-8"?>\n<suppressions xmlns="https://jeremylong.github.io/DependencyCheck/dependency-suppression.1.3.xsd">' + generatedSuppressions + '\n' + suppression.trim() + '\n</suppressions>', function (err) {
+              fs.appendFileSync('suppressions/publishedSuppressions.xml',
+              `<?xml version="1.0" encoding="UTF-8"?>
+              <suppressions xmlns="https://jeremylong.github.io/DependencyCheck/dependency-suppression.1.3.xsd
+                            xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+                            xsi:schemaLocation="https://jeremylong.github.io/DependencyCheck/dependency-suppression.1.3.xsd
+                                                https://dependency-check.github.io/DependencyCheck/dependency-suppression.1.3.xsd">
+              ${generatedSuppressions}
+              ${suppression.trim()}
+              </suppressions>`,
+              function (err) {
                 if (err) throw err;
                 console.log('publishedSuppressions.xml created');
               });
--- a/.github/workflows/false-positive-ops.yml
+++ b/.github/workflows/false-positive-ops.yml
@@ -182,14 +182,14 @@ jobs:
               issue_number: context.issue.number,
               owner: context.repo.owner,
               repo: context.repo.repo,
-              body: 'Maven Coordinates\n\n```xml\n<dependency>\n   <groupId>' + process.env.GROUPID + '</groupId>\n   <artifactId>' + process.env.ARTIFACTID + '</artifactId>\n   <version>' + process.env.VERSION + '</version>\n</dependency>\n```\n\n' +
+              body: 'Maven Coordinates\n\n```xml\n<dependency>\n  <groupId>' + process.env.GROUPID + '</groupId>\n  <artifactId>' + process.env.ARTIFACTID + '</artifactId>\n  <version>' + process.env.VERSION + '</version>\n</dependency>\n```\n\n' +
                     'Suppression rule:\n```xml\n' +
                     '<suppress base="true">\n' +
-                    '   <notes><![CDATA[\n' +
-                    '   FP per issue #' + context.issue.number + '\n' +
-                    '   ]]></notes>\n' +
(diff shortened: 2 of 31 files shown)


chadlwilson:
Perhaps could do with a double-check/review that none of the internal XML parsing has ability to open schemas from external URLs and its otherwise configuring factories/builders correctly. I imagine not an issue, given that'd probably be a security issue in its own right.

jeremylong (approved):
LGTM

chadlwilson:
> Perhaps could do with a double-check/review that none of the internal XML parsing has ability to open schemas from external URLs and its otherwise configuring factories/builders correctly. I imagine not an issue, given that'd probably be a security issue in its own right.

FYI @jeremylong @nhumblot Sadly this suspicion was right and there's at least one bug on master created by this (in some setups). This is because the sax parsers do actually process the schema location hints I've added, and currently forcibly allow both file and https external schema access (if the system property is not set). 

Suppressions validation additionally has a custom entity resolver that essentially remaps https references to local resource loaded schemas, however hints parsing doesn't have this resolver.

So in a JVM environment that has disabled external schema access system wide, base hints parsing fails (e.g seemingly in Gradle environments with the Gradle plugin this property can be set to disallow https loading). 

The tests currently seem to obscure some of these real world problems as it seems an old xerces version is on the classpath which is being used in preference to the JDK bundled xerces, which adds additional complexity to tidy up, as modern bundled xerces has more secure defaults.

I'm not 100% sure why it is working for suppressions so I might have missed something in the above analysis but I'll submit a PR to resolve this, avoid interfering with system properties - and improve the testing.

chadlwilson:
The quick fix is to remove the schemalocation hints from the base/bundled suppressions/hints.

We may also have to do so for the generated/hosted suppressions (which is already published with a schemaLocation in it) if we find it is causing people's existing ODC installs to go out to the internet (when it did not before), but I'll verify this too. 

Keep an eye out if issues with hosted suppressions are reported by people in environments without internet access.

Issue #7757, opened 2025-06-19
HostedSuppressionsFile not loaded from custom URL (behind corporate proxy) - file reported as empty


klerdos:
**Precondition**
- [x] I checked the issues list for existing open or closed reports of the same problem.

**Describe the bug**
We are using OWASP Dependency-Check Maven plugin **v12.1.0** in a corporate environment **behind a proxy without internet access**.

- Audit scanners are disabled due to lack of internet connectivity.
- The vulnerability database is mirrored internally and updated nightly.
- Several **custom suppression files** are used successfully.
- We are now attempting to utilize the **hostedSuppressionsFile** (https://jeremylong.github.io/DependencyCheck/suppressions/publishedSuppressions.xml) by mirroring it to **our internal Artifactory**.

Despite updating the configuration to point to our **internal URL**, Dependency-Check **fails to load the hosted suppression file**. The logs indicate the file is empty, although it exists and is accessible.

Strangely, when using **the exact same URL** as a **regular suppression file**, it works as expected, and suppression rules are loaded properly.

**Version of dependency-check used**
The problem occurs using version **12.1.0** of the **Maven plugin**.

**Log file**
Full debug log provided as a Gist: https://gist.github.com/klerdos/cd8a4bbc39b8b52a3de4f9210affe94d

Relevant excerpts:
```
[DEBUG] Setting: hosted.suppressions.url='https://artifactory.mydomain.com/artifactory/remote-dependencycheck-cache/suppressions/publishedSuppressions.xml'
...
[WARNING] Hosted Suppressions file is empty or missing - attempting to force the update
[DEBUG] Setting: hosted.suppressions.forceupdate='true'
[WARNING] Empty Hosted Suppression file after update, results may contain false positives already resolved by the DependencyCheck project due to failed download of the hosted suppression file
```

The same URL **works correctly** when used in `<suppressionFiles>`:
```
[DEBUG] Loading suppression rules from 'https://artifactory.mydomain.com/artifactory/remote-dependencycheck-cache/suppressions/publishedSuppressions.xml'
```

**To Reproduce**
1. Configure the Maven plugin with:
   ```xml
   <hostedSuppressionsUrl>https://artifactory.mydomain.com/artifactory/remote-dependencycheck-cache/suppressions/publishedSuppressions.xml</hostedSuppressionsUrl>
2. Run mvn org.owasp:dependency-check-maven:aggregate
[...]

PR #8266, opened 2026-01-31
fix: delete empty hosted suppression file and throw exception (#7757)

Umesh042005:
Description:

"I have implemented a targeted fix in HostedSuppressionsDataSource. Changes:

Verified that SaveToFileResponseHandler could silently create 0-byte files if the server/proxy returns a 200 OK with an empty body.

Added a check in HostedSuppressionsDataSource to detect 0-byte files after download.

If a 0-byte file is detected, it is now deleted and an UpdateException is thrown.

This prevents the empty file from being cached and ensures the retry logic is triggered on the next run. This addresses the issue reported in #7757 where users had to manually force an update."


chadlwilson (changes requested):
Thanks for trying to address the issue in #7757.

There’s unrelated changes in this PR (changes to hints, additional test dependency, an additional report file committed which should not be here), and no tests to cover the new behaviour added.

Please create a new branch off master, clean up the commit history and submit a new PR that includes tests and only the relevant changes you are proposing.

PR #8267, opened 2026-01-31
test: avoid polluting test classpaths with sample dependencies to be scanned

chadlwilson:
## Description of Change

As mentioned in https://github.com/dependency-check/DependencyCheck/pull/8254#issuecomment-3812760826 currently the XML testing is a bit unreliable and difficult to debug because xerces and xml-apis dependencies are being pulled down and used during testing due to their inclusion within artifacts that are intended just for use as "sample" artifacts to scan using ODC code.

Other than the risk of classpath clashes, when code uses JDK `ServiceLoader` mechanisms (such as the XML APIs), I think this can cause production code to be tested with an environment unrepresentative of a real deployment.

We can address this by creating a maven execution that copies dependencies as artifacts to a known test resource location rather than adding to the actual test/integration test classpaths. 

Minor tweaks
- cleaned up the assertions for known errors during engine integration test (which scans these dependencies) to a minimal set which are easier to debug
- removed an ineffective Junit assumption when a test resource isn't available (already NPEing early in the code). These should probably fail the tests anyway.

## Related issues

- relates to a fix I need to submit for #8254 and why issues are not easily demonstrated by existing unit tests

yes


core/pom.xml
```diff
@@ -154,30 +154,6 @@ Copyright (c) 2012 Jeremy Long. All Rights Reserved.
                     </execution>
                 </executions>
             </plugin>
-            <plugin>
-                <groupId>org.apache.maven.plugins</groupId>
-                <artifactId>maven-javadoc-plugin</artifactId>
-                <configuration>
-                    <excludes>${project.build.directory}/generated-sources/java/**/*.java</excludes>
-                </configuration>
-            </plugin>
```
chadlwilson, line 163: This is unnecessary - as it is defined in the parent (and technically the syntax isn't correct)

core/pom.xml
```diff
@@ -154,30 +154,6 @@ Copyright (c) 2012 Jeremy Long. All Rights Reserved.
...
-                            <goal>copy-dependencies</goal>
-                        </goals>
-                        <configuration>
-                            <outputDirectory>${project.build.directory}/test-classes</outputDirectory>
-                            <includeScope>test</includeScope>
-                            <excludeArtifactIds>dependency-check-utils</excludeArtifactIds>
-                        </configuration>
-                    </execution>
-                </executions>
-            </plugin>
```
chadlwilson, line 180: I'm not 100% sure why this was needed as it was copying all the "normal" test dependencies into `test-classes` during a non-test maven phase, possibly causing them to be scanned during `EngineIT`. Could possibly add this back, but it somewhat strangely couples EngineIT to the current testing libraries used.

jeremylong (approved):
LGTM

core/pom.xml
chadlwilson: Do you recall what this might have been intended for, which I might be missing, @jeremylong ?

jeremylong: Pretty sure this was intended just to add more to scan in the very early days. This is not needed.

Issue #2073, opened 2019-07-15
Error running depndency-check with Maven/Jenkins


obotor:
Hi!
I am running into an issue with maven-dependency-check plugin.
The issue sums up to allowing XML/XSD parsing through Maven via Jenkins.
How should I set the accessExternalSchema property?
Thx

### Logs
```
...
16:49:31 [INFO] Created CPE Index (4 seconds)
16:49:31 [WARNING] Unable to parse suppression xml file 'dependency-check-suppression.xml'
16:49:31 [WARNING] org.owasp.dependencycheck.xml.suppression.SuppressionParseException: org.xml.sax.SAXException: Line=2, Column=281: schema_reference: Failed to read schema document 'dependency-suppression.1.2.xsd', because 'file' access is not allowed due to restriction set by the accessExternalSchema property. 
16:49:31 [ERROR] Exception occurred initializing CPE Analyzer.
16:49:31 [INFO] Finished CPE Analyzer (4 seconds)
16:49:31 [INFO] Finished False Positive Analyzer (0 seconds)
16:49:31 [INFO] Finished NVD CVE Analyzer (0 seconds)
16:49:31 [INFO] Finished Sonatype OSS Index Analyzer (0 seconds)
16:49:32 [WARNING] Unable to parse suppression xml file 'dependency-check-suppression.xml'
16:49:32 [WARNING] org.owasp.dependencycheck.xml.suppression.SuppressionParseException: org.xml.sax.SAXException: Line=2, Column=281: schema_reference: Failed to read schema document 'dependency-suppression.1.2.xsd', because 'file' access is not allowed due to restriction set by the accessExternalSchema property. 
16:49:32 [ERROR] Exception occurred initializing Vulnerability Suppression Analyzer.
16:49:32 [INFO] Finished Vulnerability Suppression Analyzer (0 seconds)
16:49:32 [INFO] Finished Dependency Bundling Analyzer (0 seconds)
16:49:32 [INFO] Analysis Complete (5 seconds)
...
```

### POM.xml
The plugin version is 5.1.0. A test with previous version 3.0.2 also failed in a similar manner.

```xml
<build>
  <pluginManagement>
    <plugins>
      ...
      <plugin>
        <groupId>org.owasp</groupId>
        <artifactId>dependency-check-maven</artifactId>
        <version>${dependency.check.plugin.version}</version>
      </plugin>
     ....
    </plugins>
  </pluginManagement>
</build>
....
<profile>
[...]

PR #8272, opened 2026-02-02
fix(core): correct xml schema validation handling without needing external access

chadlwilson:
## Description of Change

The changes in #8254 unfortunately exposed a few issues which this PR corrects which I alluded to at https://github.com/dependency-check/DependencyCheck/pull/8254#issuecomment-3812760826

- The `schemaLocation` hints I added actually **currently** cause the default SAX parsers to try and load from the external source, via `https://` :-(
- The previous code was trying to force-enable `javax.xml.accessExternalSchema` to `file, https` which allowed this
  - However the **first schema** to load, if done via a `schemaLocation` attribute  would do so without external permissions; because the `javax.xml.accessExternalSchema` setting was being set **after** the SaxParserFactory was created. In practice, this means on master the hints schema would fail loading with the new schema location, since it is the first Analyzer prepared. (`Caused by: org.xml.sax.SAXException: Line=5, Column=104: schema_reference: Failed to read schema document 'dependency-hint.1.4.xsd', because 'https' access is not allowed due to restriction set by the accessExternalSchema property.`)
- There was an attempt to avoid going out to the internet with a custom entity resolver, however we did not update this for the project/domain name change, so it wasn't working for `https://dependency-check.github.io` URLs - and also was only applied to suppression xsds.

Testing challenges
- [x] There was no testing/validation that the base hints can be parsed/loaded (resolved)
- [x] The tests to load base suppressions did so using a different API, without validation
- [x] There was a custom (old) Xerces version **from 2006** on the classpath which was being used in preference to the bundled loader during tests; which did not impose the same security defaults. Fixed in #8267 

This change
- retains the ability to validate XML docs by namespace **without** schema locations
- removes the original xml `javax.xml.accessExternalSchema` system property mutation, by replacing with a mechanism to load https and file references from local resources by "convention".
  - the system property setting was added originally in #2073 to handle folks setting `schemaLocation` (alongside an EntityResolver made available for suppression documents)
- the new `EntityResolver` will inspect any attempts to resolve an entity for our own domain prefixes, and match against the the same file name as a passed resource.
  - The only important coupling when publishing to GitHub pages is the filename should stay the same.
- stops interfering with System properties for XML going forward. This wasn't really safe in a multi-threaded environment anyway, with something like a Gradle Daemon and other things running.
- avoids any remote access while validating XML, except if the user has specifically enabled it for another reason to avoid us accidentally recreating such a situation
- reviewed and consolidated the SAXParser creation to ensure they have the same security settings whether validating or not

## Related issues

- relates to #8254

yes


jeremylong (approved):
lgtm

chadlwilson:
Oh wow that was quick @jeremylong thanks! i was expecting more questions given how I managed to screw this up in the first place 😅

This exercise was a good reminder about how much I have always hated XML in Java, and that only WSDL was worse 😅

chadlwilson:
The APIs really are terrible by modern standards, and Xerces/JAXP seems dead/dormant :(

PR #8273, opened 2026-02-02
chore(fp): de-duplicate/align `_project` CPE generated suppressions in base suppressions for over ~9 months

chadlwilson:
## Description of Change

As discussed at https://github.com/dependency-check/DependencyCheck/pull/8138#issuecomment-3553605096 there are a number of generated/published/hosted suppression rules which have more generic equivalents in base suppressions. 

We can tidy up a bit by removing these duplicates; and doing a once-off normalization of the remaining `_project` ones which dont have an equivalent in base suppressions.

Most of them also had been copy+pasted to base suppressions for a long time before being edited, so this shouldn't cause issues. The logic is that most folks will have had to upgrade some time in last 9 months for various reasons anyway so will already have the more generic base suppressions. If folks report FPs again, can always add them back to generated suppressions.

Methodology was to load all rules as objects; and search for identical rules after undoing the regex conversions from https://github.com/dependency-check/DependencyCheck/pull/7684

## Related issues

- relates to https://github.com/dependency-check/DependencyCheck/pull/8138
- relates to https://github.com/dependency-check/DependencyCheck/pull/7684

N/A

### DIFF
--- a/generatedSuppressions.xml
+++ b/generatedSuppressions.xml
@@ -1,12 +1,5 @@
 <suppress base="true">
     <notes><![CDATA[
-    FP per issue #4803
-    ]]></notes>
-    <packageUrl regex="true">^pkg:maven/net\.sourceforge\.htmlunit/htmlunit-cssparser@.*$</packageUrl>
-    <cpe>cpe:/a:htmlunit_project:htmlunit</cpe>
-</suppress>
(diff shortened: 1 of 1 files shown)


jeremylong (approved):
lgtm

PR #8277, opened 2026-02-03
build(deps): base docker image off latest Alpine with updated DB driver versions

chadlwilson:
## Description of Change

We use mutable tags for the golang and JVM layers which move Alpine versions over time, so we probably should be consistent with the main base image, which is a [dotnet runtime image](https://mcr.microsoft.com/en-us/artifact/mar/dotnet/runtime/tags). Right now we are using a golang and JVM built off Alpine 3.2e while running on Alpine 3.22.

While that should not generally be an issue; it might be, as I don't think Java is statically compiled; still dynamically linked to musl libc.verisins which can different across Alpine releases.

The alternative change here is to use specific Alpine versions and update them together - there are possibly ways to trick dependabot into being able to do this if we'd rather go use an Alpine-feature-mutating tag.

Also bumps the bundled driver versions to latest.

## Related issues

N/A

yes


jeremylong (approved):
LGTM

PR #8292, opened 2026-02-10
build: use maven jdk toolchains to build with Java 25; test against Java 11/17/21/25

chadlwilson:
## Description of Change

As discussed in https://github.com/dependency-check/DependencyCheck/pull/8275#issuecomment-3841308401 this change switches Dependency Check to

- run Maven and compile using the same "default" JVM (upgraded to Java 11 --> 25)
- `-release` target is still Java 11 (as before), ensuring compile time and `bootclasspath` compatibility.
- run all unit, integration and maven release tests against all of Java 11/17/21/25 to ensure runtime compatibility

This should allow us to
- update build-time plugins to later versions while maintaining confidence on a target Java version
- better support/replicate issues on newer Java versions rather than only running with the minimum

### Additional detail

In order to do this and ensure it is as safe as possible this change as a side effect:

_Dev Experience_
- everything will work as before when running from CLI or IDE - for convenience, tests and compile will run with the same JVM as Maven
- Maven is not enforcing running/compiling with a fixed JDK right now; only GitHub Actions controls that
- running tests with specific JDK versions via Maven toolchains requires opt-in with `-Dtoolchain.jdk.test.version` (and `-Dtoolchain.jdk.test.home` for maven integration tests, unfortunately) which activates a Maven profile
- fixes warnings from both Mockito and ODC's own code when running tests on newer JVM versions

_GitHub Actions_
- runs tests in a matrix in parallel
  - for simplicity, each run compiles the code itself rather than consuming artifacts. This could be changed.
  - only the "default" JDK version archives its compiled results/deploys snapshot
- switches to using `setup-java` to manage primary maven cache rather than doing manually
- further aligns build jobs between PR/`main`/`release`

_Dependencies_
- consolidates nullness annotations on `org.jspecify:jspecify` similar to other projects (rather than jetbrains annotations)
- upgrades ancient `org.json` library to supported variant rather than stale google android fork
- cleans up exclusions and dependency clashes for transitives that we don't need
  - removes unused `mockserver` dependency
  - correctly removes transitive `bouncycastle` dependency from `packager-rpm` as originally intended
  - removes compile-only dependencies from guava
  - `maven` plugin project
    - fixes mockito version clashes on `maven` project from the resolver BOM
    - aligns junit version being used
    - upgrades test groovy dependency to current
    - upgrades one test dependency to ensure it works on modern JVMs

## Related issues

N/A

yes

- validated that all surefire/failsafe tests are running with the chosen JVM
- validated that the maven plugin tests are running with the configured JVM
- validated builds aren't taking significantly longer to run
- validated the GitHub workflow runs ok on `main` ([see this run](https://github.com/chadlwilson/DependencyCheck/actions/runs/21858751201))
- diffed the `ant` and `cli` build archives to ensure any differences in dependencies are expected
- sanity checked maven `site` generation
- sanity checked with the gradle plugin and its tests

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -74,7 +74,7 @@ jobs:
           -Dtoolchain.jdk.test.version=${{ matrix.jdk_test_version }} -Dtoolchain.jdk.test.home="$JAVA_HOME_${{ matrix.jdk_test_version }}_X64"
           ${{ matrix.jdk_test_version == matrix.jdk_default_version && 'source:jar javadoc:jar site' || '' }}
           ${{ steps.install-gpg-key.outcome == 'success' && '-Prelease gpg:sign deploy' || '' }} 
-          --no-transfer-progress --batch-mode
+          --no-transfer-progress --batch-mode -Dstyle.color=always
       - name: SARIF Multitool
         uses: microsoft/sarif-actions@v0.2
         with:
@@ -144,6 +144,6 @@ jobs:
         run: >
           mvn -V -s settings.xml -pl cli -am 
           package -DskipTests=true
-          --no-transfer-progress --batch-mode
+          --no-transfer-progress --batch-mode -Dstyle.color=always
       - name: Test Docker Image
         run: ./docker-test.sh
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -70,10 +70,10 @@ jobs:
           NVD_API_KEY: ${{ secrets.NVD_API_KEY }}
         run: >
           mvn -V -s settings.xml 
-          clean package verify source:jar javadoc:jar 
-          ${{ steps.install-gpg-key.outcome == 'success' && '-Prelease gpg:sign deploy' || '' }} 
-          -PFullIntegrationTesting 
+          clean verify -PFullIntegrationTesting
           -Dtoolchain.jdk.test.version=${{ matrix.jdk_test_version }} -Dtoolchain.jdk.test.home="$JAVA_HOME_${{ matrix.jdk_test_version }}_X64"
+          ${{ matrix.jdk_test_version == matrix.jdk_default_version && 'source:jar javadoc:jar site' || '' }}
+          ${{ steps.install-gpg-key.outcome == 'success' && '-Prelease gpg:sign deploy' || '' }} 
           --no-transfer-progress --batch-mode
       - name: SARIF Multitool
         uses: microsoft/sarif-actions@v0.2
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -9,30 +9,28 @@ on:
       - '**/*.txt'
 
 permissions: {}
+
 jobs:
   build:
+    strategy:
+      matrix:
+        jdk_default_version: [ '25' ]                # Single JDK version to run Maven with and use for compilation etc
+        jdk_test_version: [ '11', '17', '21', '25' ] # JDK version to run surefire/failsafe tests using
+      fail-fast: false
+
+    name: Build and Test (JDK ${{ matrix.jdk_test_version }}${{ matrix.jdk_test_version == matrix.jdk_default_version && ' - Default' || '' }})
     permissions:
       contents: read # to fetch code (actions/checkout)
-
-    name: Build dependency-check
-    runs-on: ubuntu-latest 
+    runs-on: ubuntu-latest
     steps:
       - name: Install gpg secret key
-        if: github.repository_owner == 'dependency-check'
+        if: matrix.jdk_test_version == matrix.jdk_default_version && github.repository_owner == 'dependency-check'
         id: install-gpg-key
         run: |
           cat <(echo -e "${{ secrets.GPG_PRIVATE_KEY }}") | gpg --batch --import
           gpg --list-secret-keys --keyid-format LONG
       - uses: actions/checkout@v6
-      - name: Check Maven Cache
-        id: maven-cache
-        uses: actions/cache@v5
-        with:
-          path: ~/.m2/repository
-          key: ${{ runner.os }}-maven-${{ hashFiles('**/pom.xml') }}
-          restore-keys: |
-            ${{ runner.os }}-maven-
-      - name: Check Local Maven Cache
+      - name: Maven Integration Test Cache
         id: maven-it-cache
         uses: actions/cache@v5
         with:
@@ -47,26 +45,36 @@ jobs:
       - uses: actions/setup-dotnet@v5.1.0
         with:
           dotnet-version: '8.0.x'
-      - name: Set up JDK 11
-        id: jdk-11
+      - name: Set up JDKs
         uses: actions/setup-java@v5
         with:
-          java-version: 11
+          java-version: | # last version takes precedence as default
+            ${{ matrix.jdk_test_version }}
+            ${{ matrix.jdk_default_version }}
           distribution: 'zulu'
+          check-latest: true
+          cache: 'maven'
+          cache-dependency-path: '**/pom.xml'
           server-id: central
           server-username: ${{ secrets.CENTRAL_USER }}
           server-password: ${{ secrets.CENTRAL_PASSWORD }}
       - uses: pnpm/action-setup@41ff72655975bd51cab0327fa583b6e92b6d3061 # v4.2.0
         with:
           version: 6.0.2
-      - name: Build Snapshot with Maven
+      - name: Build/Test Snapshot with Maven${{ steps.install-gpg-key.outcome == 'success' && ' (then Deploy)' || '' }}
         id: build-snapshot
         env:
           MAVEN_USERNAME: ${{ secrets.CENTRAL_USER }}
           MAVEN_PASSWORD: ${{ secrets.CENTRAL_PASSWORD }}
           MAVEN_GPG_PASSPHRASE: ${{ secrets.GPG_PRIVATE_KEY_PASSWORD }}
           NVD_API_KEY: ${{ secrets.NVD_API_KEY }}
-        run: mvn -V -s settings.xml clean package verify source:jar javadoc:jar ${{ steps.install-gpg-key.outcome == 'success' && '-Prelease gpg:sign deploy' || '' }} -DreleaseTesting --no-transfer-progress --batch-mode
+        run: >
+          mvn -V -s settings.xml 
+          clean package verify source:jar javadoc:jar 
+          ${{ steps.install-gpg-key.outcome == 'success' && '-Prelease gpg:sign deploy' || '' }} 
+          -PFullIntegrationTesting 
+          -Dtoolchain.jdk.test.version=${{ matrix.jdk_test_version }} -Dtoolchain.jdk.test.home="$JAVA_HOME_${{ matrix.jdk_test_version }}_X64"
+          --no-transfer-progress --batch-mode
       - name: SARIF Multitool
         uses: microsoft/sarif-actions@v0.2
         with:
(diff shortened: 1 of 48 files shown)


pom.xml
```diff
@@ -559,14 +571,14 @@ Copyright (c) 2012 - Jeremy Long
...
                             <goal>enforce</goal>
                         </goals>
                         <configuration>
                             <rules>
                                 <requireMavenVersion>
-                                    <version>[3.1,]</version>
+                                    <version>3.9.0</version>
                                 </requireMavenVersion>
                             </rules>
                             <fail>true</fail>
```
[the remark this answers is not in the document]
chadlwilson: Copilot is confused here, the value is a minimum already.

.github/workflows/purge-cache.yml
```diff
@@ -31,6 +23,14 @@ jobs:
...
           path: core/target/data
           key: odc-data
+      - name: Set up JDK
+        uses: actions/setup-java@v5
+        with:
+          java-version: ${{ env.MAVEN_JDK_DEFAULT_VERSION }}
+          distribution: 'zulu'
+          check-latest: true
+          cache: 'maven'
+          cache-dependency-path: '**/pom.xml'
```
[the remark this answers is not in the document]
chadlwilson: Oops, TY

core/src/main/java/org/owasp/dependencycheck/utils/DependencyVersion.java
```diff
@@ -232,7 +232,7 @@ public boolean matchesAtLeastThreeLevels(DependencyVersion version) {
     }
 
     @Override
-    public int compareTo(@NotNull DependencyVersion version) {
+    public int compareTo(@NonNull DependencyVersion version) {
         if (version == null) {
             return 1;
         }
```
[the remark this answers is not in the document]
chadlwilson: Fixed

chadlwilson:
Sorry for the noise here. Despite running it a few times on my fork, copilot found new things here for some reason, which I addressed. 😅

I am done for now, pending feedback 🙏

jeremylong:
Just used amp to do a quick PR review. The two items it called out are below.  @chadlwilson think these are worth fixing?

| # | Severity | Location | Problem | Fix |
|---|----------|----------|---------|-----|
| 1 | MEDIUM | `maven/pom.xml:216` | `${toolchain.jdk.test.home}` is used for maven-invoker-plugin's `<javaHome>` but has no default value defined in properties. Local developer builds without the CI-specific `-Dtoolchain.jdk.test.home=...` flag will resolve an empty/unresolved property, potentially breaking maven plugin integration tests. | Define a default in `<properties>`: `<toolchain.jdk.test.home>${java.home}</toolchain.jdk.test.home>` |
| 2 | MEDIUM | `pom.xml:631` | Mockito JAR path manually constructed as `${settings.localRepository}/org/mockito/mockito-core/${mockito.version}/mockito-core-${mockito.version}.jar` in surefire/failsafe argLine. Fragile — assumes default local repo layout; breaks with custom repository managers or split layouts. | Use `maven-dependency-plugin:properties` goal to resolve the artifact path, or document the limitation |

To resolve item 2 amp's suggested approach would be to add a maven-dependency-plugin execution with the properties goal, which automatically sets a Maven property for each dependency's artifact file path. For mockito-core, it would produce a property named org.mockito:mockito-core:jar containing the full resolved path.

Here's what the implementation would look like:


<plugin>
    <artifactId>maven-dependency-plugin</artifactId>
    <executions>
        <execution>
            <id>set-dependency-properties</id>
            <goals>
                <goal>properties</goal>
            </goals>
        </execution>
    </executions>
</plugin>

Then replace the manual path in both argLines:
```xml
<!-- Before -->
<argLine>@{surefireArgLine} -javaagent:${settings.localRepository}/org/mockito/mockito-core/${mockito.version}/mockito-core-${mockito.version}.jar ${runtime.extra.jvm.args} -Dfile.encoding=UTF-8</argLine>

<!-- After -->
<argLine>@{surefireArgLine} -javaagent:${org.mockito:mockito-core:jar} ${runtime.extra.jvm.args} -Dfile.encoding=UTF-8</argLine>
```

And the same for failsafe. The properties goal runs in the initialize phase by default, so it resolves before both surefire and failsafe execute.

jeremylong (dismissed):
@chadlwilson PR looks good to me.  Let me know if you want to fix the two items from amp or not.  I'm good with it as is; up to you.

chadlwilson:
> Define a default in `<properties>`: `<toolchain.jdk.test.home>${java.home}</toolchain.jdk.test.home>`

Yup, fair enough - this is more convenient for running locally without toolchains enabled - have done this.

> Mockito JAR path manually constructed as `${settings.localRepository}/org/mockito/mockito-core/${mockito.version}/mockito-core-${mockito.version}.jar` in surefire/failsafe argLine. Fragile — assumes default local repo layout; breaks with custom repository managers or split layouts. 	Use `maven-dependency-plugin:properties` goal to resolve the artifact path, or document the limitation

I was doing it this way earlier, but it seemed a bit unnecessary to add an extra goal execution and thought it might be confusing to maintainers since the properties it creates are kind of "magic". Doing it that this way is more Windows-friendly and robust though. Have changed to do it this way now.

Also realised the maven surefire tests weren't running with the separate PR workflow "Regression Test Maven Plugin" due to me applying an incorrect suggestion from an LLM earlier and the lucene filter tests haven't been running since the JUnit 5 upgrade; both now fixed.

Apologies for the force push; did so without thinking, which was a mistake. Ready to merge/re-review in any case.

jeremylong:
amp has one more possible issue:

**Missing toolchains.xml generation for multi-JDK test matrix**

Location: .github/workflows/build.yml (around lines 48-60)

Problem: Multiple JDKs are set up via actions/setup-java, but no toolchains.xml file appears to be generated. The <jdkToolchain> configuration in maven-surefire-plugin / maven-failsafe-plugin requires a valid ~/.m2/toolchains.xml to locate non-default JDK installations.

Why this matters: actions/setup-java doesn't automatically create a toolchains.xml. Without it, the toolchain-aware test runs may silently fall back to the default JDK (25), which would invalidate the multi-JDK testing strategy (i.e., you'd think you're testing on 11/17/21 but actually always running on 25).

Suggested fix: Add a step to generate ~/.m2/toolchains.xml mapping the installed JDK paths — for example using codecentric/java-toolchains-xml-action — or verify that the current setup-java configuration handles this.

chadlwilson:
> actions/setup-java doesn't automatically create a toolchains.xml

Straight false.

Amp is hallucinating that and has no idea what it is talking about. Generating toolchains.xml is explicitly what setup-java does, and the build would fail if there was no toolchains.xml - Maven cannot fall back implicitly once it is activated.

Besides, it's easily verifiable from looking at the actions run that it is working. It logs the JDK it is using.for tests explicitly.

jeremylong:
Thanks for confirming - I was a little skeptical when it stated that as a possible problem.

.github/workflows/build.yml
```diff
@@ -47,26 +45,36 @@ jobs:
...
         id: build-snapshot
         env:
           MAVEN_USERNAME: ${{ secrets.CENTRAL_USER }}
           MAVEN_PASSWORD: ${{ secrets.CENTRAL_PASSWORD }}
           MAVEN_GPG_PASSPHRASE: ${{ secrets.GPG_PRIVATE_KEY_PASSWORD }}
           NVD_API_KEY: ${{ secrets.NVD_API_KEY }}
-        run: mvn -V -s settings.xml clean package verify source:jar javadoc:jar ${{ steps.install-gpg-key.outcome == 'success' && '-Prelease gpg:sign deploy' || '' }} -DreleaseTesting --no-transfer-progress --batch-mode
+        run: >
+          mvn -V -s settings.xml 
+          clean package verify source:jar javadoc:jar
```
jeremylong: thoughts on adding `site` here?  We've had issues with bumping reporting plugins before... Or should we just run site once instead of across every JDK?

jeremylong (dismissed):
LGTM

jeremylong:
I'm good with the PR as is. Thanks for putting this together.  Let me know about the `site` goal.

.github/workflows/build.yml
chadlwilson: It did occur to me that `site` wasn't running for PRs or main build - I was going to ask you about that.

I'll add something in for the default JDK somehow....

chadlwilson: Hmm, good call on double checking this. There was a problem generating the `site` alongside everything else since I initially upgraded the `maven-jxr-plugin` which seems to [break dependency resolution when it runs in parallel](https://github.com/apache/maven-jxr/issues/256). It was hidden for me earlier, as when you've done a maven install at some point previously it will happen resolve stale test-jars from the local repository. Sigh.

So I have downgraded the plugin back to the ancient `2014`-era version as it was before this PR, which does seem to work. I couldn't really find a workaround for the modern `maven-jxr-plugin` to aggregate for the parent project after trying for a while. The only "workarounds" that retained the source code report to some extented seemed to be
- disable JXR (checkstyle HTML reports lose web link to source code) OR 
- don't produce aggregated source reports for the parent project (retain for modules, [e.g maven](https://dependency-check.github.io/DependencyCheck/dependency-check-maven/project-reports.html))

I do wonder if we need all the [various sources reports](https://dependency-check.github.io/DependencyCheck/project-reports.html) (and checkstyle HTML reports which use them) on the website these days (kinda dates from before things like GitHub) but have left them for now.

I imagine this plugin will break soon with Maven 4, so we'll have to find some other way forward at some point.

chadlwilson: I'm not sure why it makes it _quite_ so slow (i believe the older reporting plugins possibly prevent some parallelism?) but it does add 5-6 minutes to the build which isn't ideal. I could split it out to a separate job to run in parallel without tests; or maybe even a second mvn call.

image


But then we would need to make a similar decision for splitting on other workflows which run all tests, e.g the `master` build workflow `release` workflow (doesn't just run a `deploy site:stage`).

jeremylong: Maybe we just have a completely different workflow that only runs `mvn site`

chadlwilson: Splitting it is a little bit annoying as there is some overlap with what you'd want to do in a normal build since checkstyle, spotbugs and tests need to be run (since the reports are published incl jacoco and surefire), so if we want to rework it we should probably review the contents of the site now and decide which content we actually want to generate and publish, as there is a lot of stuff there right now.

It probably makes more sense to split when there is less overlap, and it is easier to reason about with the publishing workflows.

As a user I always completely ignore that section of Maven-generated websites since it really has nothing to do with end users, and often they are just stub pages (for project info).

Would you prefer to do something here in this PR or do something as a follow-up?

jeremylong: the site goal can be added and/or fixed later.

jeremylong:
@chadlwilson I'm good with the PR as is.  We can look into fixing (e.g., improving the speed of the jobs) later.  Let me know if you want to work on this more now - or if we are good to merge.

jeremylong (approved):

.github/workflows/build.yml
chadlwilson: Ok, I'll create a discussion somewhere 🙏

I'm not very happy with how slow the build is as-is so it could do with some work.

jeremylong:
FYI - from the scarf telemetry that was added, we are seeing a lot of Java 25 usage.  However, the telemetry is only in the latest version; as such, it is only reporting for those who have upgraded. I suspect a large portion of the Java 25 is from our [docker image](https://github.com/dependency-check/DependencyCheck/blob/93abe8b3a89104848eeaa24e5a091a80a49be409/Dockerfile#L5):
image

chadlwilson:
Ok to merge as-is (and at least make the other workflows are working ok)

PR #8302, opened 2026-02-12
build(deps): allow dependabot to manage container-embedded mysql/postgres driver versions

chadlwilson:
## Description of Change

Changes docker scripting to extract versions from the Maven POM for the drivers; allowing them to be updated with dependabot for security reasons, and avoiding need for special manual PRs.

Related minor tweaks
- avoided unnecessary use of `exec-maven-plugin` 
- switched scripts to prefer modern `#!/usr/bin/env bash`, largely to avoid using ancient bash on MacOS.
- speeds up docker test workflows by only building what is necessary to scan

## Related issues

N/A

yes (already there)


jeremylong (approved):
LGTM

PR #8322, opened 2026-02-23
chore: avoid use of parent pom and maven properties where unnecessary

chadlwilson:
## Description of Change

IMHO our current way of using properties 
- adds unnecessarily confusing indirection to dependencies when tracking dependency clashes down
- sometimes for plugins separates versions from the configuration that may be specific to that version you might want to see when reviewing a dependabot PR
- tends to cause merge conflicts across dependabot PRs (minor nitpick)

Parent POM `pluginManagement`/`dependencyManagement` already serves this purpose for the most part, so I suggest should we reserve use of properties for cases where we really need to de-duplicate versions (even though dependabot can help us with updating such duplicates anyway) or add clarity for coupled versions via the properties.

Additionally, it is clearer to manage the special dependencies for the maven plugin in its own `dependencyManagement` since this has no effect on the rest of ODC, doesn't need the parent POM to reference; and has its own considerations to consider in isolation.

Appreciate this is a bit irritating to review, but can rest assured that if I accidentally downgraded anything that dependabot will re-upgrade it :-)

## Related issues

N/A

yes

### DIFF
--- a/maven/pom.xml
+++ b/maven/pom.xml
@@ -28,6 +28,8 @@ Copyright (c) 2013 Jeremy Long. All Rights Reserved.
     <description>dependency-check-maven is a Maven Plugin that uses dependency-check-core to detect publicly disclosed vulnerabilities associated with the project's dependencies. The plugin will generate a report listing the dependency, any identified Common Platform Enumeration (CPE) identifiers, and the associated Common Vulnerability and Exposure (CVE) entries.</description>
     <inceptionYear>2013</inceptionYear>
     <properties>
+        <maven.api.version>3.6.3</maven.api.version>
+
         <version.maven-plugin-plugin>3.15.2</version.maven-plugin-plugin>
         <toolchain.jdk.test.home>${java.home}</toolchain.jdk.test.home>
     </properties>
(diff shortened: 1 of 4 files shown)


maven/pom.xml
```diff
@@ -38,8 +40,102 @@ Copyright (c) 2013 Jeremy Long. All Rights Reserved.
...
+    <dependencyManagement>
+        <dependencies>
+            <dependency>
+                <groupId>org.apache.maven</groupId>
+                <artifactId>maven-core</artifactId>
+                <version>${maven.api.version}</version>
+            </dependency>
+            <dependency>
+                <groupId>org.apache.maven.shared</groupId>
+                <artifactId>maven-shared-utils</artifactId>
```
nhumblot: **note**: this dependency does not seem to be used anywhere outside this dependencyManagement. On the main branch, the `verify` phase ends up successfully without this dependency declaration. We may be able to remove it?

This dependency declaration has been added in #5095, at the same moment of some exclusions.

It is probably better to remove it in a dedicated PR.

nhumblot (dismissed):
**praise**: thank you for this change!

I just left a note, I would be happy to get your opinion on it, then I can merge this PR. Thanks!

maven/pom.xml
chadlwilson: Thanks, I'll take a look at the dependency tree - just because I'm curious 😅

Some of these were being used to manage transitive dependency versions that come from the other compile-scope maven artifacts to ensure they are consistent with the target API version and avoid mixed versions, even if the tests pass. But it might be one of the entirely unused ones, indeed.

I was trying to tidy this up to make it a bit easier to reason about the maven target API version upgrade and getting rid of some deprecated dependencies.

jeremylong (dismissed):

maven/pom.xml
chadlwilson: This was only coming from
```
+- org.apache.maven:maven-core:jar:3.6.3:provided (scope not updated to compile)
|  +- org.apache.maven.shared:maven-shared-utils:jar:3.4.2:provided (version managed from 3.2.1)
|  |  +- (org.slf4j:slf4j-api:jar:2.0.17:provided - version managed from 1.7.36; omitted for duplicate)
|  |  \- (commons-io:commons-io:jar:2.21.0:provided - version managed from 2.11.0; omitted for duplicate)
...
```

Tests seem to work fine without the override, and since it is at `provided` scope anyway, it doesn't do anything at runtime, so we can safely omit it and probably better to use the version implied by `maven-core` to compile against.

nhumblot (approved):
**praise**: thanks for the reactivity!

PR #8328, opened 2026-02-26
test: Make tests locale independent

nhumblot:
## Description of Change

This change proposes to improve assertions to make them locale independent. The updated tests are failing on a French locale system.

## Related issues

None

*no* new test as no functionality is added


chadlwilson (approved):

chadlwilson:
Makes it a bit harder to understand the nature of the tests, but I guess I understand why you want to change this. Since ODC is not localized software, If it was easier to get IDEs to pick up settings from maven surefire config I'd normally have preferred to just force the locale during tests (or use a system properties test extension to enforce a locale for duration of the test) but I think either of those will probably be a bit of a pain.

jeremylong (approved):

PR #8354, opened 2026-03-06
docs: Clarify how suppress items are matched against files

### LINKED ISSUE #8351: Clarify file separator to be used in `filePath` suppressions
In https://dependency-check.github.io/DependencyCheck/general/suppression.html there are some example to suppress by filePath. For example

```
<filePath>c:\path\to\some.jar</filePath>
```

However it is unclear how to use a relative file path (leveraging regex) in an OS agnostic way. Which file separator can/should be used? Is forward slash also supported on Windows? At the same time it would be nice if relative file paths are supported (and which path they should be relative to).

kwin:
This closes #8351

## Description of Change

Documentation clarification

## Related issues

#8351

no

### DIFF
--- a/src/site/markdown/general/suppression.md
+++ b/src/site/markdown/general/suppression.md
@@ -25,11 +25,11 @@ The following shows some other ways to suppress individual findings. Note the wa
 
 - their sha1 hash (via element `<sha1>`),
 - their [package URL (PURL)][purl] (via element `<packageUrl>`),
-- their Maven like coordinates (groupId/namespace, artifactId/name and version via element `<gav>`), or
+- their Maven-like coordinates (groupId/namespace, artifactId/name and version via element `<gav>`), or
 - their file path (via element `<filePath>`).
 
-The latter three can optionally be given as regular expression. The `<packageUrl>` value is matched against the dependency 
-specific software identifiers (can be looked up from the report) and `<gav>` against these identifiers after they have been mapped 
+Each of the latter three can optionally be specified as a regular expression. The `<packageUrl>` value is matched against the dependency
+specific software identifiers (can be looked up from the report) and `<gav>` against these identifiers after they have been mapped
 to coordinates via `PurlIdentifier.toGav()`.
 The `<filePath>` value is matched against the OS-specific absolute file path of the according dependency
 (for example a path inside the local Maven repository).
--- a/src/site/markdown/general/suppression.md
+++ b/src/site/markdown/general/suppression.md
@@ -34,8 +34,8 @@ to coordinates via `PurlIdentifier.toGav()`. The latter is not always available 
 The `<filePath>` value is matched against the OS-specific absolute file path of the according dependency
 (for example a path inside the local Maven repository).
 
-Additionally, there are several things that can be suppressed - individual CPEs, individual CVEs, or all CVE entries below a specified CVSS score. The most common
-would be suppressing CPEs based off of SHA1 hashes or filePath (regexes) - these entries can be generated using the
+Additionally, there are several ways to suppress vulnerabilities: individual CPEs, individual CVEs, or all vulnerabilities below a specified CVSS score. The most common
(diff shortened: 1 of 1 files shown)


src/site/markdown/general/suppression.md
```diff
@@ -22,8 +22,19 @@ A sample suppression file would look like:
...
-can be suppressed - individual CPEs, individual CVEs, or all CVE entries below a specified CVSS score. The most common
+
+- their sha1 hash (via element `<sha1>),
+- their package URL (via element `<packageUrl>`),
+- their Maven like coordinates (groupId/namespace, artifactId/name and version via element `<gav>`), or
+- their file path (via element `<filePath>`).
+
+The latter three can optionally be given as regular expression. The `<packageUrl>` value is matched against the dependency 
+specific software identifiers (can be looked up from the report) and `<gav>` against these identifiers after they have been mapped 
+to coordinates via `PurlIdentifier.toGav()`. The latter is not always available while the former is mandatory.
```
chadlwilson: Purls are not always available for dependencies, so it's not "mandatory", if that's what you were referring to?

Possibly also better to rephrase without using context-dependent latter/former, since it's a bit confusing given `latter` is also used a couple of lines above to refer to something different.

src/site/markdown/general/suppression.md
```diff
@@ -22,8 +22,19 @@ A sample suppression file would look like:
...
+The `<filePath>` value is matched against the OS-specific absolute file path of the according dependency
+(for example a path inside the local Maven repository).
+
+Additionally, there are several things that can be suppressed - individual CPEs, individual CVEs, or all CVE entries below a specified CVSS score. The most common
 would be suppressing CPEs based off of SHA1 hashes or filePath (regexes) - these entries can be generated using the
```
chadlwilson: Not critical, but given the above changes, we probably could rephrase this - it's pretty out of date to refer to suppressing by hash or file path as being the "most common".

The "most common" would probably be matching via purl regex, which is why we require this information when people file false positive reports - that's the most reliable way to suppress things. `gav` was most common pre-purl, when only Java artifacts really had a canonical designation.

chadlwilson (changes requested):
Thanks for this!

src/site/markdown/general/suppression.md
kwin: In which cases are Purls not available? So only sha1 is available for all files?

chadlwilson: Purls have to be guessed/inferred for many analyzers as there is not always package metadata. E.g jars without POMs, arbitrary JavaScript files,  dlls/assemblies, scanning inside packages or archives of one type which contain built artifacts of another vendored.within.

Especially when using the CLI/docker image to scan java projects.

Sometimes there just isn't sufficient metadata.

chadlwilson: Only hashes are always available, but it's usually a last resort for suppressing because it's obviously brittle, changes with version etc so can't be regexed.

src/site/markdown/general/suppression.md
jeremylong:
```suggestion
Additionally, there are several ways to suppress vulnerabilities: individual CPEs, individual CVEs, or all vulnerabilities below a specified CVSS score. The most common
would be suppressing CPEs based on the identified Package URL (PURL) - these suppression entries can be generated using the
```

ant/.gitignore
chadlwilson, line 1: These aren't necessary, and are orthogonal to this PR - there is already a gitignore in the root that covers these?

kwin: committed by mistake and now removed.

chadlwilson: Ahh, cheers!

chadlwilson: Anyway, if you could mark as resolved when you feel you’ve addressed the various comments or we have come to a conclusion, I’ll take another proper look 👍

jeremylong (approved):

chadlwilson (approved):

PR #8355, opened 2026-03-07
fix: improve configuration consistency (casing)

jeremylong:
## Description of Change

Used `amp` to check if we had any inconsistencies between configuration names and documentation.  When incorrectly cased configuration items were found, the old field/property/etc. was deprecated, and the correctly cased variant was added.


ant/src/main/java/org/owasp/dependencycheck/taskdefs/Check.java
```diff
@@ -998,23 +998,49 @@ public void setNodeAuditSkipDevDependencies(Boolean nodeAuditSkipDevDependencies
...
      * Add a regular expression to the set of retire JS content filters.
      * <p>
      * This is called by Ant.
      *
      * @param retirejsFilter the regular expression used to filter based on file
      * content
+     * @deprecated Use {@link #addConfiguredRetireJsFilter(RetirejsFilter)} instead.
      */
+    @Deprecated
     public void addConfiguredRetirejsFilter(final RetirejsFilter retirejsFilter) {
```
chadlwilson: Probably should rename the class, but is a bit annoying when working with case-insensitive file systems and git. So no big deal.

ant/src/main/java/org/owasp/dependencycheck/taskdefs/Check.java
```diff
@@ -998,23 +998,49 @@ public void setNodeAuditSkipDevDependencies(Boolean nodeAuditSkipDevDependencies
      *
      * @param retirejsFilterNonVulnerable new value of
      * retirejsFilterNonVulnerable
+     * @deprecated Use {@link #setRetireJsFilterNonVulnerable(Boolean)} instead.
      */
+    @Deprecated
     public void setRetirejsFilterNonVulnerable(Boolean retirejsFilterNonVulnerable) {
         this.retirejsFilterNonVulnerable = retirejsFilterNonVulnerable;
```
chadlwilson: Would be good if it renamed the internal variables (arguably even the parameters) to the target state.

cli/src/main/java/org/owasp/dependencycheck/CliParser.java
```diff
@@ -452,6 +452,8 @@ private void addAdvancedOptions(final Options options) {
                         "The password to authenticate to Retire JS Repository URL"))
                 .addOption(newOption(ARGUMENT.RETIREJS_FILTER_NON_VULNERABLE, "Specifies that the Retire JS "
                         + "Analyzer should filter out non-vulnerable JS files from the report."))
```
chadlwilson: Should denote in description that this is deprecated and the replacement name.

cli/src/main/java/org/owasp/dependencycheck/CliParser.java
```diff
@@ -474,6 +476,10 @@ private void addAdvancedOptions(final Options options) {
                         "Specify Retire JS content filter used to exclude files from analysis based on their content; "
                                 + "most commonly used to exclude based on your applications own copyright line. This "
                                 + "option can be specified multiple times."))
```
chadlwilson: Should denote in description that this is deprecated and the replacement name.

cli/src/main/java/org/owasp/dependencycheck/CliParser.java
```diff
@@ -1596,12 +1604,24 @@ public static class ARGUMENT {
         public static final String RETIRED = "enableRetired";
         /**
          * The CLI argument for the retire js content filters.
+         * @deprecated Use {@link #RETIRE_JS_FILTERS} instead.
          */
+        @Deprecated
         public static final String RETIREJS_FILTERS = "retirejsFilter";
```
chadlwilson: IMHO since the actual constant is not user consumed, would be better to name more descriptively(`_DEPRECATED`), not just differ by an underscore.

cli/src/main/java/org/owasp/dependencycheck/CliParser.java
```diff
@@ -1596,12 +1604,24 @@ public static class ARGUMENT {
...
+        /**
+         * The CLI argument for the retire JS content filters.
+         */
+        public static final String RETIRE_JS_FILTERS = "retireJsFilter";
         /**
          * The CLI argument for the retire js content filters.
+         * @deprecated Use {@link #RETIRE_JS_FILTER_NON_VULNERABLE} instead.
          */
+        @Deprecated
         public static final String RETIREJS_FILTER_NON_VULNERABLE = "retirejsFilterNonVulnerable";
```
chadlwilson: IMHO since the actual constant is not user consumed, would be better to name more descriptively(`_DEPRECATED`), not just differ by an underscore.

maven/src/main/java/org/owasp/dependencycheck/maven/BaseDependencyCheckMojo.java
```diff
@@ -745,24 +745,46 @@ public abstract class BaseDependencyCheckMojo extends AbstractMojo implements Ma
...
+    @SuppressWarnings("CanBeFinal")
+    @Parameter(property = "ossIndexAnalyzerUseCache")
+    private Boolean ossIndexAnalyzerUseCache;
+    /**
+     * URL of the Sonatype OSS Index service.
+     */
+    @SuppressWarnings("CanBeFinal")
+    @Parameter(property = "ossIndexAnalyzerUrl")
+    private String ossIndexAnalyzerUrl;
+
```
chadlwilson, line 765: Maven already has `alias` on `@Parameter` - probably no need to add new properties and deprecations. Rename to the target and add an alias for the deprecated capitalization and then no need to change settings logic - basically zero maintenance overhead to retaining the old capitalization, even if we never remote the aliases?

chadlwilson (changes requested):
Things missed:
- `ant/src/test/resources/build.xml` still referring to deprecated capitalizations. Should be updated.
- `cli/src/main/resources/completion-for-dependency-check.sh` still refers to deprecated capitalization, should not.
- CLI: missed to correct `--disableRetireJS`

nhumblot:
**suggestion(non-blocking):** it may be scope creep and fitting more into an additional PR, this comment is more to raise the discussion. In order to assist the migration from the deprecated parameters to the new ones before the removal of the old parameters, could we consider adding warning logs in the form of:

> [WARN] 'retirejsFilter' is deprecated and may be removed in the next major release, please migrate to 'retireJsFilter'

?

This may incentivize our users to switch prior the major version and ease their migration.

I am considering doing something similar for the ossIndex url, so it could fit in an additional change. If we prefer to add this change in a following PR, I can create a task.

ant/src/main/java/org/owasp/dependencycheck/taskdefs/Check.java
marcelstoer: Yeah, I guess it's not worth the potential hassle. Could still leave a comment on that class to that effect.

marcelstoer:
@nhumblot I fully support your proposal.

jeremylong:
Thanks for the feedback - I'll try to incorporate this over the weekend.

chadlwilson (approved):
LGTM, thanks @jeremylong !

marcelstoer (approved):

PR #8366, opened 2026-03-15
fix: warn if deprecated configs are used

jeremylong:
For Ant and the CLI - warn if deprecated configuration settings are used.  I would have done this for Maven also, but due to the use of "alias" it is more difficult.

### DIFF
--- a/maven/src/main/java/org/owasp/dependencycheck/maven/BaseDependencyCheckMojo.java
+++ b/maven/src/main/java/org/owasp/dependencycheck/maven/BaseDependencyCheckMojo.java
@@ -18,20 +18,23 @@
 package org.owasp.dependencycheck.maven;
 
 import com.github.packageurl.MalformedPackageURLException;
-import com.github.packageurl.PackageURL.StandardTypes;
 import com.github.packageurl.PackageURL;
-import java.util.stream.Collectors;
-import java.util.stream.Stream;
-
+import com.github.packageurl.PackageURL.StandardTypes;
 import org.apache.commons.lang3.StringUtils;
 import org.apache.maven.artifact.Artifact;
 import org.apache.maven.artifact.DefaultArtifact;
 import org.apache.maven.artifact.handler.DefaultArtifactHandler;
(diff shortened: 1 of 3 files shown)


marcelstoer:
👍 If we expect the logging code to live longer, you could extract a helper function. The logging message follows a pattern (DRY).

As for Maven, we could check the `MojoExecution` configuration.

```java
@Parameter(defaultValue = "${mojoExecution}", readonly = true)
private MojoExecution mojoExecution;

private void warnIfDeprecatedParamUsed(String deprecatedName, String currentName) {
    mojoExecution.getConfiguration()
        .getChildren()
        .stream()
        .filter(child -> deprecatedName.equals(child.getName()))
        .findFirst()
        .ifPresent(c -> getLog().warn("..."));
}

// invoke warnIfDeprecatedParamUsed() for each pair or maybe use reflection to find all aliased @Parameter annotation?
```

Or check the Maven session

```java
@Parameter(defaultValue = "${session}", readonly = true)
private MavenSession session;

...
// session.getUserProperties() or session.getCurrentProject().getProperties()
```

nhumblot (dismissed):

jeremylong:
@marcelstoer thank you for the idea regarding Maven!  Your suggestion doesn't exactly work because we are using an alias instead of actually deprecating the parameter. However, based on your suggestion, I was able to check for the aliases.

maven/src/main/java/org/owasp/dependencycheck/maven/BaseDependencyCheckMojo.java
```diff
@@ -18,20 +18,23 @@
 package org.owasp.dependencycheck.maven;
 
 import com.github.packageurl.MalformedPackageURLException;
-import com.github.packageurl.PackageURL.StandardTypes;
```
jeremylong, line 21: Formatting changes and import re-ordering due to Intelij auto-format....  The actual changes are the bottom two functions in this class.

marcelstoer:
LGTM

PR #8378, opened 2026-03-19
chore: remove unused properties and schemas

chadlwilson:
## Description of Change

Some minor tidies
- removes unused properties default values
- deprecates the constants that referenced them (and some others without defaults); annotated with the ODC version they became dereferenced
- removes some old schema documents that are no longer used by or relevant to ODC itself (some have been moved to open-vulnerability-clients)

N/A


jeremylong (approved):
LGTM

PR #8381, opened 2026-03-21
chore: pin GitHub actions to specific SHAs rather than mutable tags

chadlwilson:
## Description of Change

In light of (sadly many) GitHub Actions supply chain attacks ([including trivy-action causing all their mutable-tag actions users to potentially be compromised](https://www.wiz.io/blog/trivy-compromised-teampcp-supply-chain-attack)), it is generally considered better practice for both build reproducibility and security to pin Actions versions to specific SHAs for versions (at the very least for non-official community actions outside the `actions/` organisation).

I've done this here (except for [our own action](https://github.com/dependency-check/Dependency-Check_Action), which is not tagged so won't work with dependabot if it is hash-pinned).

To compensate for the noise, I have proposed grouping all actions bumps together into one weekly PR rather than the current, rather noisy, daily frequency. Frequency could also be even less frequent, or can choose different grouping strategies, if you'd rather. (**edit:** subsequently changed to _monthly_ grouped update)

Note that `actions/create-release` and `actions/upload-release-asset` are deprecated/EOL and need replacement. Possibly just scripting around the `gh` CLI is sufficient these days if we want to avoid non-official actions, since that CLI is available on all runners.

## Related issues

N/A

N/A

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -7,7 +7,10 @@ updates:
   - package-ecosystem: "github-actions"
     directory: "/"
     schedule:
-      interval: "daily"
+      interval: "weekly"
+    groups:
+      actions-deps:
+        patterns: [ "*" ]
   - package-ecosystem: "docker"
     directory: "/"
     schedule:
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -7,7 +7,7 @@ updates:
   - package-ecosystem: "github-actions"
     directory: "/"
     schedule:
-      interval: "weekly"
+      interval: "monthly"
     groups:
       actions-deps:
         patterns: [ "*" ]
--- a/.github/workflows/build-pull-requests.yml
+++ b/.github/workflows/build-pull-requests.yml
@@ -23,18 +23,18 @@ jobs:
       contents: read
     runs-on: ubuntu-latest 
     steps:
-      - uses: actions/checkout@v6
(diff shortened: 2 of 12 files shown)


.github/dependabot.yml
```diff
@@ -7,7 +7,10 @@ updates:
   - package-ecosystem: "github-actions"
     directory: "/"
     schedule:
-      interval: "daily"
+      interval: "weekly"
```
jeremylong: If we are just looking at the actions themselves and we are concerned about supply chain attacks...  Monthly?  Thoughts?

```suggestion
      interval: "monthly"
```

jeremylong (dismissed):
LGTM - just not sure if we should do monthly vs weekly.

.github/dependabot.yml
chadlwilson: Yeah, I tend to do monthly for actions in my projects, but there are a lot more different actions here than my typical project - so you might find the grouped monthly get too "big" and prefer to review small groupings more regularly.

Since you're doing the rather aggressive/noisy "daily" at the moment I was just going one notch up 😅

There is also a cooldown setting (time for a malicious release to be detected and pulled before one updates) but I didnt introduce thet here.

jeremylong (approved):

chadlwilson:
I've been careful with the hashes here (and made sure they load into IntelliJ IDEA for syntax checking), however, there's a chance I have screwed some up for workflows that don't run as part of the PRs. Of course feel free to tag me somewhere if there are problems during release later due to something I have possibly screwed up.

PR #8386, opened 2026-03-24
fix: add corepack to docker image

dcyrille18:
## Description of Change

This merge request adds Corepack to the base OWASP Dependency-Check Docker image.
The goal is to provide native support for package managers such as Yarn and pnpm without requiring downstream images or CI pipelines to manually install Corepack.

This change enables:
- Use of yarn install or pnpm install directly inside the owasp/dependency-check image
- Better compatibility for projects using modern Node.js tooling
- Reduced setup time in CI where Dependency-Check is executed alongside Node-based dependency workflows

No breaking changes are introduced:
If Node.js tooling is not needed, the image behaves exactly as before.
Corepack is now enabled and used as recommended by the pnpm and yarn projects.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -17,6 +17,7 @@ ENV user=dependencycheck
 ENV JAVA_HOME=/opt/jdk
 ENV JAVA_OPTS="-Danalyzer.assembly.dotnet.path=/usr/bin/dotnet -Danalyzer.bundle.audit.path=/usr/bin/bundle-audit -Danalyzer.golang.path=/usr/local/go/bin/go"
 ENV ODC_NAME=dependency-check-docker
+ENV COREPACK_ENABLE_DOWNLOAD_PROMPT=false
 
 COPY --from=jlink /jlinked /opt/jdk/
 COPY --from=go /usr/local/go/ /usr/local/go/
@@ -28,10 +29,7 @@ RUN apk update                                                                  
     apk add --no-cache git ruby ruby-rdoc npm                                                        && \
     gem install bundler-audit                                                                        && \
     bundle audit update                                                                              && \
-    mkdir /opt/yarn                                                                                  && \
-    curl -Ls https://yarnpkg.com/latest.tar.gz | tar -xz --strip-components=1 --directory /opt/yarn  && \
-    ln -s /opt/yarn/bin/yarn /usr/bin/yarn                                                           && \
-    npm install -g pnpm                                                                              && \
+    npm install -g corepack                                                                          && \
     unzip dependency-check-${VERSION}-release.zip -d /usr/share/                                     && \
     rm dependency-check-${VERSION}-release.zip                                                       && \
     cd /usr/share/dependency-check/plugins                                                           && \
@@ -49,6 +47,11 @@ RUN apk update                                                                  
 ### remove any suid sgid - we don't need them
 RUN find / -path /proc -prune -perm +6000 -type f -exec chmod a-s {} \;
 USER ${UID}
+
+### preload pnpm and yarn
+RUN corepack enable                                                                                  && \
+    corepack prepare pnpm@latest --activate                                                          && \
+    corepack prepare yarn@1.22.22 --activate
 
 VOLUME ["/src", "/report"]
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -24,12 +24,11 @@ COPY --from=go /usr/local/go/ /usr/local/go/
 
 ADD cli/target/dependency-check-${VERSION}-release.zip /
 
-RUN apk update                                                                                       && \
-    apk add --no-cache --virtual .build-deps curl tar                                                && \
-    apk add --no-cache git ruby ruby-rdoc npm                                                        && \
-    gem install bundler-audit                                                                        && \
-    bundle audit update                                                                              && \
-    npm install -g corepack                                                                          && \
+RUN apk upgrade --no-cache                                                                           && \
+    apk add --no-cache --virtual .build-deps curl                                                    && \
+    apk add --no-cache git ruby npm                                                                  && \
+    gem install --no-document bundler-audit                                                          && \
+    npm install --global --ignore-scripts corepack                                                   && \
     unzip dependency-check-${VERSION}-release.zip -d /usr/share/                                     && \
     rm dependency-check-${VERSION}-release.zip                                                       && \
     cd /usr/share/dependency-check/plugins                                                           && \
@@ -42,16 +41,18 @@ RUN apk update                                                                  
     mkdir /report                                                                                    && \
     chown -R ${user}:0 /report                                                                       && \
     chmod -R g=u /report                                                                             && \
-    apk del .build-deps
+    apk del .build-deps                                                                              && \
+    rm -rf /tmp/* /root/.cache /root/.npm
 
 ### remove any suid sgid - we don't need them
 RUN find / -path /proc -prune -perm +6000 -type f -exec chmod a-s {} \;
 USER ${UID}
 
-### preload pnpm and yarn
-RUN corepack enable                                                                                  && \
-    corepack prepare pnpm@latest --activate                                                          && \
-    corepack prepare yarn@1.22.22 --activate
+### Cache pieces needed for the specific run user
+RUN bundle audit update                                                                              && \
+    corepack prepare pnpm@latest yarn@latest yarn@1 --activate                                       && \
+    printf "enableTelemetry: false\nenableScripts: false\n" >> ${HOME}/.yarnrc.yml                   && \
+    rm -rf /tmp/*
 
 VOLUME ["/src", "/report"]
 


Dockerfile
```diff
@@ -31,7 +31,7 @@ RUN apk update
     mkdir /opt/yarn                                                                                  && \
     curl -Ls https://yarnpkg.com/latest.tar.gz | tar -xz --strip-components=1 --directory /opt/yarn  && \
     ln -s /opt/yarn/bin/yarn /usr/bin/yarn                                                           && \
-    npm install -g pnpm                                                                              && \
+    npm install -g pnpm corepack                                                                     && \
```
chadlwilson: It's likely not correct to leave pnpm there.if installing corepack.

chadlwilson:
> Corepack remains disabled until explicitly invoked (e.g., corepack enable), following Node.js conventions.

This description looks LLM generated, because it mKes.assertions about the change that are already already true before the change, but i don't believe this particular point is actually correct.

Did you test this before/after?

When you install the package globally, from memory im pretty sure the shims are all crested per the package.json and on the path. IIRC `corepack enable` is just for the node-prepackaged version. If that were the case this change wouldn't be needed as corepack is already bundled with Node < 25 and should already be there ready to be enabled.

I'll have another look later, but this looks messy to me, given yarn and pnpm are still being manually installed. Should probably use only one 9r the other install approach - not both.

chadlwilson (changes requested):

Dockerfile
dcyrille18: Depending on the user project configuration, the analyzers could use pnpm or corepack, no ?

dcyrille18:
> This description looks LLM generated, because it mKes.assertions about the change that are already already true before the change, but i don't believe this particular point is actually correct.

Well seen, I ask to copilot to help me to write this change description because we turn around since many hours on how to get owasp working with our NodeJS project using yarn berry and corepack.


> Did you test this before/after?

Yes, of course, I tested and I thought it works but after more tests, it seems not totally in this first try. Let me add a short description on what we have before:
```
/src $ yarn --version
1.22.22
/src $ yarn
error This project's package.json defines "packageManager": "yarn@4.13.0". However the current global version of Yarn is 1.22.22.

Presence of the "packageManager" field indicates that the project is meant to be used with Corepack, a tool included by default with all official Node.js distributions starting from 16.9 and 14.19.
Corepack must currently be enabled by running corepack enable in your terminal. For more information, check out https://yarnpkg.com/corepack.
/src $ corepack
/bin/sh: corepack: not found
```

And after the first MR try (not fully automated for CI/CD):
```
/src $ yarn --version
! Corepack is about to download https://repo.yarnpkg.com/4.13.0/packages/yarnpkg-cli/bin/yarn.js
? Do you want to continue? [Y/n] y

4.13.0
```

After further test, it seems that adding the environnement variable COREPACK_ENABLE_DOWNLOAD_PROMPT=false make the yarn version download totally transparent:

```
/src $ yarn --version
4.13.0
```

Dockerfile
chadlwilson: corepack is a way to install/run pnpm (and yarn) - optionally with locked versions. When you install both you likely make a mess of the PATH.

We shouldn't have multiple ways to do the same thing in a container image, and in my view nor should it be a bag of different tools/approaches that various people like to bootstrap their tools.

We should pick one approach that we think is best for most users.

chadlwilson:
In my opinion if we're going to do this we should  probably switch the whole image to use corepack and only corepack. I'm not sure if there is value in both, and probably only downsides.

I guess corepack isn't there currently because the image is installing npm directly; not node.

Dockerfile
dcyrille18: So perhaps we can just install directly the latest NodeJS LTS and let corepack configure theirs environments (yarn, npn, pnpm) ?

chadlwilson: Maybe - npm -> corepack -> yarn | pnpm

No choice if we want to use corepack , since corepack isn't bundled with Node on Alpine, as you showed earlier.

But we might want to pre-cache a version for those not using package manager directive so it's not dynamically downloading in every single case (which is not easy to decide).

And need to actually research and see where pnpm and yarn are going now that corepack is no longer distributed with NodeJS. If they are ditching corepack and won't recommend use with it there's no point adding it here and we'll just be storing up problems for later.

dcyrille18: It seems corepack is no more distributed by the NodeJS from >25 but already recommended to manage properly the packages managers.

dcyrille18:
Update done as requested with prepached images for yarn (v1) and pnpm (latest):

Current:
```
/src $ pnpm --version
10.27.0
/src $ yarn --version
1.22.22
```

After MR:
```
/src $ pnpm --version
10.33.0
/src $ yarn --version
1.22.22
```

This build will now allow users to install custom package managers if needed.

marcelstoer (commented):
I do not support this proposal. Specific extensions should be handled in downstream images that extend from ours.

chadlwilson:
> I do not support this proposal. Specific extensions should be handled in downstream images that extend from ours.

Not sure how familiar you are with the node/npm ecosystem but it’s a bit more complex than “specific extensions” in my view. We already pre-package node, npm, yarn and pnpm using disparate approaches. The ODC code for these analyzers is already “corepack compatible/aware”.

For many years (prior to Node 25) corepack was pre-distributed with all Node versions (like with npm) and for all of these alternate package managers was the recommended approach to either bootstrap or install them, and provide a mechanism to “lock” ones version manager for reproducibility. Corepack shimmed the entry point scripts and would look at the `packageManager` field of the adjacent package.json, look for a config and determine whether it could run a locked version from its cache, already in source control or downloadable on-the-fly, plus verifying shas etc.

In other words, prior to Node 25, one could consider corepack being installed and enabled to be a sensible default for a “Swiss army knife” image like this which already pre-packages all sorts of tools. For some reason node voted not to package it anymore, which adds friction for every single alternate package manager. It’s worth noting that this change will not graduate to LTS until April/May 2027 so for now, corepack is still “LTS”ish.

According to OP, there appears to be a gap in the ODC image at the moment that causes yarn and pnpm to not work in a corepack aware way, and our image was possibly not “caught up” when the Yarn 2+ support was finally fixed in ODC itself. I think we need to confirm what these problems are. This is important now because yarn/pnpm are directly invoked by ODC and failure to use the correct version implied by package.json will cause scanning failures.

It’s likely only an accidental side effect of Alpine’s non-standard packaging process for node/npm that corepack has not been included earlier. Most distros follow nodes official packaging and include it. Some other distros also split corepack from node, but they include a proper package for it, so you’re not installing into the npm global space (Arch, Wolfi, newer Debian). For some reason Alpine has no distinct corepack package, however provides corepack in a compliant way with the `nodejs-current` packages -however these `current` versions are not locked to lts and are otherwise unsuitable.

Now, to summarise my earlier messages, the messy bit is determining how to move forward and requires some research into whether pnpm, yarn and others will continue to recommend corepack. It doesn’t make sense for us to add it if everyone is moving away.

In retrospect it’d probably have been better to open an issue to discuss the problem before submitting a solution, but it’s not the end of the world, as there’s not too much effort here I think.

In my personal view, if corepack still has a role, rather than using npm global space, I would suggest
- moving away from Alpine towards Wolfi
  - which is similarly a minimal security focused rolling undistro, but is glibc+based
  - with fewer breaking changes than Alpine, with much higher volume throughput on packaging
  - supports patching old versions of packages without abandoning like Alpine does
  - Also uses apk, with most packages named consistently with Alpine, making migration path easy for child packages
- install [corepack package](https://github.com/wolfi-dev/os/blob/main/corepack.yaml) in place of npm/yarn/pnpm

But I think we need to figure out exactly which combinations are not working right now with the pre-installed versions and whether corepack still has a role - before we decide to do anything.

chadlwilson (changes requested):

marcelstoer:
👍 Thanks for the effort that went into this write-up! Much appreciated.

chadlwilson:
Yeah, sorry for the novel.

I am interested in finding a consistent way forward as I recently ended up using corepack in the build to ensure we can run tests reliably since right now most of such integration/functional tests are coded to “don’t run test if yarn/pnpm etc not found” which is basically useless in ensuring we don’t break functionality since tests won’t even run if we set up the build incorrectly.

https://github.com/dependency-check/DependencyCheck/blob/bc07974f5d47de48a1f9d5107d3b6082f40b34ab/.github/workflows/build.yml#L65

I used corepack for simplicity so I could start ensuring these tests actually have a version of yarn to run, but will change the approach for consistency with whatever we choose for docker, if it makes sense to do so.

dcyrille18:
Hi everyone,

Due to security and performance requirements, we must upgrade to newer Yarn versions in our CI pipelines. With our current setup, this makes the official Docker image incompatible for us in its current form.

If the preferred direction is for users to maintain their own Docker images, that’s acceptable for us, but we would appreciate some high‑level guidance on the recommended way forward.

Once we align on the approach, I’m willing to handle the required development work so we can unblock usage on recent environments.

Let us know how you’d like to proceed.

chadlwilson:
I haven't had a chance to look at this further, just yet - but the blocker right now from my perspective is just research (does it make sense to introduce corepack given yarn+pnpm plans) and possibly a bit of sanity testing.

Dockerfile
```diff
@@ -28,10 +29,7 @@ RUN apk update
     apk add --no-cache git ruby ruby-rdoc npm                                                        && \
     gem install bundler-audit                                                                        && \
     bundle audit update                                                                              && \
-    mkdir /opt/yarn                                                                                  && \
-    curl -Ls https://yarnpkg.com/latest.tar.gz | tar -xz --strip-components=1 --directory /opt/yarn  && \
-    ln -s /opt/yarn/bin/yarn /usr/bin/yarn                                                           && \
-    npm install -g pnpm                                                                              && \
+    npm install -g corepack                                                                          && \
```
abhi-markan: Is there a reason why this is not in `package-lock.json` as you are not safe-guarding yourself against a supply-chain attack when building your image.

chadlwilson: Since we already do a pnpm global install, this is not specifically relevant to this PR, so not really the right place for this discussion, but global installs don't have package locks so a different mitigation approach is needed if indeed this is something we are concerned about.

Please open a dedicated issue or PR to discuss this, especially if you suggest alternate approaches which are sufficiently maintainable (when the goal is to make the latest version available to consumers).

Dockerfile
```diff
@@ -50,6 +48,11 @@ RUN apk update
 RUN find / -path /proc -prune -perm +6000 -type f -exec chmod a-s {} \;
 USER ${UID}
 
+### preload pnpm and yarn
+RUN corepack enable                                                                                  && \
+    corepack prepare pnpm@latest --activate                                                          && \
+    corepack prepare yarn@1.22.22 --activate
+
 VOLUME ["/src", "/report"]
```
chadlwilson: Suggested a change to this in https://github.com/dcyrille18/DependencyCheck/pull/1 to simplify

chadlwilson (changes requested):

chadlwilson:
After having looked at this a bit further and testing locally, I think it makes sense to go ahead with this with some tweaks suggested in https://github.com/dcyrille18/DependencyCheck/pull/1 to clean-up cache cruft left behind in the image layers.

**Current image**
- Yarn 1/classic is installed manually; yarn berry (2+) is not
- pnpm latest is installed
- npm from Alpine is installed

**corepack direction of tools**
- **pnpm**: corepack is still recommended https://pnpm.io/installation#using-corepack
  - community opinion [still seems to be that corepack is the best option](https://github.com/orgs/pnpm/discussions/9044)
- **yarn**: corepack still recommended approach https://yarnpkg.com/getting-started/install
  - users can still source control their specific yarn versions following yarn team advice if they like
  
**Advantages**
- consistency between yarn and pnpm
 - we _could_ also enable it for npm as well, to be totally consistent and better support folks using `package.json` `packageManager`/`devEngines` directives for npm?

**Disadvantages**
- need to do a global npm install since corepack is not packaged for Alpine.
  - as noted in https://github.com/dependency-check/DependencyCheck/pull/8386#issuecomment-4162178270 we could address this by moving to something non-Alpine, but that's essentially a breaking change so out of scope

@dcyrille18 if you could review/merge my PR-to-your-PR and go through the comments to resolve them here it will help make it easier to review for others.

dcyrille18:
Hi @chadlwilson,
Thanks for the time you spent on this.
I’ve reviewed and merged your proposal.

Dockerfile
dcyrille18: Done.

Dockerfile
```diff
@@ -17,21 +17,18 @@ ENV user=dependencycheck
...
-    bundle audit update                                                                              && \
-    mkdir /opt/yarn                                                                                  && \
-    curl -Ls https://yarnpkg.com/latest.tar.gz | tar -xz --strip-components=1 --directory /opt/yarn  && \
-    ln -s /opt/yarn/bin/yarn /usr/bin/yarn                                                           && \
-    npm install -g pnpm                                                                              && \
+RUN apk upgrade --no-cache                                                                           && \
+    apk add --no-cache --virtual .build-deps curl                                                    && \
+    apk add --no-cache git ruby npm                                                                  && \
+    gem install --no-document bundler-audit                                                          && \
+    npm install --global --ignore-scripts corepack                                                   && \
```
chadlwilson, line 31: I made these changes, just for cleanup while we are here
- we are using `--no-cache` so think `update` was intended as `upgrade`
- `tar` no longer needed
- `bundle audit update` needs to be done in the user context, not cached as root
- can avoid need for rdoc by using `--no-document` ruby install

Dockerfile
```diff
@@ -44,12 +41,19 @@ RUN apk update
     mkdir /report                                                                                    && \
     chown -R ${user}:0 /report                                                                       && \
     chmod -R g=u /report                                                                             && \
-    apk del .build-deps
+    apk del .build-deps                                                                              && \
+    rm -rf /tmp/* /root/.cache /root/.npm
```
chadlwilson, line 45: Cleaning up, since Node and Ruby leave unnecessary cruft in /tmp and `root`'s  home.

Dockerfile
```diff
@@ -44,12 +41,19 @@ RUN apk update
...
 
 ### remove any suid sgid - we don't need them
 RUN find / -path /proc -prune -perm +6000 -type f -exec chmod a-s {} \;
 USER ${UID}
 
+### Cache pieces needed for the specific run user
+RUN bundle audit update                                                                              && \
+    corepack prepare pnpm@latest yarn@latest yarn@1 --activate                                       && \
+    printf "enableTelemetry: false\nenableScripts: false\n" >> ${HOME}/.yarnrc.yml                   && \
+    rm -rf /tmp/*
```
chadlwilson, line 55: Being a good citizen be removing undocumented internet access by yarn, and disable scripts by default during `yarn install` for supply chain security good practices.

Also removes cache cruft that results from running `node` and `v8`.

chadlwilson (approved):
Adding some comments to explain the changes I personally added on, for maintainer review benefit.

This LGTM now and I think is a worthwhile contribution 👍

```shell
$ docker run -it --entrypoint='' owasp/dependency-check:12.2.1-SNAPSHOT sh -c "pnpm --version && yarn --version && corepack yarn@latest --version"
10.33.0
1.22.22
4.13.0
```

Dockerfile
```diff
@@ -44,12 +41,19 @@ RUN apk update
...
+    apk del .build-deps                                                                              && \
+    rm -rf /tmp/* /root/.cache /root/.npm
 
 ### remove any suid sgid - we don't need them
 RUN find / -path /proc -prune -perm +6000 -type f -exec chmod a-s {} \;
 USER ${UID}
 
+### Cache pieces needed for the specific run user
+RUN bundle audit update                                                                              && \
+    corepack prepare pnpm@latest yarn@latest yarn@1 --activate                                       && \
```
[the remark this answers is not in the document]
abhi-markan: Worth tagging to a specific SHA or a version to avoid a supply-chain attack please. Thanks

Dockerfile
abhi-markan: Worth installing `corepack` from `package-lock.json` and then running `npm ci --ignore-scripts`. Thanks

Dockerfile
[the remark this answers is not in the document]
chadlwilson: In my opinion - no. That adds additional maintenance burden that dependabot cannot help us with and will likely lead to stale versions without security fixes in the underlying software, which is of bigger risk to users.

Whose supply chain? What's the attack vector?
- For ODC's own supply chain this is within an ODC container so it poses minimal threat  because the npm install is within a docker build running inside a docker buildx container which doesn't have access to anything privileged.
- For user's supply chains it poses little risk because ODC is not released constantly. It'd require extremely bad luck for the latest version at the time of a release build to contain a vulnerable version

If yarn or pnpm publishing to `npm` are breached the world has MUCH bigger problems than ODC as an entrance. If you can suggest a way to keep the versions updated which dependabot can handle then it’d be OK to do this, but otherwise I do not think it is worth the trade off of additional unpaid OSS maintainer work and risk of unpatched versions staying around for longer.

As I already said earlier, this PR is really not the place for this discussion since we already have unpinned npm installs here.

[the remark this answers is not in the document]
chadlwilson: If you use this image for your paid work for your employer and are still concerned and feel you can mitigate supply chain risks better yourself - you can always build your own image?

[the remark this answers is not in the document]
chadlwilson: PR description needs updating.

Dockerfile
```diff
@@ -44,12 +41,19 @@ RUN apk update
...
+    printf "enableTelemetry: false\nenableScripts: false\n" >> ${HOME}/.yarnrc.yml                   && \
```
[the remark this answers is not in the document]
chadlwilson: Not an issue.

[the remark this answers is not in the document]
chadlwilson: Not an issue.

Dockerfile
chadlwilson: That’s not how global installs work?

Dockerfile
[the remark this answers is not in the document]
dcyrille18: Done.

jeremylong (approved):

Issue #7889, opened 2025-08-26
Replace the Legacy search.maven.org usage in Dependency download with Portal central.sonatype.com


nico-arianto:
**Is your feature request related to a problem? Please describe.**
Recently the Legacy Search - search.maven.org status is unstable (or having Major Outage)
<img width="830" height="1167" alt="Image" src="https://github.com/user-attachments/assets/de28b49e-c614-401f-9bd1-4ce658b6ffd0" />

**Describe the solution you'd like**
Replace https://github.com/dependency-check/DependencyCheck/blob/2dd7324527fb998c0ee1f8080a1ccf24b74169e9/core/src/main/java/org/owasp/dependencycheck/dependency/Dependency.java#L566 from "Legacy Search - search.maven.org" with "Portal - central.sonatype.com"

nico-arianto:
Related to https://github.com/dependency-check/DependencyCheck/issues/5827

aggeboe:
Related to #7686

marcelstoer:
Of course this affects FP processing as well: https://github.com/dependency-check/DependencyCheck/actions/runs/17265909782/job/48997818646#step:14:38

dhensche:
Is there a channel or thread somewhere that we can stay updated on where the central analyzer is headed? We have integrated ODC into our CI pipelines, but even with our own [vuln db](https://dependency-check.github.io/DependencyCheck/data/database.html) and persistent cache volume the central analyzer is taking up to 15 minutes on medium size projects. We use the CLI (and therefore the central analyzer) because we don't have a standardized gradle/maven build across our portfolio of applications so we cannot use either of those plugins that remove the necessity of running the central analyzer. The communications coming from Sonatype have not been very helpful. 

We are currently looking at standing up a nexus based proxy of maven central purely so we can have improved scan time that isn't impacted by the flakiness of the search API. The discussion on #8407 was an interesting read

chadlwilson:
Not other than this thread/issue. That PR is about all there is, but is really just an adjacent discussion to what is already here.

But using your own Nexus is worth a go. Really all bigger orgs should have something like Nexus or Artifactory to reduce their load on Maven Central even for regular artifact traffic - putting aside hash searches like this analyzer relies upon.

Do you cache the entirely data dir centrally? That will cache the central results as well.

As a side note, I think people overestimate how standardized a Maven or Gradle build needs to be to use ODC on it externally without having every project edit its POMs. 
[...]

PR #8407, opened 2026-04-09
feat: Allow Central Search to use JSON responses

B4ckslash:
## Description of Change

Maven central search supports JSON responses alongside the currently-used XML responses, which is also the preferred way of interacting with central.sonatype.com (the eventual/planned successor to the "legacy" central search). This is therefore also a prerequisite to eventually moving to that endpoint (while XML is also supported there, the schema is basically "JSON, but make it XML", i.e. the JSON properties are just converted to XML tags and that's it).

Unfortunately, the schema is currently undocumented, so the implementation is based on sample responses from the JSON endpoint.

So far, I have implemented the actual parsing and the necessary switches/settings for `core`, `cli`, and `ant`. I am marking this PR as draft for now as I have to add the new parameter to the Maven plugin as well.

## Related issues

- relates to #7889 

*no*

I wasn't entirely sure whether this is required, as I have found no tests relating to the XML format either. I'll be happy to add some JSON tests however!

### DIFF
--- a/ant/src/main/java/org/owasp/dependencycheck/taskdefs/Check.java
+++ b/ant/src/main/java/org/owasp/dependencycheck/taskdefs/Check.java
@@ -139,6 +139,10 @@ public class Check extends Update {
      */
     private Boolean centralAnalyzerUseCache;
     /**
+     * Whether or not the Central Analyzer should use JSON responses.
+     */
+    private Boolean centralAnalyzerUseJson;
+    /**
      * Whether or not the nexus analyzer is enabled.
      */
     private Boolean nexusAnalyzerEnabled;
@@ -1109,6 +1113,15 @@ public class Check extends Update {
     }
 
     /**
+     * Set the value of centralAnalyzerUseJson.
+     *
+     * @param centralAnalyzerUseJson new value of centralAnalyzerUseJson
+     */
+    public void setCentralAnalyzerUseJson(Boolean centralAnalyzerUseJson) {
+        this.centralAnalyzerUseJson = centralAnalyzerUseJson;
+    }
+
+    /**
      * Set the value of nexusAnalyzerEnabled.
      *
      * @param nexusAnalyzerEnabled new value of nexusAnalyzerEnabled
@@ -1599,6 +1612,7 @@ public class Check extends Update {
         getSettings().setBooleanIfNotNull(Settings.KEYS.ANALYZER_LIBMAN_ENABLED, libmanAnalyzerEnabled);
         getSettings().setBooleanIfNotNull(Settings.KEYS.ANALYZER_CENTRAL_ENABLED, centralAnalyzerEnabled);
         getSettings().setBooleanIfNotNull(Settings.KEYS.ANALYZER_CENTRAL_USE_CACHE, centralAnalyzerUseCache);
+        getSettings().setBooleanIfNotNull(Settings.KEYS.ANALYZER_CENTRAL_JSON, centralAnalyzerUseJson);
         getSettings().setBooleanIfNotNull(Settings.KEYS.ANALYZER_NEXUS_ENABLED, nexusAnalyzerEnabled);
         getSettings().setBooleanIfNotNull(Settings.KEYS.ANALYZER_ARCHIVE_ENABLED, archiveAnalyzerEnabled);
         getSettings().setBooleanIfNotNull(Settings.KEYS.ANALYZER_ASSEMBLY_ENABLED, assemblyAnalyzerEnabled);
--- a/cli/src/main/java/org/owasp/dependencycheck/App.java
+++ b/cli/src/main/java/org/owasp/dependencycheck/App.java
@@ -652,6 +652,8 @@ public class App {
                 !cli.isDisabled(CliParser.ARGUMENT.DISABLE_CENTRAL, Settings.KEYS.ANALYZER_CENTRAL_ENABLED));
         settings.setBoolean(Settings.KEYS.ANALYZER_CENTRAL_USE_CACHE,
                 !cli.isDisabled(CliParser.ARGUMENT.DISABLE_CENTRAL_CACHE, Settings.KEYS.ANALYZER_CENTRAL_USE_CACHE));
+        settings.setBoolean(Settings.KEYS.ANALYZER_CENTRAL_JSON,
+                !cli.isDisabled(CliParser.ARGUMENT.ANALYZER_CENTRAL_JSON, Settings.KEYS.ANALYZER_CENTRAL_JSON));
         settings.setBoolean(Settings.KEYS.ANALYZER_OSSINDEX_ENABLED,
                 !cli.isDisabled(CliParser.ARGUMENT.DISABLE_OSSINDEX, Settings.KEYS.ANALYZER_OSSINDEX_ENABLED));
         settings.setBoolean(Settings.KEYS.ANALYZER_OSSINDEX_USE_CACHE,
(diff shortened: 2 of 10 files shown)


chadlwilson:
Hi there

I'm just a fellow contributor but wonder why do you feel it'd be valuable to potentially introduce client side instability, complexity and two different implementations here - if it's not Sonatype's medium term goal for this endpoint to stay around in either form? Is it demonstrably faster or preferred by Sonatype in some public comms? Has it had less downtime than the XML part?

If it's not reliable enough to immediately replace XML, personally I cannot really see the value. We shouldn't need to expose such internal detail to user config and maintaining two deprecated impls rather than one doesn't seem like a good idea to me.

Which replacement are you referring to that this would help.towsrds.moving to? I havent seen any replacement documented such that id conclude it'll have overlap with the legacy search.

B4ckslash:
Hi @chadlwilson,

I guess I should have been more clear in the elevator pitch: I want to change ODC over to the `central.sonatype.com` Portal endpoint, which has a far better uptime according to [the status page](https://status.maven.org/). This PR is just a sort of preparatory intermediate step, as the legacy search endpoint happens to use the same JSON schema that the portal endpoint does, whereas the XML schemas differ between them.

I figured that instead of cutting over to the Portal endpoint and switching to JSON at the same time, it might be a good idea to stick with the same endpoint but switch the schema. You're right though, it's probably a better idea to replace the XML outright if this is deemed a sensible course of action. I'm just more used to do things gradually and configurably from my day job, so that's what I defaulted to.

So, in summary: 
* I want ODC to switch over to `central.sonatype.com` as it has a better uptime, according to Sonatype themselves
* This requires JSON (or a change to the XML schema which effectively makes it JSON)
* I figured submitting the JSON change as a separate PR would be easier to review than doing both endpoint + schema at once
* You're right that if the above is what ODC wants to do, it's more sensible to just replace XML entirely. I will update this PR accordingly.

chadlwilson:
Which JSON search API are you referring to that actually exists on a `central.sonatype.com` endpoint though?

There are no search APIs listed at https://central.sonatype.com/api-doc and when you go to https://central.sonatype.org/search/ the API examples all still point to legacy search. The FAQ [still refers to there being no replacement](https://central.sonatype.org/faq/what-happened-to-search-maven-org/#what-about-the-searchmavenorg-api). There is nothing on [NEWS](https://central.sonatype.org/news/archive/).

According to https://github.com/dependency-check/DependencyCheck/issues/7686#issuecomment-3198494864 and some emails to their support there is a "compatiiblity API" at https://central.sonatype.com/solrsearch/select but it's not documented anywhere. 

Fishing against a completely undocumented API is not a good use of anyone's time, and it's already been done and found not compatible: https://github.com/dependency-check/DependencyCheck/issues/7686#issuecomment-3239200705 If they documented it and said "this is how it is going to be, and this is what is stable vs subject-to-change", then someone would probably put the effort in to test+migrate.

It's great to do things incrementally where possible (and not confusing to users) but I don't think there is any reason to think this is any kind of "intermediate step"; unless magically when using `json` the two APIs look identical in response. (I highly doubt that as it looks like they are directly exposing SOLR in both cases, and the underlying search indexing approach is different)

I'm honestly quite confused, because over the last 9 months of instability, people keep referring to the "migration" in #7889 as if there is something to migrate to, which there doesn't seem to be? :-)

chadlwilson:
In any case, I'd suggest hanging off a bit and seeing what the (real) maintainers think before putting too.mich more effort in, just to see if they think it's a direction worth pursuing against `search.maven.org`.

B4ckslash:
All fair points, and I agree on holding off on much more effort until a maintainer has weighed in.

However, two things I'd like to point out:

> Fishing against a completely undocumented API is not a good use of anyone's time

There's evidently frustration in the community of ODC users regarding the flakiness of `search.maven.org`. In my case, we use ODC on our CI pipeline at my job, and when central search has a bad day, everything screeches to a halt while we try to get a handle on things. I figured it'd be at least worth a shot trying to get ODC talking to `central.sonatype.com`, which as stated seems to be a much more stable endpoint, even if the `/solrsearch` API is undocumented.

> unless magically when using json the two APIs look identical in response

While the responses are not fully identical, the JSON response content that matters for the central analyzer is mostly the same, at least in the few example searches I tried. It may well not be good enough to just have something that is _mostly_ the same though, I leave that judgment to people more qualified in that matter.

 In any case, I only hacked this code together in a couple of hours, so it's not too much time wasted if this avenue ultimately doesn't get pursued. I mostly wanted to see if this is something worth doing, and I do think it could be, if it turns out to be a way to deal with the rather poor stability of the maven central search.

As an aside, I would also prefer it if Sonatype made some official guidelines wrt. the whole legacy/portal search thing. If they intended to keep central search around in the medium-long term, then I do not understand why the undocumented compatibilty API even exists on portal search and why central search's flakiness is deemed acceptable. And if they didn't plan on keeping central search around, then I don't get why they don't communicate their plans officially so that the community can take the steps necesary to be compatible when the time comes.

chadlwilson:
> There's evidently frustration in the community of ODC users 

Not as much frustration as if it completely breaks because we switch ODC to rely on an undocumented API with literally no suggestion to use in public (that anyone can point to) by the people that maintain it.

Many ODC users often don't seem to even cache/restore data directory to isolate themselves from bouts of instability, so a lot of the user noise can be a bit perplexing.

I haven't looked, but perhaps the search APIs are used directly by the website search via the browser - so it exists for that purpose. Which doesn't mean it constitutes a stable *external* API since Sonatype can update the website any time to stay consistent with the search API.

I suspect all of this is downstream from the challenges of operating a service of Maven Central size, funding (or lack) thereof and Sonatype's other trade-offs in terms of deploying resources towards their commercial products while the whole dev tool part of the industry is suffering right now. Just a guess though.

marcelstoer:
> Not as much frustration as if it completely breaks because we switch ODC to rely on an undocumented API with literally no suggestion to use in public (that anyone can point to) by the people that maintain it.

I fully share your concerns. The best we can do for our users in terms of stability is to rely on documented APIs only. I think we can safely assume that the API provider wants us to use the documented API over the undocumented one for a reason.

At the same time I of course share your frustration over how Sonatype is handling this. This has come up in a couple of issues in this project.

chadlwilson:
Possibly a small set of us could approach Sonatype Central support and see if we can get a technical/product contact for something more official, if Jeremy doesn't already have such a contact.

Anecdotally someone had heard from support that they were intending to look at Search APIs after the major publishing and search portal migration was complete, but that's has been completed for quite a while now, and things seem reasonably stable on the so perhaps it's time to see if they had a change of plans (maybe motivated by their Sonatype Guide product launch).

B4ckslash:
> Anecdotally someone had heard from support that they were intending to look at Search APIs after the major publishing and search portal migration was complete, but that's has been completed for quite a while now, and things seem reasonably stable on the so perhaps it's time to see if they had a change of plans (maybe motivated by their Sonatype Guide product launch).

AFAIK Sonatype is currently migrating users of the OSS Indexer to use the new Guide service (which also affects ODC in a roundabout way), so perhaps that's what's keeping them. I do get the feeling that they're not super invested in the Search API functionality as a whole.

jeremylong:
Honestly - the reason we have not migrated to this yet is the lack of support and documentation. I have tried reaching out to my contacts at Sonatype but I have not gotten a response yet.

B4ckslash:
@jeremylong thank you for the information. 
I think I'll close this PR, a proper migration will probably look a lot different from this tacked-on configuration in here.

PR #8409, opened 2026-04-10
fix(fp): suppress non-golang false positives for `distribution` project

### LINKED ISSUE #8408: [FP]: CVE-2026-33540 detected in private DTO Jar
### Package URl

pkg:maven/de.finkonsens.kdialog.bno.biene.biene-bno/biene-ka-bno-shared-distribution@1.0.0.5

### CPE

cpe:2.3:a:distribution_project:distribution:1.0.0.5:*:*:*:*:*:*:*

### CVE

CVE-2026-33540

### ODC Integration

{"label" => "Maven Plugin"}

### ODC Version

12.2.0

### Description

In our project we aggregate JARs from several developer teams within our agency.

One of these JARs, which only includes a collection of DTOs and interfaces, was flagged with CVE-2026-33540. From my understanding of the CVE this should be a false positive.

chadlwilson:
## Description of Change

The distribution project: https://github.com/distribution/distribution is a golang toolkit with an extremely generic name likely to generate false positives in many different places.

This change proposes a generic suppression against anything that's not golang, as a workaround for lack of support for the extended attributes in CPEs (`:go:`) as denoted in NVD (surprisingly given I think it only has one language).

## Related issues

- fixes #8408

N/A

### DIFF
--- a/generatedSuppressions.xml
+++ b/generatedSuppressions.xml
@@ -2090,3 +2090,12 @@ only pkg:maven/org.clojure:clojure@.* is the CPE cpe:/a:clojure:clojure
     <packageUrl regex="true">^pkg:maven/dev\.detekt/detekt-report-checkstyle@.*$</packageUrl>
     <cpe>cpe:/a:checkstyle:checkstyle</cpe>
 </suppress>
+<suppress base="true">
+    <notes><![CDATA[
+    FP per issue #8408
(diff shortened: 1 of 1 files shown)


jeremylong (approved):

jeremylong:
publishing here: https://github.com/dependency-check/DependencyCheck/actions/runs/24285348794

PR #8419, opened 2026-04-14
feat: support and prefer githubID vuln identifiers from RetireJS

### LINKED ISSUE #8402: Vulnerability description is used as vulnerability name
**Precondition**

**Describe the bug**
For some specific vulnerabilities, it seems that Dependency Check uses the description of the vulnerability as vulnerability name. This looks pretty strange in the [logs](https://gist.github.com/GunoH/ed48f53331da655658d640f07cd80932) and reports (for `org.springdoc:springdoc-openapi-starter-webmvc-ui:2.8.6`):

Image

What's worse though, is that I cannot seem to suppress these vulnerabilities. Using the _suppress_  button gives me an xml fragment with a multiline value for _vulnerabilityName_, which leads to parse errors at runtime, like for example:

> Unable to parse suppression xml file 'suppressions.xml'
org.owasp.dependencycheck.xml.suppression.SuppressionParseException: org.xml.sax.SAXException: Line=33, Column=570: Open quote is expected for attribute "onclick" associated with an  element type  "img".

The problem occurs using version 12.2.0 of the gradle plugin.

**Log file**
https://gist.github.com/GunoH/ed48f53331da655658d640f07cd80932

**To Reproduce**
1. Clone https://github.com/GunoH/dependency-check-suppressions.
2. Run dependencyCheckAnalyze.
3. Have a look at the logs and the html report.
4. Add suppression using xml fragment from _suppress_ button.
5. Run dependencyCheckAnalyze again.

**Expected behavior**
* Vulnerability name concise instead of unbounded
*  The _suppress_ button gives me usable xml

chadlwilson:
## Description of Change

Currently the RetireJS support does not understand the newer `githubID` identifier type within the retireJS repository schema, so if there is no CVE ID known, it will use a generic identifier rather than a stable, global known GHSA ID as the vulnerability name.

Additionally if a `summary` identifier is present (and theres no other "better" id to use) it uses it as the vulnerability name even if it is some massive multi-line string, as is the case for many GHSAs where it is raw markdown text. `summary` is mainly intended as a description value, and is just intended as a fallback in RetireJS land.

This PR modernizes the RetireJS support to reflect the current RetireJS DB and usage:
  - <a href="https://raw.githubusercontent.com/Retirejs/retire.js/master/repository/jsrepository.json">Latest raw data </a>
  - <a href="https://github.com/RetireJS/retire.js/blob/700590ffc92f993dfe15af1b89e364b443bd9bfa/node/src/types.ts#L23-L37">TypeScript types for the identifiers</a>
  - <a href="https://github.com/RetireJS/retire.js/blob/700590ffc92f993dfe15af1b89e364b443bd9bfa/node/src/repo.ts#L29-L57">Repo validation</a>
  
### Behaviour changes when determining vulnerability name
- if a `githubID` is present, it is preferred to the secondary legacy project-specific identiifiers (`issue`, `PR`, `bug` etc)
- removes looking at `osvdb` identifiers. OSVDB is dead, and these IDs help no-one :-)
- `summary` identifiers are ignored if > 100 chars OR multi-line

Before:
```js
CVE 
  --> osvdb-id 
    --> indeterminateChoiceOf(project-issue, project-bug, project-PR) 
      --> summary 
        --> "Vulnerability in <library>"
```

After:
```js
CVE
  --> GHSA-id
    --> project-issue
      --> project-bug
        --> project-PR
          --> ifSensible(summary)
            --> "Vulnerability in <library>"
```

These changes mean folks suppressing via `vulnerabilityName` may need to update their suppressions. I think this is minor because the RetireJS DB is not particularly large, and the same happens already if RetireJS add/update IDs to existing vulns, due to the vulnerability naming preference logic. I checked our base suppressions and we don't seem to have any that refer to RetireJS.

## Related issues

- fixes #8402

yes


core/src/test/resources/javascript/angular.safe.js
chadlwilson, line 1: I removed and replaced all these massive files for RetireJS testing with stubs that match the RetireJS rules. Don't need the whole things checked in since RetireJS doesn't work off hashes, and makes it easier to search the code without getting matches inside these test resources.

jeremylong (approved):
LGTM

PR #8427, opened 2026-04-16
fix: bump open-vulnerability-clients to resolve NVD timestamp parsing errors

### LINKED ISSUE #8424: Dependency check fails with Error updating the NVD data due to Cannot deserialize value of type java.time.ZonedDateTime from String error
**Precondition**

**Describe the bug**
When running the `mvn dependency-check:aggregate` command within any repository, using version 12.2.0 or 12.2.1, I receive the following error:

```
org.owasp.dependencycheck.data.update.exception.UpdateException: Error updating the NVD Data
```

The problem occurs using version 12.2.0 and 12.2.1 of the maven plugin.

**Log file**
https://gist.github.com/commadelimited/1a4dcd5f0c99463fb69f102344a69544

**To Reproduce**
1. Visit a repository containing a pom.xml file which uses wither 12.2.0 or 12.2.1 of the maven dependency check plugin
2. Run `mvn dependency-check:aggregate`
3. Receive error

**Expected behavior**
I expect this aggregate check to complete.

**Additional context**
I was able to run this command just a few hours prior and it finished successfully.
Another coworker on a separate machine, in a different state, is also experiencing the same behavior.

chadlwilson:
## Description of Change

Bumps clients to include fix for https://github.com/jeremylong/open-vulnerability-clients/issues/106 / https://github.com/jeremylong/open-vulnerability-clients/issues/107

Also bumps logback. We must have a dependabot ignore rule that needs resetting.

## Related issues
- fixes #8424

Yes

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -912,7 +912,7 @@ Copyright (c) 2012 - Jeremy Long
             <dependency>
                 <groupId>io.github.jeremylong</groupId>
                 <artifactId>open-vulnerability-clients</artifactId>
-                <version>9.0.4</version>
+                <version>9.0.5</version>
             </dependency>
             <dependency>
                 <groupId>org.anarres.jdiagnostics</groupId>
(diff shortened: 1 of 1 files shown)


jeremylong (approved):
LGTM

Issue #8423, opened 2026-04-15
Remove YarnAuditAnalyzer (classic/v1 mode) dependency on NPM audit /quick API


chadlwilson:
**Is your feature request related to a problem? Please describe.**

The YarnAuditAnalyzer when running in Yarn v1/classic mode currently relies on the NPM /quick security audit API. This has had some recent stability problems, but as of 15 April 2026 it has started largely returning HTTP 410 errors with `This endpoint is being retired. Use the bulk advisory endpoint instead. See the following docs for more info: https://api-docs.npmjs.com/#tag/Audit`

At time of writing, there appears to be nothing official on GitHub/npm blog : https://github.blog/tag/npm/

**Describe the solution you'd like**

We'll have to evaluate using the bulk advisories endpoint, or whether we should be directly integrating at all for Yarn v1.

https://api-docs.npmjs.com/#tag/Audit/operation/bulkAudit

Yarn v1 CLI `yarn audit` is likely completely broken now and seems unlikely to be updated. It may be we should drop yarn v1 support entirely. 

https://endoflife.date/yarn
> Yarn Classic (v1) [entered maintenance mode in January 2020](https://dev.to/arcanis/introducing-yarn-2-4eh1#what-will-happen-to-the-legacy-codebase) and will eventually reach end-of-life.
> Yarn Classic only receives critical and security fixes.

**Additional context**

- PNPM upstream needs to migrate: #8421 
- Node Audit Analyzer needs to be migrated (and may be able to share code if we choose to support this): #8422
- Yarn Audit Analyzer (in classic mode) would also need to migrate
- Yarn Audit Analyzer (in berry mode) seems OK as Yarn already uses the bulk API

If the `410`s continue, in the meantime there may be no choice but for users to disable the analyzer. We may need to add config to support disabling for v1/Classic without disabling for v2+/berry, however - if are executing scans for projects that contain both types of projects.

dpfrederick:
Is there any idea of what a timeline could be for shipping a fix to use the bulk advisory endpoint? We have yarn1 and yarn4 projects so ideally a fix here would apply to both analyzers but I understand it it's not worth it for yall to update the old yarn analyzer.

chadlwilson:
No timeline.

There's not even an announcement from GitHub/npm (a commercial operation from one of the richest companies in the world) that this either will happen, or has happened - so I certainly wouldn't expect any commitments from a small open source project no-one pays for (*).

[...]

PR #8428, opened 2026-04-16
Remove legacy NodeAuditAnalyzer dependency on NPM audit v1 API

### LINKED ISSUE #8422: Remove legacy NodeAuditAnalyzer dependency on NPM audit legacy /quick API

The NodeAuditAnalyzer currently relies on the NPM v1 security audit API. This has had some recent stability problems, but as of 15 April 2026 it has started largely returning HTTP 410 errors with `This endpoint is being retired. Use the bulk advisory endpoint instead. See the following docs for more info: https://api-docs.npmjs.com/#tag/Audit`

At time of writing, there appears to be nothing official on GitHub/npm blog : https://github.blog/tag/npm/

We'll have to evaluate whether we should use the bulk advisories endpoint, or whether we should be directly integrating at all anymore. https://api-docs.npmjs.com/#tag/Audit/operation/bulkAudit

I suspect we should just drop the code entirely though, and use `npm` CLI directly like the `pnpm` analyzer does.

**Additional context**

- PNPM upstream needs to migrate: #8421 
- Yarn Audit Analyzer (in classic mode) would also need to migrate: #8423
- Yarn Audit Analyzer (in berry mode) seems OK as Yarn already uses the bulk API

marcelstoer:
## Description of Change

Stop going to https://registry.npmjs.org/-/npm/v1/security/audits and parse the audit locally instead for Yarn classic (v1, non-berry).

## Related issues

Fixes #8422

yes

### DIFF
--- a/core/src/main/java/org/owasp/dependencycheck/analyzer/YarnAuditAnalyzer.java
+++ b/core/src/main/java/org/owasp/dependencycheck/analyzer/YarnAuditAnalyzer.java
@@ -19,21 +19,21 @@ package org.owasp.dependencycheck.analyzer;
 
 import org.apache.commons.collections4.MultiValuedMap;
 import org.apache.commons.collections4.multimap.HashSetValuedHashMap;
-import org.apache.commons.io.IOUtils;
 import org.apache.commons.lang3.StringUtils;
 import org.json.JSONException;
 import org.json.JSONObject;
+import org.jspecify.annotations.NonNull;
 import org.owasp.dependencycheck.Engine;
 import org.owasp.dependencycheck.analyzer.exception.AnalysisException;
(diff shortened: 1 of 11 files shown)


chadlwilson:
I don’t understand this. You can’t use npm CLI to interact with a yarn project?

marcelstoer:
🙈 I kept storming forward with my head down on a path that leads nowhere. Hasn't happened in quite a while that I ignored "if you want to be fast, you have to go slow". I'll need to start over.

chadlwilson:
Well, that might be the right path for NodeAuditAnalyzer for the NPM projects?

And we can have a separate conversation in #8423 about whether we should support Yarn v1 at all, so just drop the yarn related changes from here for now?

I haven't actually looked at the rest, just confused by the Yarn stuff.

PR #8479, opened 2026-05-05
fix: only use packaged hosted suppression file if using the default URL

### LINKED ISSUE #8478: Hosted suppressions no longer respects force update setting since 12.2.0
**Precondition**

**Describe the bug**
We use a custom hosted suppressions file via `hostedSuppressionsUrl`, which worked fine until version 12.1.9.

Since 12.2.0, our own suppressions are not taken into account anymore. dependency-check actually downloads/creates a suppressions file in `~/.m2/repository/org/owasp/dependency-check-data/11.0/`  with the correct filename but the file lacks all our custom suppressions.

The problem occurs using version 12.2.0, 12.2.1 and 12.2.2 of the maven plugin.

**To Reproduce**
Create your own hosted suppressions file e.g. at /tmp/test-suppressions.xml with this content:
```
<suppressions xmlns="https://jeremylong.github.io/DependencyCheck/dependency-suppression.1.3.xsd"
              xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
              xsi:schemaLocation="https://jeremylong.github.io/DependencyCheck/dependency-suppression.1.3.xsd
                                  https://dependency-check.github.io/DependencyCheck/dependency-suppression.1.3.xsd">
<suppress base="true">
<notes>Test suppression</notes>
<cve>CVE-2026-00000</cve>
</suppress>
</suppressions>
```

Run:
```
mvn org.owasp:dependency-check-maven:12.2.0:aggregate -DhostedSuppressionsUrl=file:///tmp/test-suppressions.xml
```

Check the content of `~/.m2/repository/org/owasp/dependency-check-data/11.0/test-suppressions.xml. It should contain the above suppression, e.g. `CVE-2026-00000`, but it will not.

**Expected behavior**
The custom suppression should be used.

jeremylong:
resolves #8478

### DIFF
--- a/core/src/main/java/org/owasp/dependencycheck/analyzer/AbstractSuppressionAnalyzer.java
+++ b/core/src/main/java/org/owasp/dependencycheck/analyzer/AbstractSuppressionAnalyzer.java
@@ -267,8 +267,8 @@ public abstract class AbstractSuppressionAnalyzer extends AbstractAnalyzer {
             }
             final File repoFile = new File(getSettings().getDataDirectory(), fileName);
             boolean repoEmpty = !repoFile.isFile() || repoFile.length() <= 1L;
-            if (repoEmpty) {
-                // utilize the snapshot hosted suppression file
+            if (repoEmpty && HostedSuppressionsDataSource.DEFAULT_SUPPRESSIONS_URL.equals(configuredUrl)) {
+                // utilize the snapshot hosted suppression file only for the default URL
                 URL hostedSuppressionSnapshotURL = getPackagedFile(HOSTED_SUPPRESSION_SNAPSHOT_FILE);
                 try (InputStream in = hostedSuppressionSnapshotURL.openStream()) {
                     Files.copy(in, repoFile.toPath(), StandardCopyOption.REPLACE_EXISTING);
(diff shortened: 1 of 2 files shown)


chadlwilson:
Can you replicate the issue as described? I cannot....

This doesn't seem entirely desirable to me - but maybe it highlights a different issue. Won't it disable all of the suppressions which we migrated from base -> hosted on the assumption that they will be included in snapshots? These would lead to people reporting false positives whenever they use a custom URL (even if it's a proxy) because they are now missing a chunk of suppressions that are actually distributed with ODC but not inside the "legacy" base suppressions file.

This will be very difficult to reason about for support purposes. When reviewing i expect people to have at least base + last-release-hosted so I can ask people to upgrade or update their hsited suppressions.

In my opinion, the logic to use the hosted suppression snapshot file should not be guarded behind `hosted.suppressions.enabled` either - in my view these are essentially a second set of base suppressions that happen to be from a different file and I don't think its good that users can override/remove them due to the reasons above. ODC doesn't allow people to disable "normal" base suppressions does it?

I think users should only be able to disable the remote sourcing of suppressions, not the use of the snapshotted suppressions, since these are no different to our maintained base suppressions which are "part of the product".

chadlwilson:
Additionally, is it actually the goal of setting a custom hosted suppression url to override/remove a whole chunk of our curated suppressions? It seems to me it's supposed to be for pass through proxies or caches of our curated ones, not for what seems to be done in #8478 as a way to source extra user suppressions from a remote URL?

jeremylong:
I've been testing several things with this - I created this PR too quickly while I was still testing.  Closing this now as things appear to be working as expected.

chadlwilson:
Ahh ok. It’s possible there is something weird happening here in certain sequence of events, I just can’t quite recreate the problem as described right now.

If you agree that we should generally “always include/consider/load the snapshot suppressions if the file can be loaded from within in the jar, as per the legacy/base suppressions” I might look at a change to alter this a bit so the “enabled” flag only guards the remote retrieval and the two sources are dealt with more independently so configuration for the remote data source (enabled, url etc) doesn’t interact with the snapshotted version.

If possible. 😅

cpfeiffer:
To give a bit of context, we're using a hosted suppressions file in order to quickly fix false positives among many similar projects. We pull your curated publishedSuppressions.xml and extend that with our own suppressions.

jeremylong:
you don't need to use the hosted suppression file. just configure a normal suppression file.

PR #8484, opened 2026-05-06
fix(analyzer): detect OpenSSL 3.x in OpenSSLAnalyzer

shawnjohnjr:
## Summary
- `OpenSSLAnalyzer` cannot detect OpenSSL 3.x sources today: 3.x defines `OPENSSL_VERSION_NUMBER` as a macro expression (not a literal hex constant), the existing regex never matches, and the dependency is dropped via `engine.removeDependency()`. As a result a scan over an OpenSSL 3.x source tree produces no CPE and never matches NVD CVEs.
- This adds a second pattern that captures `OPENSSL_VERSION_STR "X.Y.Z"` (introduced in OpenSSL 3.0) and tries it before falling back to the legacy `OPENSSL_VERSION_NUMBER` hex pattern, so 1.x detection is preserved.
- Adds a test fixture (`core/src/test/resources/openssl-3x/opensslv.h` from OpenSSL 3.5.6) and a parallel test case asserting `3.5.6` is extracted.

## Reproducer (before this change)
Scan an OpenSSL 3.5.6 source tree:
```
dependency-check --scan path/to/openssl-3.5.6 --enableExperimental --format JSON ...
```
The resulting JSON has no dependency for `opensslv.h` and no `cpe:2.3:a:openssl:openssl:3.5.6:*` identifier. After this change the same scan produces:
```
filePath: .../include/openssl/opensslv.h
versionEvidence.Version Constant = 3.5.6   [HIGH]
softwareIdentifiers: [pkg:generic/openssl@3.5.6]
vulnerabilityIds: [cpe:2.3:a:openssl:openssl:3.5.6:*:*:*:*:*:*:*]
```

## Test plan
- [x] `mvn -pl core -am test -Dtest=OpenSSLAnalyzerTest -Dsurefire.failIfNoSpecifiedTests=false` — 5/5 pass (existing 4 + new `testOpenSSL3xVersionHeaderFile`)
- [x] End-to-end CLI scan against an OpenSSL 3.5.6 source tree confirms `cpe:2.3:a:openssl:openssl:3.5.6` is now attached to the `opensslv.h` dependency


chadlwilson:
This analyzer is super curious to me given how specific it is. I wonder how many people use it - are you a user? 😅

shawnjohnjr:
> This analyzer is super curious to me given how specific it is. I wonder how many people use it - are you a user? 😅

Yes. 💯

jeremylong (approved):

PR #8496, opened 2026-05-11
fix(cli): correct displayed application name and deprecations in help text

chadlwilson:
## Description of Change

Currently the help text tends to display `usage: Dependency-Check Core` from the default `Settings.KEYS.APPLICATION_NAME` which is mildly confusing. Generally the actual entry point is `dependency-check` or `dependency-check.sh`.

The Application Name was overridden lateish in the process from env for telemetry purposes; but it is actually used earlier so moved the value setting earlier. In any case, the help/usage should probably indicate something approximating the script/executable name. I haven't gone so far as to detect the _actual_ name from shell script; nor to disambiguate between Homebrew variant without `.sh`, or running within Docker - so this is "better but not perfect".

Also migrated to the newer non-deprecated `HelpFormatter`. it looks basically the same as the old one:

```
$ cli/target/release/bin/dependency-check.sh
 usage:  dependency-check [--advancedHelp] [--disableVersionCheck] [--enableExperimental] [--exclude
 <pattern>] [-f <format>] [--failOnCVSS <score>] [-h] [--junitFailOnCVSS <score>] [-l <file>] [-n]
    [--nvdApiKey <apiKey>] [-o <path>] [--prettyPrint] [--project <name>] [-s <path>] [--suppression
    <file>] [-v]

 Dependency-Check can be used to identify if there are any known CVE vulnerabilities in libraries
    utilized by an application. Dependency-Check will automatically update required data from the
    Internet, such as the CVE and CPE data files from nvd.nist.gov.

          Options                                           Description
 --advancedHelp                Print the advanced help message.
 --disableVersionCheck         Disables the dependency-check version check
 --enableExperimental          Enables the experimental analyzers.
 --exclude <pattern>           Specify an exclusion pattern. This option can be specified multiple
                                times and it accepts Ant style exclusions.
 -f, --format <format>         The report format (HTML, XML, CSV, JSON, JUNIT, SARIF, JENKINS, GITLAB
                                or ALL). The default is HTML. Multiple format parameters can be
                                specified.
 --failOnCVSS <score>          Specifies if the build should be failed if a CVSS score above a
                                specified level is identified. The default is 11; since the CVSS
                                scores are 0-10, by default the build will never fail.
 -h, --help                    Print this message.
 --junitFailOnCVSS <score>     Specifies the CVSS score that is considered a failure when generating
                                the junit report. The default is 0.
 -l, --log <file>              The file path to write verbose logging information.
 -n, --noupdate                Disables the automatic updating of the NVD-CVE, hosted-suppressions
                                and RetireJS data.
 --nvdApiKey <apiKey>          The API Key to access the NVD API.
 -o, --out <path>              The folder to write reports to. This defaults to the current
                                directory. It is possible to set this to a specific file name if the
                                format argument is not set to ALL.
 --prettyPrint                 When specified the JSON and XML report formats will be pretty printed.
 --project <name>              The name of the project being scanned.
 -s, --scan <path>             The path to scan - this option can be specified multiple times. Ant
                                style paths are supported (e.g. 'path/**/*.jar'); if using Ant style
                                paths it is highly recommended to quote the argument value.
 --suppression <file>          The file path to the suppression XML file. This can be specified more
                                then once to utilize multiple suppression files
 -v, --version                 Print the version information.
 ```

## Related issues

N/A

yes

### DIFF
--- a/cli/src/main/java/org/owasp/dependencycheck/App.java
+++ b/cli/src/main/java/org/owasp/dependencycheck/App.java
@@ -130,11 +130,11 @@ public class App {
             cli.parse(args);
         } catch (FileNotFoundException ex) {
             System.err.println(ex.getMessage());
-            cli.printHelp();
+            cli.printHelp(System.out);
             return 1;
         } catch (ParseException ex) {
             System.err.println(ex.getMessage());
-            cli.printHelp();
+            cli.printHelp(System.out);
             return 2;
         }
         final String verboseLog = cli.getStringArgument(CliParser.ARGUMENT.VERBOSE_LOG);
@@ -234,7 +234,7 @@ public class App {
                 settings.cleanup();
             }
         } else {
-            cli.printHelp();
+            cli.printHelp(System.out);
         }
         return exitCode;
     }
--- a/cli/src/main/java/org/owasp/dependencycheck/App.java
+++ b/cli/src/main/java/org/owasp/dependencycheck/App.java
@@ -24,6 +24,7 @@ import ch.qos.logback.classic.filter.ThresholdFilter;
 import ch.qos.logback.classic.spi.ILoggingEvent;
 import ch.qos.logback.core.FileAppender;
 import org.apache.commons.cli.ParseException;
+import org.apache.commons.lang3.StringUtils;
 import org.apache.tools.ant.DirectoryScanner;
 import org.apache.tools.ant.types.LogLevel;
 import org.owasp.dependencycheck.data.nvdcve.DatabaseException;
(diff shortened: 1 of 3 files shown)


marcelstoer (approved):
LGTM

chadlwilson:
I forgot to mention that I increased the width to 100 (from 74). Doesn't seem to have ability to adjust with shell, and 74 makes the advanced help more difficult to make sense of than it already is 😅

I retained the "sort by key" that was there in the deprecated formatter to make diff easier, but we could change that, or change to retain the native ordering with the new formatter. The "disable" options end up not grouped next to their analyzers, and the inconsistent naming of some options makes it sort a bit weirdly.

jeremylong (approved):

PR #8509, opened 2026-05-18
chore(fp): use more conservative CPE 22 prefix suppression syntax to avoid false negatives

chadlwilson:
## Description of Change

CPE 2.2 URI suppressions are string prefix based off a fully formed URI; without the colon we risk matching against the wrong product, sometimes sub-products.

https://github.com/dependency-check/DependencyCheck/blob/c82a59c28f2e275c51464cbc0b591d1c45fa5427/core/src/main/java/org/owasp/dependencycheck/xml/suppression/SuppressionRule.java#L778-L786

It's possible some of these are intentional product prefixes; but probably better to address these one-by-one if so. The current ones risk creating false negatives, especially over time as new sub-products are added.

## Related issues

- #8510 for the hosted/generated suppressions
- tests failing due to unrelated issue https://github.com/dependency-check/DependencyCheck/pull/8511

N/A


chadlwilson:
Thank for dealing with the conflicts for me, you didn't have to do that, so it's appreciated 🙏🏻

jeremylong (approved):

PR #8516, opened 2026-05-19
fix: publish-suppressions action trigger

### LINKED ISSUE #8512: Improve publish-suppressions action trigger

The [Publish Suppressions action](https://github.com/dependency-check/DependencyCheck/blob/5b8a493b4d78c524005d907af72f774cacc3eda5/.github/workflows/publish-suppressions.yml) is not automatically triggered when integrating code on the `generatedSuppressions` branch.This requires an additional manual step when integrating a change on this branch to be rolled out to Dependency-Check users.

The Publish Suppressions action is triggered automatically when a push is happening on the `generatedSuppressions` branch.

Docs: [Running your workflow only when a push to specific branches occurs](https://docs.github.com/en/actions/reference/workflows-and-actions/events-that-trigger-workflows#running-your-workflow-only-when-a-push-to-specific-branches-occurs)

Keep triggering the workflow manually

**Additional context**

- https://github.com/dependency-check/DependencyCheck/pull/8504#issuecomment-4471175486
- https://github.com/dependency-check/DependencyCheck/pull/8499#issuecomment-4472596189

jeremylong:
resolves #8512

### DIFF
--- a/.github/workflows/publish-suppressions.yml
+++ b/.github/workflows/publish-suppressions.yml
@@ -2,6 +2,9 @@ name: Publish Suppressions
 
 on:
   workflow_dispatch:
+  push:
+    branches:
+      - generatedSuppressions
 
 permissions: {}
 jobs:


chadlwilson:
I think this will double-trigger/double-publish since the approvals workflow pushes to the branch and then triggers publish directly?

I was part way through unpicking that, but was a bit confused about the logic 😅

chadlwilson:
Hmm, it didn't seem to double-trigger after an FP approval just now, so perhaps GH have some protection in place for internal actions pushes. Maybe my fears were misplaced 👍🏻

chadlwilson:
Unfortunately this doesn’t seem to be working - merge of https://github.com/dependency-check/DependencyCheck/pull/8522 didn’t trigger anything. That’s probably also why it didn’t double trigger.

https://github.com/dependency-check/DependencyCheck/actions?query=workflow%3A%22Publish%20Suppressions%22

I suspect this is because `push` events run within, and use workflow definition of, the relevant branch’s context, yet the `generatedSuppressions` branch is not a proper branch, and has no workflows. Workflow dispatch makes you select the context, which is why it works there.

I don’t think it’s a good idea to merge the workflows over to that branch as they will potentially become out of date, or hard to maintain. I’ll have a think about how we might be able to do this maintainable with minimal changes?

Issue #8521, opened 2026-05-20
[FP]: isorelax (via msv group) being misattributed as Nu HTML Checker/validator


chadlwilson:
### Package URl

pkg:maven/msv/isorelax@20050913

### CPE

`cpe:2.3:a:validator:validator:*:*:*:*:*:*:*:*`

### CVE

CVE-2025-15104

### ODC Integration

{"label" => "Gradle Plugin"}

### ODC Version

12.2.2

### Description

Same as https://github.com/dependency-check/DependencyCheck/issues/8243 - different package groupId

Root cause is the same SHA1 being connected to multiple versions and groupIDs via Maven Central: https://search.maven.org/search?q=1:b21859c352bd959ea22d06b2fe8c93b2e24531b9

<img width="1213" height="262" alt="Image" src="https://github.com/user-attachments/assets/b47ade73-4672-4ecc-b79f-ab9dbd7509f0" />

I am not sure why this wasn't showing up as an FP earlier though as nothing seems to have changed in NVD at least. Possibly Maven Central re-indexed their search or something.

chadlwilson:
approved

chadlwilson:
This might need another look; as it should have been covered by https://github.com/dependency-check/DependencyCheck/pull/8251 I think.

chadlwilson:
This is a regression due to https://github.com/dependency-check/DependencyCheck/pull/8510 - for CVEs like https://nvd.nist.gov/vuln/detail/CVE-2025-15104 that don't have a version in the "affected versions", the URI once normalised becomes `cpe:/a:validator:validator` due to logic in the `CpeParser` to remove trailing `:`:

```java
return sb.toString().replaceAll("[:]*$", "");
```

Will revert or workaround on hosted suppressions pending a fix, I think.

PR #8522, opened 2026-05-20
fix(fp): convert hosted/generated suppressions to conservative regex prefix matchers

chadlwilson:
## Description of Change

As discovered myself in https://github.com/dependency-check/DependencyCheck/issues/8521#issuecomment-4494622767 unfortunately my change in #8510 caused some unintentional regression of suppression applicability for hosted suppressions, since the trailing bits of the CPE are removed for vulnerabilities without versions in them (typically ones with startsWith/endsWith ranges at NVD such as https://nvd.nist.gov/vuln/detail/CVE-2025-15104) 
 
image
 
The existing code does not normalise the suppression rule CPE URI from the suppressions before comparing the string representation prefix. This means vulns with affected versions like `cpe:2.3:a:validator:validator:*:*:*:*:*:*` will no longer match as they are correctly normalized by the code to `cpe:/a:validator:validator` without a trailing `:` before matching to the prefix rule. I'd forgotten about this CPE 2.2 vs 2.3 difference :-(

This change converts them to regexes - so it's possible to express the matching across ODC versions. An alternative would be to just revert things and leave the false negative possibilities.
- It's probably not great (complexity, readability, performance) to have to do regex matching everywhere here, however since hosted suppressions are supported across older versions, the alternative would leave us stuck without consistent conservative matching, as requires code change to fix.
- Happy to take feedback though - WDYT @jeremylong @marcelstoer?
- I'll follow up with a "fix" of some sort on master as well once we've agreed approach; to improve the non-regex matching; and/or to improve the regex matching speed.

## Related issues

- caused #8521
- caused by #8510 

no

### DIFF
--- a/generatedSuppressions.xml
+++ b/generatedSuppressions.xml
@@ -3,330 +3,330 @@
     FP per issue #4806
     ]]></notes>
     <packageUrl regex="true">^pkg:maven/org\.vaadin\.addon/easyuploads@.*$</packageUrl>
-    <cpe>cpe:/a:vaadin:vaadin:</cpe>
+    <cpe regex="true">cpe:/a:vaadin:vaadin(:.*)?$</cpe>
 </suppress>
 <suppress base="true">
     <notes><![CDATA[
     FP per issue #4790
     ]]></notes>
     <packageUrl regex="true">^pkg:maven/ch\.qos\.logback/logback-classic@.*$</packageUrl>
-    <cpe>cpe:/a:qos:slf4j:</cpe>
+    <cpe regex="true">cpe:/a:qos:slf4j(:.*)?$</cpe>
 </suppress>
 <suppress base="true">
     <notes><![CDATA[
     FP per issue #4781
     ]]></notes>
     <packageUrl regex="true">^pkg:maven/org\.eclipse\.microprofile\.config/microprofile-config-api@.*$</packageUrl>
-    <cpe>cpe:/a:payara:payara:</cpe>
+    <cpe regex="true">cpe:/a:payara:payara(:.*)?$</cpe>
 </suppress>
 <suppress base="true">
     <notes><![CDATA[
     FP per issue #4777
     ]]></notes>
     <packageUrl regex="true">^pkg:maven/org\.apache\.james/apache-mime4j@.*$</packageUrl>
-    <cpe>cpe:/a:apache:james:</cpe>
+    <cpe regex="true">cpe:/a:apache:james(:.*)?$</cpe>
 </suppress>
 <suppress base="true">
     <notes><![CDATA[
     FP per issue #4755
     ]]></notes>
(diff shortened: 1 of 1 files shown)


chadlwilson:
Sorry forgot to tag @nhumblot for feedback 🙏🏻

We might get a swathe of FP reports today because of my mistake here. I'll handle them if so.

jeremylong (approved):
LGTM

chadlwilson:
Sorry for the hassle. 😓

chadlwilson:
@jeremylong unfortunately https://github.com/dependency-check/DependencyCheck/pull/8516 didn’t seem to lead to the publish workflow being triggered.

https://github.com/dependency-check/DependencyCheck/actions?query=workflow%3A%22Publish%20Suppressions%22

would you be able to trigger the workflow manually for now so we can get the fix out?

ArkanEPITA:
Hello,

Since the new way of handling the packageUrl with regex, the matching seems to be incrorrect. The package "pkg:maven/org.hibernate.validator/hibernate-validator@8.0.2.Final" should be matched by the following suppression in the generatedSuppression file: 
```
<suppress base="true">
    <notes><![CDATA[
    FP per issue #8243, #8244, #8247, #8249
    ]]></notes>
    <packageUrl regex="true">^pkg:(?!maven/nu\.validator/validator@|npm/vnu-jar)(:.*)?$</packageUrl>
    <cpe regex="true">cpe:/a:validator:validator(:.*)?$</cpe>
</suppress>
```
Yet, the CVE-2025-15104 still fails the check.

The packageUrl regex changed from `^pkg:(?!maven/nu\.validator/validator@|npm/vnu-jar).*` to `^pkg(?!:maven\/nu\.validator\/validator@|npm\/vnu-jar)(:.*)?$` [here](https://github.com/dependency-check/DependencyCheck/pull/8522/changes/43d53e91e6c094fb45a3e8bf66bbc931bd853fe8#diff-2299dc44c2e9f7e35eaf3774daea64f1efe2426f4c7cdc482b6e0c57b8ba80bbR1915)

Other package url doesnt seem to be touched (I didnt check each of them though).

Reading the comment on this issue and the linked one, I don't know why pkgUrl matching changed as the topic seems to be around cpe matching. Is it an oversight ?

+ AFAIK, the package URL spec doesnt specify multiple "semi-colon followed by char" sequences and only a semi-colon after the pkg scheme. So I don't know what this commit tried to achieve there.

chadlwilson:
See #8525 for the fix (already linked above) - apologies for the hassle.

chadlwilson:
None of these suppression rules have any automated tests whatsoever, so occasionally screw ups will happen, unfortunately (including systemic issues such as the widespread false negatives that this PR and #8510 are attempting to fix for the `<cpe>` matchers)

chadlwilson:
Copilot would probably have caught this, but I forgot to ask it for review. Oops.

ArkanEPITA:
Right ! Have not seen that, thanks a lot :)

PR #8530, opened 2026-05-20
fix: allow prefix matchers for CPE 2.2 URIs to strictly match a whole part only

chadlwilson:


chadlwilson:
I intended to open this against my fork, as it's still WIP, but no harm here for now.

PR #8531, opened 2026-05-21
ci: fix FP ops workflow to generate CPE regex for optional suffix matches

chadlwilson:
## Description of Change

As changed already for the existing suppressions in #8522; converts the workflow to generate regexes for CPE matches, to reduce false negatives due to matching only partial CPE `product` fields.

The previous strategy with trailing `:` doesn't work, because current ODC prefix matches will fail to match any CPE identifiers that use "up to/starts with" type constraints and thus typically have no values in `version`/`target_sw` etc etc fields. In CPE 2.2 URIs these get truncated, so naive string prefix matches became indeterminate.

I'll introduce support to do stricter CPE part matches in a PR, but we won't be able to use that for hosted suppressions unless we make some breaking change/force update at some point, so need to keep using regex for now.

## Related issues

- relates to #8522

no

### DIFF
--- a/.github/workflows/false-positive-ops.yml
+++ b/.github/workflows/false-positive-ops.yml
@@ -171,12 +171,9 @@ jobs:
             }
             purl += '@.*$';
             var cpe = process.env.CPE.trim().replaceAll(/^`|`$/g,'').split(':');
-            var cpe22UriPrefix;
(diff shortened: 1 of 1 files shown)


jeremylong (approved):
lgtm

Issue #8497, opened 2026-05-12
[FP]: CVE-2026-42154 matched against Micrometer instead of Prometheus


marcelstoer:
### Package URl

pkg:maven/io.micrometer/micrometer-registry-prometheus@1.15.11

### CPE

cpe:2.3:a:prometheus:prometheus:1.15.11:*:*:*:*:*:*:*

### CVE

CVE-2026-42154

### ODC Integration

{"label" => "Maven Plugin"}

### ODC Version

12.2.1

chadlwilson:
Approved 

We probably need to consolidate suppressions here as there are dozens of these, but will add this for now.

marcelstoer:
> dozens of these

Which ones do you refer to? We currently have 7 for `cpe:/a:prometheus:prometheus`.

chadlwilson:
There are a whole lot across base suppressions and hosted suppressions, incl some for individual CVEs. And there are many more Maven libraries that are likely incorrectly matching - just not appearing as FPs just yet due to luck.

Can't recall if I've done so already, but with some research into whether `prometheus:prometheus` is only ever used for the server component - the CPE could possibly be suppressed across all non-golang ecosystems. It's possible it's indeterminate though, as is often the case when the main artifact is the server and peripheral language artifacts are clients or API types.

chadlwilson:
Based on https://nvd.nist.gov/vuln/search#/nvd/home?cpeFilterMode=applicability&cpeName=cpe:2.3:a:prometheus:prometheus:*:*:*:*:*:*:*:*&resultType=records the CVEs are only for the server it's probably safe to do something more generic; noting that CVEs like https://nvd.nist.gov/vuln/detail/CVE-2022-21698 have their own CPE.

joeha11:
Newbie question: I'm still getting CVE-2026-42154 reported against  `pkg:maven/io.micrometer/micrometer-registry-prometheus@1.16.5`, even on version 12.2.2 of the maven plugin. 
Is there anything else I need to do as a user?

chadlwilson:
@joeha11 Thanks for raising. There has been a regression from some subsequent changes (intended to both reduce more false positives as well as some false negatives) made by me in #8522 . Sorry about this.

Should be resolved after https://github.com/dependency-check/DependencyCheck/pull/8542 is merged.

chadlwilson:
This should be fixed now; apologies for the hassle.

PR #8542, opened 2026-05-22
fix(fp): correct hosted suppression CPE matching for regexes which previously had trailing colons

chadlwilson:
## Description of Change

Fixes yet another regression from #8522 after the regex conversion - regexes which already had a trailing `:` in their rule were failing to match some CPE rules from NVD because the new optional suffix was added regardless; requiring it to end in `::.*` to match (two colons), which would no longer match anything.

Sorry @nhumblot 😅

## Related issues

- relates to #8522
- relates to #8497

no

### DIFF
--- a/generatedSuppressions.xml
+++ b/generatedSuppressions.xml
@@ -1196,7 +1196,7 @@
     FP per issue #7043
     ]]></notes>
     <packageUrl regex="true">^pkg:maven/org\.apache\.ftpserver/ftplet-api@.*$</packageUrl>
-    <cpe regex="true">cpe:/a:apache:mina:(:.*)?$</cpe>
+    <cpe regex="true">cpe:/a:apache:mina(:.*)?$</cpe>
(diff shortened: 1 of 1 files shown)


marcelstoer (approved):

chadlwilson:
I don't fully understand how these suppressions were working before for some of the vulns to be honest. An equivalent suppression for different libs using non-regex was failing to match for some CVEs when it had a trailing `:` after normalisation of the CPE 2.2 URI.

Possibly some quirk in how VulnerableSoftwareIdentifiers are matched in the code and whether they end up with the version in the CPEIdentifier when matched (or not). Perhaps it depends on whether the version in use by the user actually has a CPE entry defined for that specific version in NVD or not, or it's just matching based on a `*` and ODC's logic.

Probably need to debug or look at the raw DB data coming through for the vulns....

jeremylong (approved):

PR #8548, opened 2026-05-25
fix(fp): allow stricter matching of suppression CPE 2.2 URI prefixes to only whole parts

chadlwilson:
## Description of Change

Follows up on #8509 to allow rules with a trailing `:` to indicate a strict match of the last part is intended; rather than potentially matching only a substring of the part (usually an application name). Without this, we'd also need to correct all of the base suppressions to regex, like done in #8522.

e.g.
https://github.com/dependency-check/DependencyCheck/blob/c5f5f681eedab927ff3cad2ed5c229cbce02b206/core/src/main/resources/dependencycheck-base-suppression.xml#L13-L19

Main goal as with #8509 is to reduce false negatives from other products with the same "prefiix" in the name.

Some `CpeIdentifiers` don't include the version at all; making the changes in #8509 fail to suppress some `VulnerableSoftwareIdentifier`s. 

This change 
- allows us to avoid needing regex
- when using regex, memoize the pattern to avoid O(n^2) regex compiling during matching.
  - makes `PropertyType` immutable to do so 

I can split up the regex stuff if preferred; I kept them together since they both touch the same tests.

## Related issues

- required by #8509
- relates to #8522

yes

### DIFF
--- a/core/src/main/java/org/owasp/dependencycheck/xml/hints/HintHandler.java
+++ b/core/src/main/java/org/owasp/dependencycheck/xml/hints/HintHandler.java
@@ -17,15 +17,16 @@
  */
 package org.owasp.dependencycheck.xml.hints;
 
-import java.util.ArrayList;
-import java.util.List;
-import javax.annotation.concurrent.NotThreadSafe;
 import org.owasp.dependencycheck.dependency.Confidence;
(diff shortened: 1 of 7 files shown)


jeremylong (approved):
LGTM

jeremylong (approved):

Issue #1693, opened 2019-01-18
Document multi-module configuration with suppression file


sworisbreathing:
I had the same issue as reported in #711. I tried @jeremylong's suggestion of following the approach recommended by maven-checkstyle-plugin. It worked for `mvn verify` but failed on `mvn -B dependency-check:update-only`

```
[ERROR] No plugin found for prefix 'dependency-check' in the current project and in the plugin groups [org.apache.maven.plugins, org.codehaus.mojo] available from the repositories [local (/path/to/user/dir/.m2/repository), central (https://repo1.maven.org/maven2)] -> [Help 1]
```

It'd be awesome if the proper approach for handling multi-module configuration with a suppression file was documented on the usage page.

anand1416:
Hi,

I'd like to work on this documentation issue if it's still available.

I plan to review the current suppression file documentation and Maven multi-module configuration, then propose documentation updates with examples for multi-module projects.

Please let me know if there are any preferred locations or guidelines for the documentation update.

Thanks!

PR #8564, opened 2026-06-01
docs(maven): clarify suppression file usage for multi-module projects

anand1416:
## Description of Change

Added documentation clarifying how suppression files can be used in multi-module Maven builds.

The update explains that suppression files may be configured in a parent POM and shared across child modules, and highlights the importance of configuring file paths so they resolve correctly for all participating modules.

## Related Issues

Related to #1693

No - documentation update only.

### DIFF
--- a/maven/src/site/markdown/configuration.md
+++ b/maven/src/site/markdown/configuration.md
@@ -42,11 +42,6 @@ The following properties can be set on the dependency-check-maven plugin.
 | enableRetired                    | Enable the [retired analyzers](../analyzers/index.html). If not enabled the retired analyzers (see below) will not be loaded or used.                                                                                                                                                                                                                                                                                                                                                                                                                                                                     | false                                                                                                                                                       |
 | versionCheckEnabled              | Whether dependency-check should check if a new version of dependency-check-maven exists.                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                  | true                                                                                                                                                        |
 
-### Multi-module Maven Projects
-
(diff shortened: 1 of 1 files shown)


chadlwilson (changes requested):
This doesn't render properly (breaks the heading below it). Did you run `mvn site:stage` and look at what was generated?

The location of the suggested doc fragment doesn't make sense to me either, it's in between specific docs.

In any case, i'm not sure it really helps very much, as it's giving only generic guidance that is common for all multi-module Maven projects, and not really specific to ODC's configuration. Have you tried to configure a multi-module.proejxt work with ODC such that you can give specific example config fragments?

anand1416:
Thanks for the review and feedback.

You're right that the current text is fairly generic and doesn't provide a Dependency-Check-specific example. I also didn't validate the generated site output with `mvn site:stage` before opening the PR.

I'll take another look at the original issue and the existing Maven configuration examples to see whether I can provide a concrete multi-module Dependency-Check configuration example that adds value. If I can't identify a useful project-specific example, I'll consider closing the PR rather than adding generic Maven guidance.

Thanks for taking the time to review it.

anand1416:
Thanks for the review and feedback.

After reviewing the existing aggregate and multi-module examples in the project, I agree that my original addition was too generic and not sufficiently specific to Dependency-Check behavior.

I've removed the generic guidance from the PR and am continuing to investigate whether there is a concrete aggregate or multi-module Dependency-Check configuration example that would be valuable to document.

Thanks again for taking the time to review the change and provide guidance.

PR #8565, opened 2026-06-02
fix: improve Java 17+ compatibility by removing jdiagnostics

### LINKED ISSUE #513: JDK 21 requires undocumented --add-opens/--add-exports
**Description**

When running dependency-check-gradle on JDK 21, the build fails due to illegal access to JDK internal XML classes (Xalan/Xerces).

To make it work, the following JVM arguments are required:

--add-exports=java.xml/com.sun.org.apache.xalan.internal.xsltc.trax=ALL-UNNAMED
--add-opens=java.xml/com.sun.org.apache.xalan.internal.xsltc.trax=ALL-UNNAMED
--add-exports=java.xml/com.sun.org.apache.xerces.internal.jaxp=ALL-UNNAMED
--add-opens=java.xml/com.sun.org.apache.xerces.internal.jaxp=ALL-UNNAMED

**Observed behavior**

Without these flags:
- IllegalAccessException is thrown
- Caused by org.anarres.jdiagnostics accessing internal JDK XML classes

**Expected behavior**

The plugin should either:
- Not rely on internal JDK APIs
  OR
- Document required JVM arguments for Java 17+ / 21 users

**Notes**

This appears to originate from dependency on org.anarres.jdiagnostics, which uses reflection on internal XML implementation classes.

Even if this cannot be easily fixed, documenting the requirement would save time for users running on modern JDKs.

chadlwilson:
## Description of Change

jdiagnostics seems unmaintained, and it does reflection on class internals which is completely disallowed by default on Java 17+ - so at the very time we need it (when there is `LinkageError`) it's most likely to fail and make things even worse. I don't think this is a good idea on modern JVMs.

## Related issues

- fixes https://github.com/dependency-check/dependency-check-gradle/issues/513

N/A

### DIFF
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -162,10 +162,6 @@ Copyright (c) 2012 Jeremy Long. All Rights Reserved.
             <artifactId>open-vulnerability-clients</artifactId>
         </dependency>
         <dependency>
-            <groupId>org.anarres.jdiagnostics</groupId>
-            <artifactId>jdiagnostics</artifactId>
-        </dependency>
-        <dependency>
             <groupId>com.kichik.pecoff4j</groupId>
             <artifactId>pecoff4j</artifactId>
         </dependency>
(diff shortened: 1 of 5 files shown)


nhumblot (approved):
Thank you for your help!

Issue #7566, opened 2025-03-31
Pulling in old vulnerable version of maven-core


johanblumenberg:
**Precondition**
- [x] I checked the issues list for existing open or closed reports of the same problem.

**Describe the bug**

`dependency-check-maven` contains a compile dependency to `org.apache.maven.shared:maven-artifact-transfer:0.13.1` which contains a compile dependency to `org.apache.maven:maven-core`.
`dependency-check-maven` contains a provided dependency to `org.apache.maven:maven-core:3.6.3`.
This means that `dependency-check-maven` will pull in `org.apache.maven:maven-core:3.6.3` as a compile dependency.

`org.apache.maven:maven-core:3.6.3` contains the critical vulnerability CVE-2021-26291: https://mvnrepository.com/artifact/org.apache.maven/maven-core/3.6.3

The recommendation is to have a provided dependency to `maven-core`, not compile, and the `org.apache.maven.shared:maven-artifact-transfer` artifact is deprecated.

**Version of dependency-check used**
The problem occurs using version 12.1.0 of the maven plugin.

**Log file**
N/A

**To Reproduce**
1. Create a simple project containing `dependency-check-maven`, like this: https://gist.github.com/johanblumenberg/72d74aa8520292e1fb220a06de1490af
2. Run `mvn verify`
3. Open `target/dependency-check-report.html` and check the critical items reported.

**Expected behavior**

No critical vulnerabilities coming from the `dependency-check-maven` plugin itself.

**Additional context**

N/A

aikebah:
It does not drag in the library at compile scope, but at provided scope as can be observed when running `mvn dependency:tree -Dincludes=org.apache.maven` on the maven plugin:

```
1387 [main] [INFO] org.apache.maven.cli.event.ExecutionEventLogger - --- dependency:3.8.1:tree (default-cli) @ dependency-check-maven ---
1690 [main] [INFO] org.apache.maven.plugins.dependency.tree.TreeMojo - org.owasp:dependency-check-maven:maven-plugin:12.1.1-SNAPSHOT
1690 [main] [INFO] org.apache.maven.plugins.dependency.tree.TreeMojo - +- org.apache.maven:maven-plugin-api:jar:3.6.3:provided
1690 [main] [INFO] org.apache.maven.plugins.dependency.tree.TreeMojo - +- org.apache.maven:maven-settings:jar:3.6.3:provided
[...]

PR #8567, opened 2026-06-03
fix: resolve overridden plugin dependencies correctly (#8570)

### LINKED ISSUE #8570: Plugin dependencies overrides are ignored and may lead to FPs
**Precondition**

**Describe the bug**
When `scanPlugins` is enabled, dependency-check looks for CVEs in Maven plugins and their dependencies. If one of these dependencies is vulnerable, overriding the dependency to a more recent version usually fixes the vulnerability but dependency-check still sees the vulnerable version and reports false positives.

The problem occurs using version 12.2.x of the maven plugin.

**To Reproduce**
1. Create an empty maven project with the following configuration:
```xml
	<build>
		<pluginManagement>
			<plugins>
				<plugin>
					<artifactId>maven-dependency-plugin</artifactId>
					<version>3.11.0</version>
				</plugin>

				<plugin>
					<artifactId>maven-clean-plugin</artifactId>
					<version>3.5.0</version>
				</plugin>
				
				<plugin>
					<groupId>org.owasp</groupId>
					<version>12.2.2</version>
					
					<configuration>
						<scanPlugins>true</scanPlugins>
					</configuration>
				</plugin>
			</plugins>
		</pluginManagement>	
	</build>
```
2. Run `mvn dependency-check:check` and see CVEs in plexus-utils-4.0.2:
```
One or more dependencies were identified with known vulnerabilities in dependency-check-plugin-cve:
[...]
plexus-utils-4.0.2.jar (pkg:maven/org.codehaus.plexus/plexus-utils@4.0.2, cpe:2.3:a:codehaus-plexus:plexus-utils:4.0.2:*:*:*:*:*:*:*, cpe:2.3:a:utils_project:utils:4.0.2:*:*:*:*:*:*:*) : CVE-2025-67030
[...]
``` 
See that this dependency is pulled by maven-clean-plugin (and others) in the HTML report:

Image

3. Override the version of plexus-utils pulled by maven-clean-plugin in `pluginManagement`:
```xml
				<plugin>
					<artifactId>maven-clean-plugin</artifactId>
					<version>3.5.0</version>
					<dependencies>
						<dependency>
							<groupId>org.codehaus.plexus</groupId>
							<artifactId>plexus-utils</artifactId>
							<version>4.0.3</version>
						</dependency>
					</dependencies>
				</plugin>
```
4. Confirm that maven sees the correct version at runtime using `mvn dependency:resolve-plugins` (you need a recent version of `maven-depedency-plugin`, I used 3.11.0):
```
% mvn dependency:resolve-plugins
[...]
[INFO]    org.apache.maven.plugins:maven-clean-plugin:jar:3.5.0
[INFO]       org.apache.maven.plugins:maven-clean-plugin:jar:3.5.0
[INFO]       org.codehaus.plexus:plexus-utils:jar:4.0.3
[...]
```
5. Run dependency-check again and see in the report that it still believes maven-clean-plugin is using a vulnerable version of `plexus-utils`

**Expected**: plugins overrides should fix vulnerabilities.
**Actual**: dependency-check reports FPs once non-vunerable transitive dependencies are forced in `pluginManagement`.

bjansen:
## Description of Change

When a pom.xml overrides a plugin's dependencies like this, dependency-check does not take the overrides into account, which might lead to false positives:

```xml
<build>
	<pluginManagement>
		<plugin>
			<artifactId>maven-clean-plugin</artifactId>
			<version>3.5.0</version>
			<dependencies>
				<dependency>
					<groupId>org.codehaus.plexus</groupId>
					<artifactId>plexus-utils</artifactId>
					<version>4.0.3</version>
				</dependency>
			</dependencies>
		</plugin>
	</pluginManagement>
</build>
```

In this particular case, dependency-check reports the following error:

> [ERROR] plexus-utils-3.0.24.jar (pkg:maven/org.codehaus.plexus/plexus-utils@3.0.24, cpe:2.3:a:codehaus-plexus:plexus-utils:3.0.24:*:*:*:*:*:*:*, cpe:2.3:a:utils_project:utils:3.0.24:*:*:*:*:*:*:*): CVE-2025-67030(8.8)

The idea in this PR is to mimic what's done in `mvn dependency:resolve-plugins`, that is to say to enhance the `CollectRequest` with a list of plugin dependencies declared in the pom.xml being scanned:

* a list of plugin dependencies is [created here](https://github.com/apache/maven-dependency-plugin/blob/3712611/src/main/java/org/apache/maven/plugins/dependency/utils/ResolverUtil.java#L239)
* and then passed to the [`CollectRequest` here](https://github.com/apache/maven-dependency-plugin/blob/371261124f9cff27cf0439fe3abaec4e322fae7e/src/main/java/org/apache/maven/plugins/dependency/utils/ResolverUtil.java#L185)
 
## Related issues

This should fix #8570

Not yet, I'm just making sure I'm heading towards the right fix for now.

### DIFF
--- a/maven/src/main/java/org/owasp/dependencycheck/maven/BaseDependencyCheckMojo.java
+++ b/maven/src/main/java/org/owasp/dependencycheck/maven/BaseDependencyCheckMojo.java
@@ -34,6 +34,7 @@ import org.apache.maven.artifact.versioning.VersionRange;
 import org.apache.maven.doxia.sink.Sink;
 import org.apache.maven.execution.MavenSession;
 import org.apache.maven.model.License;
+import org.apache.maven.model.Plugin;
 import org.apache.maven.plugin.AbstractMojo;
 import org.apache.maven.plugin.MojoExecution;
 import org.apache.maven.plugin.MojoExecutionException;
@@ -1545,6 +1546,14 @@ public abstract class BaseDependencyCheckMojo extends AbstractMojo implements Ma
         final CollectRequest collectRequest = new CollectRequest();
         collectRequest.setRoot(new org.eclipse.aether.graph.Dependency(rootArtifact, null));
         collectRequest.setRepositories(project.getRemoteProjectRepositories());
+
+        final Plugin projectPlugin = project.getPlugin(rootArtifact.getGroupId() + ":" + rootArtifact.getArtifactId());
+
+        if (projectPlugin != null) {
+            for (org.apache.maven.model.Dependency dep : projectPlugin.getDependencies()) {
+                collectRequest.addDependency(RepositoryUtils.toDependency(dep, session.getRepositorySession().getArtifactTypeRegistry()));
+            }
+        }
 
         final DependencyResult dependencyResult = repoSystem.resolveDependencies(
                 session.getRepositorySession(), new DependencyRequest(collectRequest, null));


bjansen:
@jeremylong before I polish this PR, do you think this approach looks sensible and is worth working on?

chadlwilson:
Without a test to demonstrate what is wrong about the previous behaviour and how this change actually alters that behaviour, it's a bit difficult to understand why this works or whether it actually works. Also - any reason for merging into the `upgradeMaven` branch? That branch is stale, I believe.

I also am finding it difficult to correlate the change to the discussions in #7566 - the root of the issue seems to be that it is including irrelevant `provided` dependencies which are always unwanted, because they are not consumed at runtime. Seems we should be tackling the root cause directly?

bjansen:
> any reason for merging into the upgradeMaven branch? That branch is stale, I believe.

From what I can see, the branch was updated 2 days ago and `main` does not contain changes related to the use of Maven Resolver.

> Without a test to demonstrate what is wrong about the previous behaviour and how this change actually alters that behaviour, it's a bit difficult to understand why this works or whether it actually works

Fair point, I ran the plugin on my machine and it fixed my problem but you have no way to verify that :). I just wanted a quick feedback from you guys before taking more time to work on tests etc. I don't want to waste your time nor mine if this PR is silly and won't be accepted. Consider it a draft for now.

I'm not entirely sure it's the same root cause as #7566, in my case dependency-check is reporting FPs on plugins, not on `provided` dependencies, so I'll probably open a new ticket for my specific problem.

bjansen:
FYI I created a separate issue for my problem: #8570

chadlwilson:
Ahh, my bad, I hadn’t realized https://github.com/dependency-check/DependencyCheck/pull/8560 was only merged into a branch, as that’s unusual for the project. That would also explain why using SNAPSHOTs would make no difference for you 😓

bjansen:
I checked out the `upgradeMaven` branch to test version 13.0.0-SNAPSHOT, and it did not fix my problem. That's why I proposed this PR and used `upgradeMaven` as a baseline.

How do tests work in the `dependency-check-maven` module? I see a bunch of poms in `maven/src/it` but I have no clue how they are run and how to make assertions on the build result to validate my fix.

chadlwilson:
Canonical published snapshots only come from `main`, not arbitrary branches, hence the confusion when you say you tested with a snapshot.

There are some kind of unit tests there (possibly limited, haven’t paid much attention to the maven plugin), but the integration tests can be run with `-PFullIntegrationTesting` similar to the build. They invoke maven in an isolated process.

https://github.com/dependency-check/DependencyCheck/blob/6ab9d0801d063a9969a8c88407b2f751b5dc6c4d/.github/workflows/build-pull-requests.yml#L144-L146

If your fix is not dependent on using the maven resolver it would be better to not be building on top of WIP from a separate experimental draft branch though. I imagine there’s no guarantee that branch will be merged, or when?

chadlwilson:
@jeremylong deleting the target branch before merging this caused this to get closed. Would you like a PR to be retargetted against master for this?

chadlwilson:
@bjansen woukd you mind rebasing and retargetting this PR at `main` now? It’d be better if you can edit and reopen the same PR to retain the conversation - it got auto-closed by GitHub because the target branch was auto-deleted after merge.

PR #8656, opened 2026-07-12
fix(windows): use JAVA_HOME when java is not on PATH

### LINKED ISSUE #8631: dependency-check.bat does not use JAVA_HOME
**Precondition**

**Describe the bug**
If `java` is not on the `PATH` then the CLI fails on Windows, even though `JAVA_HOME` is correctly configured.

The problem occurs using version 12.2.2 of the cli

**Log file**
```
java : The term 'java' is not recognized as the name of a cmdlet, function, script file, or operable program. Check the spelling of the name, or if a path was included, verify that the path is correct and try again.
At line:1 char:1
+ java
+ ~~~~
    + CategoryInfo          : ObjectNotFound: (java:String) [], CommandNotFoundException
    + FullyQualifiedErrorId : CommandNotFoundException
```

**Expected behavior**
The script should check, and indeed prefer, `JAVA_HOME` to find the java executable.

**Workaround**
Adjust either `PATH` or `JAVACMD` variables before calling the script.

mohan485:
## Description of Change

Updated the Windows launcher template to use JAVA_HOME to locate the Java executable when JAVACMD is not set. If %JAVA_HOME%\bin\java.exe exists, it is used; otherwise, the launcher falls back to the default Java executable (java) on the system PATH.

This change allows the Dependency-Check CLI to run on Windows when JAVA_HOME is configured but java.exe is not present on the PATH, while preserving the existing behavior for users who set JAVACMD or rely on PATH.

## Related issues

**Fixes #8631**
This fixes issue #8631, where the CLI fails on Windows if JAVA_HOME is configured but java.exe is not on PATH.

This change updates the Windows launcher script only. The behavior was verified manually by testing the following scenarios:

JAVA_HOME configured and java.exe not on PATH → CLI starts successfully.
JAVACMD configured → launcher uses JAVACMD.
JAVA_HOME not configured but java.exe available on PATH → existing behavior is preserved.

### DIFF
--- a/cli/src/main/conf/windowsBinTemplate.bat
+++ b/cli/src/main/conf/windowsBinTemplate.bat
@@ -50,7 +50,13 @@ for %%i in ("%~dp0..") do set "BASEDIR=%%~fi"
 set REPO=
 #ENV_SETUP#
 
-if "%JAVACMD%"=="" set JAVACMD=#JAVA_BINARY#
+if "%JAVACMD%"=="" (
+    if exist "%JAVA_HOME%\bin\java.exe" (
+        set "JAVACMD=%JAVA_HOME%\bin\java.exe"
+    ) else (
+        set "JAVACMD=#JAVA_BINARY#"
+    )
+)
 
 if "%REPO%"=="" set REPO=%BASEDIR%\#REPO#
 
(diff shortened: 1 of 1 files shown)


chadlwilson (approved):

nhumblot (approved):
**praise:** thank you for your help!

Issue #8668, opened 2026-07-15
[FP]: Kotlin runtime libraries detected as JetBrains runtime


OrangeDog:
### Package URl

`pkg:maven/org.jetbrains.kotlin/kotlin-parcelize-runtime@1.7.20`

### CPE

`cpe:2.3:a:jetbrains:runtime:1.7.20:*:*:*:*:*:*:*`

### CVE

CVE-2025-29 903

### ODC Integration

{"label" => "Gradle Plugin"}

### ODC Version

12.2.2

### Description

Above details are just one example. I've gone for this suppression:

```xml
<suppress>
    <packageUrl regex="true">pkg:maven/org\.jetbrains\.kotlin/.*</packageUrl>
    <cpe>cpe:/a:jetbrains:runtime</cpe>
</suppress>
```

chadlwilson:
Approved

If you want something more generic (or to, like, help), please submit a PR to the `generatedSuppressions` branch that amends the suppression added here.

OrangeDog:
@chadlwilson I wasn't sure if you'd be comfortable with my broad pattern.

chadlwilson:
Well, if there are say 3+ artifacts under that group with `runtime` in the name sufficient to cause FPs it's probably worth widening the suppression rather than adding them one-by-one. Doesn't seem that risky to conclude that nothing in the kotlin group should match their JRE fork.

chadlwilson:
No harm starting with an automated suppression and then siding it via a PR anyway. Makes it easy to start with a reviewed example since the automation produces actual evidence (which people would usually omit before the project had the automation, making it take forever to evaluate if it's a bug, config issue etc etc)

OrangeDog:
Aside: what's the reason for using `(:.*)?$` in generated `<cpe>` elements? 

As well as it being equivalent to `.*`, wouldn't it speed everything up for them to be `regex="false"`?

chadlwilson:
#8531 has an approximation of an explanation, but it’s not equivalent to `.*`. (If there are any chars after the product name, they must be prefixed by a colon)

Basically the goal is to avoid a whole swathe of false negatives from the earlier “generated suppression” approach.

Here’s another summary since it was spread across a few PRs unfortunately.
- Non regex matches are prefix-matched. That creates false negatives because it’d end up suppressing, for e.g, `cpe:/a:jetbrains:runtime_kotlin` as well as the intended cpe
[...]

PR #8670, opened 2026-07-16
fix(fp): Extend Kotlin/JBR suppression scope

OrangeDog:
As discussed in #8668

### DIFF
--- a/generatedSuppressions.xml
+++ b/generatedSuppressions.xml
@@ -2185,6 +2185,6 @@
     <notes><![CDATA[
     FP per issue #8668
     ]]></notes>
-    <packageUrl regex="true">^pkg:maven/org\.jetbrains\.kotlin/kotlin-parcelize-runtime@.*$</packageUrl>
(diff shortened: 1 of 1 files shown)


chadlwilson:
If you prefix the PR title with `fix(fp): ` the linter will pass.

What are some of the other artifacts in that group which are matching incorrectly?

chadlwilson (approved):

OrangeDog:
> What are some of the other artifacts in that group which are matching incorrectly?

The other one I'm using is `kotlin-android-extensions-runtime`, but also in there are `kotlin-script-runtime`, `kotlinx-atomicfu-runtime`, `kotlin-annotation-processing-runtime`, and `kotlin-power-assert-runtime`.

chadlwilson:
Might need a trivial resolve of conflicts here. Your change (or the editor) added a new line at the end which conflicts with the automated suppressions being appended to the end.

jeremylong (approved):

PR #8678, opened 2026-07-21
feat: add v1 Golang Vulncheck Analyzer

cvas22:
## Description of Change

Adds a new v1 analyzer that runs the Go team's govulncheck against Go modules and reports the vulnerabilities it finds, sourced from the curated Go vulnerability database.

Unlike the CPE-matching Golang Mod Analyzer, govulncheck performs call-graph reachability analysis in source mode (govulncheck -json ./...), so it reports primarily the vulnerabilities actually reachable from the scanned code - substantially reducing false positives/negatives for Go.

Design:

Source mode only, keyed off go.mod (reuses existing module detection); binary mode is intentionally out of scope for v1.
Native -json (OSV) output; streaming parser joins osv advisories with finding records and selects the most precise (reachable) finding.
De-duplicates against other data sources via OSV aliases: reuses an existing NVD vulnerability when a CVE alias matches, otherwise synthesizes one under the GO-YYYY-NNNN id.
Reports against the vulnerable module as a synthetic pkg:golang dependency.
Disabled by default; requires go and govulncheck to be installed.

New:

GolangVulncheckAnalyzer, GovulncheckProcessor
GovulncheckJsonParser, GovulncheckResult
Settings keys analyzer.golang.vulncheck.{enabled,path}
Vulnerability.Source.GOVULNCHECK
Config wiring: CLI (--enableGolangVulncheck, --govulncheck), Ant (golangVulncheckEnabled, pathToGovulncheck), Maven (golangVulncheckEnabled, pathToGovulncheck).

Tests: unit tests for the parser and analyzer; an integration test that runs real govulncheck end-to-end (self-skips when the tool is absent).

Docs: analyzer page, analyzers index, CLI/Ant/Maven configuration, and bash completion updated.

## Related issues

attempts to address https://github.com/dependency-check/DependencyCheck/issues/8470

*yes*


jeremylong:
Won't merge until CI errors are corrected:

```
[ERROR] Failures: 
[ERROR]   EngineIT.testEngine:122 Analysis threw exceptions that weren't expected
Expected: an empty collection
     but: <[org.owasp.dependencycheck.exception.InitializationException: Exception from govulncheck process: java.io.IOException: Cannot run program "govulncheck" (in directory "/home/runner/work/DependencyCheck/DependencyCheck/core/target/temp/dctemp60248e31-a9b8-4d66-b1ac-d9c9553ef241"): error=2, No such file or directory. Disabling Golang Vulncheck Analyzer]>
```

core/src/main/java/org/owasp/dependencycheck/analyzer/GolangVulncheckAnalyzer.java
```diff
@@ -0,0 +1,231 @@
...
+
+    /**
+     * The name of the analyzer.
+     */
+    private static final String ANALYZER_NAME = "Golang Vulncheck Analyzer";
+
+    /**
+     * The phase that this analyzer runs in.
+     */
+    private static final AnalysisPhase ANALYSIS_PHASE = AnalysisPhase.PRE_INFORMATION_COLLECTION;
```
jeremylong: This should be:
```suggestion
    private static final AnalysisPhase ANALYSIS_PHASE = AnalysisPhase.INFORMATION_COLLECTION;
```

jeremylong (commented):
In addition, the configuration needs to be added to the ant task and maven plugin.

core/src/test/java/org/owasp/dependencycheck/analyzer/GolangVulncheckAnalyzerTest.java
```diff
@@ -0,0 +1,77 @@
...
+    }
+
+    @Test
+    void testGetName() {
+        assertThat(analyzer.getName(), is("Golang Vulncheck Analyzer"));
+    }
+
+    @Test
+    void testGetAnalysisPhase() {
+        assertThat(analyzer.getAnalysisPhase(), is(AnalysisPhase.PRE_INFORMATION_COLLECTION));
```
jeremylong:
```suggestion
        assertThat(analyzer.getAnalysisPhase(), is(AnalysisPhase.INFORMATION_COLLECTION));
```

PR #8680, opened 2026-07-27
docs: document NVD 2.0 data feeds

### LINKED ISSUE #7514: Support for NIST 2.0 data feeds

Eljees:
## Description of Change

Document the existing support for the official NVD 2.0 JSON data feeds. The new section includes CLI and Maven examples, explains the `{0}` filename placeholder, and makes clear that configuring a data-feed URL switches updates away from the REST API.

## Related issues

- Fixes #7514

No; this is a documentation-only change.


src/site/markdown/data/mirrornvd.md
```diff
@@ -4,6 +4,29 @@ If an organization blocks the servers performing dependency-check scans from
 downloading content on the internet they will need to mirror two data sources:
 The NVD API and the Retire JS repository.
```
chadlwilson:
```suggestion
downloading content on the internet they may need to mirror multiple data sources as documented at [Remote Data Sources](./index.html), however access to data from the NVD API is currently mandatory.
```

src/site/markdown/data/mirrornvd.md
```diff
@@ -4,6 +4,29 @@ If an organization blocks the servers performing dependency-check scans from
...
+<configuration>
+    <nvdDatafeedUrl>https://nvd.nist.gov/feeds/json/cve/2.0/nvdcve-2.0-{0}.json.gz</nvdDatafeedUrl>
+</configuration>
+```
+
+The `{0}` placeholder is replaced with a year or `modified` when
+dependency-check retrieves feed data. Configuring this URL switches NVD updates
+from the REST API to the data feeds.
+
 Creating an offline cache for the NVD API
```
chadlwilson: Suggest completing this by documenting ODC's own mirror/cache, as well as ensuring users are aware of the downsides of this approach.

Suggest to not document Maven configuration, as there are many integrations and these are documented elsewhere in their own documentation which will become out of sync; and Maven itself has multiple ways URLs can be configured.

````suggestion

Mirrors are made available by setting a special URL template in the `nvdDatafeed`/`nvdDatafeedUrl` configuration. Configuring this URL switches NVD updates from the REST API to a data feeds in the [NVD JSON 2.0 schema format](https://csrc.nist.gov/schema/nvd/api/2.0/cve_api_json_2.0.schema). The `{0}` placeholder is replaced with a year or `modified` when dependency-check retrieves feed data. 

Switching to data feeds from the API has some advantages and disadvantages you should consider:
- ✅ reliability: insulates from API-specific rate limits and availability issues
- ✅ reliability: can be safely cached & proxied via a private centralized forward proxy
- ❌ performance: increases data & CPU usage. Entire feed file is downloaded for each update and updates every entry in Dependency-check database
- ❌ support: most options lack a documented SLA.
- ❌ complexity: relies on data feed to correctly reflect all NVD API data changes, depending on additional tools or NVD's own proprietary feed creation process

Consult your specific Dependency-check integration's documentation on how to configure this.

Using Dependency-check's own 2.0 data feed mirror
------------------------------------------------------------

Dependency-check [maintains its own mirror/cache](https://github.com/dependency-check/DependencyCheck_Builder/actions/workflows/cache.yml), built directly from the NVD API using the [vulnz CLI](https://github.com/jeremylong/open-vulnerability-cli).

If using the Dependency-Check CLI, set the NVD Datafeed URL to the feed filename pattern:

```shell
dependency-check.sh --nvdDatafeed \
    'https://dependency-check.github.io/DependencyCheck_Builder/nvd_cache/nvdcve-{0}.json.gz'
```

This is updated every 24 hours on a best-effort basis. When the NVD API is unavailable, it will contain stale data.

Using the NVD 2.0 data feeds
------------------------------------------------------------

Dependency-check can use the [official NVD 2.0 JSON data feeds](https://nvd.nist.gov/vuln/data-feeds#divJson20Feeds) instead of the
NVD REST API. 

If using the Dependency-Check CLI, set the NVD Datafeed URL to the feed filename pattern:
```shell
dependency-check.sh --nvdDatafeed \
    'https://nvd.nist.gov/feeds/json/cve/2.0/nvdcve-2.0-{0}.json.gz'
```

Note that the official NVD 2.0 data feeds have sometimes been stale for extended periods (1-2 weeks), and at high load periods can be aggressively rate-limited. This can lead to extremely slow download speeds, frequent timeouts & interrupted downloads. Users should consider using a caching forward proxy between themselves and the NVD to improve reliability.

Creating your own offline cache from the NVD API
````

chadlwilson (commented):
Thanks for looking at this. Suggest these changes to improve completeness and simplify the docs for easier maintainability.

Eljees:
Thanks for the detailed suggestions. Addressed both in 79e280239: the introduction now points to all remote data sources; the page documents the project's own mirror/cache, trade-offs and stale-data caveats; examples are CLI-only so integration-specific configuration stays in the relevant docs.

chadlwilson (approved):

jeremylong (approved):

Issue #8700, opened 2026-08-03
Bundled suppression regressed in 13.0.0


OrangeDog:
**Precondition**
- [x] I checked the issues list for existing open or closed reports of the same problem.

**Describe the bug**
After upgrading to 13.0.0, #8257 has returned.

**Version of dependency-check used**
The problem occurs using version 13.0.0 of the maven plugin.

**Log file**
```
jna-5.3.1.jar (pkg:maven/net.java.dev.jna/jna@5.3.1, cpe:2.3:a:oracle:java_se:5.3.1:*:*:*:*:*:*:*, cpe:2.3:a:sun:java_se:5.3.1:*:*:*:*:*:*:*) : CVE-2009-2476, CVE-2009-2689, CVE-2009-2475, CVE-2009-2716, CVE-2009-2676, CVE-2009-2717, CVE-2009-2690, CVE-2009-2719, CVE-2009-2720
```

**To Reproduce**
```xml
<dependency>
  <groupId>net.java.dev.jna</groupId>
  <artifactId>jna</artifactId>
  <version>5.3.1</version>
</dependency>
```

**Expected behavior**
Previous internal suppressions should still work.

chadlwilson:
Suppressions are refactored for good reasons all the time. Unless you know the internal specifics, open a false positive report, please, not a bug.

A suppression is an implementation detail. Describe the problem, so the automation runs that reproduces them.

jeremylong:
My guess is I approved changes to the generated suppression and had not forced a publication of the updated file....  This should be fixed in the next release.

I just [republished](https://github.com/dependency-check/DependencyCheck/actions/runs/30811852472) the generated suppression file - can you check if this is an issue with the published version?

chadlwilson:
The generated suppressions auto-publishes via the workflow now though, so it shouldn't be that, unless something got messed up.

OrangeDog:
@chadlwilson I thought it best that you check for systematic issues, rather than just blindly add the suppression again, as a refactoring by definition shouldn't have changed the observable behaviour.

@jeremylong no, that didn't change anything (with `hostedSuppressionsForceUpdate`)

chadlwilson:
I never blindly add suppressions from such reports.

An FP report is a report of a specific problem that comes with structured data capture and thus a reproducer and a suggestion for a way to fix it (which this report does not come with). The whole point is because issues like this never come with an easy way to reproduce them, never an ODC report and puts the effort back onto maintainers.

[...]

PR #8704, opened 2026-08-03
fix(fp): tweak JDK false positive suppressions for exact product matches to `java_se`

chadlwilson:
## Description of Change

As discussed at https://github.com/dependency-check/DependencyCheck/issues/8700#issuecomment-5166462596 the old code was doing a lazy prefix match here. Not sure if we should be doing these filepath suppressions, but adding the missing "exact" CPE here for now anyway.

`j_se` isn't a valid CPE according to the dictionary (probably renamed at some point)

## Related issues

- relates to #8700
- "caused" by changes in #8509

No.


marcelstoer (approved):
Approved

jeremylong (approved):

PR #8705, opened 2026-08-03
chore: remove the broken/inoperable Yarn v1 Analyzer

### LINKED ISSUE #8423: Remove YarnAuditAnalyzer (classic/v1 mode) dependency on NPM audit /quick API

chadlwilson:
## Description of Change

The Yarn v1/classic Analyzer has been broken since NPM removed the underlying API which Yarn v1 uses via its CLI. Yarn v1 is EOL, and it won't be fixed. People need to move to Yarn Berry (v2/v3/v4 etc).

This changes ODC to ignore Yarn v1/classic rather than failing; but could equally throw `AnalysisException`s.

Should have been done in v13, but since the analyzer has been broken upstream for several weeks now, I don't think this is a breaking change as there's no change to config properties. Previously failing builds will now succeed with only logging warnings though.

## Related issues

- fixes #8423

yes


marcelstoer (approved):
Approved

jeremylong (approved):

Issue #5658, opened 2023-04-11
dependency-check report shows in column 'Highest severity' MEDIUM while CVSSv2 has value 8.5 (CVSSv3 has value 6.6)


GitMarco:
**Describe the bug**
The Gradle plugin fails with a found vulnerability with value greater than 7.0 during the analyze step:

_Dependency-Analyze Failure:
One or more dependencies were identified with vulnerabilities that have a CVSS score greater than '7.0': CVE-2021-42550
See the dependency-check report for more details._

When I then look at the dependency-check report, I only see vulnerabilities with as 'Highest severity' the value MEDIUM. While I would expect a value of HIGH  (because the analyze failed).

See screenshot:
![image](https://user-images.githubusercontent.com/340814/231198799-6fb87ea1-1c73-4b1f-8cd0-29e3b1181c30.png)

When clicking on the logback-core-1.3.0.jar link, it shows (amongst others):

_CVSSv2:
Base Score: HIGH (8.5)
Vector: /AV:N/AC:M/Au:S/C:C/I:C/A:C
CVSSv3:
Base Score: MEDIUM (6.6)
Vector: CVSS:3.1/AV:N/AC:H/PR:H/UI:N/S:U/C:H/I:H/A:H_

See screenshot:
![image](https://user-images.githubusercontent.com/340814/231199120-2112833d-e077-4bac-8b66-ebcdda0652a1.png)

Note that the CVSSv2 score is 8.5 and the CVSSv3 is 6.6.   It seems the plugin checks only for the status of the CVSSv3 value?  That would be fine if it wouldn't fail the plugin analyze phase on reporting there is a vulnerability with score greater than 7.0 (which indicates a severity of value HIGH).

A solution could be: take the highest of the two CVSS scores and show that in the report column "Highest severity"?  So in this case it would show HIGH (not MEDIUM)

**Version of dependency-check used**
The problem occurs using version 7.4.4 and 8.1.2 of the gradle plugin.

**Log file**
Only for now adding the reason why the job (plugin) is failing:

_Dependency-Analyze Failure:
One or more dependencies were identified with vulnerabilities that have a CVSS score greater than '7.0': CVE-2021-42550
See the dependency-check report for more details._

**To Reproduce**
1. Have logback-core-1.3.0.jar as dependency in your project and configure the plugin like this:

_dependencyCheck {
  failBuildOnCVSS = 7
  failOnError = true
}_

[...]

PR #8727, opened 2026-08-09
fix(maven): report the CVSS score that failed the build

Eljees:
## Description of Change

`checkForFailure` fails the build when **any** of the CVSS scores of a vulnerability reaches `failBuildOnCVSS`, but the score it prints is the one of the newest CVSS version available. When an older version is the one that crossed the threshold, the message contradicts itself:

```
[ERROR] One or more dependencies were identified with vulnerabilities that have a CVSS score greater than or equal to '7.0':
[ERROR] logback-core-1.2.3.jar: CVE-2021-42550(6.6)
```

for a CVE whose CVSSv2 score is 8.5, which is what actually failed the build.

The score to print is now chosen by the same comparison that decides whether the vulnerability fails the build: the first CVSS version to reach the threshold (v4, then v3, then v2, then the estimate for an unscored severity) is both the reason for the failure and the score shown, and a vulnerability that reaches none of them is not listed. With no threshold in play (`failBuildOnCVSS <= 0`, i.e. report everything) the newest available version is still used, so nothing changes for vulnerabilities where the newest version is already above the threshold.

I kept the reported score rather than switching to the maximum score (as suggested in the issue thread) because the maximum can come from a version that did not trigger the failure at all.

## Related issues

- related to #5658 - this PR does not fix that issue; unlinked after review

yes - three tests in `BaseDependencyCheckMojoTest` drive `checkForFailure` and assert on the failure message it throws: the case from the issue (v2 8.5 / v3 6.6 at a 7.0 threshold), the unchanged no-threshold behaviour, and a vulnerability below the threshold not failing the build at all. Against the code before this PR the first of the three fails, quoting 6.6 in a message that states 7.0 was exceeded.

### DIFF
--- a/maven/src/main/java/org/owasp/dependencycheck/maven/BaseDependencyCheckMojo.java
+++ b/maven/src/main/java/org/owasp/dependencycheck/maven/BaseDependencyCheckMojo.java
@@ -2907,14 +2907,9 @@ public abstract class BaseDependencyCheckMojo extends AbstractMojo implements Ma
                         || unscoredCvss >= failBuildOnCVSS
                 ) {
                     String name = v.getName();
-                    if (cvssV4 >= 0.0) {
-                        name += "(" + cvssV4 + ")";
-                    } else if (cvssV3 >= 0.0) {
-                        name += "(" + cvssV3 + ")";
-                    } else if (cvssV2 >= 0.0) {
-                        name += "(" + cvssV2 + ")";
-                    } else if (unscoredCvss >= 0.0) {
-                        name += "(" + unscoredCvss + ")";
+                    final double reportedScore = scoreToReport(cvssV2, cvssV3, cvssV4, unscoredCvss, failBuildOnCVSS);
+                    if (reportedScore >= 0.0) {
+                        name += "(" + reportedScore + ")";
                     }
                     if (addName) {
                         addName = false;
@@ -2941,6 +2936,44 @@ public abstract class BaseDependencyCheckMojo extends AbstractMojo implements Ma
             }
             throw new MojoFailureException(msg);
         }
+    }
+
+    /**
+     * Determines the CVSS score to display for a vulnerability listed by
+     * {@link #checkForFailure(org.owasp.dependencycheck.dependency.Dependency[])}.
+     * <p>
+     * The build is failed when <em>any</em> of the CVSS scores of a vulnerability reaches the
+     * configured threshold, so the score that actually reached it is the one to display; showing
+     * the score of the newest CVSS version instead quotes a score below the threshold that the
+     * failure message itself states. When no threshold applies the newest CVSS version is used.
+     *
+     * @param cvssV2 the CVSS v2 base score, or -1 when the vulnerability has no v2 score
+     * @param cvssV3 the CVSS v3 base score, or -1 when the vulnerability has no v3 score
+     * @param cvssV4 the CVSS v4 base score, or -1 when the vulnerability has no v4 score
+     * @param unscoredCvss the score estimated from an unscored severity, or -1 when not applicable
+     * @param threshold the configured failBuildOnCVSS threshold
+     * @return the score to display, or -1 when the vulnerability carries no score at all
+     */
+    static double scoreToReport(double cvssV2, double cvssV3, double cvssV4, double unscoredCvss, float threshold) {
+        if (threshold > 0.0) {
+            if (cvssV4 >= threshold) {
+                return cvssV4;
+            } else if (cvssV3 >= threshold) {
+                return cvssV3;
+            } else if (cvssV2 >= threshold) {
+                return cvssV2;
+            } else if (unscoredCvss >= threshold) {
+                return unscoredCvss;
+            }
+        }
+        if (cvssV4 >= 0.0) {
+            return cvssV4;
+        } else if (cvssV3 >= 0.0) {
+            return cvssV3;
+        } else if (cvssV2 >= 0.0) {
+            return cvssV2;
+        }
+        return unscoredCvss;
     }
 
     /**
--- a/maven/src/main/java/org/owasp/dependencycheck/maven/BaseDependencyCheckMojo.java
+++ b/maven/src/main/java/org/owasp/dependencycheck/maven/BaseDependencyCheckMojo.java
@@ -2900,28 +2900,45 @@ public abstract class BaseDependencyCheckMojo extends AbstractMojo implements Ma
                 final boolean useUnscored = cvssV2 == -1 && cvssV3 == -1 && cvssV4 == -1;
                 final double unscoredCvss = (useUnscored && v.getUnscoredSeverity() != null) ? SeverityUtil.estimateCvssV2(v.getUnscoredSeverity()) : -1;
 
-                if (failBuildOnCVSS <= 0.0
-                        || cvssV2 >= failBuildOnCVSS
-                        || cvssV3 >= failBuildOnCVSS
-                        || cvssV4 >= failBuildOnCVSS
-                        || unscoredCvss >= failBuildOnCVSS
-                ) {
-                    String name = v.getName();
-                    final double reportedScore = scoreToReport(cvssV2, cvssV3, cvssV4, unscoredCvss, failBuildOnCVSS);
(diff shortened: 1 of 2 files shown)


maven/src/main/java/org/owasp/dependencycheck/maven/BaseDependencyCheckMojo.java
```diff
@@ -2943,6 +2938,44 @@ protected void checkForFailure(Dependency[] dependencies) throws MojoFailureExce
...
+            }
+        }
+        if (cvssV4 >= 0.0) {
+            return cvssV4;
+        } else if (cvssV3 >= 0.0) {
+            return cvssV3;
+        } else if (cvssV2 >= 0.0) {
+            return cvssV2;
+        }
+        return unscoredCvss;
```
chadlwilson: the logic could be much simpler as it is, essentially there are now two places that are applying the threshold checks rather than one, which increases the chance of more bugs.

maven/src/main/java/org/owasp/dependencycheck/maven/BaseDependencyCheckMojo.java
```diff
@@ -2943,6 +2938,44 @@ protected void checkForFailure(Dependency[] dependencies) throws MojoFailureExce
...
+     * the score of the newest CVSS version instead quotes a score below the threshold that the
+     * failure message itself states. When no threshold applies the newest CVSS version is used.
+     *
+     * @param cvssV2 the CVSS v2 base score, or -1 when the vulnerability has no v2 score
+     * @param cvssV3 the CVSS v3 base score, or -1 when the vulnerability has no v3 score
+     * @param cvssV4 the CVSS v4 base score, or -1 when the vulnerability has no v4 score
+     * @param unscoredCvss the score estimated from an unscored severity, or -1 when not applicable
+     * @param threshold the configured failBuildOnCVSS threshold
+     * @return the score to display, or -1 when the vulnerability carries no score at all
+     */
```
chadlwilson: The comment is more complex than the actual code. Seems completely unnecessary.

chadlwilson (changes requested):
My 2c is that notwithstanding whether the functional change is desirable or not - this doesn't address the linked issue so should be unlinked.

The linked issue is
1) reported against Gradle initially, not Maven (so changing the Maven text output won't help, by definition - later comments are unrelated comments conflating different issues)
2) is reporting a mismatch in the report contents "**Highest Severity**" column independent of what it prints on the command line, which is caused by inconsistenct between the way scores are chosen from each vulnerability for sorting (prefer V4 to V3 to V3 to unscored, regardless of score, like the PR printing logic) with the failure limits (if any of the CVSS scores for an individual vuln is > threshold --> fail)

The fundamental problem in the linked issue is not that it doesn't report the score in the Maven output, it's that the methodology is inconsistent between integration and report and there hasn't been a discussion on what the best/correct methodology actually is - in order to make the sorting for report consistent with the command line failure output.

While this PR does ensure the error-reported score is the one that triggered the failure for Maven (which has special logic here inconsistent with other integrations), it does so with some unnecessary complexity.

To _actually_ address the linked issue, we probably need to do this holistically, and ideally have a reusable component/helper methods on the `Vulnerability` itself which
  - are consistent with sort logic
  - will minimise the chance of behavioural differences between integrations
  - will minimise the chance of sort vs choice logic accidentally deviating
  - is reused by the integrations to then decide how to display that logic rather than mixing logic and display/error-building semantics into a near-duplicated function

maven/src/main/java/org/owasp/dependencycheck/maven/BaseDependencyCheckMojo.java
```diff
@@ -2907,14 +2907,9 @@ protected void checkForFailure(Dependency[] dependencies) throws MojoFailureExce
...
                     String name = v.getName();
-                    if (cvssV4 >= 0.0) {
-                        name += "(" + cvssV4 + ")";
-                    } else if (cvssV3 >= 0.0) {
-                        name += "(" + cvssV3 + ")";
-                    } else if (cvssV2 >= 0.0) {
-                        name += "(" + cvssV2 + ")";
-                    } else if (unscoredCvss >= 0.0) {
-                        name += "(" + unscoredCvss + ")";
+                    final double reportedScore = scoreToReport(cvssV2, cvssV3, cvssV4, unscoredCvss, failBuildOnCVSS);
```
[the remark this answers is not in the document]
chadlwilson: Essentially what I said in https://github.com/dependency-check/DependencyCheck/pull/8727#pullrequestreview-4891673243 :-)

Eljees:
You're right about the linked issue, and I'm removing the `fixes #5658` link. That issue is about the **Highest severity** column in the report and its sort value; this PR only changes the Maven failure message, so claiming it as a fix was wrong.

That leaves a narrower Maven bug: `checkForFailure` fails on any CVSS score crossing the threshold but prints the newest-version score, so the message can quote a score below the threshold it says was exceeded. Two questions before I put more into it:

1. Is that worth fixing on its own, or would you rather park it until the methodology question is settled? I'm fine closing it.
2. If it is worth keeping, your point about two places applying the threshold is the real objection. I'd drop `scoreToReport` and pick the score inside the existing comparison rather than re-deriving it, and the javadoc goes with it — it is longer than the code it describes.

On the holistic fix, I'd rather not attach it to this PR without agreement on the methodology. The shape I'd propose is two methods on `Vulnerability` — one returning the score used for ordering, one returning the score that crossed a given threshold — with the report template, the sort and each integration calling those instead of repeating the v4 → v3 → v2 → unscored chain. Happy to open that separately once there is a decision on what the correct methodology is, but I don't want to guess it.

Eljees:
Posted twice by mistake - sorry about the noise. The original is above.

Replying to your points one by one in the threads, as asked.

chadlwilson:
> You're right about the linked issue, and I'm removing the fixes #5658 link.

Are you? You have still not done it.

> the message can quote a score below the threshold it says was exceeded

IMHO its worth fixing on its own (but only since this extra inclusion of the score is only in the Maven plugin integration, so it doesn’t make the inconsistency between Maven and other integrations worse than it already is).

And only if you actually properly respond to PR comments. One-by-one within context, please, not with a wall of slopped LLM text, spammed twice for no reason.

> On the holistic fix […] Happy to open that separately once there is a decision on what the correct methodology is, but I don't want to guess it.

yes, don’t guess it.

maven/src/main/java/org/owasp/dependencycheck/maven/BaseDependencyCheckMojo.java
Eljees: Agreed. I'll drop `scoreToReport` and choose the score inside the comparison that already applies the threshold, so it is decided in one place instead of two.

maven/src/main/java/org/owasp/dependencycheck/maven/BaseDependencyCheckMojo.java
Eljees: Agreed - removing it.

maven/src/main/java/org/owasp/dependencycheck/maven/BaseDependencyCheckMojo.java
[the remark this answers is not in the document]
Eljees: Unlinked. The description now refers to #5658 without a closing keyword, so merging this will not close it.

maven/src/main/java/org/owasp/dependencycheck/maven/BaseDependencyCheckMojo.java
Eljees: Done in ddb9b48. The score is now chosen by the same comparison that decides the failure, and a vulnerability that reaches none of the thresholds is skipped, so `scoreToReport` is gone along with the second copy of the check.

maven/src/main/java/org/owasp/dependencycheck/maven/BaseDependencyCheckMojo.java
Eljees: Removed in ddb9b48, together with the method it described.

chadlwilson (approved):

jeremylong (approved):

PR #8728, opened 2026-08-09
fix(core): parse single letter version suffixes without a period

### LINKED ISSUE #4139: Incorrect DependencyVersion parsing for libjpeg library
**Describe the bug**
libjpeg library uses specific version scheme with one digit and one letter (e.g., 8a, 9d).
parseVersion function from DependencyVersionUtil uses two regex patterns: RX_VERSION and RX_SINGLE_VERSION. Unfortunately, none of them is able to fully parse such version scheme.  For example for version "9d", DependencyVersionUtil::parseVersion extracts '9' as a version without a suffix with RX_SINGLE_VERSION pattern. Such incorrect parsing leads to FPs, because CPEAnalyzer uses all CPEs that are related to version 9 and have vulnerabilities ("cpe:2.3:a:ijg:libjpeg:9a:*:*:*:*:*:*:*, cpe:2.3:a:ijg:libjpeg:9c:*:*:*:*:*:*:*" in "CPE" column in the report). 

Valid versions for libjpeg can be found here: https://jpegclub.org/reference/reference-sources/ or https://www.ijg.org/files/
 
Libjpeg source code archive contains another version scheme in configure.ac file: AC_INIT([libjpeg], [9.5.0]). I use following hints.xml file to provide version string which is compatible with CPEs from NVD database:
```xml
<hints xmlns="https://jeremylong.github.io/DependencyCheck/dependency-hint.1.3.xsd">
  <hint>
    <given>
      <evidence type="product" source="configure.ac" name="package" value="libjpeg" confidence="HIGHEST"/>
    </given>
    <add>
      <evidence type="product" source="hint analyzer" name="product" value="libjpeg" confidence="HIGHEST"/>
      <evidence type="version" source="hint analyzer" name="version" value="9e" confidence="HIGHEST"/>
      <evidence type="vendor" source="hint analyzer" name="vendor" value="ijg" confidence="HIGHEST"/>
    </add>
  </hint>
</hints>
```

The problem occurs using version 7.0.0 of the cli.
I assume previous versions have the same behavior.

**To Reproduce**
1. cd ~/Downloads
2. wget https://www.ijg.org/files/jpegsr9e.zip
3. unzip jpegsr9e.zip
4. wget https://github.com/jeremylong/DependencyCheck/releases/download/v7.0.0/dependency-check-7.0.0-release.zip
5. unzip dependency-check-7.0.0-release.zip
6. create hints.xml with content provided in description section
7.  ./dependency-check/bin/dependency-check.sh --enableExperimental --hints hints.xml --format CSV -s jpeg-9e
8. dependency-check-report.csv contains at least 6 rows. CPE column contains "cpe:2.3:a:ijg:libjpeg:9.5.0:*:*:*:*:*:*:*, cpe:2.3:a:ijg:libjpeg:9a:*:*:*:*:*:*:*, cpe:2.3:a:ijg:libjpeg:9c:*:*:*:*:*:*:*"

**Expected behavior**
DependencyVersionUtil::parseVersion should be able to parse versions like 8a, 9b, etc.
CPE section in dependency-report should contain: "cpe:2.3:a:ijg:libjpeg:9.5.0:*:*:*:*:*:*:*, cpe:2.3:a:ijg:libjpeg:9e:*:*:*:*:*:*:*"

Eljees:
## Description of Change

`DependencyVersionUtil` has two expressions for pulling a version out of a name: `RX_VERSION`, which requires a period, and `RX_SINGLE_VERSION`, which does not. Only the first one accepts a single trailing letter, so libjpeg's scheme of one digit and one letter (`8a`, `9d`, `9e`) never reaches it — the name falls through to `RX_SINGLE_VERSION`, the letter is dropped and the version comes out as `9`. The CPE analyzer then matches every CPE for version 9, among them the vulnerable `9a` and `9c`, which is the false positive described in the issue. The reporter had to work around it with a hint file.

The change adds the missing alternative to `RX_SINGLE_VERSION` so that both expressions accept the same suffixes. I left the pattern flags alone on purpose: `RX_VERSION` is `CASE_INSENSITIVE` and `RX_SINGLE_VERSION` is not, and adding the flag would also have changed how the other alternatives match, which is outside the scope of this issue.

Before changing anything I ran the current and the proposed expression side by side over the names used in the existing tests (`commons-lang3-3.12.0.jar`, `spring-core-5.3.9.RELEASE.jar`, `openssl-1.0.2k`, `lib1.5r4-someflag-R26.jar`, `6`, `utf8`, ...). Only inputs of the "digit followed by a single letter" shape change; everything else parses exactly as before.

## Related issues

- fixes #4139

yes — `DependencyVersionUtilTest.testParseVersion_singleLetterSuffixWithoutPeriod`, covering `9e`, `9d`, `8a` and the two names from the reporter's reproducer. On `main` it fails with `expected: <9e> but was: <9>`.

### DIFF
--- a/core/src/main/java/org/owasp/dependencycheck/utils/DependencyVersionUtil.java
+++ b/core/src/main/java/org/owasp/dependencycheck/utils/DependencyVersionUtil.java
@@ -45,7 +45,7 @@ public final class DependencyVersionUtil {
      * version number using the previous regex.
      */
     private static final Pattern RX_SINGLE_VERSION = Pattern.compile(
-            "\\d+(\\.\\d+){0,6}([._-]?(snapshot|release|final|alpha|beta|rc$|[a-zA-Z]{1,3}[_-]?\\d{1,8}))?");
+            "\\d+(\\.\\d+){0,6}([._-]?(snapshot|release|final|alpha|beta|rc$|[a-zA-Z]{1,3}[_-]?\\d{1,8}|[a-zA-Z]\\b))?");
 
     /**
      * Regular expression to extract the part before the version numbers if
--- a/core/src/test/java/org/owasp/dependencycheck/utils/DependencyVersionUtilTest.java
+++ b/core/src/test/java/org/owasp/dependencycheck/utils/DependencyVersionUtilTest.java
@@ -21,6 +21,7 @@ import org.junit.jupiter.api.Test;
 import org.owasp.dependencycheck.BaseTest;
 
 import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertNotEquals;
 import static org.junit.jupiter.api.Assertions.assertNull;
 
 /**
@@ -80,6 +81,23 @@ class DependencyVersionUtilTest extends BaseTest {
             }
             assertEquals(expected[i], result, "Failed extraction on \"" + names[i] + "\".");
         }
+    }
+
+    /**
+     * The user visible consequence of dropping the suffix: 8a and 8b both collapse onto "8", so
+     * the version CPEAnalyzer derives from the evidence compares equal to the version of a CPE
+     * for a different release. Its exact match is
+     * {@code parseVersion(evidence).equals(parseVersion(cpe.getVersion()))}, so a dependency at
+     * 8a matched the CPE of 8b, and the identified version printed in the report was "8".
+     */
+    @Test
+    void testParseVersion_singleLetterReleasesAreNotConflated() {
+        final DependencyVersion evidence = DependencyVersionUtil.parseVersion("libjpeg-turbo-8a.tar.gz", true);
+
+        assertEquals("8a", evidence.toString());
+        assertEquals(DependencyVersionUtil.parseVersion("8a"), evidence);
+        assertNotEquals(DependencyVersionUtil.parseVersion("8b"), evidence);
+        assertNotEquals(DependencyVersionUtil.parseVersion("9e"), DependencyVersionUtil.parseVersion("9d"));
     }
 
     /**
--- a/core/src/test/java/org/owasp/dependencycheck/utils/DependencyVersionUtilTest.java
+++ b/core/src/test/java/org/owasp/dependencycheck/utils/DependencyVersionUtilTest.java
@@ -61,6 +61,28 @@ class DependencyVersionUtilTest extends BaseTest {
     }
 
     /**
+     * Test of parseVersion method, of class DependencyVersionUtil, for versions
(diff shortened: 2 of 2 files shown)


chadlwilson (approved):
Probably OK if the tests all pass? 

Not really sure though, hard to understand what the consequence of changing this logic is downstream and whether it will lead to correct matching since the (presumably naive AI generated) PR does not demonstrate any end-user functionality improvement (being able to match libjpeg correctly to vulns), and the version parsing has additional risk given it is used in all sorts of places - sometimes needing consistency with separate CPE version sorting/parsing rules for version "piece" or range matching and also compatibility with weird hacks in `CPEAnalyzer` like:

https://github.com/dependency-check/DependencyCheck/blob/a318a7e06b40029e8a05fcff46279082feefa9bf/core/src/main/java/org/owasp/dependencycheck/analyzer/CPEAnalyzer.java#L822-L838

Eljees:
Fair point — the tests asserted the parser output, not the thing that matters. Measured on this branch against its parent:

| | before | after |
|---|---|---|
| `parseVersion("libjpeg-turbo-8a.tar.gz")` | `8` | `8a` |
| `parseVersion("8b")` | `8` | `8b` |
| 8a evidence `.equals()` 8b CPE | `true` | `false` |

The last row is the user-visible effect. `CPEAnalyzer`'s exact match is `parseVersion(evidence).equals(parseVersion(cpe.getVersion()))`, and both sides went through the same expression — so dropping the suffix never stopped libjpeg from matching, it made the releases indistinguishable. A dependency at `8a` matched the CPE of another release (`9a`/`9c` in the reporter's case), and the identified version printed in the report was `8`. So it is a false positive going away, not a new match appearing. Pinned by a test in 995b8b3.

On the downstream risk: any version that parsed before parses identically — `RX_SINGLE_VERSION` only gained one alternative for a trailing single letter, so the only inputs whose result changes are the ones that previously lost their suffix. I also checked the `CPEAnalyzer` block you linked: it peels a `v`/`rc`/`20xx`-style update off the last version part, and `8a` does not match that pattern, so it takes the same path as before.

Eljees:
Note that `995b8b3` (the test asserting single-letter releases are no longer conflated) landed after your approval, so GitHub reset the review state — nothing else changed. Could you take another look when convenient?

chadlwilson (approved):
Probably OK for me if @jeremylong agrees. Hell know this area much better, especially any "gotchas" on the parsing here vs CPE range matching.

jeremylong (approved):

jeremylong:
This has the potential to create some noise - but I understand the need.  I'll go ahead and merge this, but will need to monitor the FP reports a bit...

PR #8729, opened 2026-08-09
fix(maven): report the CVSS score that actually failed the build

Eljees:
## Description of Change

`checkForFailure` fails the build when **any** of a vulnerability's CVSS scores reaches `failBuildOnCVSS`, but the score it prints is the one from the newest CVSS version present. When an older version is what actually crossed the threshold, the message quotes a score below the threshold it states in the same sentence:

    One or more dependencies were identified with vulnerabilities that have a CVSS
    score greater than or equal to '7.0': CVE-2021-42550(6.6)

There it is the CVSSv2 score of 8.5 that failed the build, not the 6.6 shown — so the message reads as a bug in the threshold rather than as a finding.

The score selection moves into a small helper that returns the score which reached the threshold, and keeps the previous order of preference (v4, then v3, then v2) for the case where no threshold applies. **Which vulnerabilities fail the build is unchanged** — only the number shown in the message changes.

## Related issues

- relates to #5658

Yes. `BaseDependencyCheckMojoTest` gains cases for a v2-only threshold hit, a hit on a newer version, and the no-threshold path. They fail on current `main` — the assertion sees the newest-version score — and pass with the change; `mvn -pl maven test` goes from `BUILD FAILURE` to `Tests run: 6, Failures: 0, Errors: 0`.

AI-assisted (LLM used for drafting); the runs above are mine.

### DIFF
--- a/maven/src/main/java/org/owasp/dependencycheck/maven/BaseDependencyCheckMojo.java
+++ b/maven/src/main/java/org/owasp/dependencycheck/maven/BaseDependencyCheckMojo.java
@@ -2907,14 +2907,9 @@ public abstract class BaseDependencyCheckMojo extends AbstractMojo implements Ma
                         || unscoredCvss >= failBuildOnCVSS
                 ) {
                     String name = v.getName();
-                    if (cvssV4 >= 0.0) {
-                        name += "(" + cvssV4 + ")";
-                    } else if (cvssV3 >= 0.0) {
-                        name += "(" + cvssV3 + ")";
-                    } else if (cvssV2 >= 0.0) {
-                        name += "(" + cvssV2 + ")";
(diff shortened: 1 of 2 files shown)


Eljees:
Closing this as a duplicate of #8727, which carries the same change — the same two files and the same diff — and was opened first. Apologies for the noise; I opened this without checking that the fix was already submitted, and that's on me.

Nothing is lost by closing it: #8727 has the identical patch and the same test cases. Please review there. If any wording in this description reads better than what's on #8727, say so and I'll fold it in rather than open anything further.

chadlwilson:
Why don’t you close it yourself? you’re not in control of your bot-slop, you risk getting banned from projects for time wasting.

PR #8731, opened 2026-08-10
Release/13.0.1 0.1

andrewnewsome:


andrewnewsome:
Apologies for this. This PR was incorrectly created against the parent repository, rather than my fork (created to trial some changes before formally requesting merges/feature requests).

Incidentally, most of the files changed here relate to a change to the Maven group ID as to avoid collisions when using in some personal projects.

PR #8758, opened 2026-08-20
fix: Add eternal configuration for NODEAUDIT cache results.

### LINKED ISSUE #8757: CENTRAL - NODEAUDIT - POM  Cached results never expire
**Precondition**
- [X] I checked the issues list for existing open or closed reports of the same problem.

**Describe the bug**
When caches are activated for CENTRAL - NODEAUDIT and POM, results are cached but with the internal flag isEternal set to true.

The problem occurs using version 13.0.0 of the maven plugin, but it affects the core module

**Log file**

**To Reproduce**
1. Run a first dependency check, on a module containing node module and specially file package-lock.json, with default configuration. Cache should be created, in folder org/owasp/dependency-check-data/11.0/cache under maven respository
2. Wait 24h and expect there is new vulnerability on the module
3. Run again the check
4. The new vulnerability is not thrown

**Expected behavior**
Cache results should expire after the configuration time

**Additional context**
Difficult to check with log file or module, as it depends on known vulnerabilities.

The error is in the file dependencycheck-cache.properties

This will apply to "default" configuration
```
jcs.default.elementattributes.IsEternal=false
```

But the configuration is not set on region CENTRAL - POM - NODEAUDIT

But in the lib commons-jc3-core lib, default values are not applied as default value on region configuration. It is used when the configuration for the region does not provide the key jcs.region.<REGION>.elementattributes, usually with value org.apache.commons.jcs3.engine.ElementAttributes.

A temporary fix for the moment is to add a system variable in Maven command, like 
```
-Djcs.region.CENTRAL.elementattributes.IsEternal=false
-Djcs.region.POM.elementattributes.IsEternal=false
-Djcs.region.NODEAUDIT.elementattributes.IsEternal=false

```
But if caches already contain result, the eternal flag is already set to true in cached result. Arguments from command line do not solve the issue.

Second solution, update the file dependencycheck-cache.properties with:
```
jcs.region.CENTRAL.elementattributes.IsEternal=false
jcs.region.POM.elementattributes.IsEternal=false
jcs.region.NODEAUDIT.elementattributes.IsEternal=false
```
But it will need to delete the existing cache, for the same reason as previous fix.

ejouvin:
## Description of Change

Update configuration file to set isEternal flag value to false, only on NODEAUDIT cache.
Without this, the cache is never refresh.

In documentation for configuration nodeAuditAnalyzerUseCache
Sets whether the Node Audit Analyzer will cache results. Cached results expire after 24 hours.

But it is never the case, results in the cache never expire as they are flagged as eternal.

## Related issues

fixes #8757

yes

### DIFF
--- a/core/src/main/resources/dependencycheck-cache.properties
+++ b/core/src/main/resources/dependencycheck-cache.properties
@@ -47,6 +47,7 @@ jcs.region.NODEAUDIT.cacheattributes=org.apache.commons.jcs3.engine.CompositeCac
 jcs.region.NODEAUDIT.elementattributes=org.apache.commons.jcs3.engine.ElementAttributes
 jcs.region.NODEAUDIT.cacheattributes.MaxObjects=0
 jcs.region.NODEAUDIT.cacheattributes.DiskUsagePattern=UPDATE
+jcs.region.NODEAUDIT.elementattributes.IsEternal=false
 #24 hour default cache life
 jcs.region.NODEAUDIT.elementattributes.MaxLife=86400
 jcs.region.NODEAUDIT.elementattributes.IsSpool=true


chadlwilson (changes requested):
This may be valid, but as noted at https://github.com/dependency-check/DependencyCheck/issues/8757#issuecomment-5353621727 NodeAuditAnalyzer is not functional right now due to https://github.com/dependency-check/DependencyCheck/issues/8422, so there's no way to validate whether this is actually working or whether it's the correct approach or even the fix to the cited problem.

And still no reproducer supplied with logs or anything else to make it easier to validate when NodeAudit is reinstated?

PR #8777, opened 2026-08-31
fix(fp): remove unnecessary generated/hosted suppression

chadlwilson:
## Description of Change

I clicked too fast in approving this; user had stale suppressions as I had handled this through a generic suppression a couple of months ago.

## Related issues

- relates to https://github.com/dependency-check/DependencyCheck/issues/8774#issuecomment-5480325634

N/A


nhumblot (approved):
Thank you for your help!

Issue #8781, opened 2026-09-02
[FP]: CVE-2004-1304 does not apply to Karambolo.Extensions.Logging.File:4.0.0


git-hub-user:
### Package URl

pkg:nuget/Karambolo.Extensions.Logging.File@3.6.0 

### CPE

cpe:2.3:a:file:file:3.6.0:*:*:*:*:*:*:*
cpe:2.3:a:file_project:file:3.6.0:*:*:*:*:*:*:* 

### CVE

CVE-2004-1304

### ODC Integration

CLI

### ODC Version

13.0.0

### Description

I do not think `[CVE-2004-1304](https://nvd.nist.gov/vuln/detail/cve-2004-1304)` could apply to `Karambolo.Extensions.Logging.File:4.0.0`. This CVE has this description:
`Stack-based buffer overflow in the ELF header parsing code in file before 4.12 allows attackers to execute arbitrary code via a crafted ELF file.` I assume this is about the linux tool `file`, which should be completely unrelated to `Karambolo.Extensions.Logging`.

chadlwilson:
Please edit the Package URL/PURL so it matches what's in your report: https://github.com/package-url/purl-spec

chadlwilson:
There is also already a suppression that should cover this.

https://github.com/dependency-check/DependencyCheck/blob/572d4111afd70a7ee3f64936e97ccdcac35593b0/core/src/main/resources/dependencycheck-base-suppression.xml#L600-L604

I dont think you have filled in your OWASP Dependency Check version correctly either, perhaps that is a jenkins plugin - that is not maintained here. We need to know the actual Dependency Check version.

git-hub-user:
I reran with the CLI, version 12.2.0 and 13.0.0. By command line: `dependency-check.bat -s [.Net solution directory] --nvdApiKey *** -f JSON`

Both report the vulnerability.

|Dependency|Karambolo.Extensions.Logging.File:3.6.0|
|---|---|
|Vulnerability IDs|cpe:2.3:a:file:file:3.6.0:*:*:*:*:*:*:*; cpe:2.3:a:file_project:file:3.6.0:*:*:*:*:*:*:*|
|Package|pkg:nuget/Karambolo.Extensions.Logging.File@3.6.0|
|Highest Severity|HIGH|
|CVE Count|9|
|Confidence|Low| 
|Evidence Count|8|

[...]

PR #8784, opened 2026-09-03
fix(fp): consolidate file:file nuget suppressions

chadlwilson:
## Description of Change

This is a C library, not a nuget package.

## Related issues

-relates to #8781

N/A


nhumblot (approved):
Thank you for your help!

